PR #80, opened 2026-07-16
feat: add self-update command

### LINKED ISSUE #79: Add a self-update command
## Problem

Updating a standalone deja installation currently requires rerunning the shell installer or downloading and replacing the binary manually. Other agent CLIs such as Codex and Grok Build expose a native update command.

## Proposed solution

Add `deja update` to:

- resolve the latest stable GitHub release for the current OS and architecture
- download the matching release archive and verify it against `checksums.txt`
- replace the running standalone binary in place while preserving its permissions
- report already-current versions and avoid downgrading a newer build
- handle Unix and Windows replacement safely, with rollback on replacement failure
- direct Homebrew and npm installations to their package managers

## Acceptance criteria

- Supports the Linux, macOS, and Windows amd64/arm64 assets produced by GoReleaser.
- Checksum, HTTP, or extraction failures leave the existing binary unchanged.
- The command and its explicit GitHub network access are documented.
- Tests cover tar and ZIP assets, current/newer versions, failed downloads, checksum failures, permissions, and platform naming.

shgpavel:
## Summary

- add `deja update` for standalone installations using the latest stable GitHub release for the current OS and architecture
- download with Go's native HTTPS client, enforce HTTPS redirects and response limits, and remove the external `curl` dependency
- verify `checksums.txt`, preserve executable permissions, avoid downgrades, and replace binaries with Unix atomic rename or Windows rollback/backup handling
- build and test with Go 1.25, the oldest supported Go line that emits the Darwin `LC_UUID` required by current macOS

## Test plan

- Go 1.25.12: `go mod verify`, `go vet ./...`, and race tests with coverage
- Go 1.25.12: `sh scripts/e2e.sh`
- golangci-lint 2.12.2 and actionlint 1.7.12
- all six release cross-builds plus Darwin and Windows test-binary compilation
- verify `LC_UUID` in both Darwin release architectures
- GoReleaser 2.17.0 snapshot release
- source and release-binary `govulncheck`
- live update from v0.9.2 to v0.10.0 with an empty `PATH`, plus an already-current v0.10.0 check

## Notes

Go 1.24 is the technical `LC_UUID` boundary, but it is end-of-life and its final patch has reachable standard-library vulnerabilities. Go 1.25 is therefore the oldest safe supported build line.

The updater intentionally targets standalone binaries. Homebrew and npm installations should continue updating through their package managers so installed-version metadata remains consistent.

Closes #79

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-go@v6
         with:
-          go-version: '1.22.x'
+          go-version: '1.25.x'
           cache: false
       - run: go mod verify
       - run: go vet ./...
@@ -45,7 +45,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-go@v6
         with:
-          go-version: '1.22.x'
+          go-version: '1.25.x'
           cache: false
       - run: sh scripts/e2e.sh
 
@@ -56,7 +56,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-go@v6
         with:
-          go-version: '1.22.x'
+          go-version: '1.25.x'
           cache: false
       - name: Build all release targets
         run: |
@@ -74,7 +74,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-go@v6
         with:
-          go-version: stable
+          go-version: '1.25.x'
           cache: false
       - run: go install golang.org/x/vuln/cmd/govulncheck@latest
       - run: govulncheck ./...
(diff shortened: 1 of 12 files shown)


vshulcz (approved):
This is how a self-update command should be written. Went through it strictly: HTTPS enforced including redirects, size limits on every download, checksum verified against checksums.txt before anything touches disk, zip/tar extraction to memory only (no path traversal surface), atomic rename on unix, rollback + stale-backup cleanup on windows, downgrade refusal, perms preserved. Also verified the LC_UUID rationale — Go 1.24+ emits it and macOS 15 needs it, so the toolchain bump is justified (golang/go#68678).

Ran it live: built the branch, `deja update` took the dev binary to v0.10.0 with the checksum verified, single atomic replace. Tests and e2e green here too.

Notes for later, none blocking:
- checksums.txt comes from the same release as the archive, so this verifies integrity, not provenance. Artifact signing (sigstore/minisign) would close that if we ever care.
- a dev build updates itself without a prompt (version check is skipped for "dev"). Reasonable default, just stating it so it's a decision and not an accident.

Thanks — between this and the grok parser you've basically become the co-maintainer.

PR #81, opened 2026-07-16
redact: catch modern provider keys, env/JSON shapes, and more token families

suthakamal2:
## Problem

The redactor misses the most common ways an API key appears in an agent transcript, so those keys reach the local index — and therefore `share` digests and `sync export` batches, which re-run the same pass. That undercuts the core "the cache is safe to keep / hand to a colleague" promise.

## Gaps closed

- **Modern provider keys.** `providerRE`'s `sk-[A-Za-z0-9]{20,}` stopped at the first hyphen — matching legacy `sk-<alnum>` keys but not `sk-ant-…` (Anthropic) or `sk-proj-…` (OpenAI project). Now allows internal `-`/`_`.
- **More token families.** Added Groq `gsk_`, xAI `xai-`, HuggingFace `hf_`, GitLab `glpat-`, Stripe `sk_live_/sk_test_/rk_*`, and GitHub server/user/refresh `ghs_/ghu_/ghr_`, each labeled by kind. `awsAccessKeyRE` now also matches temporary `ASIA…` IDs, not only `AKIA…`.
- **Env-var and JSON key shapes.** `genericKVRE` (the labeled-key fallback) missed `ANTHROPIC_API_KEY=…` (no word boundary when the key is embedded in a larger identifier) and JSON `"api_key": "…"` (the key's closing quote sits between the name and the colon). Both are now tolerated; same delimiter fix applied to the AWS-secret pattern.
- **Password containing `@`.** `connURLRE` redacted only up to the first `@`, leaking the rest of a `user:p@ss@host` password into the preserved host span. Password is now matched greedily and split on the last `@`.
- **Short JWTs.** `jwtRE` required all three segments ≥8 chars, missing valid tokens with a near-empty-claims payload. Payload/signature floor is now 4; the `eyJ` header anchor stays at 8.

## Tests

New table tests cover the env-var/JSON/bare/modern-key forms, the added token families, `ASIA`, and a password-containing-`@` URL, plus a false-positive guard that ordinary prose mentioning "token"/"secret"/"password" is **not** redacted. Fixtures are assembled at runtime, so no token literal enters the source (keeps push protection happy). `go build ./...`, full `go test ./...`, `gofmt`, `go vet`, and `staticcheck` are all clean.

Reviewed cross-family (a GPT-class reviewer against the redaction contract) — its findings (Stripe/GitHub token families, `ASIA`, the `@`-in-password split, short JWTs) are folded in above.

Measured impact on a real multi-harness corpus (~6.5k sessions): redaction hits rose 5→9 on one store and 19→38 on another — real credential occurrences that previously reached the index.

### DIFF
--- a/internal/redact/redact.go
+++ b/internal/redact/redact.go
@@ -23,14 +23,23 @@ func (c Counts) Total() int {
 }
 
 var (
-	awsAccessKeyRE = regexp.MustCompile(`AKIA[0-9A-Z]{16}`)
-	awsSecretRE    = regexp.MustCompile(`(?i)\b(aws[_-]?secret[_-]?access[_-]?key)(\s*[:=]\s*)(['"]?)([A-Za-z0-9/+=_-]{32,})(['"]?)`)
-	genericKVRE    = regexp.MustCompile(`(?i)\b(api[_-]?key|secret|token|passwd|password|authorization)(\s*[:=]\s*)(['"]?)([A-Za-z0-9/+=._-]{16,})(['"]?)`)
-	bearerRE       = regexp.MustCompile(`(?i)\b(Bearer)(\s+)([A-Za-z0-9._~+/=-]{16,})`)
-	pemPrivateRE   = regexp.MustCompile(`(?s)-----BEGIN [A-Z0-9 ]*PRIVATE KEY-----.*?-----END [A-Z0-9 ]*PRIVATE KEY-----`)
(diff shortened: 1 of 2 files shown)


vshulcz (approved):
Reviewed line by line and ran it against an adversarial probe set (JSON key shapes, ANTHROPIC_API_KEY, sk-ant/sk-proj, glpat, xai-, stripe live/test, password-with-@, two URLs on one line) — everything redacts the way the comments promise, and clean text stays untouched. The inline comments explaining each regex decision made this easy to review, thanks for that.

Two data points from testing, neither blocking:
- the wider genericKV prefix costs some regex throughput; on my real 3.3 GB corpus a cold rebuild goes 14.1s -> 15.4s. Fine for what it buys. A cheap strings.Contains prefilter before the regex could win it back later.
- `https://host:443/path@something` (port + @ later in the token) now gets treated as credentials and over-redacted. Safe direction, just noting the tradeoff exists.

PR #82, opened 2026-07-16
feat: respect CLAUDE_CONFIG_DIR for read and install paths

globalundo:
## Summary

`deja` hard-coded `~/.claude`, ignoring Claude Code's `CLAUDE_CONFIG_DIR` — the standard way to run separate profiles (personal/work).
With `CLAUDE_CONFIG_DIR` set, `deja install` writes the MCP server, hook and statusline into the wrong (inactive) profile, and indexing reads an empty `~/.claude/projects`.

This honors `CLAUDE_CONFIG_DIR` for both the read and install paths, falling back to `~/.claude`. `DEJA_CLAUDE_ROOT` still takes precedence — the same pattern `grok.go` already uses for `GROK_HOME`. Existing setups are unaffected.

## Test plan

- New tests: the path helpers follow `CLAUDE_CONFIG_DIR` (with `DEJA_CLAUDE_ROOT` still winning); `install` lands in the configured profile and leaves the default `~/.claude*` untouched.
- `go build ./cmd/deja`, `go test -race ./...`, `gofmt -l`, `go vet ./...` — all clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ Credentials are redacted at index time: AWS keys, generic `api_key=`/`token=` as
 | Antigravity | `~/.gemini/antigravity*/brain/*/.system_generated/logs/transcript.jsonl` | ✅ |
 | Grok Build | `~/.grok/sessions/**/updates.jsonl` | ✅ |
 
-Custom locations via `DEJA_CLAUDE_ROOT`, `DEJA_CODEX_ROOT`, `DEJA_OPENCODE_DB`, `DEJA_AIDER_ROOTS`, `DEJA_GEMINI_ROOT`, `DEJA_CURSOR_ROOT`, `DEJA_CURSOR_CLI_ROOT`, `DEJA_ANTIGRAVITY_ROOT`, `DEJA_GROK_ROOT`, `DEJA_INDEX_DIR`. Grok's native `GROK_HOME` is also honored.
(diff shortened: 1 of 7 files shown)


vshulcz (changes requested):
Solid fix and the .claude.json-moves-inside-the-config-dir subtlety is handled exactly right, tests included. Ran the branch and a rebase onto main — suite green both ways, no conflicts.

One gap, from code that landed after you branched: `cmd/deja/doctor.go` now duplicates the install paths for its MCP-wiring report, and line ~165 still hard-codes `~/.claude.json`. Reproduced on your branch rebased onto main:

    CLAUDE_CONFIG_DIR=/tmp/probe-profile deja doctor
    # MCP wiring:
    #   claude-code  wired  /Users/shulcz/.claude.json   <- default profile, not the active one

Swapping that line to `sources.ClaudeJSONPath()` closes it (the harness-stores section already picks up your change through ClaudeRoot()). Rebase + that one line and this is good to merge. If you'd rather I fold it in, say the word.

vshulcz (approved):
Took you up on finishing it maintainer-side: rebased onto main (your changes now sit alongside #103, which applies your exact pattern to the other harnesses — CODEX_HOME, GEMINI_CLI_HOME, CURSOR_CONFIG_DIR and friends, credited to this PR's approach) and folded the doctor.go line in during conflict resolution. Verified live: with CLAUDE_CONFIG_DIR set, doctor now reports the active profile. Nice work — this closed a real multi-profile gap and set the pattern for the whole fleet.

vshulcz:
Linking #101 for context — your CLAUDE_CONFIG_DIR approach became the template for every other agent's relocation variable.

PR #156, opened 2026-07-19
fix(redact): stop xai- pattern matching kebab-case prose

Sora-bluesky:
## Problem

`providerRE`'s xAI arm — `xai-[A-Za-z0-9-]{20,}` — allows internal hyphens, so any long kebab-case slug that happens to start with `xai-` is redacted as a key: branch names (`xai-oauth-correction-loop-retry`), doc file names (`xai-rate-limit-troubleshooting-notes.md`), issue titles, and similar prose.

This isn't hypothetical: running this exact pattern over my real agent-transcript corpus produced **35 xai- redactions, and all 35 were false positives** (hyphenated prose, zero real keys).

## Fix

Real xAI keys are alphanumeric after the `xai-` prefix, so drop `-` from the character class: `xai-[A-Za-z0-9]{20,}`. The `sk-` arm keeps its hyphens intentionally (sk-ant-…, sk-proj-… are real formats) — a comment now records why the two differ.

Added two kebab-case slugs to `TestTextKeepsOrdinaryProse`; they fail against the old regex and pass with the fix. The existing positive case (`fake("xai-", 32)`, alphanumeric) still passes. `go test ./...`: 565 passed.

## Context (partial field report for #9)

Found while verifying deja v0.12.0 on a real Windows 11 machine, which may be useful for #9:
- store discovery: `%USERPROFILE%\.claude\projects` (138 sessions), `.codex` (328), `.grok` (101), `.gemini\antigravity-cli` (7) — all found, index built fine
- search/doctor/stats work in Git Bash and PowerShell (Windows Terminal)
- MCP `deja mcp` handshake + `recall` verified over stdio (registered via `cmd /c deja mcp` in Claude Code)

### DIFF
--- a/internal/redact/redact.go
+++ b/internal/redact/redact.go
@@ -34,8 +34,11 @@ var (
 	pemPrivateRE = regexp.MustCompile(`(?s)-----BEGIN [A-Z0-9 ]*PRIVATE KEY-----.*?-----END [A-Z0-9 ]*PRIVATE KEY-----`)
 	// Provider prefixes. sk- allows internal hyphens/underscores so modern
 	// hyphenated formats (sk-ant-…, sk-proj-…) are covered, not just legacy
-	// sk-<alnum> keys.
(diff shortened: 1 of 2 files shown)


vshulcz (approved):
Thanks — good catch and exactly the right scope: real xai keys are alphanumeric, so the hyphen class only bought false positives on kebab-case slugs. The prose regression tests are appreciated.

PR #158, opened 2026-07-19
fix(search): quote context messages per-term, never return a bare header

### LINKED ISSUE #157: recall_context returns a header-only digest for multi-word queries
## What happens

MCP `recall_context` (and `deja context <query>` on the CLI) can return just the header line with no digest body:

```
# deja context: claude · apps\playground · b5351d10-… · updated 2026-07-19
```

Observed on v0.12.0 (Windows 11, real corpus): `recall` finds the session fine, but `recall_context` with the query `Turso Platform API rotate token Vercel` returned only the header. Re-querying the same session with a token that literally appears in one message (`filter-repo raw/transcripts invert-paths`) returned a proper digest.

## Why

`PrintContext` decides which messages to quote with a full-phrase substring check ([search.go#L341](https://github.com/vshulcz/deja-vu/blob/595d9bc/internal/search/search.go#L341)):

```go
matched := qlow != "" && strings.Contains(strings.ToLower(m.Text), qlow)
if !matched && m.Role != "user" {
    continue
}
```

Search itself (`search.Run` → `MatchesParts`) matches per-message on ANDed terms, so a multi-word query happily selects a session — but the digest side then requires the entire query to appear as one contiguous substring in a single message, which multi-word queries almost never satisfy. Every non-user message is dropped, and if the session's user messages have no printable prose either (common for agent sessions that are mostly pasted terminal output), nothing but the header is written.

Two problems fall out of this:

1. **Asymmetric matching** — the digest should quote messages using the same term semantics the search used to pick the session.
2. **No floor** — even when nothing matches, a bare header is returned with no hint of what went wrong or what to try instead.

## Suggested fix

- Reuse `QueryParts` + `MatchesParts` in `PrintContext` so a message qualifies when it contains all query terms (in any order), keeping the existing full-phrase check as a fast path.
- If nothing gets written after the message loop, fall back to a session overview (opening messages) with an explicit note, so a header-only digest becomes impossible.
- Nudge the MCP tool description toward verbatim-token queries.

I have a small patch with tests ready and will open a PR.

Sora-bluesky:
Fixes #157

## Problem

`PrintContext` quotes a non-user message only when the **entire query** appears as one contiguous substring of that message, while session selection (`search.Run` → `MatchesParts`) matches ANDed terms. A multi-word query like `Turso Platform API rotate token Vercel` therefore selects a session but drops every assistant/tool message from the digest, and when the session's user messages have no printable prose (agent sessions are often mostly pasted terminal output), `recall_context` returns literally one header line. Reproduced on a real v0.12.0 corpus; details in #157.

## Fix

- `PrintContext` now qualifies a message via `QueryParts` + `MatchesParts` — the same term semantics search used to pick the session — keeping the old full-phrase `strings.Contains` as a fast path.
- If the message loop writes nothing, fall back to a session overview (opening messages, any role) with an explicit note, so a header-only digest is structurally impossible. The empty-query CLI path (`deja context` on a session with no query) keeps its current behavior and gains the same fallback without the note.
- The `recall_context` MCP tool description now tells the calling model to query with tokens that appear verbatim in the transcript.

## Tests

`TestPrintContextTokenMatchAndFallback` covers both halves: out-of-order tokens in one message (fails on the old regex-free full-phrase check with exactly the bare-header output) and terms spread across messages (expects the overview fallback). Verified the new test fails against the old implementation before applying the fix. `go test ./...`: 566 passed.

### DIFF
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -84,7 +84,7 @@ func handleMCP(req rpcRequest) (any, int, string) {
 			},
 			{
 				"name":        "recall_context",
-				"description": "Return a full markdown digest (~8KB) of the single best-matching prior session — problem, decisions, outcome — when a bare recall hit is not enough and you need the reasoning behind it. Use after recall, or directly when the user asks 'remind me how we handled X' or 'what was the whole story with Y'. Query with the error string, function name, or flag that identifies the session. Not for browsing many sessions — use recall for that; this returns one deep digest.",
+				"description": "Return a full markdown digest (~8KB) of the single best-matching prior session — problem, decisions, outcome — when a bare recall hit is not enough and you need the reasoning behind it. Use after recall, or directly when the user asks 'remind me how we handled X' or 'what was the whole story with Y'. Query terms are matched against transcript text, so use tokens likely to appear verbatim: an error string, function name, or flag. Not for browsing many sessions — use recall for that; this returns one deep digest.",
 				"annotations": map[string]any{"title": "Digest one past session", "readOnlyHint": true, "openWorldHint": false},
 				"inputSchema": map[string]any{"type": "object", "properties": map[string]any{"query": map[string]any{"type": "string", "description": "Search terms identifying the session to digest."}, "harness": map[string]any{"type": "string", "description": "Optional harness filter."}}, "required": []string{"query"}},
 			},
--- a/internal/search/search.go
+++ b/internal/search/search.go
@@ -376,14 +376,31 @@ func PrintContext(w io.Writer, s model.Session, query string) {
 	}
 	fmt.Fprintln(w)
 	qlow := strings.ToLower(query)
+	terms, phrases := QueryParts(query)
 	budget := 8000
+	written := printContextChunks(w, s, budget, func(m model.Message) (bool, bool) {
+		matched := qlow != "" && (strings.Contains(strings.ToLower(m.Text), qlow) || MatchesParts(m.Text, terms, phrases, nil))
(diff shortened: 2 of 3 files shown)


vshulcz (approved):
Thanks for the report and the fix — reusing MatchesParts closes the asymmetry properly, and the overview fallback with an explicit note is the right floor. Nice refactor into printContextChunks.

vshulcz:
Heads-up: #155 rewrote the MCP tool descriptions and landed first, so this now conflicts in cmd/deja/mcp.go (the recall_context description). The resolution that keeps both intents — my rewrite plus your verbatim-token guidance:

```
"description": "Return a full markdown digest (~8KB) of the single best-matching prior session — problem, decisions, outcome — when a bare recall hit is not enough and you need the reasoning behind it. Use after recall, or directly when the user asks 'remind me how we handled X' or 'what was the whole story with Y'. Query terms are matched against transcript text, so use tokens likely to appear verbatim: an error string, function name, or flag. Not for browsing many sessions — use recall for that; this returns one deep digest.",
```

(keep the annotations line from main as is). Rebase with that and this merges — everything else applies cleanly, full suite is green on my side with the resolution.

Sora-bluesky:
Rebased onto main with exactly your proposed resolution — your #155 description plus the verbatim-token sentence, annotations line kept as is. Everything else applied cleanly; `go test ./...`: 582 passed. Thanks for the review and for #156!

vshulcz:
Merged — thanks for the quick rebase and for the pair of reports. Both fixes ship in 0.13.1.

PR #194, opened 2026-07-19
fix(doctor): warn when index is stale relative to source stores

### LINKED ISSUE #184: deja doctor: warn when the index is stale relative to the stores
`deja doctor` reports store discovery, MCP wiring and the index location, but not whether the index is actually up to date. A common confusion (seen in a field report) is a user searching and getting old results because the index hasn't been rebuilt since new sessions landed.

**What's needed**
- In `cmd/deja/doctor.go`/`doctor_report.go`, compare the newest source file mtime against the index manifest's build time and add one line: e.g. `index: up to date` or `index: 3 stores changed since last build — run deja index`.
- Include the same field in `deja doctor --json`.
- Keep it read-only and offline (no network); it's just an mtime comparison against the manifest.

Good first issue — the doctor report already gathers store file lists and the manifest; this adds one derived line. Hermetic tests, 95% coverage on the touched package.

syf2211:
## Summary

`deja doctor` now compares each harness store's newest source file against the index manifest build time and reports whether the index is up to date.

## Motivation

Users often search and get stale results because the index was not rebuilt after new sessions landed. The doctor report already gathered store paths and manifest metadata, but did not surface freshness.

Fixes #184

## Changes

- Add `index.ManifestBuiltAt` to read the manifest `BuiltAt` timestamp (fallback: `manifest.gob` mtime).
- Count stores whose newest file is newer than the manifest build time.
- Human output: `freshness up to date` or `N store(s) changed since last build — run deja index`.
- JSON output: `index.state` (`ok` / `stale`) and `index.stale_stores`.

## Tests

- `go test ./cmd/deja/... ./internal/index/...`
- Added `TestManifestBuiltAt`, expanded `TestDoctorIndexStates`, and `TestDoctorIndexFreshnessOutput`.

## Notes

Read-only and offline: only mtime comparison against the on-disk manifest, no network or index rebuild.

### DIFF
--- a/cmd/deja/coverage_recover_test.go
+++ b/cmd/deja/coverage_recover_test.go
@@ -99,10 +99,14 @@ func TestDoctorIndexBuiltBranch(t *testing.T) {
 		t.Fatal(err)
 	}
 	var out bytes.Buffer
-	doctorIndex(&out)
+	report := collectDoctorReport(nil)
+	doctorIndex(&out, report.Index)
 	got := out.String()
 	if !strings.Contains(got, "status   built") || !strings.Contains(got, "size=") || !strings.Contains(got, "updated=") {
 		t.Fatalf("doctorIndex built = %q", got)
+	}
+	if !strings.Contains(got, "freshness up to date") {
+		t.Fatalf("doctorIndex freshness = %q", got)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


vshulcz:
Merged — thanks! Clean multi-store handling and the mtime fallback for old manifests is a nice detail. This is a genuinely useful signal in `doctor`; people hit stale indexes and have no idea.

PR #195, opened 2026-07-19
feat: add shell completions

### LINKED ISSUE #185: Shell completions for bash, zsh and fish
deja has a growing set of subcommands and flags (`search`, `recall`, `blame`, `remember`, `forget`, `stats`, `install`, `bench`, …). Shell completions would make it noticeably nicer to use and are the kind of polish people notice.

**What's needed**
- A `deja completion <bash|zsh|fish>` subcommand that prints a completion script to stdout (no external dependency — deja is a zero-dep binary, so hand-write the scripts rather than pulling in a framework).
- Complete subcommands and their flags; for `install <harness>` complete the harness names; for `blame <path>` fall back to file completion.
- A short README section under Install with the one-liner per shell (e.g. `deja completion zsh > ...`).

Hermetic tests asserting the generated script contains the subcommands and is syntactically loadable where feasible. 95% coverage on the touched package. Keep the scripts small and readable.

floze-the-genius:
## Summary
- add `deja completion <bash|zsh|fish>` with dependency-free scripts
- complete commands, command flags, install targets, and file paths for `blame`
- document shell setup under Install

Closes #185

## Test plan
- `go test ./cmd/deja -count=1`
- `go vet ./cmd/deja`
- `go build ./cmd/deja`

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,14 @@ npx @vshulcz/deja-vu "query"                            # npm, no install
 brew install vshulcz/tap/deja-vu                        # Homebrew
 ```
 
+### Shell completion
+
+```sh
+deja completion bash >> ~/.bashrc
+deja completion zsh >> ~/.zshrc
+deja completion fish > ~/.config/fish/completions/deja.fish
+```
+
 Wire it into the agents you use (edits config, keeps a `.bak`):
 
 ```sh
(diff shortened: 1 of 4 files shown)


vshulcz:
Merged, thanks! Hand-written completions across all three shells with `sh -n` parse tests — exactly right for a zero-dep binary. I pushed one small follow-up commit dropping the internal `hook-context`/`hook-precompact` from the bash top-level list so it matches zsh/fish (they're config-invoked, not interactive). Appreciate the work.

PR #198, opened 2026-07-19
feat(stats): rebrand the shareable stats card

vshulcz:
`deja stats --card` now renders in the deja palette (gradient hero, teal harness bars, activity heatmap) with thousands-formatted numbers and a footer link back to the site — so when someone posts their card it advertises the repo. Same SVG contract, tests unchanged.


vshulcz:
Parking this — a stats card is table stakes that specialized tools already do better, not what makes deja worth using. Focusing on what's actually unique to deja instead.

Issue #186, opened 2026-07-19
deja search --json: stable, documented output schema for scripting


vshulcz:
People want to build on deja (dashboards, editor plugins, git hooks). `deja search --json` and `deja stats --json` exist, but the shapes aren't documented as a contract, so scripting against them is fragile.

**What's needed**
- Document the JSON shapes for `search --json`, `stats --json`, `doctor --json` and `blame --json` in `docs/` (a `docs/json-output.md` with a small example per command).
- Add a `schema_version` field to each top-level JSON envelope so consumers can detect breaking changes, and a note in the doc about the stability policy (additive changes are safe within a version).
- A test that pins each schema (golden or field-presence assertions) so an accidental shape change fails CI.

This is mostly wiring + docs + tests, no new behavior. Coverage 95% on touched packages; keep the envelopes backward compatible with existing fields.

PR #207, opened 2026-07-20
feat(last): add --since and --role filters matching search

### LINKED ISSUE #187: deja last: filter by date range and role, matching search
`deja last` takes a count and `--project`/`--harness` filters (from #10). `deja search` also supports `--since` and `--role`. Making `last` accept the same two filters removes an inconsistency people hit — 'show my recent sessions from this week' currently needs a full search.

**What's needed**
- Add `--since <duration|date>` and `--role <user|assistant>` to `deja last`, parsed the same way `search` parses them (reuse the existing parsing so the semantics match exactly).
- Update the usage text and the README `last` line.
- Tests covering the new filters and their combination with the existing `--project`/`--harness`; 95% coverage on the touched package.

Well-scoped: the filtering primitives already exist in the search path, this wires them into `last`.

syf2211:
## Summary
Add `--since` and `--role` flags to `deja last`, using the same duration parsing and filter semantics as `deja search` and `deja stats`.

## Motivation
Users often want recent sessions from a time window or role (e.g. "sessions from this week with user messages") without running a full search. Issue #187.

## Changes
- Extend `parseLast` to accept `--since` and `--role`
- Apply since/role filtering in `filterRecentSources` (indexed path uses `SearchWithRecovery` when role is set, matching stats)
- Update bash/zsh/fish completions, usage text, README, and docs guide
- Add unit and integration tests for the new filters and combinations

## Tests
- `go test ./cmd/deja/ -race -count=1 -run TestLast` — pass
- `go test ./... -race -count=1` — pass
- `go vet ./...` — pass
- `go build ./cmd/deja` — pass

## Notes
- Role filtering on the indexed path loads full session records (same trade-off as `deja stats --role`)
- Fixed in-place slice reuse in `filterRecentSources` that could corrupt the input slice during filtering

Fixes #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ $ deja "jwt refresh token"
 | `deja stats` | Headline counts, totals, per-harness split, top projects, monthly sparkline. `--json` too. |
 | `deja doctor [--json]` | Self-diagnosis: store parse state, sqlite3 presence, MCP wiring per agent, index health, version; `--json` emits the same checks for scripts. |
 | `deja sync export <dir> [--full]` / `import <dir>` / `ssh <host> [--pull]` | Move memory between machines — via a shared folder or one ssh command. Watermarked, append-only, idempotent. |
-| `deja show <id>` / `deja last [n]` | Read one session / list recent ones. |
+| `deja show <id>` / `deja last [n]` | Read one session / list recent ones (`--project`, `--harness`, `--since`, `--role`). |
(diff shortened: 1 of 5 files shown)


vshulcz:
Merged — thanks! Nice touch replacing the aliasing `ss[:0]` with a fresh slice while you were in there, and completions across all three shells is exactly the standard. Second merged PR from you — much appreciated. #186 (documented `--json` schema) is still open if you want another.

PR #217, opened 2026-07-20
fix(sources): resolve Claude project dirs encoded with a Windows drive letter

### LINKED ISSUE #216: bug(claude): project dirs encoded with a Windows drive letter never resolve — "domain-manage" is displayed as "domain\manage" and splits one project in two
## Environment

- deja built from `main` (a `go build ./cmd/deja`, CGO off) on Windows 11 Pro
  10.0.26200, Go 1.25
- Real corpus: 445 sessions / 144,752 messages indexed from
  `~/.claude/projects` + `~/.codex/sessions`

## Symptom

`deja stats` lists the same project twice:

```
Top projects
  domain-manage      ################## 180     <- codex source (stores real cwd)
  history            ########## 101
  domain\manage      #####  54                  <- claude source (Windows)
```

`domain\manage` is not a real directory. The project is
`C:\Users\daoti\Downloads\domain-manage`, and its Claude project dir is
`C--Users-daoti-Downloads-domain-manage`.

## Root cause

On Windows, Claude Code encodes the **drive root** as well, so the project dir
is `C--Users-x-app`, not `-Users-x-app`. `resolveEncodedPath`
(`internal/sources/claude.go`) starts with:

```go
if !strings.HasPrefix(base, "-") {
	return ""
}
```

so the Windows form bails immediately. `decodeProjectBase` then falls back to
the dash heuristic, which cannot distinguish an encoded separator from a
literal hyphen — `domain-manage` becomes `domain\manage`.

Proof on this machine (`ResolveEncodedPath` is exported):

```
ResolveEncodedPath("C--Users-daoti-Downloads-domain-manage") = ""
ResolveEncodedPath("-Users-daoti-Downloads-domain-manage")   = "\Users\daoti\Downloads\domain-manage"
```

The second only works because `C:` happens to be the current drive.

## Why the Windows path audit missed it

#54 skipped the resolution test on Windows with the rationale *"claude encodes
unix absolute paths, so directory recovery is a no-op on windows"*. That
assumption does not hold: Claude Code on Windows does create encoded project
dirs — recovery is not a no-op there, it is simply unimplemented. Because the
test is skipped, nothing on Windows exercises the resolver.

## Fix (PR attached)

Peel the encoded drive off and reuse the existing segment walk with `C:` as
the root:

```go
root := ""
if drive, rest, ok := splitEncodedWindowsDrive(base); ok {
	root, base = drive, rest
} else if !strings.HasPrefix(base, "-") {
	return ""
}
...
return try(root, parts[0], 1)
```

Verified end-to-end on the corpus above: after the fix the claude sessions
resolve to the real directory and the hyphenated name survives. Windows-gated
regression tests added for both the resolver and the decoded display name;
`go test ./...` is green on windows/amd64.

## Separate, lower-confidence observation

Even with the path resolved, the claude source labels a project
`<parent>\<leaf>` (`decodeProjectBase` joins the last two segments) while
every other source goes through `projectName()` in `internal/sources/util.go`,
which is `filepath.Base` — the leaf only. So the same project still shows as
`Downloads\domain-manage` (claude) vs `domain-manage` (codex), and cross-agent
grouping — the headline feature — splits.

This may well be intentional (the extra segment disambiguates same-named
leaves), so I left it alone in the PR. If leaf-only is preferred for the
claude source too, or a normalization at grouping time is wanted, happy to
follow up.

tiendungdev:
Fixes #216.

## Problem

Claude Code on Windows encodes the drive root too, so a project directory is
named `C--Users-x-Downloads-domain-manage`, not `-Users-x-...`.
`resolveEncodedPath` required a leading `-` and returned `""` for that form,
so `decodeProjectBase` fell through to the dash heuristic — which cannot tell
an encoded separator from a literal hyphen and renders `domain-manage` as
`domain\manage`.

On a real corpus (445 sessions / 144,752 messages) this split one project
across two entries in `deja stats`:

```
domain-manage   ################## 180    <- codex source (real cwd paths)
domain\manage   #####  54                 <- claude source (Windows)
```

## Fix

Peel the encoded drive off and hand the remainder to the existing segment
walk with `C:` as the root, so the on-disk probe recovers hyphenated names on
Windows exactly as it already does on unix. Non-Windows encodings take the
unchanged code path.

## Verification (Windows 11, go 1.25, windows/amd64)

- `ResolveEncodedPath("C--Users-daoti-Downloads-domain-manage")` before: `""`;
  after: the real directory.
- Re-indexed the corpus with the patched binary: the claude sessions resolve
  to the real path and `domain-manage` keeps its hyphen.
- Added windows-gated regression tests for the resolver and for the decoded
  display name.
- `go test ./...` green (cmd/deja, index, search, sources, redact, embed,
  bench, usage).

## Note on #54

#54 skipped the resolution test on Windows on the grounds that "claude
encodes unix absolute paths, so directory recovery is a no-op on windows".
That assumption does not hold on a Windows host — Claude Code does create
encoded project dirs there — which is why the resolver has had no Windows
coverage. The tests added here are windows-gated and exercise it directly.

### DIFF
--- a/internal/sources/claude.go
+++ b/internal/sources/claude.go
@@ -162,7 +162,13 @@ func decodeProjectBase(base string) string {
 // at. Segments are re-joined with "/" or "-" and pruned by checking each
 // completed directory prefix on disk.
 func resolveEncodedPath(base string) string {
-	if !strings.HasPrefix(base, "-") {
+	// Windows encodes the drive root too: C:\Users\x\app becomes
+	// "C--Users-x-app". Peel the drive off and let the shared segment walk
+	// resolve the rest from "C:" instead of from an empty root.
+	root := ""
+	if drive, rest, ok := splitEncodedWindowsDrive(base); ok {
+		root, base = drive, rest
+	} else if !strings.HasPrefix(base, "-") {
 		return ""
 	}
 	// pi uses "--" prefix/suffix (e.g. --Users-x-app--), Claude uses
@@ -198,7 +204,21 @@ func resolveEncodedPath(base string) string {
 	if parts[0] == "" {
 		return ""
 	}
-	return try("", parts[0], 1)
+	return try(root, parts[0], 1)
+}
+
+// splitEncodedWindowsDrive recognises the "C--Users-x-app" form Claude Code
+// writes on Windows and returns the drive root ("C:") plus the remainder in
+// the unix-style encoding the resolver already understands ("-Users-x-app").
+func splitEncodedWindowsDrive(base string) (root, rest string, ok bool) {
+	if len(base) < 4 || base[1] != '-' || base[2] != '-' {
+		return "", "", false
+	}
+	c := base[0]
+	if (c < 'A' || c > 'Z') && (c < 'a' || c > 'z') {
+		return "", "", false
+	}
+	return string(c) + ":", base[2:], true
 }
 
 // ClaudeProjectName derives the display project name using the same rules as
--- a/internal/sources/claude_drive_split_test.go
+++ b/internal/sources/claude_drive_split_test.go
@@ -1,0 +1,30 @@
+package sources
+
+import "testing"
+
+// splitEncodedWindowsDrive's whole job is to fire on the Windows "C--…" form
+// and NEVER on unix or pi encodings — a false positive would rewrite good
+// project paths. This runs on every platform (unlike the windows-gated
+// resolver tests) because that invariant is what regresses silently.
+func TestSplitEncodedWindowsDrive(t *testing.T) {
+	cases := []struct {
+		in   string
+		root string
+		rest string
+		ok   bool
+	}{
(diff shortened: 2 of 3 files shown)


vshulcz:
Merged — thanks, this is a clean fix for a real corpus-splitting bug, and the diagnosis (drive root encoded too, dash heuristic can't tell a separator from a literal hyphen) is exactly right. Good catch on the #54 assumption too — you're correct that Claude Code does create encoded project dirs on a Windows host, so the resolver genuinely had no coverage there. I pushed one small follow-up: a cross-platform table test for `splitEncodedWindowsDrive` itself, since your resolver tests are windows-gated and the no-false-positive invariant (never hijacking unix/pi encodings) is the part most likely to regress silently on the machines most contributors run. Appreciate the thorough writeup and the field verification on a 445-session corpus.

PR #233, opened 2026-07-21
docs: document JSON output schemas and add schema_version

### LINKED ISSUE #186: deja search --json: stable, documented output schema for scripting

syf2211:
## Summary

Document the stable JSON shapes for `deja search --json`, `stats --json`, `doctor --json`, and `blame --json` in `docs/json-output.md`, and add a `schema_version` field to object-shaped CLI JSON envelopes.

## Motivation

Fixes #186. Scripting against `--json` output was fragile because the shapes were undocumented and had no version marker for breaking changes.

## Changes

- Add `docs/json-output.md` with examples per command and a stability policy
- Introduce `internal/jsonout.Version` (currently `1`)
- Emit `schema_version` on object envelopes: `stats --json`, `doctor --json`, and fuzzy/stemmed/semantic `search --json`
- Keep exact `search --json` and `blame --json` as top-level JSON arrays for backward compatibility
- Add schema contract tests; update doctor JSON golden fixture

## Tests

```sh
go test ./... -race -count=1
go vet ./...
go build ./cmd/deja
```

All pass locally.

## Notes

Exact search and blame intentionally remain arrays (not object envelopes) so existing `jq` scripts and MCP consumers keep working. Only additive fields inside hit/session objects are permitted within schema version 1.

### DIFF
--- a/cmd/deja/doctor_report.go
+++ b/cmd/deja/doctor_report.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/vshulcz/deja-vu/internal/embed"
 	"github.com/vshulcz/deja-vu/internal/index"
+	"github.com/vshulcz/deja-vu/internal/jsonout"
 	"github.com/vshulcz/deja-vu/internal/model"
 	"github.com/vshulcz/deja-vu/internal/sources"
 )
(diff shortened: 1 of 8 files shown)


vshulcz (approved):
Thanks — this closes a real gap, and keeping exact search/blame as bare arrays for backward compat was the right call. Checked every documented field against the actual struct tags and pushed one commit on top: the recall block documented `recalls`/`injected_sessions` but the real envelope is `recalls_served`/`recall_sessions`/`bytes`/`injected_bytes`, the index path example now matches the real default (~/.cache/deja), and the blame example gains `count`/`score`. Also pinned `recalls_served` in the schema test so that field can't drift silently. Tests green with -race locally.

vshulcz:
Merged — thanks for the contribution! docs/json-output.md now ships as the contract for anyone building on --json.

Issue #337, opened 2026-07-24
search: CJK text collapses into single tokens — proposal: dictionary-free bigram indexing


AliceLJY:
## Problem

The index tokenizer keeps CJK characters (they pass the letter/digit/underscore rune test) but CJK text has no spaces between words — so an entire run of Chinese/Japanese/Korean collapses into **one giant token**. A message like `刷新令牌怎么实现` is indexed as a single term.

Consequences, observed on a real corpus (~6,100 sessions across two machines, heavily mixed Chinese/English):

- **Exact tier never fires for CJK.** A natural query like `装订` (2 chars) can only match via the substring tier — it works (nice ladder design!), but it walks the token catalog instead of postings, and ranking loses BM25/proximity quality.
- Multi-term CJK AND (`邮轮 反馈`) also survives only via `close` expansion.
- English/identifier queries on the same corpus are first-class: exact postings, BM25, `identifierParts` — the gap is purely the CJK side.

Retroactive indexing is the headline feature, and for CJK-speaking users a large share of their history is exactly this kind of text.

## Proposal: CJK bigram indexing (dictionary-free, zero-dependency)

The standard approach from Lucene's CJKAnalyzer family — no dictionary, no new dependency, language-agnostic (Han/Kana/Hangul all benefit):

**Index side** — in `indexKeys`, add a third emitter next to `tokens()` and `identifierParts()`:

- scan the text for runs of CJK runes (`unicode.Is(unicode.Han, r)` etc.);
- within each run, emit overlapping bigrams: `装订计数` → `装订`, `订计`, `计数`;
- a single-rune run emits its unigram (`茶` → `茶`);
- runs never cross a non-CJK boundary, so ASCII paths are untouched.

**Query side** — mirror the same expansion in query tokenization (`Tokens` / `RelevanceTerms`): a CJK query segment becomes its bigram set, and the existing AND-intersection semantics just work (`装订` = one exact bigram; longer phrases = several ANDed bigrams, which is the classic precision/recall sweet spot for bigram indexes).

**Everything downstream is free.** Bigrams are ordinary tokens: postings, buckets, BM25, proximity/title/freshness boosts, and the ladder all apply unchanged. `stem`/`fuzzy`/`identifierParts` simply don't match CJK runs — same as today, harmless.

## Impact

- **Index format**: needs a version bump → one automatic rebuild (~10 s on my corpus; consistent with "the index is a rebuildable cache").
[...]

PR #338, opened 2026-07-24
feat(search): CJK bigram indexing (#337)

vshulcz:
Implements #337 exactly as proposed — thanks @AliceLJY for a model spec. Bigrams are ordinary tokens, so postings, ranking and the ladder apply unchanged; the issue's test table is in verbatim (`TestCJKBigramSearch`). Index version bumps 10->11 for one automatic rebuild. English benchmarks are byte-identical before/after (LongMemEval 84.9/94.3, LoCoMo guard steady).

@AliceLJY review welcome — and if you can run your before/after on the bilingual corpus (hit-rate on CJK queries, index size delta), the numbers would go straight into the release notes with credit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- CJK bigram indexing (#337, design by @AliceLJY): Chinese/Japanese/Korean text gets first-class exact search and full ranking; index version bumps for one automatic rebuild. Note: a single-character query against a longer run (`茶` in `喝茶`) resolves via the close tier, not exact.
+
 ## [0.15.5] - 2026-07-24
 
 ### Added
--- a/internal/index/cjk.go
+++ b/internal/index/cjk.go
@@ -1,0 +1,76 @@
+package index
+
+import (
+	"unicode"
+)
+
+// CJK text carries no spaces, so the base tokenizer collapses whole phrases
+// into one giant token and the exact tier never fires for it. The classic
+// dictionary-free fix (Lucene's CJKAnalyzer family, proposed in #337) indexes
+// overlapping bigrams inside every CJK run: 装订计数 -> 装订, 订计, 计数.
+// A single-rune run keeps its unigram. Runs never cross a non-CJK boundary,
+// so ASCII and Cyrillic paths are untouched, and bigrams are ordinary tokens
+// downstream — postings, ranking and the ladder apply unchanged.
+
+func isCJK(r rune) bool {
+	return unicode.Is(unicode.Han, r) || unicode.Is(unicode.Hiragana, r) ||
+		unicode.Is(unicode.Katakana, r) || unicode.Is(unicode.Hangul, r)
+}
+
+// cjkBigrams emits the bigram set for every CJK run in s.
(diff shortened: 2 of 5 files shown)


AliceLJY:
Reviewed the diff and ran the before/after on my corpus — numbers below. That was fast!

## Review

Implementation matches the spec exactly, and all four wiring points are where they should be: `indexKeys` third emitter, `queryKeys` mirror, `RelevanceTerms` expansion (glad the NL fallback tier got it too — that was the one I worried would be missed), version 10→11. Asserting cross-boundary bigrams at the **postings level** (`postingsFor("tc装")`) instead of behavior level is stronger than what I sketched — nice. Built the branch locally: `TestCJKBigramSearch` / `TestCJKBigramsUnit` pass, `internal/index` + `internal/search` suites green.

Two small notes, neither blocking:

1. **Fuzzy tier × bigrams** — do fuzzy variants ever get generated for a 2-rune CJK token, and if so is the variant space bounded? If the fuzzy generator is ASCII-only (I suspect it is), a one-line comment saying so would close the question.
2. **Docs** — worth one line in the changelog: a single-char query against a longer run (`茶` vs `喝茶`) resolves via the close tier, not exact. The test comment says it; users will wonder.

## Before/after on a real bilingual corpus

Corpus: **2,943 sessions / ~56k messages** (claude + codex + kimi, heavily mixed Chinese/English coding transcripts), **2.88 GB** source. M-series MacBook.

| | v10 (0.15.4) | v11 (PR #338) |
|---|---|---|
| Full index build | — | **15.5 s** |
| Index size | 161 MB (5.5% of source) | 195 MB (**6.6%** of source) |
| 10 real Chinese queries, target found | 10/10 (mixed tiers, see below) | 10/10 (**all exact**) |

Size: **+21% relative, +1.1pp of source** on a CJK-heavy corpus — this should be close to the upper bound; mostly-English corpora won't move (your 2.4% stays put).

Hit-rate needs a nuance that might be worth a line in the release notes: **v10 Chinese wasn't uniformly dead** — real transcripts are full of fullwidth punctuation, which splits long runs into "sentence-segment tokens". So a short query sometimes exact-matched by luck (`装订` sat between a space and a comma), and otherwise fell to the close tier with weak ranking (`邮轮` buried mid-sentence). What the bigrams change is the *guarantee*: every query is now a first-class exact match regardless of where punctuation happened to fall.

The depth difference shows it clearly — matches counted per query, same corpus:

| query | v10 matches | v11 matches |
|---|---|---|
| `装订` | 14 | 24 |
| `脱敏` | 25 | 44 |
| `记忆 检索` (multi-term AND) | 3 | **93** |

Multi-term Chinese AND goes from "whatever punctuation left us" to full-text reach.

(Methodology note: today's session on this machine discusses these exact terms, so freshness puts it at rank 1 for several queries — hit judgment above counts the true historical target appearing in results, which it did in all cases, both versions.)

Happy to re-run anything on a later revision. Ship it 🚀

vshulcz:
Thank you for a review this thorough — running the corpus numbers same-day is above and beyond, and they're going into the release notes with credit. Both notes addressed in the latest commit: (1) confirmed the fuzzy generator's 4-rune floor keeps bigrams out of variant generation entirely, now stated in a comment right at the gate; (2) the single-char/close-tier nuance is in the changelog. Merging on green.

PR #357, opened 2026-07-25
perf(index): deflate the tail of large records

vshulcz:
Tool output and file dumps are about **1% of messages but 22.6% of the text** on a real corpus, and they are the most compressible part of it. Records whose payload exceeds 8 KB are now deflated; smaller ones are stored as they were.

Measured on a 900-session corpus with a realistic dump tail:

| | before | after |
|---|---|---|
| records.bin | 33.7 MB | **12.1 MB** |
| whole index | 44.6 MB | **22.3 MB** |
| index build | 1185 ms | 1236 ms |
| search (materializes many records) | 77 ms | 76 ms |

**The floor is high on purpose, and that is the whole finding.** Compressing every message instead (a 256-byte floor) cut the record log by 24% but **doubled index build time and tripled** the latency of a query that materializes many records — the payload is touched once per message on write and once per message shown. Restricting it to the tail keeps essentially all of the size win at no measurable cost, because dumps are simultaneously the bulk of the bytes, the most compressible, and the rarest to be displayed.

The layout is `[keyID][pathID][flag][payload]`. The ids stay outside the compressed part, so the record walk still peeks a key and skips without inflating anything — the hot path never touches a compressed byte.

Round-trip is covered for all four shapes: short, highly compressible, incompressible, and empty; plus an end-to-end test that a compressed dump is still searchable and comes back in full. Both fail if inflation is disabled. Index version 13 to 14.

Full `-race` suite green, lint clean. Stacked on #356.

### DIFF
--- a/internal/index/compress_test.go
+++ b/internal/index/compress_test.go
@@ -1,0 +1,98 @@
+package index
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	search "github.com/vshulcz/deja-vu/internal/query"
+)
+
+// Large payloads are deflated; everything else is stored as it is. Whatever
+// the encoding, the text must come back byte-identical and stay findable.
+func TestCompressedRecordsRoundTrip(t *testing.T) {
+	dir := t.TempDir()
+	p := filepath.Join(dir, "records.bin")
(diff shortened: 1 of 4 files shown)


vshulcz:

PR #358, opened 2026-07-25
perf(index): drop the derivable offset from bucket directories

vshulcz:
A bucket directory stored, per token, an 8-byte absolute offset and a 4-byte length. Posting blocks are written back to back in directory order, so the offset is just the running sum of the lengths before it — 8 bytes per token to record a number the reader can add up itself. The length is now a uvarint too.

On a 216 010-token corpus (a real store has ~180k):

| | before | after |
|---|---|---|
| buckets | 7.27 MB | **4.89 MB** |
| query falling through the ladder | 118 ms | 113 ms |
| query at the exact tier | 85 ms | 84 ms |

Latency moves slightly the right way — there is less directory to read on every catalog build.

The saving is per token, so it is invisible on a corpus with a small vocabulary and largest exactly where it matters: a real store with many identifiers, paths and hashes.

Two test updates this forced, both real rather than cosmetic: the old "truncated fixed block" fixture is valid under the new layout, so it now truncates before the block length; and a length that would overflow uint32 is rejected as corruption instead of being silently wrapped. `TestBucketOffsetsAreDerivedCorrectly` covers multi-token round trip including an empty block, and fails if the running sum is not accumulated.

Index version 14 to 15. Full `-race` suite green, lint clean. Stacked on #357.

### DIFF
--- a/internal/index/compress_test.go
+++ b/internal/index/compress_test.go
@@ -96,3 +96,45 @@ func TestCompressedDumpsStaySearchable(t *testing.T) {
 		t.Fatalf("compressed dump came back truncated: %d bytes, want >= %d", len(full), len(dump))
 	}
 }
+
+// A bucket directory no longer stores each token's offset; it is the running
+// sum of the block lengths. Every token must still land on its own postings.
+func TestBucketOffsetsAreDerivedCorrectly(t *testing.T) {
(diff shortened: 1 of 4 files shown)


vshulcz:

PR #359, opened 2026-07-25
perf(search): bucket the token catalog by length for fuzzy matching

vshulcz:
Fuzzy matching ran Damerau-Levenshtein against **every token in the corpus** — 267 268 of them on the measured store — once per query term. It was 266 ms of a 385 ms query, the single largest piece of search latency.

An edit changes a string's length by at most one and a transposition not at all, so the length difference is already a lower bound on the distance: only tokens within the edit limit of the query's length can possibly match. The cached catalog now also holds its tokens bucketed by rune length, so matching walks a few short slices instead of the whole vocabulary.

Measured on a 900-session corpus with a realistic length spread (short words, identifiers, paths):

| | before | after |
|---|---|---|
| tokens scanned per term | 267 268 | **72 000** |
| `closeTokens` | 168 ms | **36 ms** |

Results are identical, which is the part worth checking rather than assuming. `TestFuzzyLengthBucketsMatchBruteForce` compares against a direct scan over a mixed Latin/Cyrillic catalog including tokens past the bucket cap, and fails on each way the bookkeeping can go wrong:

- narrowing the band by one → misses `a` for query `ab`
- skipping the overflow bucket → a 72-rune query finds nothing
- bucketing by byte length instead of runes → `уо` returns `[о у]` instead of eight Cyrillic matches

That last one is why the bucket key is the rune count: byte length would silently break every non-Latin corpus while looking fine in an ASCII test.

Full `-race` suite green, lint clean. Stacked on #358.

### DIFF
--- a/internal/index/coverage_extra_test.go
+++ b/internal/index/coverage_extra_test.go
@@ -77,7 +77,7 @@ func BenchmarkFuzzyTokenEnumeration(b *testing.B) {
 	}
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		_ = closeTokens("synthetic-token-12345", catalog)
+		_ = closeTokens("synthetic-token-12345", newTokenIndex(catalog))
 	}
 }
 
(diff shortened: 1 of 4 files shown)


vshulcz:

PR #368, opened 2026-07-26
fix(search): treat Traditional and Cantonese function words as grammar

ithiria894:
`cjkFunctionRunes` lists Simplified Mandarin forms only, so the same closed class written in Traditional or Cantonese reaches the relevance tier as content. The guarantee `TestCJKFunctionBigramsAreNotQueryTerms` gives Simplified input does not hold for the other two ways of writing the same language.

Reproduced on `main` — both new tests fail before the change and pass after:

```
RelevanceTerms("復旦大學在哪個城市？")
  before → [復旦 旦大 大學 學在 哪個 個城 城市]      # 哪個 is grammar
  after  → [復旦 旦大 大學 學在 個城 城市]

RelevanceTerms("我點解唔用 postgres 嘅索引")
  before → [我點 點解 解唔 唔用 postgres 嘅索 索引]  # 5 of 7 terms are grammar
  after  → [點解 解唔 唔用 postgres 嘅索 索引]
```

The Cantonese case matters because 點解 ("why") and 我哋 ("we") appear in nearly every session written by a Cantonese speaker, so the tier ranks on the question's wording rather than on the entity being asked about.

### What the patch adds

- Traditional forms of entries already in the map: 與 還 們 這 麼 嗎 讓 給 於 從 對 則 並 為 個 會
- The Cantonese closed class: 嘅(的) 係(是) 喺(在) 唔(不) 咩/乜(什麼) 點(怎) 哋(們) 嗰(那) 啲(些) 樣 冇(沒) 嚟(來), plus the sentence-final particles 咪 囉 喎 啦 咁

The both-runes rule is untouched, so words whose characters are function runes individually still survive — covered by tests: 關係, 重點, 個人, 樣本, 冇錢, 中國, 目的, 這個關係. Query-time only, so the index is unchanged and no reindex is needed.

### Scope, deliberately narrow

The tests assert **parity with the Simplified baseline**, not a stricter standard. Boundary bigrams pairing a particle with a content character survive in both scripts (`的索` in Simplified, `嘅索` in Cantonese) — inherent to bigram tokenization. Two things I did not do:

- `解` is not added, even though 點解 is the Cantonese "why": adding it would filter 了解, a real Simplified word, under the both-runes rule.
- Japanese and Korean function words are still unlisted. `isCJK` accepts Hiragana/Katakana/Hangul but the map is Chinese-only — out of scope here.

### Verification

- `go test ./internal/index/ -run CJK` — 10/10 pass including the two new tests
- `go test ./...` — full suite green
- `gofmt -l` and `go vet` clean
- Built and run against a real 3,300-session / 106k-message mixed-script corpus: grammar noise drops (a Cantonese question's match count 549 → 532, 202 → 132) and on-topic sessions stay on top. To be straight about it: this is a noise reduction, not a dramatic ranking change.

I have a second, independent change for Simplified↔Traditional folding — separate PR, since it needs a reindex and this one does not.

### DIFF
--- a/internal/index/cjk.go
+++ b/internal/index/cjk.go
@@ -108,7 +108,13 @@ func expandCJKTokens(toks []string) []string {
 // bigram that merely contains one would delete real words. This filter runs
 // at query time only — the index keeps every bigram, so nothing about
 // ingestion changes.
+//
+// The list covers three scripts of the same closed class. Simplified Mandarin
+// alone is not enough: a Traditional writer's 哪個 and a Cantonese writer's
+// 我點 / 唔用 / 嘅索 are the same grammar, and without them a question ranks on
+// its own wording instead of on the entity it asks about.
 var cjkFunctionRunes = map[rune]bool{
+	// Simplified Mandarin
 	'的': true, '了': true, '是': true, '在': true, '和': true, '与': true,
 	'或': true, '而': true, '但': true, '就': true, '都': true, '也': true,
 	'还': true, '又': true, '再': true, '只': true, '才': true, '很': true,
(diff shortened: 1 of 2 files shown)


vshulcz (approved):
Thanks — this is the right fix, and the framing is right too: the "every rune is a function rune" guarantee was only ever true for Simplified input, which made it a guarantee for one third of the users who write this language.

I checked the risk that matters here, since widening a closed class can start eating content words: 點 樣 對 會 個 從 all carry meaning in Mandarin compounds. Ran 地點, 重點, 樣本, 機會, 會議, 對象, 從前, 點擊 through `RelevanceTerms` — all survive, because the all-runes rule holds them up. Full `internal/index` and `internal/search` suites pass, and the recall floor is unchanged (recall@5 1.0, median 0.216 ms).

Merging.

PR #369, opened 2026-07-26
feat(search): fold Traditional to Simplified so CJK recall is bidirectional

ithiria894:
CJK bigrams are sequences of raw codepoints, so 距離 and 距离 are unrelated index keys and a query written in one script cannot reach content written in the other. For anyone whose corpus mixes the two — the normal case for a Traditional writer, whose own text is Traditional while pasted docs, library READMEs and model output are frequently Simplified — recall is silently one-directional. Nothing signals it: the query just returns less, or falls through to the relevance tier and returns unrelated sessions.

Measured on a real 193-session / 72k-message Traditional corpus, before:

| query | Traditional | Simplified |
|---|---|---|
| 簽證 / 签证 | 15 hits | **2 hits** |
| 升學計劃 / 升学计划 | 11 hits | **1 hit** |
| 刺蝟的距離 / 刺猬的距离 | 15 hits | **1 hit** |

After, both columns are identical for every pair.

### How

`internal/cjkfold` holds a generated Traditional→Simplified table: 6,460 pairs from Unihan `kSimplifiedVariant`, keeping only entries with a single target that differs from the source, so the fold is a pure function. It ships as two parallel rune strings (~45 KB of source, no new dependency) built into a map lazily, so a run that never touches CJK never pays for it. `tools/gen_cjk_fold.py` regenerates it.

The fold has to be applied in three places, and all three must agree or a record the postings legitimately found gets dropped further down:

1. `indexKeys` and `queryKeys` — both sides meet on one key
2. `recordMatchesQueryVariants` — re-verifies against surface text
3. `search.Run` and `RelevanceHits` — count and rank on surface text

(3) is where I first got it wrong: postings matched, then the counting pass scored the record 0 and it vanished from the results. The retry only runs when the raw match scores zero and the query contains CJK, so non-CJK queries are untouched.

**Bigrams themselves stay unfolded.** The query-time function-word filter reads those runes, and 係 folds to 系 — a content character in 系統 and 關係 — so a filter applied after folding could not tell grammar from content.

**Direction matters.** Traditional→Simplified is effectively many-to-one and therefore deterministic; the reverse is ambiguous (Simplified 干 covers Traditional 乾, 幹, 干). Folding only ever runs one way.

### Compatibility

Stored text is untouched, so snippets still show exactly what the user wrote. Posting keys change, so **this needs a reindex** — likely an index format version bump on your side, or a release note. Happy to add whichever you prefer; I left it out rather than guess.

### Verification

- New tests: cross-script search both directions, Katakana/Hangul/Latin untouched, folding an already-Simplified string is a no-op, and index keys agree across scripts while bigrams deliberately do not
- `go test ./...` — full suite green; `gofmt`/`go vet` clean
- Real corpus after the change: CJK queries 66–103 ms, English unchanged, index 165 MB

Independent of #368 (Traditional/Cantonese function words). If both land, the Traditional half of that list becomes redundant — folding normalizes those runes before the filter sees them — but stays harmless; the Cantonese half is still needed, since 嘅 係 喺 唔 咩 乜 哋 嗰 啲 have no Simplified counterparts.


vshulcz (approved):
This is a strong contribution — the measurement on a real mixed corpus, the choice to fold one direction only because Simplified→Traditional is ambiguous, and deliberately leaving bigrams unfolded so the function-word filter still sees 係 rather than 系. That last one is the trap, and you avoided it.

Verified here: full suite green, recall floor unchanged (recall@5 1.0), English latency 0.212–0.218 ms before vs 0.216–0.232 after, binary +66 KB for the table.

One gap, which you correctly flagged and left to us: posting keys change, so an index built before this holds keys the new build never queries — silent zero CJK recall until a manual reindex. We gate that on the index version, so I pushed a commit bumping it to 16 with a test to keep the two tied together. Upgrades now rebuild on their own.

0.16.0 needs a reindex regardless, so the timing works out.

PR #370, opened 2026-07-26
fix(hooks): let auto-recall fire for CJK prompts

ithiria894:
`promptSearchTerms` drops every CJK prompt on the floor, so `UserPromptSubmit` auto-recall can never fire for a Chinese, Japanese or Korean user:

```
promptSearchTerms("韓國簽證點樣申請") → []
promptSearchTerms("刷新令牌怎么实现") → []
promptSearchTerms("パスワードの再設定") → []
```

Two things combine. `FieldsFunc` treats every rune above U+0400 as wordy, so a CJK phrase — which carries no spaces — arrives as a single field. Then `techTerm` returns false on the first rune above 127. `runHookPrompt` bails at its "fewer than two terms" guard and stays silent, permanently. The feature is effectively ASCII-only, which is easy to miss because the search path got bigram handling in #337 and this path did not.

### Fix

When the prompt contains CJK, fall back to `index.RelevanceTerms` — the same per-run bigrams the relevance tier already ranks on, with pure-grammar pairs (`什么`, `在哪`) dropped by the existing `cjkFunctionBigram` filter. A bigram is about as specific as the identifiers `techTerm` looks for, and the caller still demands `matched >= 2` before claiming a déjà vu, so the "one lucky rare word is not a déjà vu" contract is unchanged.

ASCII prompts take the original path untouched — the CJK branch only runs when the prompt actually contains Han, Hiragana, Katakana or Hangul, and a test pins that.

### Verification

- three new tests: CJK prompts now yield terms, grammar bigrams stay out, ASCII behaviour is unchanged
- `go test ./...` — full suite green; `gofmt`/`go vet` clean
- End to end on a real corpus, same prompt, before and after:

```
before → (no output at all)
after  → injects 844 chars, and the user sees
         deja-vu: you have been here — "我想follow up一下韓國嗰個visa嗰度囉…" (Jul 14 · via: 韓國, 國簽, 簽證)
```

That is the session where the visa follow-up actually happened.

### One thing I deliberately did not touch

Cyrillic has the same shape of problem: `techTerm`'s `r > 127` rejects it too, so a Russian prompt with no ASCII identifier also yields no terms. That is your language and your judgement about what counts as a "tech term" in it, so I left it alone rather than guess. Happy to extend the fix if you want it.

Independent of #368 and #369 — this one touches only `cmd/deja/hook_prompt.go`. With #368 also applied, Cantonese grammar bigrams (`我點`) drop out of the term list too.

### DIFF
--- a/cmd/deja/hook_prompt.go
+++ b/cmd/deja/hook_prompt.go
@@ -173,7 +173,48 @@ func promptSearchTerms(prompt string) []string {
 			}
 		}
 	}
+	// Cyrillic hits the same wall for the same reason: techTerm rejects every
+	// rune above 127, so a Russian prompt without an ASCII identifier yields
+	// nothing and auto-recall stays silent. Unlike CJK it is space-separated,
+	// so the words are already there — they just need a length floor and the
+	// closed class removed, and the index matches their inflected forms.
+	for _, f := range fields {
+		if seen[f] || !cyrPromptTerm(f) {
+			continue
+		}
+		if add(f) {
+			break
+		}
+	}
 	return out
+}
+
+// cyrPromptTerm reports whether a field is a Cyrillic word specific enough to
+// search on. Short words carry no signal and the closed class is noise, so
+// both are dropped; what is left is roughly what techTerm keeps for ASCII.
+func cyrPromptTerm(f string) bool {
+	n := 0
+	for _, r := range f {
+		if r < 0x400 || r > 0x4ff {
+			return false
+		}
+		n++
+	}
+	return n >= 5 && !cyrPromptStop[f]
+}
+
+// The closed class plus the handful of verbs that open half of all questions.
+var cyrPromptStop = map[string]bool{
+	"который": true, "которая": true, "которые": true, "потому": true,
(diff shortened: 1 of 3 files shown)


vshulcz (approved):
Good catch, and the diagnosis is exact: `FieldsFunc` hands back the whole phrase because every rune above U+0400 counts as wordy, then `techTerm` rejects it on the first rune above 127. The feature was ASCII-only and nothing said so. Falling back to the bigrams the relevance tier already ranks on is the right move — same specificity, and the `matched >= 2` contract upstream is untouched.

Verified here: full suite green after rebasing onto main with #368 and #369, ASCII term extraction byte-identical, and CJK prompts now yield terms with grammar bigrams filtered out.

On the Cyrillic gap you flagged and left alone — you were right that it is the same hole, so I pushed a commit closing it. Russian is space-separated, so the words are already there; they only needed a length floor and the closed class removed, and the index already matches inflected forms. `почему падает индексация кириллицы` now yields terms instead of nothing, and a mixed prompt keeps both the identifier and the prose.

Thanks for all three of these — the CJK path had been quietly half-working since #337 and none of it was visible from the English side.

PR #376, opened 2026-07-27
feat(sources): index Hermes Agent sessions

### LINKED ISSUE #355: Feature: Hermes Agent session support (new source)
## Feature Description

Add Hermes Agent as a new session source for indexing, similar to existing claude, opencode, cursor, codex, and qwen sources.

## Background

Hermes Agent (NousResearch/hermes-agent) stores conversation history in a SQLite database with FTS5 full-text search at `~/.hermes/profiles/<profile>/state.db`. The `messages` table schema:

```sql
CREATE TABLE messages (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    session_id TEXT NOT NULL,
    role TEXT NOT NULL,
    content TEXT,
    tool_call_id TEXT,
    tool_calls TEXT,
    tool_name TEXT,
    timestamp REAL NOT NULL,
    token_count INTEGER,
    finish_reason TEXT,
    -- ... other fields
);
```

## Proposed Implementation

A new source file `internal/sources/hermes.go` that:

1. **Discovers profiles:** scans `~/.hermes/profiles/*/state.db` for all profiles
2. **Reads sessions:** queries the `messages` table, grouped by `session_id`
3. **Exports to Deja-Vu model:** maps `role`/`content`/`timestamp` to `model.Session`

### Environment variable control

```go
func HermesDB() string {
    if p := os.Getenv("DEJA_HERMES_DB"); p != "" {
        return p
    }
    return filepath.Join(Home(), ".hermes", "profiles", "architect", "state.db")
}
```

Allowing `DEJA_HERMES_PROFILES_ROOT` to point to a custom profiles directory.

### Data volume

On our system: ~56 sessions, ~4000 messages from a single profile (architect) — manageable size.

## Integration with Hermes Agent

Hermes Agent already has a built-in `session_search` tool (FTS5-backed), but Deja-Vu would provide cross-agent memory — a single query across Hermes + OpenCode + Claude + Cursor sessions.

## Technical notes

- The DB uses `sqlite3` CLI (same as opencode.go pattern)
- Message ordering by `timestamp` (REAL type, Unix epoch)
- `session_id` from messages table maps to model.Session ID
- Content may include tool calls, reasoning, and markdown — handle as plain text

## References

- Hermes Agent: https://github.com/NousResearch/hermes-agent
- Existing sources in Deja-Vu: claude.go, opencode.go, cursor.go, codex.go, qwen.go
- Schema dump available on request

vshulcz:
Hermes keeps one SQLite store per profile at `~/.hermes/profiles/<profile>/state.db`: a flat `messages` table grouped by `session_id`, with `timestamp` as REAL epoch seconds. All profiles are picked up, not just the default one, and `DEJA_HERMES_PROFILES_ROOT` / `DEJA_HERMES_DB` override the location.

Two details the schema forces:

- `role = 'tool'` rows and rows with a null `content` carry no prose and are skipped.
- There is no working directory anywhere in the schema, so the profile name stands in for the project.

The fractional timestamp is what the first test caught: the shared time parser handles integer epochs and RFC 3339, and `1785015600.25` fell through to the zero time, which sorts as ancient and never surfaces in recall.

Indexed, searched and listed end to end against a real store — `deja doctor`, `deja sources`, and a query that returns the Hermes session alongside Claude and Codex hits. Registry entry, fixture, format doc and the published matrix are in the same change.

Closes #355.


Jelloeater:
Awesome 😎

PR #394, opened 2026-07-27
Ship the MCP server in the plugin bundle

### LINKED ISSUE #393: Plugin bundle installs into six harnesses but carries no MCP server
The plugin bundle in this repository installs into six harnesses now — Claude Code, Codex, Cursor, Qwen Code, OpenClaw and Copilot CLI all read the Claude plugin format, and each was installed from it against the real client. What it carries is not the same everywhere:

| | hooks | skill | `/deja` | MCP |
| --- | :-: | :-: | :-: | :-: |
| Claude Code | yes | yes | yes | no |
| Codex | — | yes | — | yes (own bundle) |
| Cursor | untested | yes | yes | no |
| Qwen Code | no | yes | yes | no |
| OpenClaw | no | yes | yes | no |
| Copilot CLI | run, but cannot inject | yes | yes | no |

The gap is MCP. The Claude plugin manifest supports an `mcpServers` field and so do the harnesses that read this format, but our bundle does not declare one — so installing from a marketplace wires the hooks and the command while the recall tools, which are the part that works everywhere, still need a separate `deja install`.

Adding it makes one marketplace install the whole integration on six harnesses.

Worth recording alongside: Copilot CLI does run plugin hooks — a `SessionStart` hook fired and received a full payload — but `additionalContext` never reaches the model, so recall there stays MCP plus the skill.

vshulcz:
Closes #393.

The bundle installs into six harnesses now — Copilot CLI and OpenClaw both read the Claude plugin format, same as Cursor and Qwen — but it carried only hooks and the slash command, so a marketplace install still needed a separate `deja install` to get the recall tools.

Adds an `mcpServers` entry pointing at a small launcher in the bundle. It resolves the binary the way the hook bridge does but does not stand down when deja is installed locally: clients key MCP servers by name, so the two collapse into one.

Verified live: `✔ Connected` in Claude, a recall call answering in Copilot, and OpenClaw listing the server as coming from the bundle. Also worth recording — Copilot does run plugin hooks, but their `additionalContext` never reaches the model, so recall there is MCP plus the skill.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ claude plugin marketplace add vshulcz/deja-vu
 claude plugin install deja-vu@deja-vu
 ```
 
-The plugin wires the same three hooks and the `/deja` command. It stands down on its own if `deja install` already wired them, so having both does not recall twice.
+The plugin wires the same three hooks, the `/deja` command and the MCP server. It stands down on its own if `deja install` already wired them, so having both does not recall twice.
 
 Codex, Cursor and Qwen read the same bundle from their own registries:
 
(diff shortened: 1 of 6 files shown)


vshulcz:
Rebased onto main and fixed one thing #391 shipped: `openclaw plugins install <URL>` does not work — that version takes a path, an npm spec or a ClawHub package, and a URL comes back as `unsupported npm spec: URLs are not allowed`. Docs now show the path form. Copilot is the sixth registry that reads the bundle.

PR #414, opened 2026-07-27
Roo: install target and CLI storage root

### LINKED ISSUE #413: Roo: CLI sessions unindexed, no install target
Roo was a store deja read, with no install target — and the store it reads is missing a root.

Verified against the real thing: extension 3.54.0 installed in VS Code, and `@roo-code/cli` 0.1.17 installed to a scratch prefix and run headless.

- **The CLI's history is invisible.** `roo` runs the same extension against a VS Code shim, so its tasks land in `~/.vscode-mock/global-storage/tasks/<uuid>/`, not under any editor's globalStorage. I ran a task and confirmed the directory and the file layout — `api_conversation_history.json`, `history_item.json`, `ui_messages.json`, exactly what `ParseRooTask` already handles.
- **No hooks.** Every `hook` match in the bundle is a React one; there is no lifecycle event and nothing shells out.
- **MCP** lives in `<storage>/settings/mcp_settings.json`, per editor host — and the CLI keeps its own copy under the shim root.
- **Global rules** (`~/.roo/rules/`) are read verbatim into the system prompt for every mode, which is the only always-on channel.

Still open, filed so it is not lost: with the CLI root added, `deja doctor` sees the store and the file, and `LoadRoo` parses it into a session — but the harness never appears in `deja stats`, so something between the file kind and ingest is dropping it. Not diagnosed yet.

Also unhandled: the `roo-cline.customStoragePath` VS Code setting relocates the whole store, and deja does not read it.

vshulcz:
Closes #413.

Verified against Roo 3.54.0 in VS Code and `@roo-code/cli` 0.1.17 run headless in a scratch prefix.

**Install.** Roo has no hooks — every `hook` in the bundle is a React one. Its channels are MCP settings per editor host and the global rules directory, read verbatim into the system prompt for every mode. `deja install roo` writes the MCP server into each host Roo has actually run in, skipping editors it hasn't, and the guidance goes to `~/.roo/rules/deja.md` rather than a mode-specific directory where it would appear in Code mode and vanish in Ask.

**CLI history.** `roo` runs the extension against a VS Code shim, so its tasks live in `~/.vscode-mock/global-storage/tasks/` — deja never looked there.

**And the reason none of it showed up:** Roo and Cline write the same filename in the same layout, and Cline's file kind matched *any* `api_conversation_history.json` regardless of root. Registered first, it claimed every Roo task, so Roo history has been filed under cline all along. Scoping that matcher to Cline's own roots fixes it — in my index the probe session moved from cline to roo, and cline's message count dropped by exactly that session's 5703.

`deja "roows2"` now returns `[roo] scratchpad/roows2`. Regression test pins the kind for a CLI task path.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -52,6 +52,12 @@ signs:
     args: ["sign-blob", "--yes", "--output-signature", "${signature}", "--output-certificate", "${certificate}", "${artifact}"]
     artifacts: checksum
 
+# The nightly publishes a rolling `nightly` tag. Left visible, it becomes the
+# most recent tag and every version template downstream fails on it.
+git:
+  ignore_tags:
+    - nightly
+
 snapshot:
   version_template: "{{ incpatch .Version }}-next"
 
--- a/cmd/deja/guidance.go
+++ b/cmd/deja/guidance.go
@@ -47,6 +47,10 @@ func guidancePath(harness string) string {
 		return filepath.Join(sources.GrokHome(), "GROK.md")
 	case "opencode":
 		return filepath.Join(opencodeConfigHome(), "opencode", "AGENTS.md")
+	case "roo":
+		// Global rules are read for every mode and every task, which is the
+		// only always-on channel Roo has.
+		return rooRulesPath()
 	default:
 		return ""
 	}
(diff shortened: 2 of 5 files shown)


vshulcz:
Landing this as-is is no longer possible: both commits are already on main — I pushed them there directly by mistake, which is the process failure, not the code. Force-pushing main back is blocked at my end, so rewriting history is not something I can do here.

Closing rather than leaving a PR that would be an empty diff. The change is on main and green: CI, Lint, CodeQL and Scorecard all passed on it, and the regression test for the cline/roo file-kind collision went with it.

PR #433, opened 2026-07-27
doctor: find the MCP server whatever it was named

### LINKED ISSUE #363: doctor: MCP wiring check misses servers registered under a name other than "deja"
Setup: Windows 11, deja built from main at 98b43e1 (v0.15.7). The MCP server has been wired into Claude Code since 0.12.x, registered manually in `~/.claude.json` under the key `deja-vu`.

`deja doctor` says:

```
MCP wiring:
  claude-code  not wired      guidance missing     C:\Users\<user>\.claude.json
```

The server is running and answers recall/remember calls daily, so this is a false negative. Cause: `doctorJSONWired` in cmd/deja/doctor.go checks for a server literally named "deja" under the config key, so any other name reads as not wired. `doctorTOMLWired` and `doctorOpenClawWired` share the assumption.

For me this is cosmetic, I can rename the key. But doctor is the first thing to run when something looks off, and telling a user their working setup is not wired points the debugging in the wrong direction. One option: treat an entry as wired when its command points at a deja binary, whatever the key is called. Happy to test whatever direction you pick on this machine.

vshulcz:
Closes #363.

Reported from a real setup: the server registered by hand as `deja-vu`, answering recall calls daily since 0.12.x, and `deja doctor` calling it `not wired`. doctor is the first thing someone runs when memory goes quiet, so a false negative there sends the debugging in the wrong direction.

What identifies the server is the command it runs, not the key it was filed under. The JSON and TOML probes now accept any entry whose command is a deja binary, in the shapes clients actually use: bare command, command plus args, a nested `transport` object, and the Windows `cmd /c deja mcp` shim. Paths are compared with the separator normalised, since these configs are read on the platform they were not written on.

An unrelated server still reads as not wired — there is a test for that too, or the check would tell everyone they are fine.

### DIFF
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"path"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


Sora-bluesky:
Confirmed on the machine that hit this: built from main at 111df2e and `deja doctor` now reports claude-code wired for the `deja-vu` key behind the `cmd /c` shim. Thanks for the quick turnaround.

PR #476, opened 2026-07-28
Add versioned session read contracts

adamsitar:
## What changed

- adds an explicit bounded search limit
- adds schema-v1 JSON for recent session metadata
- adds exact composite session reads with bounded message windows and offsets
- adds source origin and optional instance provenance without fabricating imported peer identity
- documents and tests the additive machine contract

## Why

Downstream tools need stable machine-readable recent and exact-read operations. Search alone cannot safely recover one exact session, and path-only provenance is insufficient to distinguish local from imported material.

## Compatibility

The additions are opt-in and preserve existing human output and JSON search shapes. Exact reads require both session ID and harness to avoid ambiguous cross-harness identity.

## Validation

- go test ./... -race -count=1
- go vet ./...
- Linux build
- Windows amd64 cross-build
- final go test ./... and coverage-floor checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,14 +136,14 @@ $ deja "jwt refresh token"
 
 | Command | What it does |
 | --- | --- |
-| `deja <query>` | Search all histories. Multi-word = AND, common English filler words are ignored, substrings match (`code` finds `opencode`), and double-quoted phrases require contiguous text; zero-result queries try word forms before close spellings. `--re`, `--harness`, `--project`, `--since 30d`, `--role`, `--json`. |
+| `deja <query>` | Search all histories. Multi-word = AND, common English filler words are ignored, substrings match (`code` finds `opencode`), and double-quoted phrases require contiguous text; zero-result queries try word forms before close spellings. `--re`, `--harness`, `--project`, `--since 30d`, `--role`, `--limit`, `--json`. |
 | `deja ctx <query>` | Compact markdown digest of the best match — pipe it into a prompt. |
 | `deja blame <path>` | Find sessions that discussed a file, newest and most specific first. `--json`, `--all`, and the usual filters are supported. |
 | `deja share <id>` | Sanitized session digest for a colleague: secrets redacted, tool noise stripped. |
 | `deja stats` | Headline counts, totals, per-harness split, top projects, monthly sparkline. `--json` too. |
 | `deja doctor [--json]` | Self-diagnosis: store parse state, sqlite3 presence, MCP wiring per agent, index health, version; `--json` emits the same checks for scripts. |
 | `deja sync export <dir> [--full]` / `import <dir>` / `ssh <host> [--pull]` | Move memory between machines — via a shared folder or one ssh command. Watermarked, append-only, idempotent. |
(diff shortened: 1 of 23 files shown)


vshulcz (commented):
Thanks for this — it is a well-shaped contribution, and the parts that usually go wrong here went right. Requiring `--harness` alongside the id for `--json` reads is the correct call: session ids collide across harnesses and a machine reader has no way to notice. Bounding `--limit`, clamping `offset` against `total` and returning `returned` next to `total` means a caller can page without guessing. And withholding `instance` on imported sessions rather than filling it with the local name is exactly the restraint the field needs — a provenance value that is sometimes a lie is worse than an absent one.

Two things before it comes out of draft.

**`source` is emitted empty on the paths that do not set it.** `Source` has no `omitempty` and `SetSource` is only called in `search.Print` and the two machine helpers, so every other marshal of a `model.Session` now carries a provenance field that states nothing:

```
{"id":"x","harness":"claude",...,"source":{"origin":""}}
```

That is not hypothetical — `cmd/deja/mcp.go:271` does `json.Marshal(hits)` on the recall path, so this shape reaches every MCP client today. Since the whole point of the field is that a reader can trust it, `"origin": ""` is the one value it must never have. Either set the source where sessions are built rather than where they are printed, or mark the struct `omitempty` so an unset source is visibly absent instead of visibly blank. I would prefer the former.

**`DEJA_SOURCE_INSTANCE` is read inside the printers.** `os.Getenv` in `printRecentJSON`, `printSessionJSON` and `search.Print` puts configuration lookup three layers below where the command is set up, and makes the output of a pure-looking function depend on ambient state. Read it once at startup and pass it down.

Smaller notes, none blocking:

- `parseShow` returns six values, and every error path repeats `"", "", false, 0, 0`. A small options struct would make the next flag cheaper to add.
- `--limit` on search caps at 100 while `show --limit` caps at 200. Both are defensible, but the difference should be deliberate and documented in `docs/json-output.md`, which currently mentions neither ceiling.
- Worth a test that `--limit` is applied after `--harness`/`--project` scoping, not before. That exact ordering was a real bug here (#348), and this is the code path where it would come back.

The contract itself I would take as designed. Once `source` cannot be blank, mark it ready and I will run the full matrix.

vshulcz (approved):
Both points are properly closed, and the second one better than I asked for.

`Source` is a pointer with `omitempty` now, so an unset source is absent rather than blank — I re-ran the probe against your branch and the MCP marshal path is clean:

```
{"id":"x","harness":"claude","project":"","started":"...","updated":"..."}
```

The env lookup is read once and threaded through as `SourceInstance` instead of being called from inside the printers.

The ordering test is the part I want to call out. You did not just assert that `--limit` and scoping coexist — you built the case where the out-of-scope sessions score *higher* (`needle needle needle` against a single `needle`) so a limit applied before scoping would return the wrong session rather than merely a different count. That is the shape #348 actually had. Thanks for taking it seriously.

Approving. The Hermes root-store commit rode along in the same branch; it is small and correct, so I am not going to make you split it.

vshulcz:
Merged. This lands in 0.17.0 as the documented machine contract — thanks for the care, especially on the identity and paging edges.

Issue #492, opened 2026-07-28
CJK bigrams cost 17.5% of full-build time


vshulcz:
Measured by @AliceLJY on a 6,409-session / 304,420-message / 12.2 GB corpus, three versions built back-to-back in one run into isolated `DEJA_INDEX_DIR` paths ([full report](https://github.com/vshulcz/deja-vu/issues/492#issuecomment-5111334417)).

**The original +45% in this issue was wrong and has been retracted by its author.** It came from one pair of builds where the older version ran against a cold page cache and the newer against a warm one. Five interleaved rounds put the real figure at:

| span | build time |
|---|---|
| 0.15.5 → 0.15.6 — bigrams alone | **+17.5%** |
| 0.15.6 → 0.16.1 — everything after | −2.9% |
| 0.15.5 → 0.16.1 — full span | +14.1% |

Index size, byte-identical across every round, so no noise at all:

| | v0.15.5 | v0.15.6 (bigrams only) | v0.16.1 (shipped) |
|---|---|---|---|
| `buckets/` | 202.5 MB | 340.8 MB (**+68.3%**) | 305.1 MB |
| `records.bin` | 488.5 MB | 488.5 MB (**±0 bytes**) | 300.6 MB |
| `manifest.gob` | 1.52 MB | 1.52 MB (**±0 bytes**) | 2.53 MB |
| **total** | 698.1 MB | 836.5 MB (**+19.8%**) | 613.9 MB |

So the isolated cost of bigrams is postings and nothing else: `records.bin` and `manifest.gob` are unchanged to the byte. **The manifest growth this issue originally listed as somewhere to look is not the bigrams** — it happens entirely in #354/#357/#358/#360, after the bigram PR. That item is struck.

The +19.8% isolated index cost independently reproduces the "+21%" from #338 on a 2.2× corpus and a different machine. The −12% on the shipped release does not contradict it; it adds #360 on top.

Where to look, corrected:

- Bigrams roughly double the token count for a run of CJK runes, and unlike English there are no spaces to bound them. The cost is in generating and inserting those postings — split the phase timings before optimising, rather than guessing which stage owns it.
- ASCII is untouched by construction: the control queries return identical counts to the digit across all three versions, and a Latin/Cyrillic corpus measured −29% build time and −22% index size on the shipped release. Whatever this costs, only CJK content pays it.
- ~~The +66.8% manifest growth is worth a look on its own.~~ Not the bigrams; see above.

[...]

PR #493, opened 2026-07-29
docs(cli): list the search flags in --help

AliceLJY:
## Summary

`printUsage` documents flags for every subcommand — `deja show [--offset n] [--limit n]`, `deja last [--project name] [--role ...]`, `deja blame [--all] [--json] ...` — but the bare query form is printed as `deja [flags] <query>` and those flags are never listed anywhere.

`parseSearch` accepts nine: `--harness`, `--project`, `--since`, `--role`, `--limit`, `--all`, `--re`, `--json`, `--no-embed`. Three appear incidentally under `Examples:`; the other six are discoverable only by reading `cmd/deja/main.go`.

`--all` is the one that costs the most to miss. It is the only way to lift the default 15-session cap (`internal/search/search.go:229`), which makes it the only way to measure recall on a corpus instead of measuring the cap. While benchmarking CJK retrieval for #492 I needed exactly that, did not find it in `--help`, and concluded from the help text that no such control existed — I said so in that thread before reading the source, then had to correct myself. `--limit` compounds it: it appears in `--help` only on the `deja show` line, so it reads as a `show`-only flag even though the main query has accepted it since v0.16.1.

This adds the list. No behaviour change.

## Test plan

```
gofmt -s -l cmd/deja/          # clean
go vet ./cmd/deja/             # clean
go build ./cmd/deja            # ok
go test ./... -race -count=1   # 12 packages, no failures
```

New test `TestUsageDocumentsSearchFlags` asserts, for each documented flag, that it appears in the usage text **and** that `parseSearch` accepts it. The list therefore cannot drift from the parser in either direction: dropping a flag from the parser, or adding one without documenting it, fails the test.

## Notes

- Added one example for `--all`, since "no cap" is the behaviour least guessable from the flag name.
- Flag descriptions were read off the parser and the consumers (`search.go:229` for `--all`, `main.go:404` for `--no-embed`, `search.go:126` for `--re`) rather than inferred from the names.
- Happy to drop the example line or reword any description to match your voice.

### DIFF
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -962,11 +962,23 @@ Usage:
   deja install <claude-code|codex|opencode|cursor|gemini|antigravity|grok|qwen|kimi|cline|statusline|--all|--auto>
   deja uninstall <claude-code|codex|opencode|cursor|gemini|antigravity|grok|qwen|kimi|statusline|--all|--auto>
 
+Search flags (the bare "deja [flags] <query>" form above):
+  --harness <name>              only sessions from one harness (claude, codex, ...)
+  --project <name>              only sessions from one project
+  --since <duration>            only sessions newer than e.g. 30d, 12h
+  --role <user|assistant|tool>  only match turns from one role
+  --limit <1-100>               max sessions to return (default 15)
+  --all                         return every match, no cap
+  --re                          treat the query as a regular expression
+  --json                        machine-readable output
+  --no-embed                    skip the semantic (embedding) tier
+
 Examples:
   deja "jwt refresh token bug"
   deja '"connection pool exhausted"'
   deja "exhaustd"  # zero exact results try close spellings
   deja --harness claude --since 30d "panic in indexer"
+  deja --all "connection pool"  # every match, not just the first 15
   deja last 20 --harness codex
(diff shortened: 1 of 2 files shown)


vshulcz (approved):
Taking it. Three things I want to name, because they are why this is a good patch rather than a tidy one.

You did not just list the flags — you checked each description against the code that consumes it rather than guessing from the name. `--all` really is the only way past the cap at `internal/search/search.go:230`, and `--limit` really has been accepted by the bare query form since 0.16.1 while appearing in `--help` only on the `deja show` line, which is exactly why it reads as a `show` flag. I verified both.

The test is the part I would have missed. Asserting that every documented flag is *also* accepted by `parseSearch` means the list cannot drift in either direction: drop a flag from the parser and it fails, add one without documenting it and it fails. Documentation that is checked is documentation that survives.

And the cost you describe is real and specific. You needed `--all` to measure recall depth rather than measure the cap, could not find it, and said so in #492 before reading the source. That is a documentation bug with a measurable price — someone benchmarking us drew a wrong conclusion from our own help text. That framing is worth more than the diff.

Keeping the `--all` example; "no cap" is the behaviour least guessable from the name, as you said.

Issue #547, opened 2026-07-29
Index the work, not just the conversation


vshulcz:
Measured while validating the idea backlog, and it reframes most of it.

**deja indexes prose and throws the work away.** The index holds three roles — `user`, `assistant`, `developer` — and nothing else. Records that look like a command run: **280 of 70,982 (0.4%)**, and those only because someone quoted a command in a sentence.

Meanwhile, in the raw Claude transcripts on this machine (478 files, 644 MB):

| | |
|---|---|
| `tool_use` blocks | **40,502** |
| `tool_result` blocks | **40,487** |
| Bash calls | 26,956 |
| Read / Edit / Write | 5,151 / 4,259 / 1,233 |
| tool_result payload | 110 MB of 626 MB scanned (**18%**) |
| tool_result size | median 315 B · p90 3.2 KB · max 1 MB |

Every parser we have drops all of it. So does every competitor: the closest one mines transcripts in a mode literally called `convos`.

## Why this is the interesting asset

The pitch has been "we keep the verbatim record rather than extracted facts". That is only half true today: we keep the verbatim *talk about* the work. The work itself — what was run, what it printed, what got edited to what — is the part no summarizing system can reconstruct and the part we are also missing.

It is also what six of the ideas in the backlog actually need, and the measurements say the material is good:

- **#541 verify the agents claims** — 2,996 test invocations against 1,274 prose claims of having run or verified something. The comparison is now possible; today it is not.
- **#537 recover what the agent overwrote** — 4,259 `Edit` calls, **every one carrying `old_string`**, median 159 B, p90 568 B, across 912 distinct files. Not whole files, which is what I assumed: the *exact replaced span*, which is better targeted for "it clobbered my function".
- **#527 repeated dead ends** — errors in tool output: 1,732 distinct, **405 in 2+ sessions, 209 in 3+**. From prose alone the same probe finds 843 and 119. Tool output nearly doubles the strong signal and the strings are far cleaner.
- **#539 command history as a time series** — 20,164 clean Bash results against 7,240 error-ish ones. That is the series.
- **#546 automation candidates** and **#526 the decision trail** both get the attempts themselves instead of the description of them.

## What it costs, honestly

Records go from ~44 MB to ~150 MB on this corpus, and the whole index roughly triples. Two consequences:

[...]

Issue #546, opened 2026-07-29
Spot the work being done by hand twelve times and offer to automate it


vshulcz:
The same sequence of commands, typed across sessions, weeks apart, by hand. Nobody notices, because each time it only costs two minutes.

The history knows. This is the one idea here where the output is not knowledge but a tool.

## Shape

```
$ deja repeat
  run 14x across 9 sessions:
      go test ./internal/index/ -count=1 -cover
      go tool cover -func=/tmp/idx.cov | tail -1
    → deja repeat --script coverage   writes scripts/coverage.sh

  run 11x:  gh pr checks $N; gh pr merge $N --squash --delete-branch
```

And the higher-value variant: emit it in the shape the harness already understands — a Claude Code skill, an `AGENTS.md` snippet, a slash command — so the automation is available to the agent, not just to the shell.

## Why us

It needs the command record across sessions and across agents. A per-agent tool sees a fraction; shell history sees commands but not what they were for, and not the ones the agent ran itself.

## Hard parts

- **A repeated command is not automatically worth automating.** `git status` runs a thousand times. The unit is a *sequence* with a stable shape and a purpose, which is a harder pattern than frequency.
- Paths and arguments vary between runs; turning those into parameters is the difference between a useful script and a broken one.
- Generated scripts that fail are worse than no scripts. This should propose, never install.

## Depends on

#539 — recognising command runs and their output inside sixteen transcript formats is the shared foundation. If that does not work, this does not either.

## Kill condition

Mine my own corpus and look at the top ten proposals. If they are dominated by single ubiquitous commands rather than meaningful sequences, the pattern detection is not there.

vshulcz:
Measured from the command role (#561, #589) rather than from raw transcripts.

1,635 distinct commands; **212 run in three or more sessions**. Ranked by frequency, the top of the list is:

| runs | sessions | shape |
|---|---|---|
| 848 | 23 | `git status --short` |
| 650 | 264 | a wrapper's `git status` |
| 595 | 190 | a wrapper's `git diff` |
| 569 | 7 | `kubectl --context <cluster>` |
| 275 | 14 | `git log --oneline` |
| 229 | 74 | a wrapper's `go test` |

**Frequency alone ranks the wrong thing.** Five of the top six are inspection — status, diff, log — which nobody wants automated; they are how a person looks at what is happening. Offering to script `git status` would make the feature look silly on its first run, which for a suggestion feature is fatal.

[...]

Issue #541, opened 2026-07-29
Check whether the agent actually did what it said it did


vshulcz:
An agent says "I ran the tests and they pass". Sometimes it did. The transcript knows which.

This is the anxiety everyone working with agents has right now, and nothing addresses it: not memory tools, not linters, not CI. The evidence is in the tool-call record, and we keep the tool-call record verbatim.

## Shape

```
$ deja audit                     # the session that just ended
claude 2fc1d1ef · 2026-07-29
  claimed  "ran the full suite, green"       no test invocation in this session
  claimed  "verified the benchmark unchanged" ✓ scripts/longmemeval ran 20:14, output quoted
  claimed  "fixed the flake"                 ✓ file edited, but no run afterwards
```

Also as an MCP tool and a hook, so it lands at the end of a session rather than being something you remember to check.

## Why us

It needs the assistants claim and the tool record side by side, in order, verbatim. Every system that stores extracted facts keeps the claim and discards the evidence — the exact inversion of what this needs. And it needs to work across harnesses, because the framing of a tool call differs in all sixteen.

## Why it might be the strongest idea in this batch

- It is not about memory at all, which is why nobody in the memory category will build it.
- The output is about the tool the person is already paying for, and it produces a finding on the first run.
- It is one command with no configuration, and it is checkable — every claim links to a session and a line.
- The headline writes itself, and it is true.

## Hard parts

- **The claim side is natural language.** "Tests pass" appears fifty ways. Detection has to be narrow: a small set of high-confidence claim shapes (ran/verified/checked/tested, plus the object) and silence everywhere else.
- **Accusing an honest agent is worse than missing a lie.** A test run through an unusual path, a suite run in a prior session, a claim about work done by a subagent — all produce false alarms. This has to under-report by design and say "no evidence in this session" rather than "it lied".
- Subagents and background tasks make the record incomplete in ways that look like absence of evidence.

## Kill condition

Take twenty sessions, hand-label the claims, and measure precision. If we cannot get near-zero false accusations, ship it as "claims with no evidence found in this session" — a prompt to look, never a verdict.

vshulcz:
Measured overnight. **This one does not survive its own measurement, and I would reframe rather than build it.**

478 Claude transcripts, 27,196 Bash invocations. Claims about tests, checked against the invocations in the same session:

| detector | claims | substantiated | no run at all | run predates a covered edit |
|---|---|---|---|---|
| naive | 588 | 568 | 20 | — |
[...]

Issue #539, opened 2026-07-29
Your build and test history as a time series, from the transcripts


vshulcz:
Agent transcripts contain thousands of command runs with their output — every test run, every failed build, every stack trace, every `curl` that returned the wrong thing. Nothing indexes that as a **history of what worked when**, and git cannot: most of those failures were local and were never committed.

## Shape

```
$ deja when "TestChangedTaskFilesReadsGitState"
  passed   Jul 22 → Jul 28   (11 runs, claude + codex)
  failed   Jul 29 07:12      windows CI, "got []"
  passed   Jul 29 20:40      after the budget fix

$ deja first "unexpected EOF: record length"
  first seen Jul 29 17:58 in claude 2fc1d1ef — during the span-read rewrite
```

Bisect by transcript rather than by commit. "When did this start failing" is answerable for local, uncommitted, long-since-overwritten states.

## Why us

It needs the tool output kept verbatim and attributed to a moment — the exact thing summarizing memory discards first, because a stack trace looks like noise until you need the date it appeared.

We also already have most of the pieces pointed the wrong way: the paste penalty (#510) currently *demotes* command output in ranking, because as an answer to a question it is noise. This feature is the case where it is the answer, which means it wants a separate retrieval path rather than a tweak to the ranking.

## Hard parts

- Recognising a command run and its output inside a transcript, across sixteen formats that each frame tool calls differently.
- Deciding pass versus fail without running anything: exit codes are sometimes there, and otherwise it is the shape of the output.
- Volume: this is the largest slice of the corpus and the least searched. It may need its own narrower index rather than riding the main one — the opposite of #513, which was about reading less.

## Kill condition

Pick five things I remember breaking and see whether the transcripts can date them. If the answer needs me to already know which session to look in, the retrieval is not there.

vshulcz:
Measured, now that commands are a first-class role (#561, #589) rather than something to be re-derived from raw transcripts.

**The material is there.** Test invocations indexed as `--role command` on a 1150-session store:

| command | records |
|---|---|
| `go test` | 2,301 |
| `make test` | 160 |
| `pytest` | 153 |
| `npm test` | 88 |

Each carries a timestamp and a session, which is the series this issue asks for.

**The outcome is the hard part.** Pairing each `go test` command with the tool output that follows it and looking for a verdict:

| | |
|---|---|
| go-test runs found | 1,180 |
| **with a readable outcome** | **205 (17%)** |
[...]

PR #549, opened 2026-07-29
feat(sources): index the commands that say what happened

vshulcz:
The other half of #547, narrowed after your point that a flag nobody enables is a feature nobody has.

## What changed since the first version

It indexed **every** Bash call: 23,774 commands, 13.1 MB, +27% of the index. Three quarters of that is navigation — `ls`, `cat`, `grep`, `cd` — which is not merely cheap to store but actively harmful: `cat internal/index/index.go` matches a query about the index and answers nothing.

Now it keeps the commands that say what happened — test runs, builds, deploys, `git`, `gh`, `docker`, `deja` itself:

| | commands | text |
|---|---|---|
| every Bash call | 23,774 | 13.1 MB |
| **kept** | **5,051** | **3.5 MB (27%)** |

| index | |
|---|---|
| without commands | 63.90 MB |
| **with the narrowed set** | **68.83 MB (+7.7%)** |

**On by default**, `DEJA_INDEX_TOOLS=0` to turn it off.

## Ranking

**2 of 260 positions move** (`scripts/rankdiff`), down from 11 in the every-command version. Commands are still kept out of ordinary results and served by role, so a command line cannot lift a session because it contains the words of a question.

## What it is for

- **#539** — "when did this test last pass", "when did this error first appear": needs exactly the test and build invocations this keeps.
- **#546** — the work being done by hand twelve times: the repeated sequences I measured were `go test` + `go tool cover` and `gh pr checks` + `gh pr merge`, both inside the kept set.
- **#541** is not on this list. Measured and closed: the agent ran what it claimed 97% of the time.

Tool *output* has always been indexed — 39,836 of 40,933 results are strings the decoder already reads. Only the invocation was missing, so output sat in the index with nothing saying what produced it.

Full suite green, decisionbench 8/8 · 8/8 · 4/4 · 3/3. Index version bumps, so an existing index rebuilds once.

### DIFF
--- a/internal/index/coverage_extra_test.go
+++ b/internal/index/coverage_extra_test.go
@@ -1078,3 +1078,52 @@ func TestFuzzySearchWithPhraseTokens(t *testing.T) {
 		t.Fatalf("reversed phrase result=%#v err=%v", result, err)
 	}
 }
+
+// TestToolRecordsStayOutOfOrdinaryRanking pins the rule that keeps command
+// records from polluting search: they answer "what did I run", never "what did
+// we decide", and left in the ordinary ranking they lift a session because a
+// command line contained the words of a question.
+func TestToolRecordsStayOutOfOrdinaryRanking(t *testing.T) {
+	tmp := hermeticIndexEnv(t)
+	dir := filepath.Join(tmp, "idx")
+	ss := []model.Session{{
+		ID: "s1", Harness: "claude", Project: "app",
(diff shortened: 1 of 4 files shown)


vshulcz:
Superseded by a clean re-application on top of #558 — same change, narrowed selection, default on. The rebase across the paths work tangled two versions of the same file, and re-applying was safer than untangling. New PR below.

Issue #533, opened 2026-07-29
Continue another agent's work across harnesses


vshulcz:
People switch agents mid-task — Claude Code for the design, Codex for the grind, opencode when a subscription runs out — and the context dies at the boundary. Every time, the new agent is told the story again by hand.

We are the only tool positioned for this: 16 native parsers. The nearest competitor reads three harnesses. This is the one place where breadth is not a vanity number but the whole feature.

#524 asks for `deja handoff --to agy` — this is the general shape of it.

## Shape

```
$ deja handoff --to codex
picked up: claude session 8a41c02, "index build allocation profile", 2h ago
  what was being done, in its own words (quoted, not summarized)
  files touched, with the last state of each
  what had already been ruled out
  the exact command that was about to run
```

The receiving agent gets material, not a précis: the turns that matter, verbatim, plus the file list. A summary is where handoffs already fail today — the human writes one and leaves out the constraint that mattered.

## Why not just recall

Recall answers a question. A handoff has no question: the state to transfer is "the last hour of work on this thing", which is a different selection problem — recency and continuity within one thread, not relevance to a query.

## What exists

Sessions carry harness, project, files and time; `deja share` sanitizes; the MCP server can serve the result to whichever agent is receiving. The missing piece is the selection: what belongs in a handoff, and how much.

## Risks

- **Size.** An hour of agent work is far larger than any context budget. This is a ranking problem inside one session, which is not something we do today — we rank sessions, not turns.
- Secrets in the tail of a working session are common; redaction is already there but this path sends the *most recent* material, which is the least reviewed.
- Two agents working the same repo at once. A handoff implies a baton, and there is no locking anywhere in this design.

## Kill condition

Do ten handoffs by hand from real history, measure whether the receiving agent got going without the human re-explaining. If a person still has to fill in the gap, the selection is wrong and the feature is theatre.

vshulcz:
Implemented: deja handoff --to <agent> continues another agent's work across harnesses (cmd/deja/handoff.go).

Issue #451, opened 2026-07-28
handoff calls three harnesses paste-only that take a prompt


vshulcz:
`deja handoff --to cline` prints the prompt for the user to paste, and so do goose and kimi. All three take a prompt from the command line:

- `cline "prompt"` runs the extension headlessly — I ran one and it answered.
- `goose run -t "prompt"` — used it earlier today while verifying the goose integration.
- `kimi -p "prompt"` — "Run one prompt non-interactively and print the response" straight from its own help.

Handoff exists so being stuck in one agent costs one command. Making someone copy and paste into an agent that would have taken the prompt is the difference between a feature and a suggestion.

The four that genuinely cannot stay as they are: antigravity is a GUI, openclaw has no run command, hermes only opens a chat, roo's tasks start from the extension.

PR #550, opened 2026-07-29
feat(handoff): start Antigravity through its CLI, as agy

### LINKED ISSUE #524: Support deja handoff --to agy
We need to support the command `deja handoff --to agy` to allow handing off the session context to the agy agent.

shgpavel:
Closes #524.

`deja handoff --to agy` starts Antigravity with the digest as its first prompt.

Antigravity was listed paste-only on the reading that it is a GUI — #451 named
it first among the four that "genuinely cannot" take a prompt. But it has a
terminal client, `agy`, and it is already in this repo: `deja resume` has always
called `agy --conversation <id>` for antigravity sessions. `agy -i <prompt>`
seeds a prompt into an interactive session, the same shape gemini and qwen use,
so handoff was the one path never wired to it.

On my machine antigravity is the harness deja indexes most — 217 sessions
against claude's 75 — and it was the one deja could not hand work to.

## Behavior

- `--to agy` and `--to antigravity` are the same target; `agy` resolves through
  a one-entry alias table, so the target can be spelled the way its own CLI is
  invoked without a second registry row.
- Registry `antigravity.capabilities.handoff` goes `paste` → `exec`; the README
  and site matrices are regenerated with `go run ./scripts/genmatrix`.
- One deletion worth flagging in review: the paste branch was keyed on the
  literal string `"antigravity"`, so removing it from there leaves `--exec` with
  a single reason to refuse — no `--to`. The second error message,
  `%s has no CLI prompt entry`, became unreachable and is gone.
- hermes, openclaw and roo are deliberately untouched. They still answer
  `don't know how to hand off` rather than printing the digest, exactly as on
  main — that predates this change and belongs in its own issue.

## Verification

Against the real CLI, agy 1.1.7:

| Check | Result |
| --- | --- |
| `agy -p "<prompt>"` | answers — the CLI takes a prompt non-interactively |
| `agy -i "<prompt>"` with no TTY | flag parses, fails only on `could not open TTY` |
| `--exec` through a shim on `PATH` | argv is exactly `agy -i <single 6 KB element>`, no shell |
| `--exec` with `agy` absent | `antigravity is not installed (looked for "agy" in PATH)` |
| `--to notepad` | target list includes `antigravity` |
| `completion bash`/`zsh` | `-n` clean, `agy` present |

End to end: handed a real 6251-byte digest to live `agy` in print mode from a
scratch directory. It reported back what work it was picking up and the next
step, with nothing re-explained — the check #533 asks for.

## Coverage

Per CONTRIBUTING, measured per package rather than trusting the aggregate:

```
cmd/deja                 88.4%   (floor 88)
shipped packages total   88.0%   (floor 87.5)
```

`go test ./...` is green; `go vet ./...` clean. Tests stay hermetic — the new
`--to agy` case runs through `captureRun` on the existing fixture store.


vshulcz:
Reviewed this properly rather than just watching CI go green — thank you, it is a well-argued change and the reasoning in the description is the part I appreciate most. The observation that `deja resume` was already invoking `agy` is exactly the kind of internal inconsistency that is easy to walk past for months.

**What I verified locally**, on your branch:

- The three harnesses that remain paste-only behave exactly as before: `--to openclaw`, `--to hermes` and `--to roo` still error out rather than silently trying to exec anything.
- `--to antigravity` and `--to agy` both resolve to the same target and reach the launch path.
- `--to nosuch` still errors with the target list.
- `go test ./cmd/deja/ -run "Handoff|Capability"` passes, so the registry and the code agree after the `+4` → `+3` change.

**What I could not verify:** `agy -i <prompt>`. Antigravity is indexed on this machine but `agy` itself is not installed, so that flag rests on your report of 1.1.7 — which is fine, it is stated plainly in the test comment, and that is the right way to record it.

**One small thing, not a blocker and arguably a follow-up.** Before this change, antigravity was the only target that reached the friendlier error: *"X has no CLI prompt entry — run `deja handoff --to X` and paste the digest into a new chat."* Now that antigravity has a real entry point, that branch is gone, and the three harnesses still marked `paste` in the table get *"don't know how to hand off to openclaw"* — which reads like a typo rather than a capability. They were already getting that message before your change, so nothing regressed; it is just that the good message no longer exists anywhere. Worth restoring for the remaining three, keyed off the registry rather than a hardcoded name.

Not touching the merge button: Vlad reviews harness changes himself, and he will see this first thing. On his own corpus antigravity is a small harness, so the detail that it is the one deja indexes most on yours is a useful argument for the change.

vshulcz:
Verified on a second machine before merging — installed the CLI rather than taking the flag on trust.

`agy 1.1.8`, straight from `agy --help`:

```
-i                    Short alias for --prompt-interactive
--prompt-interactive  Run an initial prompt interactively and continue the session
--conversation        Resume a previous conversation by ID
```

So `-i` is exactly what you said, and `--conversation` — which `deja resume` has been calling all along — is real too. Worth noting for anyone who checks the docs first: **neither flag appears in the published CLI reference**, which lists only slash commands, so the binary is the authority here.

End to end on my side:

```
deja resume <id>                → agy --conversation 3a2aa020-…
deja handoff --to agy --exec    → launched agy, which then failed only on
                                  "could not open TTY" (no terminal in a subprocess)
```

That second line is the one that matters: deja built the digest, invoked `agy -i <digest>`, and the CLI accepted it. Antigravity indexes here too — 8 sessions, 393 messages — so the whole row is now green on a machine that is not yours.

Merging. Thank you: the observation that `deja resume` was already calling `agy` while handoff called it a GUI is the kind of internal contradiction that survives for months precisely because each half looks right on its own.

vshulcz:
Ready to go in, one rebase away. **0.16.3 shipped while this was open**, so `CHANGELOG.md` now conflicts — the released section landed above your `Unreleased` entry. Nothing else conflicts.

I rebased it locally to check, and the resolution is trivial: your two entries stay under `## [Unreleased]`, the `## [0.16.3]` section follows them. After that, everything builds and `go test ./cmd/deja/ -run "Handoff|Capability"` passes.

I could not push the rebase to your fork from here, so it needs a `git rebase main && git push --force-with-lease` from you.

**Your second commit is the right fix and I should have caught the case myself.** Routing paste-only harnesses through the paste path — rather than leaving them on the "don't know how to hand off" error — is what my review note was gesturing at, and keying it off the registry instead of a hardcoded list is better than what I suggested.

Verification from my side, since it may be useful to you: I installed the CLI rather than trusting the docs. `agy 1.1.8`, from `agy --help`:

```
-i                    Short alias for --prompt-interactive
--prompt-interactive  Run an initial prompt interactively and continue the session
--conversation        Resume a previous conversation by ID
```

Both real, and neither is in the published CLI reference — it lists only slash commands. End to end here: `deja handoff --to agy --exec` launched `agy` and got as far as `could not open TTY`, which is what a subprocess without a terminal should get, so the wiring is right.

vshulcz:
Merged as #562 — your two commits, rebased onto main, authorship intact. 0.16.3 shipped while this was open and took the changelog with it; I could not push the rebase to your fork from this machine, so carrying it forward was the way to get it in without making you redo work you had already finished.

Both changelog entries name you.

For the record, since it may be useful elsewhere: I installed `agy` here to check rather than take the flags on trust. Version 1.1.8 confirms both — `-i` is `--prompt-interactive` ("run an initial prompt interactively and continue the session") and `--conversation` resumes by ID. **Neither appears in the published CLI reference**, which documents only slash commands, so anyone reading the docs first would conclude this was impossible. Your report was right and the documentation is wrong.

And the second commit — routing the registry's paste-only harnesses through the paste path — was the review note I was in the middle of writing when you pushed it. Keying it off the registry rather than a list in the code is better than what I was going to suggest.

Thank you. Antigravity is the fourteenth harness deja can start directly.

PR #554, opened 2026-07-30
fix(search): report the pool the relevance window was taken from

### LINKED ISSUE #497: Relevance results are capped at 50 but report capped: false
Found while checking #495 against this repo's own corpus generator.

On the exact path the new fields are right, and I checked them independently: `deja "the" --json` reports `total: 391, capped: true` with 15 hits, and `--all` counts 391 with the cap lifted. Two paths, one number.

On the relevance path both fields describe a result set that is not the one being returned.

## Reproduce

```sh
go run ./scripts/demo -out /tmp/demo-home
go build -o /tmp/deja ./cmd/deja
export HOME=/tmp/demo-home DEJA_INDEX_DIR=/tmp/idx.db DEJA_EMBED=off
/tmp/deja the >/dev/null 2>&1                     # build the index first
/tmp/deja --json "keepalive column" | jq -c '{match, total, capped, hits: (.hits|length)}'
```

```json
{"match":"relevance","total":50,"capped":null,"hits":50}
```

(Small aside picked up on the way: `corpus.go`'s own doc comment says `go run ./scripts/demo/corpus`, which is not a package path that exists — the working one is `./scripts/demo`.)

Every relevance query I tried on this corpus returns exactly 50. Against the same terms OR-ed together on the exact path, via `--all --re "keepalive|column"`:

| query | relevance `total` | exact-path control | hidden |
|---|---|---|---|
| `keepalive column` | 50 | 82 | 32 |
| `middleware keepalive` | 50 | 117 | 67 |
| `keepalive retry column` | 50 | 143 | 93 |

The individual terms hold 52, 47, 87 and 88 sessions, so 50 is not a property of the data. Neither `--all` nor `--limit=100` lifts it:

```
--json --all       "keepalive column"  ->  total=50  capped=null  hits=50
--json --limit=100 "keepalive column"  ->  total=50  capped=null  hits=50
```

## What is actually wrong

The 50 is deliberate and I would not touch it. `relevanceSearch` (`internal/index/retrieval.go:247`) splits candidates into `keep` (two or more terms matched) and a `weak` single-term tail, then holds the set to 50 at `:257`, `:283` and `:294`, with `internal/search/search.go:301` doing the same downstream. The comments explain the intent and it reads as a ranking-quality decision, not an oversight.

The defect is that this cap is invisible to the two fields you added to describe caps. `total` and `capped` see the 15-session cap inside `Run` and nothing earlier in the call stack, so on the relevance path they report the survivors as the match count, and `capped: false` as "the cap hid nothing" — while 32 to 93 sessions were in fact dropped. It is the same failure mode #495 removed from exact search, one layer further up.

Worth naming alongside it: exact returns 15 rows and relevance returns 50 from the same `--json` invocation, because the relevance path never passes through `Run`'s cap at all. A consumer reading `hits` cannot assume a single window size across tiers.

## Why this blocks #492

CJK queries reach this tier disproportionately. Bigram segmentation means a query with no exact match still produces fragments that exist somewhere in the corpus, so it falls back to relevance instead of returning nothing — on my own history that is the common case rather than the edge case, and it is the behaviour I described in #337.

So any CJK retrieval change measured through the relevance path is measured against a constant 50, whichever direction recall actually moved. That is precisely the "counting the survivors measures the cap" trap #495 set out to close, still open on the tier where the CJK work will land.

## Happy to send a PR

Two shapes, and I'd rather you pick than guess:

- Propagate a pre-cap count out of `relevantMetasCounts` so `total` means the same thing on every tier, and set `capped` when the 50 bites.
- Or, if that count is not cheap at that layer, keep `total` absent on the relevance path and set `capped: true` whenever the tail was truncated — a smaller claim, but one that is never wrong.

Either way the invariant worth having is that `capped: false` means the consumer holds everything that matched, on every tier.

AliceLJY:
## What was wrong

`relevantMetasCounts` collects every candidate, sorts the whole slice, and truncates on its last line. `relevanceSearch` asks it for the top 50, and `runSearch` then took `total` from the sessions it had been handed:

```go
hits = search.RelevanceHits(ss, index.RelevanceMatchTerms(o.Query))
o.Total = len(hits)          // ss was already cut to 50, inside retrieval
```

`o.Capped` was never set on that branch at all. So the two fields whose whole job is to describe a cap were describing the cap itself: `"total": 50, "capped": false`, however deep the pool behind it went.

## The fix

Option one, and it is free exactly as you measured: the pre-cap count is `len(ranked)` one line above `ranked = ranked[:n]`. No extra scan, no second pass, no `schema_version` bump, and the exact path keeps both its numbers and its shape.

- `relevantMetasCounts` returns the pre-truncation count as a fifth value. That is the only place it exists — after the return, the pool the top 50 came out of is gone.
- `relevanceResult()` labels every relevance exit with `Total: matched, Capped: matched > len(ss)`. That includes the merged close-tier tail path, where the tail sessions are added on both sides so `capped` still reads the relevance window — the only thing that withheld anything there.
- `runSearch` takes the tier's figures, with `len(hits)` as a floor for the paths that report none. The quoted-phrase retry is one: a query re-run without its quotes is served under the relevance label but never truncated, so what arrived is the whole of it, which is also what it reported before this change.
- The literal `50` became `relevanceWindow` in the three places `relevanceSearch` used it. Same number, same behaviour; it now has somewhere to write down that `--limit` and `--all` act downstream of it.

## Reproduction, before → after

Sealed against the demo corpus — no real history, and `~/.cache/deja` untouched (fingerprinted before and after, byte-identical):

```sh
DEMO=$(mktemp -d); go run ./scripts/demo -out "$DEMO"        # 392 sessions
export HOME=$DEMO USERPROFILE=$DEMO
export DEJA_CLAUDE_ROOT=$DEMO/.claude/projects DEJA_CODEX_ROOT=$DEMO/.codex DEJA_CURSOR_ROOT=$DEMO/.cursor
export DEJA_INDEX_DIR=$(mktemp -d)/index.db DEJA_EMBED=off
deja --json --no-embed "keepalive column" | jq '{tier, total, capped, hits: (.hits|length)}'
```

| query | before | after | counted independently |
|---|---|---|---|
| `keepalive column` | total 50, capped false, 50 hits | **total 53, capped true**, 50 hits | 53 |
| `middleware keepalive` | total 50, capped false, 50 hits | **total 79, capped true**, 50 hits | 79 |
| `keepalive retry column` | total 50, capped false, 50 hits | **total 96, capped true**, 50 hits | 96 |

The last column is the exact path counting the same term sets on its own — `--all --re 'keepalive|column'` → 53, `'middleware|keepalive'` → 79, `'keepalive|retry|column'` → 96. Two unrelated code paths, the same three numbers.

The issue reported 82 / 117 / 143 against a different corpus; these are the same shape measured somewhere anyone can re-measure.

## Docs

`docs/json-output.md` carries the invariant — `capped: false` means the response holds everything that matched, on every tier — and gains a section saying the length of `hits` is not a fixed window across tiers: exact serves 15 and relevance serves 50 from the same invocation, and neither `--limit` nor `--all` reaches the relevance window because it is applied during retrieval, upstream of both.

That last part is behaviour I documented rather than changed: `--limit 5` on a relevance query returns 50 hits today, and `--all` leaves `capped: true`. Both follow from where the window lives. If you would rather `--limit` were plumbed through to it, that is a behaviour change and I would send it separately.

## Tests

- `internal/index` — `TestRelevanceReportsThePoolItsWindowTrimmed`: table-driven over a pool deeper than the window (61 scored, 50 served, capped) and one inside it (11 scored, 11 served, not capped), on a `t.TempDir()` fixture with `HOME`, `USERPROFILE` and `DEJA_*` pinned. It asserts the tier the fixture resolved through, so it fails loudly rather than silently if the ladder stops falling to relevance, and it asserts the invariant directly: not capped ⇒ `total == len(sessions)`.
- `cmd/deja` — `TestRelevanceEnvelopeReportsThePreCapTotal`: the envelope a consumer actually parses. Relevance total/capped/hits, `--all` not lifting the window, the exact tier unchanged, and a query matching nothing still answering with the envelope rather than the bare array it used to fall back to.
- Both fail against the old semantics with the bug's own signature, `total = 50 capped = false`. I checked that before keeping them.

## Verification

```
go test ./... -race -count=1      every package ok
go vet ./...                      clean
gofmt -s -l cmd internal          no output
sh scripts/e2e.sh                 e2e OK
deja bench recall --json          lexical.recall_at_5 = 1.0   (CI floor 0.85)
```

Per-package coverage, measured the way CONTRIBUTING asks:

| package | floor | measured |
|---|---|---|
| `cmd/deja` | 88 | 88.4 |
| `internal/index` | 90 | 90.6 |

Aggregate over shipped packages: 87.9% against the 87.5% floor.

## Not in this change

- `CHANGELOG.md` — you write those in their own passes; glad to add an Unreleased entry here if you would rather it travelled with the fix.
- The MCP `recall` tool's `matches N-M of T` line counts the hits it is paginating, which is a different contract from the envelope's `total`. Left alone.
- The 50 itself, and anything near #492.

Closes #497.

### DIFF
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -400,7 +400,18 @@ func runSearch(dir string, args []string, sourceInstance string) error {
 	if result.Tier == search.TierRelevance {
 		fmt.Fprintln(os.Stderr, "deja: no exact match; showing sessions ranked by relevance to the whole query")
 		hits = search.RelevanceHits(ss, index.RelevanceMatchTerms(o.Query))
-		o.Total = len(hits)
+		// This tier ranks and truncates inside retrieval, so counting the
+		// sessions it handed back measures its window, not the match: every
+		// query deeper than the window reported the window's own size and
+		// capped: false, which told a consumer to stop checking exactly when
+		// there was something to check. Take the tier's figures when it has
+		// them; the paths that report none (a quoted query retried without
+		// its quotes, served here under the relevance label) never truncated,
+		// so what arrived is the whole of it.
+		o.Total, o.Capped = result.Total, result.Capped
+		if o.Total < len(hits) {
+			o.Total = len(hits)
+		}
 	} else {
 		// RunDetailed rather than Run: the JSON envelope reports how many
 		// sessions matched before the cap, and that is not recoverable from a
--- a/cmd/deja/relevance_envelope_test.go
+++ b/cmd/deja/relevance_envelope_test.go
@@ -1,0 +1,105 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/jsonout"
+)
+
+type searchEnvelope struct {
+	SchemaVersion int    `json:"schema_version"`
+	Tier          string `json:"tier"`
+	Total         int    `json:"total"`
+	Capped        bool   `json:"capped"`
+	Hits          []struct {
+		Session struct {
+			ID string `json:"id"`
+		} `json:"session"`
+	} `json:"hits"`
+}
+
+func decodeEnvelope(t *testing.T, out string) searchEnvelope {
+	t.Helper()
+	var env searchEnvelope
+	if err := json.Unmarshal([]byte(out), &env); err != nil {
+		t.Fatalf("search json: %v\n%s", err, out)
+	}
+	if env.SchemaVersion != jsonout.Version {
+		t.Fatalf("schema_version = %d, want %d\n%s", env.SchemaVersion, jsonout.Version, out)
+	}
+	return env
+}
+
+// total and capped are the two questions a caller cannot answer from a list of
+// hits, and on the relevance tier both used to answer about the ranking window
+// instead of the match: 50 sessions returned, "total": 50, "capped": false,
+// whatever the pool behind it was. A consumer reading that stops checking at
+// the exact moment there is something to check.
+func TestRelevanceEnvelopeReportsThePreCapTotal(t *testing.T) {
+	hermeticEnv(t)
+	root := os.Getenv("DEJA_CLAUDE_ROOT")
+	// 60 sessions carry three of the four query terms and one carries only the
+	// fourth: every term resolves in the corpus, no session holds them all —
+	// which is what drops the ladder to the relevance tier — and the scored
+	// pool is 61, deeper than the window it is served through.
+	for i := 0; i < 60; i++ {
+		id := fmt.Sprintf("rel%02d", i)
+		writeClaudeFixture(t, filepath.Join(root, "relevance", id+".jsonl"), id, []string{
+			`{"type":"user","sessionId":"` + id + `","timestamp":"2026-01-02T03:04:05Z","message":{"role":"user","content":"kubernetes autoscaler telemetry notes ` + id + `"}}`,
+		})
+	}
+	writeClaudeFixture(t, filepath.Join(root, "relevance", "dash.jsonl"), "dash", []string{
+		`{"type":"user","sessionId":"dash","timestamp":"2026-01-02T03:04:05Z","message":{"role":"user","content":"dashboards"}}`,
+	})
+
+	const q = "kubernetes autoscaler telemetry dashboards"
+	out, err := captureRun(t, "--json", "--no-embed", q)
+	if err != nil {
+		t.Fatal(err)
+	}
+	env := decodeEnvelope(t, out)
+	if env.Tier != "relevance" {
+		t.Fatalf("tier = %q, want relevance — the fixture stopped exercising the path\n%s", env.Tier, out)
+	}
+	if len(env.Hits) != 50 {
+		t.Fatalf("hits = %d, want the relevance window of 50", len(env.Hits))
+	}
+	if env.Total != 61 || !env.Capped {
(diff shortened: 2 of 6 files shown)


vshulcz:
Merging this. Reproduced the bug and verified the fix on a 1,150-session corpus rather than taking CI's word:

**Before, on main:**
```
"что мы решали про блочное сжатие…"   tier=relevance total=50 capped=absent hits=50
```

**On your branch:**
```
same query                            tier=relevance total=81 capped=true  hits=50
"как мы чинили покрытие в CI…"        tier=relevance total=43 capped=absent hits=43
"keepalive column"                    tier=exact     total=1  capped=absent hits=1
```

81 behind a window of 50, reported as such; the query that never hit the cap is unchanged; the exact path untouched. Full suite green, `decisionbench` 8/8 · 8/8 · 4/4 · 3/3, `bench recall` R@5 1.00.

The part I want to name is the diagnosis, not the patch. *"The two fields whose whole job is to describe a cap were describing the cap itself"* is exactly right, and `len(ranked)` one line above the truncation being the only place the pre-cap count still exists is the observation that makes this free — no second pass, no schema bump. I had assumed this would cost a scan and was wrong.

Thank you. Second time this week that a correction of yours has been worth more than the original report — the `--limit` retraction and the interleaved re-measurement in #492 were the same instinct.

Issue #633, opened 2026-07-31
sessionFromMeta drops Touched, so Recent* callers silently get none


vshulcz:
There are two constructors turning a `SessionMeta` into a `model.Session`:

- `retrieval.go:1086` copies `Touched`
- `ingest.go:960` (`sessionFromMeta`) does not

`Recent`, `RecentMatching` and everything built on them go through the second one, so a caller reading `Session.Touched` from a recent-sessions list gets an empty slice on every session and no error — the field is simply absent.

Nothing is broken today: the two consumers (`moved.go`, MCP `blame`) both arrive through the search path that copies it. But the failure mode is silence, and it already cost me a wrong measurement — reading `Touched` off `index.Recent` reports 0 of 1153 sessions carrying files, while the manifest has them.

Same shape as the field itself: `Asked` and `Hit` are dropped by that constructor too.

PR #634, opened 2026-07-31
feat(statusline): what earlier sessions decided about the file in hand

### LINKED ISSUE #581: Ambient presence: say what memory this file has, while the work is happening
Every screen deja has is a screen someone must decide to open. The recall in the agent is invisible (#579), the brief is in a terminal, the stats page is a destination. So the tool is either forgotten or credited to the model.

`deja statusline` already exists and already runs continuously. It is the only surface that is present *during* the work, and it currently does not say the one thing that would matter there:

```
deja · 3 sessions touched store_io.go · last decision Jun 12
```

The data behind it is now cheap. Sessions carry the files they touched in the manifest as of #571, so answering "which sessions touched the file I am in" costs no record read — the same lookup that made the moved-since annotation free.

Design constraints that decide whether this is good or annoying:

- it must be silent when there is nothing, not print a zero
- it must cost nothing per refresh — a statusline runs constantly and a fork there is unacceptable
- it should name a decision, not a count, when one is available: "3 sessions" is a statistic, "you chose transaction pooling here in June" is a memory

vshulcz:
Closes #581.

```
deja · application.go · 1 earlier: "explorer TASK: Codebase reality check…" May 25 · 12 recalls · 18.1 KB ctx today
```

**The version the issue describes cannot be built, and that is worth stating.** Claude Code's status-line payload has no field naming the file under edit — no `current_file`, no `active_file`, none — so "3 sessions touched store_io.go" cannot be tied to what is open. I checked the documented schema rather than assuming.

What the payload does carry is `transcript_path`, which identifies the *current session*, and a session's busiest files have been in the manifest since #571. So the claim made here is the one the data supports: the file this session has worked on most, and what deja remembers about it.

**The three constraints from the issue:**

- *Silent when there is nothing.* No transcript path, unknown session, no index, or a file no one else touched — all print exactly the line that was there before. There is a test per case.
- *Costs nothing per refresh.* Manifest only: no record read, no lock, no fork. Measured on a 1153-session store, unchanged at ~10 ms.
- *A decision, not a count.* The earlier session's title wins when there is one; the count is the fallback, with the singular spelled right.

The usage half is shortened when a memory is present rather than the line simply growing — a status bar lives in whatever width the terminal has, and the tail is what gets lost.

**On testing.** Beyond the cases above: junk payloads (`not json`, `null`, `[]`, wrong types, 4 MB of garbage) and a mutation check — reverting the lookup, counting the current session as its own memory, and printing a zero instead of staying silent each make the suite fail.

`index.AllMeta` is new because the existing constructors drop `Touched` (#633, filed separately — nothing is broken today, but the failure mode is silence and it already cost me one wrong measurement).

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ $ deja "jwt refresh token"
 | `deja warmup` | Build/refresh the index without searching — handy in cron or shell startup. |
 | `deja index [--rebuild]` | Same as warmup; `--rebuild` forces a full rebuild. Cold builds narrate per-harness progress. |
 | `deja update` | Download the latest GitHub release, verify its checksum, and replace the current binary. |
-| `deja statusline` | One line for your status bar: recalls served to agents today. `deja install statusline` wires it into Claude Code (won't touch an existing statusline). |
+| `deja statusline` | One line for your status bar: recalls served to agents today, plus the file this session works on most and what an earlier session decided about it. Silent when there is no earlier session to name. `deja install statusline` wires it into Claude Code (won't touch an existing statusline). |
 | `deja log [n] [--last] [--json]` | Audit what deja actually served: recent recalls and injections, or the exact text of the last injected digest with `--last`. |
 | `deja` | On a terminal with an index: a living brief — today's sessions, recalls served, déjà vu moments, a question you asked in more than one session, a wall your agents keep hitting, and a search suggestion from your own history. |
 
--- a/cmd/deja/statusline.go
+++ b/cmd/deja/statusline.go
@@ -13,7 +13,7 @@ import (
 // to agents today. It must stay fast and quiet — no index access, no locks —
 // because status bars call it constantly.
 func runStatusline(dir string, stdin io.Reader, stdout io.Writer) error {
-	drainStdin(stdin)
+	in := readStatuslineInput(stdin)
 	// A first build takes a while and runs detached. The status bar is the
 	// one surface the user is already looking at, so the build reports there
 	// instead of leaving them to wonder why recall is quiet.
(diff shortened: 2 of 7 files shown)


vshulcz:
Independent review found nine issues. Seven fixed, two answered with a measurement.

**Fixed**

1. **The filename reached the bar raw** — the same hole as the title, through the other half of the segment. `/w/pool\r\x1b[31mEVIL.go` rewrote the line and leaked colour, and a 200-character path produced a 226-rune line on a bar with no horizontal scroll. Both halves now go through one function, each with its own rune cap.
2. **`unicode.IsControl` misses format characters.** U+202E reverses the rendering of everything after it; zero-width characters pad the bar invisibly. Neither is category Cc.
3. **Only `Touched[0]` was consulted.** The busiest file of a session is often the one being written from scratch, which by definition no earlier session touched. Now every file the session worked on is scanned, busiest first.
4. **Identity was guessed from a basename.** The manifest is keyed `harness:id`, so an id alone can name two sessions and Go's randomised map order picked between them — the reported file could flip between two refreshes 300 ms apart. Matching on `SessionMeta.Path` against the payload's own `transcript_path` is exact; the basename stays as a fallback. Exclusion is now the pair, so a second harness with the same id is not dropped from the count.
5. **The recency selection was untested.** Flipping `.After` to `.Before` left the whole package green, because every seeded session had identical text. They now differ and the test names the newest.
6. **`TestReadStatuslineInputIsBounded` observed nothing** — it passed with no limit at all. It counts bytes now.
7. **A zero `Updated` printed `Jan 1 0001`.** Suppressed, with no dangling separator left behind.

Plus two the review noted in passing: the `runStatusline` comment claimed "no index access" and no longer could, and `drainStdin` was dead with a test that existed only to test it. Both gone. Nested quotes (`1 earlier: "why does "x" break"`) now use an outer pair that cannot collide.

**Measured rather than taken on faith**

The review called stopping at `Touched[0]` "not an edge case, the common shape of a working session". On a 1153-session store: 614 sessions speak today, **41** are silent with a memory one entry down, 21 are correctly silent. So it is 6%, not the common case — worth fixing, and it is fixed, but the scale was overstated.

Manifest cost was confirmed independently at ~2 ms cold, which matches what I measured.

**Verification.** Six mutations, each of which now fails the suite: reversed recency sort, control characters through the title, format characters, unsanitised filename, zero-date printing, identity by id alone, and `Touched[:1]`. Three of the seven fixes initially had no test catching them — those tests were written before the fixes were kept.

PR #638, opened 2026-07-31
feat: name the memory agents keep coming back to

### LINKED ISSUE #579: The recall is invisible, so the improvement gets attributed to the agent
Once hooks are wired, deja injects recalled context into agent prompts and the user never sees it happen. The session goes better and the credit goes to the model. That is the retention problem in one sentence: a tool nobody notices working is a tool that gets uninstalled during the next cleanup.

**Correction to the first version of this issue:** I wrote that the brief reports `0 recalls` despite hooks being installed. That was an artefact of my own test — I ran it against a scratch index directory, and the usage log lives beside the index, so there was nothing to count. Against the real store the line is already good:

```
today      2 sessions · 8 recalls served (10.7 KB from 627.6 KB)
this week  145 sessions · 75 recalls · 108 déjà vu moments
```

So the gap is narrower than stated, and it is about *specificity and placement* rather than a missing counter. Two things the numbers above still do not do:

- **They do not name anything.** "75 recalls" is a rate; "most reused: *pgbouncer prepared statements*, 4×" is a memory. The second one is what a person repeats to a colleague.
- **They are only on a screen nobody has a reason to open.** The recall happens in the agent; the receipt is in a terminal the user may not visit for days. Whether a receipt belongs closer to the agent — the statusline already exists — is the real question here.

`deja stats --impact` already labels the arithmetic behind these numbers and should stay the source of truth rather than growing a second counter beside it.

vshulcz:
Closes #579.

"75 recalls" is a rate — it says deja did something seventy-five times, and a reader can neither check it nor repeat it to anyone. The issue asks for the other half: name the work that kept being worth recalling, and put it where someone actually sees it.

Both places, from data already recorded (every served recall logs which sessions it returned), so `deja stats --impact` stays the source of truth for the arithmetic:

```
reused     pgbouncer prepared statements
           4× by agents · last worked Jul 20
```

and in the session-start receipt, which is the one surface a person reliably sees:

```
deja: recalled 3 prior sessions touching brief.go, compact.go … — the agent starts
already knowing them · most re-used so far: "…", 31×
```

**The measurement that shaped it.** The most-served session on this machine had 67 recalls — and it was the session doing this work, recalling itself all afternoon. Naming that is a mirror, not a memory. So a session has to have settled for a day before its reuse counts, and there is a test that fails without the filter.

Two more floors, each with a test: a session recalled once is an answer rather than a memory, and a session whose title deja never recovered cannot be named at all — a count with no subject is exactly what the issue objects to.

**Verification.** Five mutations: dropping the settle filter, taking the minimum instead of the maximum, admitting untitled sessions, and removing the reuse floor — the floor is checked twice in the code, so it takes removing both to change behaviour, and the test catches that. Only the redundant early return is invisible to it.


vshulcz:
Independent review found eight issues. Six fixed, one confirmed as clean, one taken as wording.

**Fixed**

1. **Ambiguous ids are now dropped, not guessed.** The reviewer showed the named memory flipping between runs on identical data — 347/53 across 400 calls in one process, because the metas arrive from a map. The deeper half is that no tie-break can fix it: the usage log records a bare id, so the count under a duplicated id is the *sum* of two sessions and naming either attaches a number belonging to both. Silence is the only honest answer. Latent on this store (1153 sessions, 1153 distinct ids), and there is now a test that calls the function 50 times to prove it stays silent.
2. **Undated sessions no longer print `Jan 1 0001`.** The zero time is before any cutoff, so the settle filter admitted it. Same defect as finding 7 of the #634 review, in a new place — which is why the fix is a check, and the test builds the case through the real Codex history parser with a non-RFC3339 timestamp rather than constructing a zero time by hand.
3. **`by agents` was wrong.** `WornSessions` counts déjà-vu events, which are written when the *user's own prompt* returns to old ground. Now `re-used`.
4. **Titles are scrubbed once, in `trimBriefTitle`.** The reviewer is right that this PR added a second instance of a class it did not create — the `recent` lines have printed raw titles since they existed. Fixing it at the shared helper closes both, and matches the rule #634 set for the status bar.
5. **`reusedTitles` deleted.** Dead on arrival, 37 lines duplicating the lookup with quietly different semantics.
8. **`so far` → `recently`.** The usage log keeps a rolling window, so the stronger claim expires the first time the log rotates.

**Test quality (6).** All four mutations the reviewer found surviving now fail the suite, and I checked each one honestly — my first attempt at the title mutation "passed" only because it broke compilation, which proves nothing. `TestFindReusedMemorySkipsUntitledSessions` was indeed misnamed: it exercised the does-it-exist path, and a blank-title case now exists beside it.

**Cost (7).** Confirmed clean and independently measured — ~3 ms in a cold hook process, and it only runs inside `sessions > 0 && receiptIsNews(...)`. Matches what I measured.

Issue #551, opened 2026-07-30
Harness plumbing is indexed as if a person said it


vshulcz:
Measured while filtering the corpus for #530: **833 records, 0.82 MB — 1.2% of records and 1.9% of indexed text — are harness mechanics rather than anything a person or an agent said.**

| what | records | MB |
|---|---|---|
| `<` / `tool_use_error` fragments | 293 | 0.06 |
| `[Request interrupted by user]` | 129 | 0.02 |
| `<bash-stdout>` / `<bash-stderr>` wrappers | 124 | 0.16 |
| `<local-command-caveat>` | 87 | 0.03 |
| `<system-reminder>` | 84 | 0.13 |
| **deja's own hook payloads** | 65 | 0.41 |
| slash-command echoes (`<command-name>/clear`) | 50 | 0.01 |
| `[Your previous response had no visible output]` | 1 | 0.00 |

Two reasons this is worth an hour rather than nothing:

- It is indexed **as content**, so it can be recalled, can match a query, and counts toward the stats a user sees. `[Request interrupted by user]` is not memory.
- The hook-payload rows are the largest slice by bytes and are the same class of problem as #480 — deja indexing its own output. #488 fixed the recall blocks; the hook envelope evidently still gets through.

Credit where due: the parser already filters nearly all of this. `[Your previous response had no visible output]` appears in **103** raw subagent transcripts and exactly **one** record in the index. The remainder is a long tail of markers that were not on the list.

Suggested shape: one place that recognises harness-injected wrappers, applied at parse time across harnesses rather than per-parser, with the list of markers in a table so adding one is a two-line change. Worth checking whether it belongs next to the `<deja-recall>` filter from #488, since that is the same idea with a different marker.

Not urgent. Filing so it is measured rather than suspected.

PR #639, opened 2026-07-31
fix(codex): thread identity and injected preamble

### LINKED ISSUE #635: Codex: 811 threads collapse into 74 sessions — distinct ThreadIds are keyed on the shared SessionId
Tested on `main` @ 17864d9 (after #629), macOS arm64, a Codex-heavy store.

Codex's `SessionMeta` carries two identifiers with different cardinality ([protocol.rs#L3066-L3072](https://github.com/openai/codex/blob/aea26afaee177d3fe40721ef261a29f89879d505/codex-rs/protocol/src/protocol.rs#L3066-L3072)):

```rust
pub struct SessionMeta {
    pub session_id: SessionId,
    pub id: ThreadId,
    pub forked_from_id: Option<ThreadId>,
    pub parent_thread_id: Option<ThreadId>,
    …
}
```

One rollout file is one **thread**. A **session** groups every thread that branched from it, and `forked_from_id` / `parent_thread_id` link them. So `id` (the ThreadId) is the per-rollout identifier — it is what the filename carries — and `session_id` (the SessionId) is shared by the whole group:

```json
{"type":"session_meta","payload":{
  "session_id":"<SessionId>",   // SessionId — shared by all 143 threads below
  "id":"<ThreadId>",           // ThreadId  — this rollout, equals the filename
  "forked_from_id":"<parent ThreadId>"}}
```

`internal/sources/codex.go:75` derives a unique ID from the filename — effectively the ThreadId — and then lines 88–91 overwrite it with the SessionId:

```go
if typ, _ := m["type"].(string); typ == "session_meta" {
    if id, _ := payload["session_id"].(string); id != "" {
        s.ID = id
    }
```

Every thread in a session is therefore keyed as the same deja session.

## Scale

Counted over this store's rollouts:

| | |
|---|---|
| rollout files carrying `session_meta` | 811 |
| distinct `ThreadId` (`payload.id`) | 811 |
| distinct `SessionId` (`payload.session_id`) | **74** |
| threads that collapse into another | **737 (91%)** |
| threads carrying `forked_from_id` | 293 |
| largest single session | 191 threads (then 143, 137, 100, 35) |

In the 191-thread session, all 191 ThreadIds are distinct and 150 of them have a `forked_from_id` naming another ThreadId in the same group — the fork tree is fully recorded in the store and entirely discarded by the parser.

The two counters deja prints disagree because of it: `deja index` reports `codex: 865 sessions`, and `deja stats` then reports **268** sessions in total across nine harnesses.

## Reproduction

Three threads, one session, distinct ThreadIds:

```sh
R=/tmp/deja-fx; rm -rf $R; mkdir -p $R/sessions/2026/07/31
mk(){ cat > "$R/sessions/2026/07/31/rollout-2026-07-31T00-00-0$1-$2.jsonl" <<EOF
{"timestamp":"2026-07-31T00:00:0$1Z","type":"session_meta","payload":{"session_id":"$3","id":"$2","cwd":"/tmp/fxproj"}}
{"timestamp":"2026-07-31T00:00:0$1Z","type":"response_item","payload":{"type":"message","role":"user","content":[{"type":"input_text","text":"unique marker $4 in thread $1"}]}}
EOF
}
mk 1 aaaaaaaa-1111-4111-8111-111111111111 aaaaaaaa-1111-4111-8111-111111111111 ZEBRAFIRST
mk 2 bbbbbbbb-2222-4222-8222-222222222222 aaaaaaaa-1111-4111-8111-111111111111 ZEBRASECOND
mk 3 cccccccc-3333-4333-8333-333333333333 aaaaaaaa-1111-4111-8111-111111111111 ZEBRATHIRD

export DEJA_CODEX_ROOT=$R DEJA_INDEX_DIR=/tmp/deja-fx-idx
deja index --rebuild | grep codex
deja last 10 --harness codex
deja show bbbbbbbb-2222 --harness codex
```

Observed:

```
deja: codex: 3 sessions, 3 messages
[codex · fxproj · 2026-07-31 · aaaaaaaa-1111-…] unique marker ZEBRAFIRST in thread 1
deja: no session matches "bbbbbbbb-2222"
```

One entry listed instead of three, and the other two ThreadIds are not addressable by any id.

## What it costs

1. **Listings hide most threads.** `deja last` shows 1 of 3 in the fixture; 74 of 811 on the real store.
2. **Provenance is wrong.** `deja show aaaaaaaa-1111-…` returns all three transcripts merged, but reports `session.path` as the **third** file. On the real store the 143-thread session reports one arbitrary rollout path for 14.6 MB of merged content. Anything reading that path — `deja resume` in particular — reopens the wrong thread.
3. **Ranking is dominated by the merged sessions.** Match counts aggregate across every thread, so the 143-thread session reports `6976 matches` and takes rank 1 for unrelated queries. The sessions a user branched most are exactly the ones that crowd everything else out.
4. `started`/`updated` become the union of the group, so `--since` and recency ranking see one long session instead of many short threads.

## Suggested fix

Key on the ThreadId, falling back to the SessionId:

```go
if id, _ := payload["id"].(string); id != "" {
    s.ID = id
} else if id, _ := payload["session_id"].(string); id != "" {
    s.ID = id
}
```

`payload.id` is present in **811/811 (100%)** of the rollouts here and equals the filename id in **811/811 (100%)**, so the fallback is only there for older formats.

If grouping branched threads is wanted later, `session_id`, `forked_from_id` and `parent_thread_id` are the fields for it — as a thread-group key beside a unique per-thread id, rather than in place of one.

### LINKED ISSUE #636: Codex: harness preamble is indexed as content and becomes the session title in 81 of 82 sessions
Tested on `main` @ 17864d9, macOS arm64.

#551 closed the Claude-side markers (`<system-reminder>`, `<bash-stdout>`, `[Request interrupted by user]`, deja's own hook payloads) and suggested "one place that recognises harness-injected wrappers, applied at parse time across harnesses rather than per-parser". Codex's wrappers were not on that list, and they are the dominant case on a Codex store.

Codex injects its preamble as `response_item` / `message` records with `role: "developer"`. `internal/sources/codex.go:109-119` stores any record with a non-empty role and text, so these land in the index as content and — being first — become the session title.

## Measured

Every Codex session title in this index:

| title prefix | sessions |
|---|---:|
| `<recommended_plugins> Here is a list of plugins…` | 41 |
| `<environment_context> <cwd>…` | 31 |
| `# AGENTS.md instructions for …` | 5 |
| something a person typed | **1** |
| **total** | **82** |

`81 of 82` (99%) titles are harness plumbing. `deja last --harness codex` is 81 identical-looking rows.

The text is searchable as content, too:

```
$ deja "recommended_plugins" --limit 2
[codex] <project> · 2d ago · <id> — 2 matches
  <recommended_plugins> Here is a list of plugins that are available but not installed. - Atlassian
```

## Why it matters beyond cosmetics

The preamble is identical across sessions and contains a broad vocabulary — plugin names, tool descriptions, and, where a user-level `AGENTS.md` exists, a persona paragraph naming every domain that user works in. That makes it a strong lexical match for a large fraction of natural-language queries, and it is present in *every* session, so it wins on aggregate match count.

Three unrelated queries on this store — a vendor error string, an infrastructure term, and a file-format name — each returning the same injected paragraph at rank 1. The matched text is a user-level `AGENTS.md` persona block, so I have replaced its contents; the counts are as measured:

```
$ deja "<vendor error string>" --limit 1
[codex] … — 111 matches
  … [redacted: ~200 chars of a subagent role description from the injected preamble] …

$ deja "<infrastructure term>" --limit 1
[codex] … — 1024 matches
  … [redacted: ~200 chars of the AGENTS.md persona paragraph naming this user's work domains] …

$ deja "<file-format name>" --limit 1
[codex] … — 342 matches
  … [redacted: the same persona paragraph as above] …
```

None of the three is a hit on anything the user or the agent said; all three are the injected block. For `--auto` this matters more than for search, since the 2 KB SessionStart digest is drawn from the same ranking.

This compounds with the fork-collapse issue filed alongside this one — the merged groups carry one copy of the preamble per fork, so the aggregate count grows with the number of forks.

## Suggested shape

Two separable parts:

1. **Do not treat `role: "developer"` records as session content.** They are the harness talking to the model, not the user or the assistant. Dropping them, or storing them under a role that ranking and title selection ignore, removes all three title classes above.
2. **Title selection should prefer the first `user` record**, falling back to the current behaviour only when a session has none.

If a marker table is preferred over a role rule, the Codex entries are `<recommended_plugins>`, `<environment_context>`, `<app-context>`, and the `# AGENTS.md instructions for <path>` header.

vshulcz:
Closes #635. Closes #636. Both reported by @hieilookie with measurements from a Codex-heavy store; both reproduce here.

**Identity (#635).** One rollout file is one *thread*; a session groups every thread that branched from it. The parser derived the ThreadId from the filename and then overwrote it with the SessionId, so every thread in a fork tree became the same deja session. Reproduced on the minimal case from the issue — three rollouts, distinct ThreadIds, one shared SessionId:

```
before   deja: codex: 3 sessions      last → 1 row  (SHARED-SESSION)
after    deja: codex: 3 sessions      last → 3 rows (aaa, bbb, ccc)
```

The reporter's store shows the scale: 811 rollouts, 811 ThreadIds, **74** SessionIds. Mine cannot — 28 rollouts, no forks — so that number is theirs, not mine. `payload.session_id` stays as a fallback for rollouts written before Codex split the two, where there are no threads to collapse.

**Preamble (#636).** Codex writes its preamble both under `role: "developer"` and, for the AGENTS.md block, **as a user turn** — which the reporter's suggested role filter would not have reached. Measured here: `developer` is the first role in 28 of 28 files, and the first *user* message is injected plumbing in 28 of 28 as well. Zero real titles.

```
before   # AGENTS.md instructions <INSTRUCTIONS>…   ×28
after    Read main.go, change the word hello…
         prepared statement errors behind pgbouncer…
```

Codex records drop from 310 KB to 28 KB — 91% of what was indexed for that harness was plumbing.

Two parts, per the issue's suggested split: `HarnessAuthored` in `internal/sources` is the one place #551 asked for, and the user-turn preamble is handled by the existing `stripSelfRecall` machinery.

**One thing worth stating.** Those tags are ordinary enough that people quote them: nine Claude sessions on this store contain `<environment_context>` and two contain `<user_instructions>` — every one a real turn discussing this very filtering. Stripping them anywhere in a message ate 6 real records; stripping them only at the *head* removes the same 282 KB of preamble and keeps those. A preamble is always the head of the message, so the restriction costs nothing.

Index version 21: existing Codex entries have the wrong identity and cannot be repaired incrementally.

Four mutations, each fails the suite: keying back on the SessionId, restoring `developer` as content, dropping the preamble strip, and relaxing the head-only rule.

### DIFF
--- a/internal/index/index.go
+++ b/internal/index/index.go
@@ -22,7 +22,10 @@ import (
 // 20: opencode tool output is parsed (#621) and each session records what it
 // tripped over (#576). An index built before this has neither, and neither can
 // be derived from what it does hold.
-const version = 20
+// 21: Codex sessions are keyed on the ThreadId rather than the SessionId
+// (#635), so every existing Codex entry has the wrong identity, and the
+// harness preamble Codex injects as a user turn is stripped (#636).
+const version = 21
 const maxIndexedText = 64 * 1024
 
 // maxRecordSize bounds a single serialized record. A record is one message
--- a/internal/index/selfrecall.go
+++ b/internal/index/selfrecall.go
@@ -42,6 +42,31 @@ var injectedBlocks = [][2]string{
 	{"<command-name>", "</command-args>"},
 }
 
+// Codex writes its preamble as a user turn rather than under its own role, so
+// a role filter does not reach it. Measured on two stores: it is the first
+// user message in 28 of 28 sessions here and titles 81 of 82 there (#636).
+// Identical in every session and lexically broad — plugin names, tool
+// descriptions, an AGENTS.md persona paragraph — so it outranks real turns on
+// aggregate match count, which makes it a ranking bug rather than a cosmetic
+// one.
+//
+// These are stripped only at the *start* of a message, unlike injectedBlocks.
+// The tags are ordinary enough that a person quotes them: on this store nine
+// Claude sessions contain `<environment_context>` and two contain
+// `<user_instructions>`, every one of them a real turn discussing the very
+// filtering this list implements. A preamble is always the head of the
+// message, so requiring that costs nothing and stops the filter from eating
+// someone's sentence.
+var injectedPrefixBlocks = [][2]string{
+	{"<environment_context>", "</environment_context>"},
+	{"<recommended_plugins>", "</recommended_plugins>"},
+	{"<permissions instructions>", "</permissions instructions>"},
+	{"<skills_instructions>", "</skills_instructions>"},
+	{"<user_instructions>", "</user_instructions>"},
+	{"<INSTRUCTIONS>", "</INSTRUCTIONS>"},
+	{"# AGENTS.md instructions", "</INSTRUCTIONS>"},
+}
+
 // Whole-line markers: a harness note that occupies its own line, with no
 // closing tag. Matched at line granularity so a line quoting one inside a real
 // message does not take the message with it.
@@ -63,6 +88,7 @@ func stripSelfRecall(text string) string {
 	for _, m := range injectedBlocks {
 		text = stripBetweenClosed(text, m[0], m[1])
 	}
+	text = stripInjectedPrefixes(text)
 	return stripInjectedLines(text)
 }
 
@@ -92,6 +118,28 @@ func stripInjectedLines(text string) string {
 // truncated <deja-recall> is still ours, but a sentence mentioning
 // <system-reminder> — a bug report, this file's own tests — is not, and
 // swallowing the rest of it loses what the person actually wrote.
+// stripInjectedPrefixes removes a harness preamble from the head of a message,
+// repeatedly: Codex stacks several of them before the first real turn.
+func stripInjectedPrefixes(text string) string {
+	for again := true; again; {
+		again = false
+		trimmed := strings.TrimSpace(text)
+		for _, m := range injectedPrefixBlocks {
+			if !strings.HasPrefix(trimmed, m[0]) {
+				continue
+			}
+			rel := strings.Index(trimmed[len(m[0]):], m[1])
+			if rel < 0 {
+				continue
+			}
+			text = trimmed[len(m[0])+rel+len(m[1]):]
+			again = true
+			break
+		}
+	}
+	return text
+}
+
 func stripBetweenClosed(text, open, close string) string {
 	for {
 		start := strings.Index(text, open)
--- a/internal/index/selfrecall.go
+++ b/internal/index/selfrecall.go
@@ -63,9 +63,23 @@ var injectedPrefixBlocks = [][2]string{
 	{"<permissions instructions>", "</permissions instructions>"},
 	{"<skills_instructions>", "</skills_instructions>"},
 	{"<user_instructions>", "</user_instructions>"},
-	{"<INSTRUCTIONS>", "</INSTRUCTIONS>"},
-	{"# AGENTS.md instructions", "</INSTRUCTIONS>"},
 }
+
+// agentsHeading is the line Codex writes above the AGENTS.md block it injects,
+// and it needs its own rule rather than a place in the list above.
+//
+// As a plain open/close pair it spans from the heading to the *next*
+// `</INSTRUCTIONS>` anywhere in the message, and the two do not have to belong
+// together. A person writing "# AGENTS.md instructions — how do I stop Codex
+// injecting these? Mine is: <INSTRUCTIONS>be terse</INSTRUCTIONS>" lost the
+// whole question, which is precisely the message someone writes when reporting
+// this bug. Requiring the opening tag to follow the heading immediately makes
+// the span the injected block and nothing else.
+const (
+	agentsHeading = "# AGENTS.md instructions"
+	agentsOpen    = "<INSTRUCTIONS>"
+	agentsClose   = "</INSTRUCTIONS>"
+)
 
 // Whole-line markers: a harness note that occupies its own line, with no
 // closing tag. Matched at line granularity so a line quoting one inside a real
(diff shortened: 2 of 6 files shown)


vshulcz:
Independent review found eight issues. Six fixed, two kept deliberately and now stated as such.

**1 — the one that mattered.** An appended rollout is parsed from an offset, so `session_meta` is never read again and the id fell back to the filename: one session became two, and every further turn landed in the second half. Reproduced exactly as reported.

```
after first build: codex:thread-x
after append:      codex:2026-07-31T00-00-00-thread-x
```

That undoes #635 for precisely the sessions someone is still talking in, and the filename form matches the real ThreadId in 0 of 28 rollouts here — so it can never coincide by luck. The head is now re-read on an offset parse, which also restores the project. Pre-existing, as noted, but this PR is about Codex identity and the version bump already forces the reindex that heals it.

**3 and 4 — the over-reach, fixed together.** `# AGENTS.md instructions` paired with `</INSTRUCTIONS>` spanned to the next closer anywhere in the message, so:

```
in   "# AGENTS.md instructions\n\nHow do I stop Codex injecting these? Mine is:\n<INSTRUCTIONS>be terse</INSTRUCTIONS>"
out  ""
```

That is the message someone writes when reporting this bug, deleted whole. The heading now only introduces the block when the block follows it immediately. And `<INSTRUCTIONS>` is dropped from the list entirely — you are right that it had neither a test nor an occurrence on either corpus, and it carried the largest blast radius of the seven.

**2** — a present-but-malformed `id` fell through to the SessionId, silently reintroducing the collapse. Absent and malformed are now different cases: absent keeps the legacy SessionId, malformed keeps the filename.

**5** — nesting is counted, so `<T>a<T>b</T>c</T>tail` no longer leaves `c</T>` behind.

**8** — the dead assertion is gone.

**6 and 7 kept, deliberately.** An unclosed head block is left alone: stripping to the end of the message would delete real text after a truncated write. A trailing preamble survives, which is the cost of head-only — and head-only is what keeps a person's quoted tags intact. Both now have a test that says so, so changing either is a decision rather than a slip.

Thank you for the cross-harness quantification — 56 records changed, all codex, zero elsewhere, is exactly the number that makes this safe to merge. Four mutations behind the new fixes: the heading rule, the nesting count, the head re-read, and the malformed-id branch.

Issue #499, opened 2026-07-29
Index build is now allocation-bound


vshulcz:
After #498 the profile of a full rebuild has a new shape. The top of it is no longer parsing:

```
17.55%  runtime.memclrNoHeapPointers
 9.66%  runtime.madvise
 6.05%  syscall.syscall
 5.77%  runtime.memmove
```

That is allocation churn and the runtime handing memory back to the OS — not work anyone asked for. `mapassign_faststr` was 10.75% before the redaction work and is still near the top, which points at the postings maps being grown one assignment at a time rather than sized once.

Worth trying, in the order the profile suggests:

- Size the per-bucket postings maps from the message count instead of letting them grow. The count is known before the walk.
- Reuse the per-message token slice across messages in a worker rather than allocating one each time. `index.tokens` was 11.30% cumulative.
- Check whether `madvise` traffic is `GOGC` fighting a build that allocates in bursts; a `debug.SetGCPercent` bump for the duration of a rebuild is a two-line experiment that either shows up in the numbers or does not.

Method that worked and should be repeated: measure with interleaved runs, minimum of several, because this machine and @AliceLJY's both showed the same build varying by 40% between a cold and a warm page cache. `DEJA_CPUPROFILE` and `DEJA_MEMPROFILE` are on the real binary now, so the corpus being profiled can be a real one.

Index size did not move in #498 and nobody has looked at it. 60.3 MB for a 3.5 GB corpus is 1.7%, which is already good, but the same profile would say where the bytes go.

vshulcz:
Measured overnight. **The premise holds, one of the three experiments is answered negatively, and the framing needs to change: this is a CI win, not a laptop win.**

## First, a correction to something I wrote in #505

I concluded there that the build is I/O bound, from "11.6 s full, 5.6 s with the opencode store hidden". That test never varied the core count, and it was wrong:

| | 1 core | 8 cores | speedup |
|---|---|---|---|
| full rebuild | 17.2 s | 11.6 s | 1.5× |
| without the opencode store | **12.9 s** | **6.3 s** | **2.0×** |
| ⇒ opencode | 4.3 s | 5.3 s | — |

The non-opencode half halves with cores — that is ~6.6 s of parallelisable CPU work. What is actually serial is opencode: an external single-threaded `sqlite3` scan, ~5 s, untouchable from Go.

## GC tuning: no effect, no code needed to find out

| GOGC | run 1 | run 2 |
|---|---|---|
[...]

PR #640, opened 2026-07-31
perf(index): skip the CJK scan on bodies that contain no CJK

AliceLJY:
Filed alongside an issue with the full measurement; this is the fix half.

## What

`indexKeys` (`internal/index/retrieval.go`) calls `cjkBigrams` on the whole message body with no CJK check in front of it. For a transcript with no CJK rune in it — diffs, paths, stack traces, English prose — every byte is decoded to a rune and put through `isCJK`, which is four `unicode.Is` range searches, to build an empty slice. A CJK rune is never a single UTF-8 byte, so a byte scan settles it before any decoding starts.

## Measurement

At `17864d9`, synthetic 5000-session corpora (Chinese and English, index-aligned vocabularies so they differ in wording and nothing structural), interleaved runs, min of 7, `GOMAXPROCS=2` per the method in #499:

| all-ASCII corpus | build |
|---|---|
| stock | 2.57 s |
| this change | 2.37 s |
| bigram path stubbed out entirely | 2.34 s |

So it recovers essentially all of it. On the all-CJK corpus the change is +0.4%, inside the noise.

CPU profile of the ASCII build, `DEJA_CPUPROFILE`:

| | `cjkBigrams` | `isCJK` | `unicode.Is` | `indexKeys` total |
|---|---|---|---|---|
| stock | 5.17% | 4.39% | 3.88% | 14.99% |
| this change | absent | absent | absent | 7.74% |

Reverse check on the CJK corpus: `isCJK` costs 0.85% there. Nothing terminates early in either case — the difference is per rune. A Han rune matches on the first of the four checks and stays inside the current run; an ASCII rune fails all four and then calls `flush()`. Every character of an English transcript takes the expensive branch, which is why the corpus with no CJK in it is the one that pays.

The three-arm design also rules out binary layout as the cause: this binary keeps the entire implementation and adds ~15 lines, yet lands with the stub rather than with stock.

## Equivalence

Provable rather than measured. For all-ASCII input the original returns nil (no rune passes `isCJK`, `out` stays nil) and so does the fast path, including the empty string and the nil-versus-empty-slice distinction. Any input with a byte >= `utf8.RuneSelf` falls through to the untouched original loop, so CJK, mixed-script and invalid-UTF-8 behaviour is unchanged by construction.

## Tests

`TestCJKBigramsOnlyFiresOnCJK` is a guard on the scan, not on the output — the output table passed before this change too, since the answer was always nil. What it catches is a scan that decides "ASCII" too eagerly: that returns nil for the CJK rows and corrupts an index. The mixed (`abc装订def`), Cyrillic-adjacent (`очередь装订`) and invalid-UTF-8 rows are the ones that would fail.

`TestCJKBigramsDoesNotAllocateOnNonCJK` covers what a correctness table cannot see: that the work is skipped, not just the result empty.

Benchmarks record the shape — 3777 ns/op, 0 allocs on a 10 KB ASCII body against 50107 ns/op, 1409 allocs on a CJK one.

`go test ./... -race -count=1` green (14 packages), `gofmt` and `go vet` clean.

Everything above is reproducible from a synthetic corpus; the generator, the runner, the profile script and the recorded outputs are in the companion issue.

### DIFF
--- a/internal/index/cjk.go
+++ b/internal/index/cjk.go
@@ -2,6 +2,7 @@ package index
 
 import (
 	"unicode"
+	"unicode/utf8"
 )
 
 // CJK text carries no spaces, so the base tokenizer collapses whole phrases
@@ -25,6 +26,21 @@ func isCJK(r rune) bool {
 // to 系, which is a content character in 系統 and 關係, so a filter applied
 // after folding could not tell the two apart.
 func cjkBigrams(s string) []string {
+	// indexKeys hands this the whole message body, so a transcript with no CJK in
+	// it would otherwise decode every rune and run all four unicode.Is searches
+	// to build nothing. A CJK rune is never a single UTF-8 byte, so one byte scan
+	// answers the question before any decoding starts.
+	ascii := true
+	for i := 0; i < len(s); i++ {
+		if s[i] >= utf8.RuneSelf {
+			ascii = false
+			break
+		}
+	}
+	if ascii {
+		return nil
+	}
+
 	var out []string
 	seen := map[string]bool{}
 	var run []rune
--- a/internal/index/cjk_ascii_test.go
+++ b/internal/index/cjk_ascii_test.go
@@ -1,0 +1,80 @@
+package index
+
+import (
+	"strings"
+	"testing"
+)
+
+// indexKeys calls cjkBigrams on every message body it sees, most of which are
+// diffs, paths, stack traces and English prose with no CJK rune anywhere. This
+// table pins the contract that lets those exit early: what comes back for a
+// string with no CJK run is nothing, whatever else the bytes happen to be.
+//
+// It is also the guard on the byte scan itself. A scan that decided "ASCII" too
+// eagerly would return nil for the CJK rows below; one that never fired would
+// leave the ASCII rows unchanged and quietly cost what it always cost. Only the
+// first failure mode can corrupt an index, and these rows catch it.
+func TestCJKBigramsOnlyFiresOnCJK(t *testing.T) {
+	cases := []struct {
+		name string
+		in   string
+		want []string
+	}{
+		{"empty", "", nil},
+		{"ascii prose", "the scheduler times out under load", nil},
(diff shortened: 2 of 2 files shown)


vshulcz:
Reviewed and verified independently. Merging.

**Equivalence — checked, not assumed.** Ran your implementation against the pre-change one on 24 hand-picked edge cases plus **20,000 random inputs** mixing ASCII, Han, kana, Hangul, Cyrillic, emoji, zero-width, fullwidth forms and spliced raw bytes including invalid UTF-8. Zero mismatches.

Then the thing that actually matters — the index itself. Built two indexes from a frozen corpus, one from `main` and one from `main` + your commit alone, and compared them by key rather than by bytes:

```
main 8060 keys · main+patch 8060 keys
only in main: 0 · only in patched: 0 · differing posting counts: 0
```

Worth recording how that check went wrong first: comparing your branch head against `main` showed 654 extra keys and 1267 differing counts, which looked alarming for about a minute. The cause was that your branch predates eight of today's merges, one of which removes duplicated Codex turns. Cherry-picking your commit onto current `main` is the only version of that comparison that means anything.

**Speed — reproduced, and larger than you claimed at the function level.** `BenchmarkCJKBigramsASCII`, 200 iterations, 3 runs:

```
main          118–158 µs/op
main+patch    5.5–5.8 µs/op     ~21× on the function
```

Your 8% figure is for the whole build, where this is one part of ingest, so the two are consistent.

**Mutations on your tests:** a scan that decides "ASCII" too eagerly fails, and one that never fires fails. The third I tried — `>=` to `>` at the boundary — survives, and I checked whether that is a gap in the tests: it is not. The two spellings differ only on inputs whose highest byte is exactly 0x80, a lone continuation byte decodes to RuneError, RuneError is not CJK, and the answer is empty either way. Semantically equivalent, not untested.

The three-arm design in the issue is what makes this reviewable — thank you for including the stubbed-out arm, since that is what rules out the "different binary layout" explanation.

PR #642, opened 2026-07-31
fix: three status messages report a state the tool is not in

### LINKED ISSUE #637: Three status messages report a state the tool is not in (zero-result count, deja last, doctor guidance)
Tested on `main` @ 17864d9, macOS arm64, index with 268 sessions across 9 harnesses.

Three places report a state that is not the state the tool is in. Grouping them because they share a shape — a count or a status is read from the nearest variable rather than from the thing being described — but they are independent fixes.

## 1. Ordinary zero-result searches print the index-corruption signature

```
$ deja "zzqqxx_nonexistent_token"
deja: no matches in 0 indexed sessions — try fewer words or --re (query "zzqqxx_nonexistent_token")

$ deja stats | head -1
268 sessions indexed · 1 questions asked more than once
```

`cmd/deja/main.go:456` passes `len(ss)`:

```go
if len(hits) == 0 {
    printNoMatches(os.Stderr, o.Query, len(ss))
}
```

On the index-backed path `ss` is not the corpus, so this prints `0` for a healthy index. The same run prints `266` or `268` on paths where sessions were loaded eagerly, so the number is not stable across queries either.

This one is worth more than its size because `internal/index/manifest.go:150-155` already reserves that exact sentence as the signature of a specific failure:

```go
// The manifest survives, so the index reads as built and
// up to date while every search answers "no matches in 0 indexed sessions"
// — the one failure a memory tool must not present as an empty result,
// because the reader concludes it is useless rather than broken.
```

`recordsIntact` guards against the cause, but the string it warns about is what every ordinary miss prints today, so the signature cannot be used to recognise the failure.

Suggested: report the manifest session count (what `deja stats` reports), or drop the count from the sentence.

## 2. `deja last` reports "nothing indexed" when filters matched nothing

```
$ deja last 3 --harness codex --role command
deja: no sessions indexed yet — run `deja index`, or `deja doctor` to see which agent stores were found
```

The index has 268 sessions and 82 of them are codex; what is absent is the `command` role for that harness. `cmd/deja/main.go:338-343` takes the empty-result branch regardless of whether filters were supplied, and the comment above it shows the message was written for the fresh-install case:

```go
// Printing nothing at all and exiting 0 leaves no way to tell whether the
// command worked, found nothing, or failed silently — which is what a
// fresh install sees.
if len(ss) == 0 {
```

The advice is actively wrong here: `deja index` will not change the result, and `deja doctor` reports the stores as found. Suggested: when any filter is set, say which filter emptied the result; keep the current text for the unfiltered case.

## 3. `doctor` reports guidance as written when only the file exists

```
$ deja doctor | grep '^  codex .*guidance'
  codex        not wired      guidance written     $HOME/.codex/config.toml
```

Nothing from deja has ever been installed on this machine — `deja install` was never run, and `grep -ci deja ~/.codex/config.toml ~/.codex/AGENTS.md` returns 0 for both. `cmd/deja/guidance.go:203-212` stats the path and stops:

```go
func guidanceStatus(harness string) string {
	path := guidancePath(harness)
	if path == "" {
		return "unsupported"
	}
	if _, err := os.Stat(path); err == nil {
		return "written"
	}
	return "missing"
}
```

So any user who has their own `AGENTS.md` is told deja's guidance is in place. Since re-running install rewrites "deja's skill or marked block", the marker is already the thing to look for. Suggested: report `written` only when that marker is present, and distinguish `absent` from `file-missing` if the difference is useful.

### Related, same cause, lower confidence

`doctor`'s hook rows label third-party files as deja's:

```
  codex-hook   untrusted   ~/.codex/hooks.json  (codex will not run it until you review it …)
  cursor       stale       ~/.cursor/hooks.json (no hook-context call — reinstall)
  kimi         stale       ~/.kimi-code/config.toml (no hook-prompt call — reinstall)
```

All three files exist and contain no deja entry; the Cursor one holds hooks belonging to four unrelated tools. "reinstall" and "codex will not run it" describe a deja hook that was never installed. I have not read that code path, so treating this as the same existence-vs-content check is a guess — but the observable output has the same shape as #3.

vshulcz:
Closes #637. All three reported by @hieilookie, all three reproduce here.

**1. Ordinary misses printed the corruption signature.**

```
before  deja: no matches in 0 indexed sessions — …      (store has 1153)
after   deja: no matches in 1153 indexed sessions — …
```

`len(ss)` is the loaded set, which on the index-backed path is empty by definition when nothing matched. `internal/index/manifest.go` reserves that exact sentence as the signature of a corrupt store — on the grounds that a reader told the index holds zero sessions concludes the tool is broken rather than that their query missed — so the signature was useless while every ordinary miss printed it. Now it reports the manifest count, and says nothing rather than a wrong number when there is no index.

**2. `deja last` blamed the index for a filter.**

```
before  deja: no sessions indexed yet — run `deja index`, or `deja doctor` …
after   deja: no sessions match harness "codex" and role "command"
```

The advice was not merely unhelpful, it was wrong: indexing changes nothing and doctor reports the stores as found. The original text stays for the unfiltered case it was written for.

**3. `doctor` read guidance from a file's existence.**

```
before  codex  guidance written   (on a machine where install had never run)
after   codex  guidance absent
```

Anyone with their own `AGENTS.md` was told deja's guidance was in place. It now looks for the marked block install writes, and `absent` is distinguished from `missing` because only the first means someone else owns the file.

One thing the issue could not have known: five harnesses get a **skill file** deja writes whole, which has no marker in it. Checking for the marker there would have reported guidance missing on a machine where it is installed and working — so `guidanceOwnsWholeFile` splits the two cases, and it now names the same set the install path branches on rather than repeating the list.

The related note about hook rows labelling third-party files is left open — it is a different code path and deserves its own measurement.

Four mutations, each fails the suite: restoring the zero count, blaming the index again, reading guidance from existence, and applying the marker check to a skill file.

### DIFF
--- a/cmd/deja/guidance.go
+++ b/cmd/deja/guidance.go
@@ -70,7 +70,7 @@ func opencodeConfigHome() string {
 }
 
 func guidanceText(harness string) string {
-	if harness == "claude-code" || harness == "claude" || harness == "antigravity" || harness == "copilot" || harness == "pi" {
+	if guidanceOwnsWholeFile(harness) {
 		body := guidanceBody
 		if harness == "copilot" {
 			body = "deja does not index Copilot history. It is a consumer: use the deja MCP tools to search memory from the other harnesses.\n\n" + guidanceBody
@@ -200,12 +200,6 @@ func guidanceHarness(harness string) string {
 	}
 }
 
-// guidanceStatus reports whether deja's guidance is in that file, not whether
-// the file exists. Anyone who keeps their own AGENTS.md was told deja had
-// written guidance there when nothing from deja had ever been installed
-// (#637). Install writes a marked block, so the marker is the thing to look
-// for — and "the file is there but ours is not" is a different answer from
-// "there is no file", because only the first means someone else owns it.
 // guidanceOwnsWholeFile reports whether install writes the whole file rather
 // than a marked block inside someone else's. A skill file lives in deja's own
 // directory and has no marker in it; AGENTS.md and its siblings belong to the
@@ -226,6 +220,10 @@ func guidanceOwnsWholeFile(harness string) bool {
 // answer from "there is no file", because only the first means someone else
 // owns it. A skill file deja writes whole has no marker and does not need one.
 func guidanceStatus(harness string) string {
+	// One form throughout: the path lookup took the raw name and the
+	// whole-file check took the normalised one, so the function contradicted
+	// itself about which it accepts.
+	harness = guidanceHarness(harness)
 	path := guidancePath(harness)
 	if path == "" {
 		return "unsupported"
@@ -234,7 +232,13 @@ func guidanceStatus(harness string) string {
 	if err != nil {
 		return "missing"
 	}
-	if guidanceOwnsWholeFile(guidanceHarness(harness)) || strings.Contains(string(b), guidanceStart) {
+	// An empty file is an interrupted write, not guidance — the one shape
+	// where "the file exists" still fails to mean "we wrote it" even in a
+	// directory deja owns.
+	if len(strings.TrimSpace(string(b))) == 0 {
+		return "absent"
+	}
+	if guidanceOwnsWholeFile(harness) || strings.Contains(string(b), guidanceStart) {
 		return "written"
 	}
 	return "absent"
--- a/cmd/deja/guidance.go
+++ b/cmd/deja/guidance.go
@@ -106,7 +106,7 @@ func installGuidance(harness string, uninstall bool) (installResult, error) {
 		return installResult{}, err
 	}
 	var next []byte
-	if harness == "claude-code" || harness == "claude" || harness == "antigravity" || harness == "copilot" || harness == "pi" {
+	if guidanceOwnsWholeFile(harness) {
 		if uninstall {
 			if len(old) == 0 {
 				return installResult{Path: path, Action: "unchanged"}, nil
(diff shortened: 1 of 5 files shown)


vshulcz:
Independent review found ten issues. Eight fixed, two answered.

**Fixed**

1. A doc paragraph landed on the wrong function — a botched edit of mine, and the first thing in the diff.
2. `--since` was echoed as `168h0m0s` where the reader typed `7d`. The whole point of the message is to name the filter they set, so it now carries the raw flag.
3. A negative `--since` named a filter that was never applied: `activeFilters` tested `!= 0` while `filterRecentSources` tests `> 0`, and `parseDur` accepts negatives. Same predicate in both places now, so the empty-store advice survives where it is the right answer.
7. The third copy of the whole-file harness list — the one in `guidanceText` that decides whether a file gets skill frontmatter or the marker pair — now calls the helper. That was the drift the helper existed to stop, and I left the copy that mattered. `guidanceStatus` also normalises once at the top rather than passing the raw name to one lookup and the normalised name to another.
9. An empty guidance file reports `absent`. It was the one shape where "the file exists" still failed to mean "we wrote it", even in a directory deja owns.
4, 5, 6. Tests for the three untested filters, for `SessionCount` counting sessions rather than files, and for every harness in the whole-file set. Each is now caught by mutation: deleting the `project` branch, returning `len(m.Files)`, and dropping `"pi"` all fail the suite.
10. The comment in `manifest.go` describing that sentence as the corruption signature is updated, since it no longer is.

**8 — you are right and I have said so in the code.** `TestGuidanceStatusForAFileDejaOwnsWhole` passes against the old behaviour; it is a guard for the carve-out, not a test of the fix. The comment now says that, so nobody reads it as proof of something it does not show.

**5's fixture needed rebuilding, and it taught me something.** One file holding three sessions does not exist for Claude: the parser overwrites the id with the last `sessionId` it sees, so a multi-session file collapses to one — the same shape as the Codex bug in #635, latent because Claude Code writes one session per file. The test now uses two files with sessions and two without, so the counts cannot be confused.

Thank you for answering the two framing questions with evidence, and for **10** in particular: confirming the signature was never the detector — `recordsIntact` is — and that both corrupt shapes are caught before any message, is what makes this a removal of a false positive rather than a hole.

PR #644, opened 2026-07-31
feat(stats): say what restore could hand back, before anyone needs it

### LINKED ISSUE #577: Tell people restore exists before they need it, not after
`deja restore` (#537) is the panic-moment feature: an agent replaced a function with something worse, the work was not committed, git has nothing. It matters entirely at one moment, and the rest of the time it is invisible.

The problem is timing. Nobody reads a command list while panicking, and nobody remembers a command they read a month earlier. The user has to learn it exists **while calm**, in a line that is about their own data:

```
3,836 spans your agents replaced are recorded
if something got clobbered:  deja restore <file>
```

The number is theirs, computed at index time, and it doubles as proof that deja keeps the part of a transcript everything else throws away.

Where it goes is the design question — the first screen after install is already carrying #574 and #576, and a fourth line makes it a brochure rather than a demonstration. Possibly this belongs in `deja stats`, or as a one-time line on the first run only.

vshulcz:
Closes #577.

```
Recover   16201 spans your agents replaced, across 1621 files
          if something got clobbered: deja restore <file>
```

`deja restore` matters entirely at one moment — an agent replaced a function with something worse and the work was not committed — and nobody reads a command list while panicking. The number is theirs, and it doubles as proof that deja keeps the part of a transcript everything else discards: 5.6 MB of exact replaced bytes on this store.

**Placement, which the issue left open.** Not the first screen: it already carries the asked line, the friction line and the suggestion, and a fourth would make it a brochure. `deja stats` is a page someone opens deliberately, which also affords the ~180 ms this costs.

**It needed its own pass, and finding out why was the useful part.** I first counted spans inside `stats.Build`, from the sessions it is already given — and got zero. Replaced spans are deliberately excluded from ordinary retrieval (`retrieval.go`: "a replaced span is the file's old contents, not a statement about anything"), so they never reach a caller loading sessions the normal way. That is correct behaviour and the reason `SpanInventory` exists; there is a test asserting the ordinary path still cannot see them, so the day that changes, this stops being necessary and says so.

Silent when there is nothing to recover — a zero here is an invitation to a command that would answer nothing.

Three mutations, each fails the suite: dropping the line, printing it at zero, and counting whole spans instead of their paths (which would report 16201 files where there are 1621).


vshulcz:
Review found the number was a slight overclaim, and chasing it turned up a real bug in `restore` itself. Both fixed here; #644 now closes #647 as well.

**The overclaim, and why it is not fixed by softening the wording.** The reviewer measured that for **41 of 1,621 paths** the session holding the span never appears in a search for that file — **76 of 16,201 spans** unreachable. Reproduced exactly:

```
$ deja restore telebot.go
no replaced spans recorded for "telebot.go"        # index holds 3
```

`findRestoreSpans` went through ranked retrieval, which samples postings per session before BM25 ranks them, so in a large session the specific edit record never reaches the candidate set. Ranking is the wrong instrument here: everywhere else in deja the reader wants the most relevant session, but restore has an exact key, the person knows what they lost, and a near-miss is worth nothing. It scans now — ~190 ms for a command run once, in a panic.

```
after   3 replaced spans recorded for telebot.go
        1  Jun 10 17:43  opencode ses_2364a0fc  16 B replaced
        …
```

`package.json` went from 1 span to 8. The stats number is now a promise rather than an estimate, which is what makes the line worth printing at all.

**Where my test stops, stated plainly.** `TestRestoreFindsSpansInACrowdedSession` does not fail against the ranked implementation. I tried to reproduce the miss synthetically — 400 edit records, then 2,000 and 9,000 speech messages carrying the needle word — and the ranked path found the span every time. The trigger is something the fixtures do not capture. The evidence is the live store: `index.Search("telebot.go", Role:"edit")` returns 0 sessions where the log holds 3 spans. The test comment says this rather than implying coverage it does not have.

**Finding 2 — the `i > 0` guard is untestable, and that is the honest answer.** Both span extractors drop an edit with no path, so no such record can be written; measured on the live store, 0 edit records with an empty first line and 0 with no newline. My first attempt at a test looked like it covered the guard and did not. It now pins the invariant the guard rests on and says why, so if a parser ever starts recording pathless spans, it fails at that moment.

**Finding 5 — the doc example.** `spans` and `span_files` added to `docs/json-output.md` with a sentence on what they mean and when they are omitted.

Thank you for the correction on my mutation claims — four of five were caught, not the three I said. And for flagging that your probe re-indexed my store: I checked, the counts are identical and `cooccur.gob` rebuilds on the next full build.

Issue #532, opened 2026-07-29
Check a plan against history before the agent executes it


vshulcz:
Recall today is reactive: it answers a question, or it fires once at session start. But the moment where memory is worth the most is not when someone asks — it is when an agent has just produced a plan and is about to spend twenty minutes on it.

Memory as a guard, not a lookup table. Nothing in this category does prospective checking.

## Shape

The agent proposes six steps. deja reads them and speaks only if it has something:

```
step 3 — "switch the pool to transaction mode"
   contradicts a decision from Jul 12 (claude 8a41c02): transaction mode
   cannot hold prepared statements; pgx was pinned instead
step 5 — attempted in 2 earlier sessions, abandoned both times (#527)
```

Silence on the other four steps. Same discipline as the prompt hook: a guard that comments on everything is a guard nobody reads.

## Why this is ours

It needs the decision itself and the rejected alternative, not a summary of the topic — the same asset #526 is built on. And it needs to run in-line, in front of an agent, with no API call: an LLM-based checker adds seconds and a key to every plan, which is exactly the cost that keeps memory out of the loop today.

## Where it plugs in

An MCP tool the agent is instructed to call before executing a multi-step plan, plus a `deja check -` that reads a plan on stdin so it works in a hook or a CI step.

## The hard part is not the matching

It is **the interruption budget**. A false contradiction is expensive twice over: it wastes the run and it teaches the agent to ignore the guard. This has to be tuned for precision at the cost of recall — nine missed contradictions and zero false ones is the right trade, and the opposite ships a tool people disable in a week.

The second hard part: a plan step is a sentence of intent, and our matching is lexical. "switch the pool to transaction mode" and the session that decided against it may share almost no words. This is where the relevance tier gets a real test, and it may simply not be strong enough.

## Kill condition

Take real plans from history — the transcripts are full of them — and measure how often a step that a past decision contradicted is actually caught, and how often we cry wolf. If precision is not overwhelming, do not ship it.

Sora-bluesky:
I implemented the shape this issue sketches and measured it before opening a PR. Sharing the numbers because they answer the kill condition, and the kill condition says no.

[...]

PR #1125, opened 2026-08-08
check: an explicit, off-by-default plan check, plus the harness to re-measure it

Sora-bluesky:
Adds `deja check -` and `deja hook-plan` as explicit, off-by-default subcommands, plus `scripts/planbench` so the measurement can be repeated on someone else's plans. No install wiring, no hook registration: nothing here fires unless it is called by name.

This is the implementation from #532, opened at your request after the measurement came back short of the kill condition. The numbers are below and they are not overwhelming. I would rather you have the harness and decide than have my summary of it.

## What it does

`deja check -` reads a plan on stdin and prints one line per finding, or nothing. A finding states a co-occurrence and stops there:

```
wall hit in 23 sessions since 2026-07-20: "…: No module named pytest"
```

No contradiction claim, no "you already tried this". #541 is why. The judgement about whether the wall matters to the step is left to whoever reads it.

`deja hook-plan` is the same core behind a PreToolUse/ExitPlanMode shape. It shares one function with `check`, so the numbers below describe the hook path too.

Both honour the auto-activation policy and the existing `.hookseen` dedupe, and neither builds an index: a missing or stale one is a silent miss, because plan submission should not block on indexing.

## The measurement

30 real plans out of my own plan-mode history, against a personal index carrying 19 recurring walls, each hit in 3 or more sessions.

| gate | fires | precision (hand-labeled) |
|---|---|---|
| 2 informative terms, idf >= 2.0 (auto-recall's floor) | 0/30 | — |
| identifier-shaped terms exempt from the floor | 10/30 | ~0.69 |
| two-tier floor: identifier >= 1.0, plain >= 2.0 | 6/30 | 0.58–0.83 |

Only the last row reproduces at this branch. The first two came from earlier gates the code no longer contains.

Read 6/30 with the caveat attached: those six plans are three documents and their revisions, and all 12 findings carry **two** distinct wall strings. So the twelve findings are three documents against two walls, and that is the real size of the sample, not twelve independent judgements. `planbench` prints the distinct-wall count for exactly this reason.

## Two things the measurement taught me

**Wall vocabulary is common by construction.** Walls come from repeated work, so the terms that could anchor one are frequent in the corpus that produced them: the anchors that fired sit at idf 1.08 and 1.22, while ordinary words inside the same wall strings sit at 0.18–0.38. The informative floor built for prompts excludes every one of them. That is why row 1 is 0/30 rather than "a few misses".

The second floor at 1.0 sits in a measured gap (identifier-shaped false anchors topped out at 0.70, true ones started at 1.08), but I picked it on the same 30 plans the precision column reports, so read 0.83 as the optimistic in-sample end. The gap is not a law either: "preview" (1.26) and "denied" (1.85) are equally ordinary words that clear the same floor and simply did not appear in these plans.

**Requiring a command-role join leaves it silent.** 84 of the 110 wall-carrier sessions hold no command records at all, and at most 1 wall of 19 could ever join. So the join is an optional strengthening clause on a finding, never a precondition.

## The harness

```
DEJA_INDEX_DIR=~/.cache/deja/index.db go run ./scripts/planbench -plans ./plans
```

It runs each file in the directory through the built binary and prints fire/silent per plan, then totals: plans, fires, findings, and distinct walls cited. It counts; it does not judge. Precision is a human label, and the tool produces the denominator, not the verdict.

It refuses any output line that is not a finding. That guard exists because an older `deja` on PATH silently answered `check -` as a search (an unknown first word falls through) and reported 47 findings for a one-line plan. A precision harness that measures `deja search` and calls it a fire rate is worse than none.

Point it at a directory of plans only; every regular file counts as one, so a stray `results.tsv` inflates the denominator. Each file is listed, so it is visible, but the headline number moves.

## What I could not verify

`go test -race` does not run here: no C toolchain on this machine, and `-race` requires cgo. CI will be the first place it runs. Everything else in CONTRIBUTING passed: `gofmt -s`, `go vet ./...`, `go build ./cmd/deja`, and the package tests.

`scripts/planbench` ships with a test, which cuts against the note in ci.yml that harnesses under `scripts/` have none by design. I kept it because it pins the search-fallthrough guard above, and because `scripts/` is excluded from the coverage profile so it costs nothing there. If you would rather the directory stay uniform, deleting `main_test.go` loses nothing else.

## Kill condition

0.83 at best, in-sample, on a sample that is really three documents. I did not think that cleared the bar in #532, and I still do not. What changed is that you asked for the code and the harness anyway, so here they are, off by default and easy to delete if the numbers do not hold up on your plans.

### DIFF
--- a/cmd/deja/check.go
+++ b/cmd/deja/check.go
@@ -1,0 +1,25 @@
+package main
+
+import (
+	"fmt"
+	"io"
+)
+
+// runCheck is the human-facing form of the plan hook. It deliberately calls
+// planFindings directly so measured precision reflects the injected path.
+// The empty session id skips the .hookseen dedupe on purpose: dedupe governs
+// when a finding is delivered twice in one session, not whether it matches,
+// and a measurement run has to see every match.
+func runCheck(dir string, args []string, stdin io.Reader, stdout io.Writer) error {
+	if len(args) != 1 || args[0] != "-" {
+		return fmt.Errorf("check needs '-' to read a plan from stdin")
+	}
+	plan, err := io.ReadAll(stdin)
+	if err != nil {
+		return fmt.Errorf("check: read plan: %w", err)
+	}
+	for _, finding := range planFindings(dir, string(plan), "") {
+		fmt.Fprintln(stdout, finding)
+	}
+	return nil
+}
--- a/cmd/deja/check_test.go
+++ b/cmd/deja/check_test.go
@@ -1,0 +1,93 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestRunCheckUsesTheHookPlanCore(t *testing.T) {
+	withPlanLookup(t, samplePlanMatches(`C:\work\repo`))
+	dir := filepath.Join(t.TempDir(), "index.db")
+	plan := "1. Run timeout around deploy_probe"
+
+	var hookOut bytes.Buffer
+	hookInput := `{
+		"hook_event_name":"PreToolUse",
+		"tool_name":"ExitPlanMode",
+		"tool_input":{"plan":"1. Run timeout around deploy_probe"},
+		"session_id":""
+	}`
+	if err := runHookPlan(dir, strings.NewReader(hookInput), &hookOut); err != nil {
+		t.Fatal(err)
+	}
+
+	var resp struct {
+		HookSpecificOutput struct {
+			AdditionalContext string `json:"additionalContext"`
+		} `json:"hookSpecificOutput"`
+	}
+	if err := json.Unmarshal(hookOut.Bytes(), &resp); err != nil {
+		t.Fatalf("hook JSON = %q: %v", hookOut.String(), err)
+	}
+	hookPlain := strings.TrimPrefix(resp.HookSpecificOutput.AdditionalContext, recallFrameHeader)
+	hookPlain = strings.TrimSuffix(hookPlain, recallFrameFooter)
+
+	var checkOut bytes.Buffer
+	if err := runCheck(dir, []string{"-"}, strings.NewReader(plan), &checkOut); err != nil {
+		t.Fatal(err)
+	}
+	checkPlain := strings.TrimSuffix(checkOut.String(), "\n")
+	if checkPlain != hookPlain {
+		t.Fatalf("check and hook differ:\ncheck: %q\nhook:  %q", checkPlain, hookPlain)
+	}
+	if strings.Contains(checkOut.String(), "<deja-recall>") {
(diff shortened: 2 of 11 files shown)


vshulcz:
Thanks for this, and especially for shipping the harness with the raw numbers instead of a summary — that's exactly the right way to hand over a marginal-precision call.

I re-read the whole thing. The plan path honours the auto-activation policy, frames findings as untrusted like the other recall surfaces, caps the payload at 2048, only answers for ExitPlanMode, and stays a silent miss when the index is missing/stale/damaged so plan submission never blocks on indexing. Off-by-default with no install wiring or hook registration means it costs nothing until someone calls it by name. `check` and `hook-plan` sharing one core keeps the measurement honest. Tests cover the parity, the silent miss, and the dash requirement.

The precision isn't overwhelming, but the design makes that the caller's decision rather than forcing it on anyone — which is the right shape for something #532's data left undecided. Merging as opt-in. Thanks again.

Sora-bluesky:
Thanks for merging it.

Two operational notes for whoever points this at a bigger corpus.

`planbench` prints distinct walls next to findings, and that is the number that says how large the sample really is. My twelve findings were three documents against two walls, so the precision column rests on far less than twelve independent judgements. If distinct walls stays flat while findings climbs, the gate is narrow rather than precise.

The floors are in-sample too. The 1.0 identifier floor sits in a gap I measured on the same 30 plans, so it is a boundary that held once, not a constant. `planIdentifierIDFFloor` is one line, and re-running `planbench` over the same plans shows what moving it costs, so it should be cheap to revisit if it misfires on vocabulary that is not mine.

Happy to follow up on #532 if the numbers move either way.

vshulcz:
Thanks — the distinct-wall count is exactly why planbench prints it, with that caveat in the summary block, so we're reading it the same way. One clarification on the identifier floor: it isn't only in-sample. `planIdentifierIDFFloor` comes from a census of 2,886 sessions — identifier-shaped false anchors topped out at idf 0.70, true ones started at 1.09, and 1.0 sits in that gap. The 30-plan set is what validates the gate's precision, not what sets the floor. Agreed it's a one-line const, though, and planbench makes re-measuring it cheap if it misfires on other vocabulary — glad to revisit on #532 if the numbers move.

Sora-bluesky:
You're right, and the code says it plainly: the comment on `planIdentifierIDFFloor` cites the 2,886-session census — identifier-shaped false anchors topping out at 0.70, true ones starting at 1.09. I wrote that I picked the floor on the same 30 plans, and that was wrong. The census sets the floor; the 30 plans measure what the gate does with it.

I also wrote 1.08 where the census says 1.09. That figure belongs to the plan corpus, where the anchors that fired sit at 1.08 and 1.22, and I carried it into the census sentence — which is how the two measurements collapsed into one.

Leaving the PR description as written rather than editing it, so your correction still reads against the thing it corrects.

PR #1128, opened 2026-08-08
fix: index hangs, import hiding local sessions, semantic policy leak, and audit-wave fixes

vshulcz:
Audit wave over the index, sources, promote/remember and MCP surfaces. Each item was reproduced on a hermetic index, fixed, and covered by a test with a mutation check.

Two are real hangs/data-loss:
- Index froze forever on a FIFO or socket in a scanned store — `Open` blocks on a pipe with no writer. Discovery now keeps only regular files.
- Import before the first `index` snapshotted the local stores into the manifest, so the next index found it "fresh" and skipped the reader's own sessions until a full rebuild. The initial manifest now starts with an empty file set.

One is a trust-policy leak:
- The semantic tier reached the whole sidecar past the policy scoping the lexical hits already had, so an imported peer's withheld content came back through the embedding fallback. Scoped in search, MCP recall and recall_context.

Rest are correctness/UX: claude timestamps now accept stringified/fractional epochs and zone-less RFC3339 (were losing the date); identical re-promote no longer inflates a note's recall weight; promote/remember say when they wrote into a tombstoned note; the shared-id hint names `harness:id` instead of a `--harness` flag those commands reject; `--` ends options in search; MCP remember accepts tags; search says when results were capped.

### DIFF
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -832,6 +832,14 @@ func runSearch(dir string, args []string, sourceInstance string) error {
 	}
 	var semantic bool
 	hits, semantic = maybeSemantic(dir, hits, o, os.Stderr)
+	if semantic {
+		// The lexical hits were scoped by the trust policy above, but the
+		// semantic tier reaches the whole sidecar and brings back sessions of
+		// its own — an imported peer's content the policy withholds from every
+		// other read path leaked straight through the embedding fallback. Scope
+		// the semantic hits the same way before they are shown.
+		hits, _ = policyFilterHitsCounted(policy.ActivationSearch, hits)
+	}
 	o.Semantic = semantic
 	// Policy scoping, reranking and the semantic tier all run after the cap, so
 	// the pre-cap count can no longer describe what is being returned. When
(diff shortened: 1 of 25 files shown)


vshulcz:
Deep review before merge — all reproduced on a hermetic index, each with a test + mutation:

- FIFO/socket in a scanned store froze index forever (blocking Open) — discovery keeps regular files only; verified the same suite passes on the three OS test jobs.
- Import before first index snapshotted local stores into the manifest and hid the reader's own sessions — initial manifest now starts empty.
- Semantic tier reached past policy scoping in search / MCP recall / recall_context — scoped all three sites; confirmed no 4th (hook-prompt is not one). Note: the extra filter discards its hidden-count, so the "N withheld" wording can undercount — cosmetic, not a leak.
- Timestamps, promote dedup, tombstone warnings, shared-id hint, `--` end-of-options, MCP tags, cap notice — all behavioural, covered.

build/vet/gofmt clean, coverage cmd/deja 90.6% · sources 88.3% · index 90.2%. Merging.

PR #1158, opened 2026-08-10
test(index): isolate the notes store from the developer's real one

### LINKED ISSUE #1141: internal/index tests write into the developer's real notes store, then fail on what they wrote
`go test ./internal/index/` appends fixture notes to the developer's own notes store, then fails 11 tests because it reads them back. My store is 25 lines and 14 of them were written by the suite.

`cmd/deja` already guards against this. `internal/index` does not.

## What happens

`sources.NotesFile()` (internal/sources/notes.go:38) resolves in this order:

```
DEJA_NOTES_FILE  ->  XDG_DATA_HOME  ->  os.UserConfigDir() on Windows  ->  Home()
```

`internal/index`'s TestMain (internal/index/grok_test.go:13) neutralizes HOME, USERPROFILE, XDG_CONFIG_HOME and the DEJA_*_ROOT set. It touches none of the first three entries of that list, so on Windows the resolution reaches `%APPDATA%\deja\notes.jsonl` — the real one. Redirecting HOME cannot get to it.

Two consequences, in order.

**It writes.** internal/index/note_state_title_test.go:21-22 sets HOME and XDG_CONFIG_HOME, then line 27 calls `sources.AppendPromotedSourced("work/api", "what pool size should the pool use", ...)`. That append lands in the real store. Mine now holds 14 of those rows beside the 11 notes I wrote myself.

**Then it reads.** Every test that builds an index picks the store up as a `deja` source, so fixtures see sessions they never created. At ebf727a, `go test ./internal/index/` gives 11 failures here, and the arithmetic is visible in them:

```
import_ord_test.go:55:  got 304 distinct ordinals for 300 sessions
index_test.go:457:      full ingest touched 31 files, want 30
deepverify_test.go:54:  counts = files=2 sessions=5 sampled=2, want 1/1/1
```

The store arrives as four `deja` sessions and one extra file, which is where the +4 and the +1 come from.

## Attribution

Redirecting only APPDATA, changing nothing else, takes it from 11 failures to 1. I do not have a cause for that last one. A probe showed `exportedProject()` (internal/index/sync_peer_test.go:181) reading `paths[0]` of a batch whose first record was a leaked note, which would explain it, but the same test also fails without any leak present — see the last section — so I cannot tell those two apart and am not claiming either.

## Reproducing it without Windows

The Windows branch is not required to hit this. XDG_DATA_HOME is checked before it and TestMain does not clear that either, so one file should be enough anywhere:

```sh
mkdir -p /tmp/xdg/deja
echo '{"ts":"2026-08-01T10:00:00Z","project":"private-client","text":"a note that must never reach a test fixture"}' > /tmp/xdg/deja/notes.jsonl
XDG_DATA_HOME=/tmp/xdg go test ./internal/index/
```

I ran that on Windows with APPDATA already pointed at an empty directory, so XDG_DATA_HOME was the only way in, and it gave 13 failures. I have not run it on Linux or macOS — the resolution order says it should behave the same there, but that part is reasoning, not a measurement.

## Why CI is green

A runner has no notes store and exports no XDG_DATA_HOME, so the fallback lands on a path that does not exist. The windows-latest job shows the code works on Windows. It cannot show anything about a Windows machine that uses deja, and that is the machine this breaks.

## The fix

cmd/deja/testmain_test.go:9 already points APPDATA at a temp path and blanks XDG_DATA_HOME and DEJA_NOTES_FILE, with a comment about the suite going red on a developer's machine only. The same three keys in `internal/index`'s TestMain close this. I ran it here: none of the eleven come back, and the store stops growing — it held at 25 lines across a full run that had been adding two rows every time.

A regression test is worth more than the keys, because this drifts back quietly. Pinning `sources.NotesFile()` under the temp root fails without them here and names the real path in the failure message. By the resolution order it should fail on windows-latest as well, since APPDATA there points at the runner profile rather than the temp root — which would give that job a way to see this. I have not run CI to confirm that part.

The two maps have drifted further than those three: cmd/deja also blanks CLAUDE_CONFIG_DIR, CODEX_HOME, GEMINI_CLI_HOME, CURSOR_CONFIG_DIR, AIDER_CHAT_HISTORY_FILE, DEJA_SOURCE_INSTANCE and DEJA_WARMUP_SENTINEL. I verified the notes ones only; whether the DEJA_*_ROOT entries already cover the rest, I have not checked.

One thing the keys do not settle. `TestImportHonorsExcludeList` and `TestExcludeAppliesToAlreadyImportedSessions` still fail here, zero to two of them per full run on identical input, while passing five out of five in isolation. Neither was among the eleven, and I have not tied them to any of this — I mention it only so a green suite is not the bar you measure the fix against.

Happy to send the PR.

vshulcz:
Thanks @Sora-bluesky for the diagnosis — it was exactly right.

internal/index's TestMain neutralized HOME/XDG_CONFIG_HOME and the DEJA_*_ROOT set, but not XDG_DATA_HOME, APPDATA or DEJA_NOTES_FILE — the three paths sources.NotesFile() resolves through before HOME. So the suite appended promoted-note fixtures to the developer's real store and then failed eleven tests reading them back as leaked `deja` sessions.

Blanks the same keys cmd/deja's TestMain already does (closing the drift) and adds a regression test that pins NotesFile() under the temp root, so this fails loudly instead of quietly the next time the maps drift apart. Confirmed with your repro: `XDG_DATA_HOME=/tmp/xdg go test ./internal/index/` now passes and leaves the notes file untouched.

Closes #1141

### DIFF
--- a/internal/index/grok_test.go
+++ b/internal/index/grok_test.go
@@ -22,6 +22,20 @@ func TestMain(m *testing.M) {
 		// developer who has it set would otherwise have the suite write into
 		// their real ~/.config/deja.
 		"XDG_CONFIG_HOME": filepath.Join(root, "config"),
+		// Notes resolve via XDG_DATA_HOME, then APPDATA on Windows, then
+		// DEJA_NOTES_FILE — none of which HOME can reach. Without these the
+		// suite appended fixture notes to the developer's real store and then
+		// failed reading them back as leaked `deja` sessions (#1141). cmd/deja's
+		// TestMain already blanks the same keys; this closes the drift.
+		"XDG_DATA_HOME":           "",
+		"APPDATA":                 filepath.Join(root, "AppData", "Roaming"),
+		"DEJA_NOTES_FILE":         "",
+		"DEJA_SOURCE_INSTANCE":    "",
+		"CLAUDE_CONFIG_DIR":       "",
+		"CODEX_HOME":              "",
+		"GEMINI_CLI_HOME":         "",
+		"CURSOR_CONFIG_DIR":       "",
+		"AIDER_CHAT_HISTORY_FILE": "",
 		// A developer with DEJA_INDEX_DIR exported reads their own index
 		// here, and the suite went red on their machine only.
 		"DEJA_INDEX_DIR":        "",
--- a/internal/index/notes_isolation_test.go
+++ b/internal/index/notes_isolation_test.go
@@ -1,0 +1,25 @@
+package index
+
+import (
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/sources"
+)
+
+// The suite writes and reads promoted notes, and sources.NotesFile() resolves
+// through XDG_DATA_HOME and APPDATA before it ever reaches HOME. When TestMain
+// left those pointing at the developer's real profile, the suite appended
+// fixtures to their store and then failed reading them back as leaked `deja`
+// sessions (#1141). This pins the store under the temp root so that drift
+// fails here, naming the real path, instead of eleven arithmetic tests.
+func TestNotesFileStaysUnderTheTestRoot(t *testing.T) {
+	nf := sources.NotesFile()
+	home := os.Getenv("HOME")
+	appdata := os.Getenv("APPDATA")
+	if (home != "" && strings.HasPrefix(nf, home)) || (appdata != "" && strings.HasPrefix(nf, appdata)) {
+		return
+	}
+	t.Fatalf("notes store resolved outside the test root: %q — TestMain is leaking to the developer's real store (HOME=%q APPDATA=%q); blank XDG_DATA_HOME/DEJA_NOTES_FILE and point APPDATA at the temp root (#1141)", nf, home, appdata)
+}
--- a/internal/index/notes_isolation_test.go
+++ b/internal/index/notes_isolation_test.go
@@ -18,7 +18,13 @@ func TestNotesFileStaysUnderTheTestRoot(t *testing.T) {
 	nf := sources.NotesFile()
 	home := os.Getenv("HOME")
 	appdata := os.Getenv("APPDATA")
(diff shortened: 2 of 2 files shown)


Sora-bluesky:
Ran this branch on the machine #1141 came from — the Windows install with a real notes store.

Two full runs of `go test ./internal/index/`: none of the eleven come back, and the store held at 11 lines through both. The only failures were `TestImportHonorsExcludeList` and `TestExcludeAppliesToAlreadyImportedSessions`, in one run of the two — the same pair #1141 flagged as unrelated, so nothing this PR owes.

One thing worth a look before merge. The guard accepts any path under `%APPDATA%` as the environment currently reads, so the drift it most needs to catch — losing the `APPDATA` redirect, which is the exact shape of #1141 — leaves it green. I tried that here: with the `"APPDATA"` line removed from TestMain, `TestManifestCacheInvalidatesOnReindex` fails again while `TestNotesFileStaysUnderTheTestRoot` stays quiet.

Anchoring to HOME alone closes it, since TestMain points HOME at the temp root on every platform and every notes path lands under it — `root\AppData\Roaming` on Windows, `root/.local/share` elsewhere:

```go
if home != "" && strings.HasPrefix(nf, home) {
	return
}
```

Measured both directions on this machine: green with the redirects intact, and with the `APPDATA` line dropped it fails naming the real store path.

vshulcz:
Thanks — the HOME-only anchor is the right call. You're right that accepting an APPDATA prefix reads the drift back from the very environment the guard is meant to catch; with the redirect dropped it stayed green where it should fail. Adopted it, and confirmed the store holds under the temp root. Agreed the two exclude tests are the unrelated flake #1141 already flagged, not this PR's.

Sora-bluesky:
Verified 2ee73f4 on the same machine: green with the redirects intact, and with the APPDATA redirect dropped the guard now fails naming the real store path — the case that stayed quiet before. Full suite came back clean on this run and the store held at 11 lines. Nothing further from my side.

Issue #1216, opened 2026-08-13
Plain-language questions return almost nothing: matching requires every content term


vshulcz:
A question asked in plain language collapses the candidate set to a handful of sessions, and usually the wrong ones. Matching requires every content term to be present, so the wording of the question decides what can be found at all, before any ranking happens.

Measured with `scripts/day0bench` over a 1910-session corpus laid down as real history (40 questions whose answers exist only in that history):

```
hit@1 21/40   hit@5 25/40   found@50 28/40
```

`found@50` counts the answer session appearing anywhere in the top 50. 12 questions never return it at all, so ranking work cannot reach them — the ceiling is 28.

The cause is conjunctive matching. Asking the full question returns almost nothing:

```
"How many playlists do I have on Spotify?"   1 result   (of 1910 sessions, not the answer)
"playlists spotify"                         15 results  (answer at rank 2)
"spotify"                                   33 results
```

Across all 12 misses, the full question returns 1–8 candidates and never the answer, while a single rare term from that same question recovers the answer in 10 of them:

```
e47becba  full=4   rare 'graduate'      df=21   -
6ade9755  full=8   rare 'yoga'          df=119  HIT
c960da58  full=1   rare 'playlists'     df=29   HIT
66f24dbb  full=3   rare 'birthday'      df=54   HIT
ad7109d1  full=1   rare 'internet'      df=55   HIT
8ebdbe50  full=1   rare 'certification' df=54   HIT
6b168ec8  full=2   rare 'bikes'         df=20   HIT
d52b4f67  full=1   rare 'cousin'        df=14   HIT
25e5aa4f  full=2   rare 'bachelor'      df=11   -
8550ddae  full=5   rare 'cocktail'      df=22   HIT
60d45044  full=2   rare 'rice'          df=123  HIT
577d4d32  full=50  rare 'emails'        df=59   HIT
```

The data is indexed and retrievable. The question's own words exclude it: "What degree did I graduate with?" requires both `degree` and `graduate`, and a session that says "I finished my BSc" has neither pair member in common.

Mechanically it is `intersectPostings` (internal/index/retrieval.go:1606) returning only offsets present in every posting list, with `MatchesParts` (internal/query/query.go:175) requiring every term to match. Stopwords are dropped first, but what survives is still conjunctive: `many`, `have`, `own`, `type` are content words to the index and hard requirements to the query.

[...]

PR #1217, opened 2026-08-13
fix(search): hang the relevance ranking under a thin AND

vshulcz:
Fixes the retrieval half of #1216.

A question asked in plain words requires every content word to be present, so on a large history it lands on a handful of sessions — often one, and often the wrong one. The answer is in the index and reachable, but the question's own wording excludes it before ranking gets a say.

When an AND comes back thin (under 10 sessions) on a store bigger than the relevance window, the relevance ranking is now hung underneath it. The strict sessions keep the front: they are why hit@1 is what it is, and relevance ranked alone scores worse at the first position. `RelevanceHits` scores by arrival, so the merged order survives to the caller — which is why the result is labelled as that tier.

On day0bench over a 1910-session corpus:

```
before   hit@1 21/40   hit@5 25/40   found@50 28/40
after    hit@1 21/40   hit@5 27/40   found@50 34/40
```

longmemeval is unchanged at 90.0% / 95.0%, and decisionbench stays green.

Two gates keep it from firing where it would only add noise. A wide AND is describing a real cluster and is left alone. And a store smaller than the window has nothing to rank — relevance would hand back most of it, so a precise query on a small history would come back with the rest of the history attached; few results out of few sessions is an answer, not a symptom.

Six of the original twelve misses remain. They reach the relevance tier now and still do not surface, so what is left is the quality of that ranking rather than access to it — plain BM25 finds them, which is the follow-up in #1216.


vshulcz:
Chased the remaining misses with the benchmark rather than by reading code, and the answer moved the numbers a lot further than the first commit did.

```
main         hit@1 21/40   hit@5 25/40   found@50 28/40
first pass   hit@1 21/40   hit@5 27/40   found@50 34/40
now          hit@1 21/40   hit@5 28/40   found@50 40/40
```

Four hypotheses died on measurement first, and they are worth recording so nobody spends the time again:

- **Missing length normalisation.** BM25 with `b=0` scores the same 39/40 as `b=0.75` on this corpus, and slightly better at rank 1. Chat sessions are not padded documents; a longer conversation is not a diluted one.
- **The scoring formula.** Swapping the relevance tier's term weight for BM25 saturation, with and without the best-message shaping, left `found@50` at 34.
- **The relevance window.** Widening it from 50 to 500 changed nothing (34 → 33).
- **Query term extraction.** Feeding BM25 deja's own `RelevanceTerms` output still gave 39/40, so the terms were never the problem.

What it actually was: the word-form fallbacks return early. Instrumenting three of the misses showed the relevance ranking placing the answer session at rank 1, 4, and 19 of its pool — while the served result was `tier="close"`, two to five incidental sessions, answer absent. `stemSearch` and `fuzzySearch` had already returned by then. They now get the same tail the strict path got.

One thing that had to change with it: the merged list is labelled as the relevance tier so its order survives to the caller, but the head arrives in posting order — the tier that produced it expected the caller to rank it, and the caller no longer can. Merging without ordering the head cost 4 positions at rank 1. Ordering it by where the same sessions sit in the relevance pool gives that back and keeps the recall.

At the larger size the gap widens rather than closes — 80 questions over 3745 sessions:

```
before   hit@1 27/80   hit@5 37/80   found@50 44/80
after    hit@1 27/80   hit@5 44/80   found@50 68/80
```

longmemeval is unchanged at 90.0% / 95.0%, decisionbench is unchanged including its noise guard (8/8 still ranked first), and the full suite passes.

vshulcz:
Closing this in favour of four PRs cut out of it. It had grown to 60 commits over 71 files while the title still described the first one, and a squash of that would have put a misleading sentence in `git log` forever.

- #1225 — the rig: a recording endpoint and a sweep that drives every installed harness. No product code.
- #1226 — this PR's original subject: the relevance tail under a thin AND, plus three performance changes.
- #1227 — the wiring: prompt-time recall for cline, openclaw, goose and gemini, a project for antigravity's CLI sessions, and seven install failures the new tests found.
- #1228 — what the user is told: the déjà vu line rate-limited per session rather than per machine, six stuttering receipts, and the citation quoting the line that matched.

Together they are exactly this branch: 71 files, none left out, none in two of them. Each is cut from current main, builds and passes the suite on its own.

PR #1220, opened 2026-08-14
perf(index): cut the CJK bigram build overhead without changing the index

Sora-bluesky:
The pass 1 rewrite sketched in [#492](https://github.com/vshulcz/deja-vu/issues/492#issuecomment-5260556062): same keys, same bucket names, byte-identical `buckets/`, less work producing them.

Two changes, both index-side only:

- `cjkIndexKeys` replaces the build-path use of `cjkBigrams`. It folds each CJK run once through `cjkfold.Rune` while walking it, dedupes folded pairs in one `map[uint64]struct{}` keyed on the rune pair, and builds each emitted key with a single allocation. The legacy path folded every emitted bigram separately and deduped the same keys twice, once on pre-fold strings inside `cjkBigrams` and again in `eachIndexKey`. The query side (`cjkBigrams`, `expandCJKTokens`) is untouched — its dedupe is what keeps a phrase query's AND slots useful, and the function-word filter still sees the runes the user typed.
- `bucket` decodes at most three leading runes in place instead of materializing `[]rune` of the whole token, and reads shard names from a 256-entry table instead of `fmt.Sprintf`. Query terms can carry bytes no tokenizer produced, so the moment decoding hits an invalid byte in that prefix the function returns to a verbatim copy of the legacy `[]rune` conversion — U+FFFD replacement and all — keeping every existing term in the bucket it is in today.

## Equivalence, verified

- `buckets/` is byte-identical old-vs-new (`diff -r`) on three corpora: the all-Japanese and English arms from the #492 measurement, plus a mixed store of both. `records.bin` hashes equal; `manifest.gob` matches in size (its bytes carry timestamps).
- `cjkindex_diff_test.go` pins the boundary classes against verbatim legacy copies: invalid UTF-8 prefixes, 1/2/3-rune ASCII edges, fold-collision pairs (關係/关系), single-rune runs, supplementary-plane Han, kana and Hangul, a run crossing the tokenizer's 64-byte flush, and seeded generative sweeps over random bytes and script mixes.
- The query side is the leg a byte-identical index cannot testify for, as the review on the issue pointed out: `cjkBigrams` stays shared and unfolded (the function-word filter must see the runes the user typed), so `TestExpandCJKTokensKeepsScript492` pushes Traditional input through `expandCJKTokens` and asserts the bigrams keep their script — a future fold inside the shared function now fails a query-side test, not just `TestCJKFoldUnit`.
- Full suite passes. The dedupe-scope change is observable only as fewer duplicate strings in `indexKeys`' return value; every caller either dedupes (`eachIndexKey`, both incremental paths) or converges through `encodePostings`' offset dedupe (`sync.go` import path), which the fold-collision test locks in.

## What it buys

Same-binary benchmarks (new implementation against the legacy copies in the same test binary, i7-14700K):

| | legacy | new | |
|---|---|---|---|
| key emission, one Japanese message | 11.4 µs, 214 allocs | 5.7 µs, 44 allocs | 2.0× |
| `bucket`, 8 mixed tokens | 510 ns, 8 allocs | 96 ns, 0 allocs | 5.3× |

One honest caveat on end-to-end numbers, in the spirit of this thread's cold-cache retraction: on this corpus's English arm (3.2 s builds), binary layout alone moved whole-build wall clock by about 15% between two builds of identical old code (default flags against `-trimpath`), interleaved in one window — more than this change's whole expected effect — and `-trimpath` builds of the old and new code land in the same band. So I am not quoting a wall-clock delta from this rig; the per-stage profile in the issue comment says where the removed work was (0.30 s dedupe maps + 0.22 s per-bigram folds + 0.28 s Sprintf out of a 2.7 s bigram delta), and the allocation counts above are layout-immune. A corpus at the scale of the original report should see it above the noise.

How much the dedupe savings buy on real text also depends on the corpus's bigram repetition rate, and a deterministic generator's repetition is not obviously that of real conversation — the offer on the issue to re-run this against a real-Chinese store answers that directly, so the harness rides along: `tools/bench492/` holds the corpus generator and the interleaved runner, both confined to isolated index and store paths. Drop that commit if `tools/` is not where you want it living.

### DIFF
--- a/internal/index/cjkindex.go
+++ b/internal/index/cjkindex.go
@@ -1,0 +1,167 @@
+package index
+
+import (
+	"hash/fnv"
+	"unicode/utf8"
+
+	"github.com/vshulcz/deja-vu/internal/cjkfold"
+)
+
+// hexBuckets fixes every shard name in a lookup table. bucket relies on these
+// exact names for on-disk compatibility while avoiding formatter work in the
+// build path addressed by #492.
+var hexBuckets = [256]string{
+	"x00", "x01", "x02", "x03", "x04", "x05", "x06", "x07",
+	"x08", "x09", "x0a", "x0b", "x0c", "x0d", "x0e", "x0f",
+	"x10", "x11", "x12", "x13", "x14", "x15", "x16", "x17",
+	"x18", "x19", "x1a", "x1b", "x1c", "x1d", "x1e", "x1f",
+	"x20", "x21", "x22", "x23", "x24", "x25", "x26", "x27",
+	"x28", "x29", "x2a", "x2b", "x2c", "x2d", "x2e", "x2f",
+	"x30", "x31", "x32", "x33", "x34", "x35", "x36", "x37",
+	"x38", "x39", "x3a", "x3b", "x3c", "x3d", "x3e", "x3f",
+	"x40", "x41", "x42", "x43", "x44", "x45", "x46", "x47",
+	"x48", "x49", "x4a", "x4b", "x4c", "x4d", "x4e", "x4f",
+	"x50", "x51", "x52", "x53", "x54", "x55", "x56", "x57",
+	"x58", "x59", "x5a", "x5b", "x5c", "x5d", "x5e", "x5f",
+	"x60", "x61", "x62", "x63", "x64", "x65", "x66", "x67",
+	"x68", "x69", "x6a", "x6b", "x6c", "x6d", "x6e", "x6f",
+	"x70", "x71", "x72", "x73", "x74", "x75", "x76", "x77",
+	"x78", "x79", "x7a", "x7b", "x7c", "x7d", "x7e", "x7f",
+	"x80", "x81", "x82", "x83", "x84", "x85", "x86", "x87",
+	"x88", "x89", "x8a", "x8b", "x8c", "x8d", "x8e", "x8f",
+	"x90", "x91", "x92", "x93", "x94", "x95", "x96", "x97",
+	"x98", "x99", "x9a", "x9b", "x9c", "x9d", "x9e", "x9f",
+	"xa0", "xa1", "xa2", "xa3", "xa4", "xa5", "xa6", "xa7",
+	"xa8", "xa9", "xaa", "xab", "xac", "xad", "xae", "xaf",
+	"xb0", "xb1", "xb2", "xb3", "xb4", "xb5", "xb6", "xb7",
+	"xb8", "xb9", "xba", "xbb", "xbc", "xbd", "xbe", "xbf",
+	"xc0", "xc1", "xc2", "xc3", "xc4", "xc5", "xc6", "xc7",
+	"xc8", "xc9", "xca", "xcb", "xcc", "xcd", "xce", "xcf",
+	"xd0", "xd1", "xd2", "xd3", "xd4", "xd5", "xd6", "xd7",
+	"xd8", "xd9", "xda", "xdb", "xdc", "xdd", "xde", "xdf",
+	"xe0", "xe1", "xe2", "xe3", "xe4", "xe5", "xe6", "xe7",
+	"xe8", "xe9", "xea", "xeb", "xec", "xed", "xee", "xef",
+	"xf0", "xf1", "xf2", "xf3", "xf4", "xf5", "xf6", "xf7",
+	"xf8", "xf9", "xfa", "xfb", "xfc", "xfd", "xfe", "xff",
+}
+
+// bucket shards a token by its opening runes. The first three runes are decoded
+// without materializing the whole token, but #492 preserves the legacy byte
+// contract for invalid UTF-8 by returning to the []rune conversion whenever
+// decoding observes an invalid byte in that prefix.
+func bucket(tok string) string {
+	var first, second rune
+	end := 0
+	count := 0
+	for count < 3 && end < len(tok) {
+		r, size := utf8.DecodeRuneInString(tok[end:])
+		if r == utf8.RuneError && size == 1 {
+			return bucketInvalidUTF8(tok)
+		}
+		switch count {
+		case 0:
+			first = r
+		case 1:
+			second = r
+		}
+		end += size
+		count++
+	}
+
+	if count >= 2 && isShardASCII(first) && isShardASCII(second) {
+		return safe(tok[:2])
+	}
+
+	h := fnv.New32a()
+	_, _ = h.Write([]byte(tok[:end]))
+	return hexBuckets[h.Sum32()%256]
+}
+
+// bucketInvalidUTF8 reproduces the legacy conversion before hashing. []rune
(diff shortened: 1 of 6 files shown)


Sora-bluesky:
The red govulncheck check is the toolchain, not this diff. Against go1.25.12 it reports five standard-library vulnerabilities (GO-2026-6218, -6091, -6090, -5972, -5026), every one of them fixed in go1.25.13. GO-2026-5972 was published on August 13 (pkg.go.dev); main's most recent CI run passed earlier that day, so the vulnerability database appears to have moved between the two runs — a fresh run on main should reproduce the failure. The example traces go through `internal/embed/client.go`, `internal/usage/snapshots.go` and `internal/index/store_io.go`, none of which this PR touches, and the one `cjkfold` trace is the `sync.Once` inside `Rune`, which `cjkfold.String` already called before this branch. A toolchain bump to go1.25.13 clears it, and that belongs in its own commit rather than riding in this one.

vshulcz:
Read it against `8ecdfe3`. The shape is right: the split stays on the ingestion side, `cjkBigrams` and `expandCJKTokens` are untouched, and the differential tests carry the weight the equivalence claim needs — the invalid-UTF-8 cases in `TestBucketDifferential492` are the ones I would have asked for, including the truncated multibyte prefix and the lone continuation byte. Same-binary numbers instead of wall clock, after the 15% layout wobble, is the right call.

One thing the byte-identical check does not reach, and it is worth writing down before this merges.

`cjkBigrams` dedupes **before** folding; `cjkIndexKeys` dedupes **after**. Two spellings of one word are distinct bigrams that fold onto one key, so today the key comes out twice:

```go
// on main
for _, bg := range cjkBigrams("關係 关系") {
    counts["t"+cjkfold.String(bg)]++
}
// key "t关系" emitted 2 times
```

Three of the four callers of `indexKeys` collapse that themselves — `eachIndexKey` and both loops in `ingest.go` keep a `seen` map — which is why `buckets/` comes out byte-identical on a rebuild. The fourth does not: `sync.go` appends a posting per key occurrence with no dedupe. And posting count is not inert — `keyTF[pp.Sid]++` counts postings, and that feeds `score[ord] += idf * (1 + 0.25*math.Log2(float64(tf[ord])))`.

So on a store built by sync, a message that writes the same word in both scripts scored as if it had said it twice, and after this PR it scores once. That is the better behaviour and it makes the two writers agree, but it is a change to the index, and the PR title says there is none. Worth a line in the description at least.

If you want to close it properly rather than describe it: an invariant test that the same input, ingested and synced, produces the same postings. Nothing pins that today, which is why the divergence has been sitting there. Happy for that to be a separate PR — this one is already doing one thing well.

Not a blocker either way. The `govulncheck` red is ours, not yours: `setup-go` was stopping at the runner image's 1.25.12 because it satisfies `1.25.x`, so it never looked at the manifest where 1.25.13 lives. Fixed on our side with `check-latest: true`; rebase once that lands and it goes green.

Sora-bluesky:
The dedupe-scope observation is exactly right at the `indexKeys` level — on main the fold collision leaves the loop twice, on this branch once — and the design review flagged the same caller for the same reason: `sync.go` is the one consumer with no `seen` map of its own. Here is where I believe the duplicate dies before it can reach `keyTF`.

The sync loop appends into an in-memory `bucketPostings` (`sync.go`, the `data[tok] = append(...)` under `appendImportedRecords`), and that map only reaches disk through `writeBucketsConcurrent` → `writeBucket` → `encodePostings` (`store_io.go:730`), whose first act is `sortedUniquePostings` (`retrieval.go:1524-1540`) — postings sorted and collapsed on `Off` alone. A fold collision inside one record produces two postings with the same offset, so one survives encoding. `keyTF[pp.Sid]++` counts postings decoded from the bucket file, after that collapse, so a synced store scores the both-scripts message once on main too.

Ran it to be sure rather than argue from the call chain:

```go
dup := map[string][]posting{"t关系": {{Off: 100, Sid: 7}, {Off: 100, Sid: 7}}}
_ = writeBucket(p, dup)
got, _ := readBucketToken(p, "t关系")
// len(got) == 1, and the bucket file is byte-identical to one written
// from the single posting
```

So unless there is a path where the pre-encode count reaches scoring that I have not found, the title's claim holds on the sync path as well — say the word and the description changes if I am missing one.

Your invariant is still worth pinning: nothing today asserts that the same input ingested and synced produces the same postings, and that absence is exactly why this took a probe to settle instead of a failing test. I'll put that up as the separate PR you suggested, with the probe above as its seed.

Will rebase for the `setup-go` fix once it lands — thanks for chasing that down.

Sora-bluesky:
The invariant test paid for itself before it ever reached CI. Writing it turned up the divergence your review said was sitting there — three of them, none the fold one: the import loop skips `dateTokens` (an imported record never matches "what did we do in may"), never sets the tool bit, and indexes tool output uncut past the `signalLines` filter. The test failed on main on exactly the seven date tokens its fixtures generate.

Fix and test are up as #1223, the separate PR you suggested — the import loop now goes through the same `eachIndexKey(tokenizedPart(...))` call ingestion makes. The parity test pins each leg its own way: the term-frequency maps catch the date tokens, explicit posting assertions catch the tool bit, a long tool result whose middle word must earn no posting on either route catches the cut, and the fold-collision fixture from this thread rides along asserting its key collapses to one posting.

vshulcz:
Merged, and my objection to the title is withdrawn — it was wrong when I made it and it is doubly wrong now.

Wrong then: `encodePostings` runs `sortedUniquePostings`, which collapses on `Off`, so the duplicate folded key never reached disk and could not have moved `tf`. Wrong now for a second reason: after #1223 the import loop goes through the same `eachIndexKey` call ingestion makes, so there is no longer a second route for the dedupe to differ on. "Without changing the index" is literally true. Leave it as it is.

I checked the claim myself rather than reading the differential tests and believing them. A corpus of seven sessions in Traditional, Simplified, Japanese, Korean and mixed ASCII, rebuilt with `index --rebuild` on main and on this branch:

```
main: 68 bucket files | pr: 68 bucket files
byte-identical: True
```

That is the whole acceptance criterion in one line, and it holds.

What I want on the record about how this went. The three things you listed in the issue — the double dedupe, the per-bigram fold, the `Sprintf` in the hot loop — were each a real cost, and you resisted the obvious version of the fix in favour of the one that survives `TestCJKFoldUnit`. The `bucketInvalidUTF8` fallback is the part I would have got wrong: reproducing the `[]rune` round-trip so an invalid byte still lands in the shard it always did, rather than deciding that invalid input does not matter. It does — those tokens are already in people's indexes.

And the 15% wall-clock swing from binary layout, caught by interleaving two builds of identical code, is the measurement discipline this repo should copy. Quoting allocation counts instead was right.

Thank you. Both this and #1223 came out of one issue thread, and the second one only exists because you tested a review comment instead of implementing it.

Issue #655, opened 2026-07-31
Work records are still missing for copilot, cline, kimi and qwen — the stores already hold them


vshulcz:
#595 covered Claude, opencode, Cursor and Codex. The audit shows four more harnesses record tool calls that deja reads past, and copilot's shape is the cleanest of any harness so far.

**copilot** — `~/.copilot/session-state/*/events.jsonl`:

```json
{"type":"tool.execution_start","data":{"toolName":"edit",
  "arguments":{"path":"…/main.go","old_str":"println(\"hello\")","new_str":"println(\"goodbye\")"}}}
{"type":"tool.execution_start","data":{"toolName":"bash",
  "arguments":{"command":"cd … && go mod init coprobe && go vet ./..."}}}
{"type":"tool.execution_complete","data":{"success":true,
  "result":{"content":"pattern ./...: directory prefix . does not contain main module…"}}}
{"type":"session.shutdown","data":{"codeChanges":{"filesModified":["…/notes.txt","…/main.go"]}}}
```

Path, old text, new text, command, output, success flag, and a ready-made list of modified files. Everything the four supported harnesses give, in one place, plus `session.shutdown` which none of them has.

Reachability today, on a session that edited two files and ran three commands:

```
$ deja "directory prefix" --harness copilot   → no matches    (the error it hit)
$ deja "go mod init"      --harness copilot   → no matches    (the command it ran)
```

Also dropped: `assistant.message` records whose `content` is empty and whose work is entirely in `toolRequests[]`.

**cline** — `tool_use` / `tool_result` blocks in `<sid>.messages.json`, skipped by the role filter and by `clineContentText` keeping only `type:"text"`.

**qwen** — `message.parts[].functionCall{name,args}` (args carry `working_dir`) and `functionResponse.response.output`.

**kimi** — `context.append_loop_event` events of type `tool.call` (name, args) and `tool.result` (output).

Order by value: copilot first — richest shape, and the only harness with a modified-files list the parser can trust without inference.

AliceLJY:
The Kimi quarter of this landed in #1221 — tool.call and tool.result now come out as work records, so `deja files` and command search reach Kimi sessions. Copilot, Cline and Qwen remain open; I only had real Kimi transcripts to verify against, so I stopped where my data ends.

vshulcz:
[...]

PR #1221, opened 2026-08-14
feat(kimi): turn tool calls and results into work records

AliceLJY:
The Kimi quarter of #655: wire.jsonl carries every call as a `context.append_loop_event` — `tool.call` with a name and args, `tool.result` with the output — and the parser read past both, so `deja files`, `deja restore` and command search were silent on every Kimi session.

Both event kinds now become the records the other harnesses emit, through the shared dialect helpers with Kimi's vocabulary:

- The tool names mirror Claude Code's — Kimi Code is built in that shape — but the file key is `path` where Claude's is `file_path` (the key Cursor uses too), only `Edit` carries a replaced span (`old_string`; there is no MultiEdit), and `ReadMediaFile` is Kimi's own path tool. The shell tool is `Bash` with its command under `command`, which is what `commandsIn` already reads.
- Read off wire transcripts the Kimi CLI had just written, per this issue's rule of checking rather than assuming: on the store here (71 main-agent wires), all 822 Bash calls carry `command`, all 99 Edits carry `path` + `old_string` + `new_string`, and every one of 1,616 results keeps its text under `result.output` — plain, with a note, truncated, or flagged `isError`. The error ones are kept on purpose: the error a command hit is exactly what a later search reaches for.
- The pending assistant stream flushes only when a call yields a record, so the words that led to a command stay ahead of it — and with every `DEJA_INDEX_*` switch off, the messages come out exactly as they did before tool events were read (`TestParseKimiToolRecordSwitchesOff` pins that).

## Measured on a real store

71 main-agent wires / 67 sessions; old and new binaries built from the same tree, same corpus, isolated index paths:

| query | old | new |
|---|---|---|
| a container name it ran in Bash | 2 sessions | 5 |
| a project directory in command paths | 1 | 3 |
| a file it edited | 1 | 9 |
| a conversation word, as control | 19 | 31 |

The control moves too because tool output mentions it; nothing the old index matched went away — for all four queries the old result set is a subset of the new one. `deja files <topic>` goes from "22 sessions mention it, none of them recorded a file near it" to a ranked file list. The index grows 2.5 → 8.9 MB on this store, nearly all of it tool-output text — the price the supported harnesses already pay.

### DIFF
--- a/internal/sources/kimi.go
+++ b/internal/sources/kimi.go
@@ -19,8 +19,9 @@ import (
 // wire.jsonl is the append-only main-agent transcript. User turns arrive as
 // context.append_message records; streamed assistant turns never do — they
 // have to be reconstructed from step.begin → content.part → step.end loop
-// events. Only the main agent is indexed; sub-agents, tool payloads and
-// think-parts are skipped by design (issue #248).
+// events. Tool calls and their results arrive as loop events too (tool.call,
+// tool.result) and become work records (#655). Only the main agent is indexed;
+// sub-agents and think-parts are skipped by design (issue #248).
 
 // KimiConfigDir is the native Kimi Code home. DEJA_KIMI_ROOT intentionally
 // does not affect it because that variable only relocates reads.
@@ -42,6 +43,19 @@ func ParseKimiFile(path string) ([]model.Session, error) {
 
 func ParseKimiFileFromOffset(path string, offset int64) ([]model.Session, error) {
 	return parseKimiFileFromOffset(path, offset)
+}
+
+// kimiDialect is Kimi Code's tool vocabulary, read off wire.jsonl transcripts
+// its own CLI had just written. The names mirror Claude Code's — Kimi Code is
+// built in that shape — but the file key is `path` where Claude's is
+// `file_path`, and only Edit carries a replaced span; there is no MultiEdit.
+// ReadMediaFile is Kimi's own path tool. The shell tool's argument key is
+// `command`, which is what commandsIn already reads.
+var kimiDialect = toolDialect{
+	pathKey:   "path",
+	pathTools: map[string]bool{"Read": true, "Edit": true, "Write": true, "ReadMediaFile": true},
+	shellTool: "Bash",
+	editTools: map[string]bool{"Edit": true},
 }
 
 // kimiState is the slice of state.json deja cares about.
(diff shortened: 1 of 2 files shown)


AliceLJY:
The govulncheck red is not this diff: all five findings are standard-library CVEs fixed in go1.25.13 (net/url, html/template, crypto/tls, encoding/asn1, net/http), the example traces run through `internal/embed/client.go`, `cmd/deja/statshtml.go` and `internal/cjkfold/fold.go` — none through `internal/sources` — and a clean checkout of main at b2cdf6c with `GOTOOLCHAIN=go1.25.12` reports the same five. Any PR opened today carries the same red until the runner's toolchain reaches 1.25.13.

vshulcz:
This is the version of the change I would have asked for. It reuses `toolDialect` instead of growing a second parser, the dialect is written from what the CLI actually emits — `path` rather than `file_path`, no MultiEdit, `ReadMediaFile` as Kimi's own — and every record sits behind its `DEJA_INDEX_*` switch. `TestParseKimiToolRecordSwitchesOff` is the test that makes the rest safe to take: with the switches off the stream comes back exactly as it was, so nobody's existing index changes shape under them.

The ordering assertion is the part I liked most. Flushing only when a call yields a record, and pinning it with `want := []string{"answer one, joined", "looking at the failing test", "$ go test ./...", "— fixed"}`, is what keeps the words that led to a command ahead of the command instead of stranded at `step.end`. That is easy to get wrong and invisible in a search until someone reads a transcript and finds the reasoning detached from the action.

Two notes, neither a blocker:

`tool.result` is indexed independently of whether its `tool.call` produced a record. A `TodoList` call yields nothing, but a result carrying output would still land as `RoleToolOutput`. That is probably what you want — tool output has its own switch and its own value — but the fixture never exercises it, since the two calls without records also have no result events. One more line in `kimiWireTools` would say which way it is meant to go.

`TestParseKimiToleratesTornTail` predates this, and the tail it tears is a text event. The parser now reads `tool.call` and `tool.result` too, and those are the longer lines — an args blob is where a write is most likely to be caught mid-flush. Tearing one of those would extend the existing guarantee to the events this PR adds.

Merging as is; both notes are follow-ups rather than changes I want before it lands.

vshulcz:
Merged. Thank you — kimi has been listed as supported for a while and was quietly the thinnest of the parsers: speech only, with the commands and files a session touched thrown away. It reads like the others now.

The `DEJA_INDEX_*` test is what made this easy to take. A change that adds records to a parser can move every existing user's index without anyone noticing until a search comes back different; proving the stream is byte-for-byte unchanged with the switches off answers that before it gets asked.

The two follow-ups from the review — a fixture where a result outlives a call that produced no record, and a torn tail on a `tool.call` rather than a text event — are yours if you want them, and fine to leave if you would rather move on. Neither blocks anything.

AliceLJY:
Both notes read the intent right — a result carrying output is meant to land even when its call yields no record; tool output has its own switch and its own value. Both follow-ups are sitting on a branch already: the fixture line saying so, and the torn-tail test running against the full tools fixture, tearing the message, a tool.call args blob and a tool.result in turn. I'll open the PR once this lands, so the diff stays just the two tests.

PR #1223, opened 2026-08-14
fix(sync): index imported records the way ingestion does

Sora-bluesky:
The invariant test suggested in #1220 review, plus the fix it turned out to need. Writing the test found the divergence you predicted was sitting there — three of them, none the fold one.

The import loop fed bare `indexKeys(r.Text)` straight into postings (`sync.go`), while ingestion pushes `tokenizedPart(role, text)` through `eachIndexKey` and stamps the tool bit. So a synced store diverged from a rebuilt one three ways:

- **No date tokens.** `eachIndexKey` appends `dateTokens(when)`; the import loop never did, so an imported record can not answer "what did we do in may" on the machine it synced to.
- **No tool bit.** The loop appended `posting{Off, Sid}`, leaving `Tool` false on every imported posting, so the per-session bound spends budget on imported tool output as if it were speech.
- **No `tokenizedPart` cut.** A long tool-output record indexed its full text on import — the exact posting explosion `signalLines` exists to stop, paid again by every synced copy.

The fix routes the loop through `eachIndexKey(tokenizedPart(r.Role, r.Text), r.Time, ...)` with `Tool: isToolRole(r.Role)` — the same call ingestion makes, so the two writers cannot drift apart silently again.

`TestIngestAndSyncProduceSamePostings` pins it: build a store, ingest it into one index, export and import it into a second, and compare per-token per-session term frequencies (offsets and ordinals are excluded — two independently built indexes assign both differently). Before the fix it failed on exactly the seven date tokens the fixtures generate; after, the maps are equal. The fixtures cover each divergence on purpose: fixed timestamps for the date tokens, a long tool result whose head, middle and tail carry distinct words (the middle must earn no posting on either route, head and tail must carry the tool bit on both), a two-turn session for tf above one, and a message writing 關係 and 关系 in one breath for the fold-collision case from the #1220 thread — which settled the other way, since duplicates already collapse inside `encodePostings`.

On existing synced stores this is a behavior change in the direction of the rebuild. This PR migrates nothing in place: re-importing the same batch dedupes and touches nothing, so the divergent postings heal on the next full rebuild, which re-derives imported sessions from `records.bin` and indexes them down the ingestion path.

### DIFF
--- a/internal/index/ingest_sync_parity_test.go
+++ b/internal/index/ingest_sync_parity_test.go
@@ -1,10 +1,12 @@
 package index
 
 import (
+	"encoding/json"
 	"os"
 	"path/filepath"
 	"reflect"
 	"sort"
+	"strings"
 	"testing"
 )
 
@@ -51,6 +53,24 @@ func TestIngestAndSyncProduceSamePostings(t *testing.T) {
 		}
 	}
 
+	// A session whose only turn is one long tool result pins the other two
+	// halves of the invariant. The text has no signal-marker lines, so
+	// signalLines keeps its head and tail and drops the middle: headonlyword
+	// sits inside the first 8 KiB, midonlyword between the head and the final
+	// 4 KiB, tailonlyword in the tail. Ingest indexes head and tail with the
+	// tool bit set and never indexes the middle; the imported copy must do
+	// exactly the same.
+	pad := strings.Repeat("plain filler that stays clear of every signal marker ", 200)
+	longTool := "headonlyword " + pad[:9000] + " midonlyword " + pad[:5000] + " tailonlyword"
+	toolJSON, err := json.Marshal(longTool)
+	if err != nil {
+		t.Fatal(err)
+	}
+	toolLine := `{"type":"user","message":{"role":"user","content":[{"type":"tool_result","content":` + string(toolJSON) + `}]},"timestamp":"2026-04-05T06:07:08Z","sessionId":"tool-session","cwd":"/tmp/parity"}` + "\n"
+	if err := os.WriteFile(filepath.Join(projectA, "tool.jsonl"), []byte(toolLine), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
 	dirA := filepath.Join(tmp, "index-a.db")
 	if err := Ensure(dirA, "", false, nil); err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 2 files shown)


vshulcz:
First: my claim in the #1220 thread was wrong, and you were right to check it instead of taking it on trust. Duplicate postings never reach disk — `encodePostings` runs `sortedUniquePostings`, which collapses on `Off`:

```go
for _, p := range out {
    if n == 0 || out[n-1].Off != p.Off {
```

So the fold collision could not have moved `tf`, and the "sync scores it twice" half of that comment does not hold. Thanks for settling it with the code rather than the argument.

The test found three things instead, which is the better outcome and the reason to ask for invariants rather than for spot checks. All three read as real to me, and the date-token one is the worst of them: a session that syncs to a second machine silently stops answering questions about when it happened, on a store that looks complete. Nothing else would have surfaced that — the sessions are all there, the text search works, and only the date axis is missing.

The fix is the right shape. Routing the import loop through the same `eachIndexKey(tokenizedPart(...))` call ingestion makes is what stops this recurring; three separate patches against three symptoms would have left the fourth divergence to be found later by someone with a broken store.

Two things I checked and liked:

`bucketErr` threading rather than returning out of the callback. `eachIndexKey` takes a `func(string)` with no error, and the obvious move — panic or ignore — would either take the process down or write a partial bucket set that looks complete. Capturing and checking after is unglamorous and correct.

The fixture design does real work: a long tool result whose head, middle and tail carry distinct words is the only way to assert that `tokenizedPart`'s cut lands in the same place on both routes, and comparing per-token per-session term frequencies while excluding offsets and ordinals is exactly right — two independently built indexes have no reason to agree on either, and asserting on them would have produced a test that fails for the wrong reason.

On migration: agreed that healing on the next full rebuild is the right call over an in-place fixup. A migration that rewrites postings is a much larger risk than a store that is slightly worse at date queries until it rebuilds.

Merging once CI is green. This is the second time in this thread that writing the test was worth more than the change it was written for.

vshulcz:
Merged. Three divergences that nothing would have surfaced until someone's synced store answered a date question with silence — and the one I predicted was not among them.

What I want to keep from this: the invariant was worth more than any of the three fixes. "Both writers produce the same postings" is a sentence someone could have written at any point in the last year, and it would have caught all three the day the second writer was added. `TestIngestAndSyncProduceSamePostings` is now the thing that stops the fourth one.

Thanks for checking the claim instead of implementing it. A review from a maintainer is a hypothesis like any other.

PR #1224, opened 2026-08-14
test(kimi): pin record-less results and tear tool lines

AliceLJY:
The two follow-ups from #1221's review (https://github.com/vshulcz/deja-vu/pull/1221#issuecomment-5292588451).

The todo-list call in `kimiWireTools` now carries a result, and the fixture says which way it is meant to go: a call that yields no record still lands its output as `RoleToolOutput` — tool.result is indexed on its own switch, independent of what the call produced.

The torn-tail test used to tear only a text event, from before the parser read tool events. It now runs against the full tools fixture and tears each shape in turn — the message, a tool.call caught mid-args, a tool.result caught mid-output — pinning the same guarantee for all three: the torn line vanishes, everything before it comes out untouched.

### DIFF
--- a/internal/sources/kimi_test.go
+++ b/internal/sources/kimi_test.go
@@ -100,21 +100,43 @@ func TestParseKimiMidStreamAndIncrementalAppend(t *testing.T) {
 }
 
 func TestParseKimiToleratesTornTail(t *testing.T) {
+	// A torn tail is what a crash mid-write leaves behind. The message tear is
+	// the original case; tool.call and tool.result lines are the longer ones —
+	// an args blob is where a write is most likely to be caught mid-flush — so
+	// each shape is torn in turn. The guarantee is the same for all three: the
+	// torn line vanishes, everything before it comes out untouched.
 	_, wire := kimiFixture(t)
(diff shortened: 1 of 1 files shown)


vshulcz:
Both follow-ups, done the way that makes them worth having. Ran the file against current main before reading it closely — all seven kimi tests pass.

The torn-tail change is the better half. Tearing each shape in turn and asserting against the untorn baseline, rather than a hardcoded count, means the test keeps its meaning when the fixture grows: someone adding a tool call to `kimiWireTools` next year does not have to remember to bump a number, and the test does not quietly stop checking anything. The tear points are chosen where a real crash would land, too — mid-`old_string` and mid-`output` are exactly the long writes.

The todo result answers the question I actually had. "A call that yields no record still lands its output" is a decision, not an accident, and now the fixture says so — the next person to read `tool.result` will not wonder whether the independence was intended.

Merging once CI is green.

vshulcz:
Merged. Two review notes turned around in an afternoon, and the torn-tail one came back better than the note asked for — comparing against the untorn baseline instead of a fixed count is what keeps it honest as the fixture grows.

Thanks. kimi went from the thinnest parser in the repo to one with the tightest tests, in three PRs, all yours.

PR #1236, opened 2026-08-14
ci: take the windows leg off the per-commit path, and drop -race there

### LINKED ISSUE #1229: Windows CI passes with seconds to spare: nineteen minutes of it is uncached -race compilation
The Windows test job passes with seconds to spare, and has for a while. It is not slow tests — it is nineteen minutes of compilation that the build cache is not saving.

## The margin

Windows `test` jobs across the last 25 CI runs, all branches:

```
 29.8m success   main
 29.4m success   split/search-perf
 29.3m failure   split/install-robustness
 29.2m success   main
 29.2m failure   disjunctive-tail
 28.9m success   split/install-robustness
 28.7m success   kimi-work-records
 28.3m success   main
 ...
14 of 23 at or past 25 minutes, against a 30-minute cap
```

The best recent success finished with twelve seconds of headroom. Whether a PR goes green is close to a coin flip, and the failures above are the same suite as the successes.

`timeout-minutes` has already been raised once (20 → 30) and coverage instrumentation dropped from Windows to buy room. Both bought a year, not a fix.

## Where the time goes

From a successful run (job 94812384182), test execution by package:

```
  333.0s  cmd/deja
  146.8s  internal/index
   11.7s  internal/usage
   11.5s  scripts/planbench
    ...
  533s total — 8.9 minutes
```

And the step's own timeline:

```
test step began         15:26:13
first package result    15:54:06   (+27.9 min)
last package result     15:54:06   (+27.9 min)
```

Every package reports in the same second, at the end. So of the 27.9 minutes, 8.9 are tests and roughly **19 are building the race-instrumented binaries**.

That also explains the shape of the failures: a cancelled job logs *no test output at all*, because nothing has been flushed yet. It reads like a hang, and there is no way to tell a genuine hang from a slow compile — `-timeout 12m` never gets a chance to fire, since no test has started.

## Why the cache is not helping

`setup-go` restores a cache and reports a hit:

```
Cache hit for: setup-go-Windows-x64-go-1.25.12-7c7c62e0…
Cache Size: ~76 MB
```

That key is derived from `go.sum`, which almost never changes. GitHub caches are immutable: once saved under a key, later runs restore it and never update it. So the entry holds whatever `GOCACHE` contained the day it was first written — the standard library and dependencies, compiled with `-race`, and none of this repository's own packages, which are the ones that change every commit and cost the most to instrument.

## What would actually move it

Roughly in order of effect per unit of risk:

1. **Cache `GOCACHE` with a rotating key.** The standard `actions/cache` pattern — save under a key carrying the run id, restore by prefix — so each run starts from the previous run's compiled objects instead of a frozen snapshot. This is the one that targets the nineteen minutes directly.
2. **Shard the Windows job.** `cmd/deja` and `internal/index` are 90% of test time and would compile in parallel rather than in series.
3. **Drop `-race` on Windows**, keeping it on Linux and macOS. Cheapest to do, and the least defensible: a data race is not usually OS-specific, but "not usually" is doing work in that sentence.

Happy to take (1). Filing rather than pushing it because it is CI cost everyone pays, and because the diagnosis is worth having written down either way: the next person to see a cancelled Windows job with an empty log should not have to rediscover that the silence is a compiler.

vshulcz:
Closes #1229.

The windows leg took 21-28 minutes against two for linux and macOS, and the cap turned its bad days into cancellations that printed no test output at all — which read like hangs.

Two attempts at making it faster, both measured, both dead:

| attempt | windows | verdict |
| --- | --- | --- |
| baseline | ~27.9 min | |
| rotating build cache | 45m17s, cancelled | **worse** |
| Defender exclusions | 26m57s | ~1 min, noise |
| no `-race` | 21m17s | 5m40s |

The cache one failed for a reason worth keeping: **a cancelled job never runs its post steps**, so the job that needed a warm cache was the only one that could seed it, and it was being cancelled at the cap every time. It could never have worked.

What is left after those: `go test` on windows is 20m32s, of which the tests themselves are 3m28s. **The other seventeen minutes are the runner compiling**, which linux does in seconds. Two runs of the identical commit came in at 21m17s and 28m4s, so it varies by a third — that variance is what tipped it over the cap, not any single slow thing.

So this stops paying for it rather than trying to speed it up:

- **No `-race` on windows.** Linux and macOS still run it. A data race lives in Go's memory model rather than the OS, so the same race surfaces there; windows is in the matrix for paths, file locking and case rules, which plain `go test` covers.
- **The leg is skipped from the inside on pull requests**, unless the PR carries the `windows` label. It still runs on every push to main, on the weekly canary, and on `workflow_dispatch`.

Skipped from the inside rather than dropped from the matrix on purpose: `test / windows-latest` is a required check, so removing it from the matrix would leave every unlabelled PR waiting forever on a report that never comes. **No branch protection change is needed.** A dynamic matrix also rendered the job name unexpanded (`test / ${{ matrix.os }}`) whenever concurrency cancelled a run mid-flight.

`labeled` had to be added to the `pull_request` trigger types, or the label would have sat there doing nothing until someone happened to push again.

Verified both directions on this PR:

- labelled → windows runs in full, 28m4s
- unlabelled → `test / windows-latest` reports success in **7s**, every step `skipped`
- whole PR run: **180s**, down from 21-28 minutes


vshulcz:
The first run with this in place was cancelled at 30m00s, and the log says why the change could not have worked as written:

```
Cache not found for input keys: go-Windows--31866288993, go-Windows--, go-Windows-
...
Run go test -race -timeout 12m ./...      05:11:36
##[error]The operation was canceled.     05:40:48
Post job cleanup.
```

Two things in that.

**A cancelled job never runs its post steps**, so `actions/cache` never wrote anything — there is no `Cache saved` line, only the cleanup. The job that needed a warm build cache was the only job that could seed it, and it was cancelled at the cap every time. The change was a no-op on Windows by construction, and would have stayed one no matter how many times it ran.

**The key had an empty hash slot** (`go-Windows--`) because this module has no dependencies and therefore no `go.sum`, and `hashFiles` of a missing file returns an empty string.

Also worth recording: `go test -timeout 12m` bounds test execution, not compilation. Twenty-nine minutes elapsed in that step without the Go timeout firing, because nothing had started running yet. That is the mechanism behind "cancellations print no test output at all" from the issue — the silence really is the compiler.

Pushed both fixes: the cap goes to 45 so one cold run can finish and seed the cache, and the key hashes `go.mod`.

vshulcz:
Measured with a warm cache, and it argues against this change.

| run | windows | cache |
| --- | --- | --- |
| baseline, before this PR | ~27.9 min | setup-go, frozen |
| cold, this PR | 30m26s | miss, then saved |
| warm, this PR | **45m17s, cancelled at the cap** | hit, restored in 9s |

The restore itself is healthy — `Cache restored from key: go-Windows-7c7c62e…`, 71 MB, nine seconds — and `go vet ./...` finished in **12 seconds**, so the ordinary build cache clearly did carry over and did help.

The race build did not. `go test -race` ran from 06:17:34 to 07:02:09 and printed nothing at all before the cancellation, so all 44 minutes were compilation again, with a hot cache sitting right there. Whatever the restored 71 MB contains, it is not the race-instrumented objects that cost the time.

So the premise in #1229 holds — 19 of 28 minutes really are spent instrumenting — but this fix does not address it, and it makes the job slower rather than faster. Marking this draft rather than merging it. The two things worth keeping are already proven separately: a cancelled job never runs its post steps, and `go test -timeout` bounds test execution, not compilation, which is why these cancellations print no output and read like hangs.

Next thing to test, in a fresh attempt: whether a race build is cacheable across runners at all here, or whether the answer is to stop rebuilding it every time — split the suite so only the packages that changed get instrumented, or drop `-race` on Windows and keep it on the two fast platforms.

vshulcz:
Three attempts, measured, and only the third moves anything.

| windows leg | test step | `cmd/deja` tests | outcome |
| --- | --- | --- | --- |
| baseline | — | — | ~27.9 min |
| rotating build cache | — | — | 45m17s, cancelled |
| Defender exclusions | 26m17s | 315.9s | 26m57s |
| **no `-race`** | 20m32s | 208.1s | **21m17s** |

The Defender step applied cleanly — no exception, no access denied — and bought about a minute. So the antivirus theory is dead and the runner really is just slow.

I also had the arithmetic wrong earlier. I subtracted test time from the step and attributed the rest to race instrumentation, which put ~20 minutes on it. Measured, `-race` costs about four minutes of compiling plus 108s of slower execution. The remaining **~16.5 minutes is plain compilation** of this tree, which linux does in seconds. That is the whole story: not caching, not scanning, just the machine.

What this change buys is smaller than I expected but real: 5m40s, and headroom against the 30m cap goes from about 3 minutes to 9. The cancellations that read like hangs stop.

What it costs: no race detection on the windows leg. The argument that this is safe is that Go's race detector works against the scheduler's memory model rather than the OS, so the same race surfaces on the linux and macOS legs, both of which still run it. Windows earns its place in the matrix on paths, file locking and case rules, and plain `go test` covers those.

Leaving this draft — dropping `-race` anywhere is a coverage decision, not a CI tweak, and it should be a deliberate one. If 21 minutes on every commit is still too much, the next lever is scope rather than speed: move the windows leg to a nightly and a pre-release gate, off the per-commit path.

PR #1239, opened 2026-08-15
fix(cursor): resolve the Windows IDE root through %APPDATA%

Sora-bluesky:
Found while wiring Cursor on a Windows machine that uses it daily. `CursorUserRoot` has a darwin branch and an XDG-or-`~/.config` fallback, but no Windows branch — and the IDE writes to `%APPDATA%\Cursor\User` there. So the root resolved to a directory Cursor has never written, and every chat store stayed invisible while `doctor` reported the harness found:

```
before:  cursor  found  C:\Users\<user>\.config\Cursor\User, C:\Users\<user>\.cursor  (1 CLI transcript, 1 indexed session)
after:   cursor  found  C:\Users\<user>\AppData\Roaming\Cursor\User, C:\Users\<user>\.cursor  (1 CLI transcript, 3 stores IDE 21.0 MB, 1 indexed session)
```

Same machine, same environment, no `DEJA_CURSOR_ROOT` — 21 MB of IDE chats in three `state.vscdb` stores that the released binary cannot see. A store that looks complete while a harness's data is silently absent has the same smell as the date tokens in #1223 (different mechanism, same silence), which is why I went looking.

The fix is the resolution `roo.go` and `cline.go` already use for the same directory family: `%APPDATA%` first, `Home()\AppData\Roaming` when it is unset. Cursor was the one source that never grew the branch. Windows also stops consulting `XDG_CONFIG_HOME`, matching those two.

Tests:

- `TestCursorUserRootWindows` pins the three legs on the Windows CI leg: `APPDATA` respected, XDG ignored even when its directory exists, and the `Home()` fallback when `APPDATA` is empty.
- The two XDG assertions in `coverage_extra_test.go` are gated off Windows, where that resolution no longer applies.
- `TestSkipReasonNamesTheMissingToolOnlyWhenTheStoreIsThere` now isolates `APPDATA` alongside `HOME`: once the root resolves correctly, a developer machine with real Cursor chats broke the test's no-store premise — the same class #1158 cleaned out of the index tests.

One caveat worth knowing: `sources` tests that fake `HOME`/`USERPROFILE` mostly do not fake `APPDATA`. CI is unaffected (no Cursor on the runners), and the one test this change actually broke is fixed here; I did not sweep the rest speculatively, but a package-level guard in a `TestMain` would close the class if you want it closed.


vshulcz:
Reproduced the reasoning end to end and it holds: `CursorUserRoot` had a darwin branch and an XDG fallback and nothing for Windows, so it resolved to `~/.config/Cursor/User`, which the IDE never writes there. `doctor` reporting the harness as found while every `state.vscdb` stayed invisible is the worst shape of that bug.

Checked the claim about `roo.go` and `cline.go` too, and the resolution matches them exactly, fallback included. No recursion risk since `Home()` reads `USERPROFILE`, not `APPDATA`. Every caller goes through this one function, so discovery, `doctor` and the registry matcher are all covered by the single change.

The test skips off Windows and the CI leg runs it natively, so it is real coverage rather than a comment. Thanks for catching this on a machine we do not have.

vshulcz:
Following up properly, because my first pass short-changed this.

I reviewed the resolution change and skipped straight past `skip_test.go`, which is the more interesting half. Isolating `APPDATA` alongside `HOME` there is exactly right: once the root resolves correctly, a developer machine with real Cursor chats breaks that test's no-store premise, and the test would have started failing for people who had nothing to do with this change. Finding the second-order effect of your own fix and fixing it in the same PR is the part that is easy to leave for someone else, and you did not.

Thank you. This came off a Windows machine we do not have, and `doctor` reporting a harness as found while every store stays invisible is the worst failure shape there is — it looks like everything works. It is shipped.

Sora-bluesky:
Thanks for the follow-up. I did not plan the `skip_test.go` change, though. Once the root pointed at the real `%APPDATA%`, that test started failing against my own Cursor stores, so it had to go into the same PR. The rest of that class (sources tests that fake `HOME`/`USERPROFILE` but not `APPDATA`) is still open, as noted in the description. If a package-level guard in `TestMain` is how you want it closed, I can send that as a small follow-up.

Issue #794, opened 2026-08-01
index says nothing when a whole harness is skipped for a missing sqlite3


vshulcz:
A machine whose only store is opencode, with sqlite3 off PATH:

```
$ deja index
deja: indexing sessions into ~/idx3 ...
$ deja "davit"
deja: no matches in 0 indexed sessions — try fewer words or --re (query "davit")
```

With sqlite3 present the same store gives `deja: opencode: 1 session, 1 message`. The index run names every harness it read and says nothing about the one it could not, so an empty deja looks like an empty history — the failure mode the per-harness lines exist to prevent.

`deja sources` in the same state already words it correctly:

```
opencode  …/opencode.db  sessions=0 messages=0 size=16.0 KB  (sqlite3 CLI not found — opencode sessions unavailable)
```

Same for cursor, grok, hermes and goose, and only when the store file is actually there — a machine that never used opencode should stay quiet.

PR #1281, opened 2026-08-17
feat(zed): index the built-in agent's threads (#183)

### LINKED ISSUE #183: Add a harness parser for Zed's assistant history
Zed's built-in AI assistant stores conversation threads locally. deja indexes ten agent histories today; Zed is a natural addition for people who use it alongside a CLI agent.

**What's needed**
- A parser in `internal/sources/zed.go` modeled on an existing one (`qwen.go` or `claude.go` are the closest templates), plus `DEJA_ZED_ROOT` for relocation.
- Wiring at the usual points — grep the tree for a recently added harness name like `qwen` to find every site (discovery, doctor store check, `deja sources`, resume, search tag, registry docs/fixtures, hermetic test env lists).
- A `docs/registry/zed.md` following the other entries, a synthetic fixture, and the conformance test row.

**Where the data lives**
Zed keeps assistant state under `~/.config/zed/` (conversations in a SQLite db or JSONL depending on version — please confirm the on-disk shape on a real install and note the version in the PR). If it's SQLite, model the reader on `opencode.go`/`cursor.go` (shell out to `sqlite3`); if JSONL, on `qwen.go`.

Tests must be hermetic (temp HOME, no real Zed state) and keep the touched packages at 95% coverage. See CONTRIBUTING.md for the house rules.

css521:
Closes #183.

## The on-disk shape, confirmed

The issue asked for this to be checked on a real install and the version noted, so before the code:

**The store is not under `~/.config/zed`.** That directory exists on every platform and holds `settings.json`, keymaps and prompts — no threads. The threads are under Zed's **data** directory, per its own `paths::data_dir()`:

| Platform | Store |
| --- | --- |
| macOS | `~/Library/Application Support/Zed/threads/threads.db` |
| Linux, FreeBSD | `${XDG_DATA_HOME:-~/.local/share}/zed/threads/threads.db` |
| Windows | `%LOCALAPPDATA%\Zed\threads\threads.db` |

It is SQLite, so `opencode.go`/`cursor.go` were the right templates to point at — with one thing the issue could not have known:

```sql
CREATE TABLE threads (
    id TEXT PRIMARY KEY, summary TEXT NOT NULL, updated_at TEXT NOT NULL,
    data_type TEXT NOT NULL, data BLOB NOT NULL,
    parent_id TEXT, folder_paths TEXT, folder_paths_order TEXT, created_at TEXT
);
```

`data` is **zstd-compressed**, not JSON in a text column. `data_type` is `zstd` for everything Zed writes today and `json` only for rows predating compression.

Read off a real install (schema and paths) and Zed 0.61 at `968a13d2` (`crates/agent/src/db.rs`, `crates/agent/src/thread.rs`, `crates/paths/src/paths.rs`).

## The decision this forces, and the one I'd like checked

deja has no runtime Go dependencies and the standard library has no zstd. So the frame goes through the **`zstd` CLI**, gated by `ZstdAvailable()` exactly as five stores are gated by `SQLite3Available()`. `SkipReason("zed")` names whichever of the two is missing, so `deja sources` says

```
zed	/…/threads/threads.db	sessions=0 …	(zstd CLI not found — Zed threads unavailable)
```

rather than reporting an empty history — the #794 failure, one layer down, since `sqlite3` alone opens this store and reads nothing out of it.

**This is the part worth a maintainer's opinion:** zed is the first harness needing a second external tool. The alternatives I rejected were handling only `data_type='json'` (indexes almost nothing — Zed always compresses now) and vendoring a zstd decoder (against the no-dependencies rule). If you'd rather not take a second CLI dependency, say so and I'll rework it.

Decompression is per frame, not one batched `zstd -d` over concatenated frames. That costs a process per changed thread and buys isolation: one corrupt frame loses its own thread instead of the store. Incremental runs only decompress what moved.

## Two thread document generations

Both parse, and both had to. Zed serialises `SerializedThread { #[serde(flatten)] thread, version }`, so `version` sits beside the thread's fields rather than wrapping them. Current documents are `0.3.0` and hold Rust's externally tagged `Message` enum (`{"User":…}`, `{"Agent":…}`, `"Resume"`); agent-1 documents name the title `summary` and carry internally tagged `{"role":…,"segments":[…]}`.

Zed rewrites a thread in the current shape only when that thread is **next saved**, so a live store mixes generations indefinitely — and history from before deja was installed is the whole pitch, so dropping the old shape would quietly drop the oldest threads.

Indexing follows cline's line: user and assistant text only. Thinking, redacted thinking, tool calls, tool results, images and the `Resume` marker are skipped. A `Mention`'s inlined `content` is kept, because that is the text Zed put in front of the model.

## One bug this would have shipped with

`updated_at` is Rust `to_rfc3339` output, so it ends in `+00:00`; a Go watermark formatted with `RFC3339Nano` ends in `Z` and drops trailing fractional zeros. Comparing them as text is wrong in a direction that never surfaces: `.` sorts below `Z`, so a row newer than a whole-second watermark by half a second compares as *older* and an incremental pass would never look at it again. The filter normalises both sides with `strftime` first, and keeps a row whose timestamp SQLite cannot parse rather than assuming it was already seen. `TestParseZedDBSinceKeepsARowNewerByAFractionOfASecond` fails against the naive version.

## What is wired

- `internal/sources/zed.go` — parser, `DEJA_ZED_ROOT` and `DEJA_ZED_DB`, both schema generations (the later columns arrive by `ALTER TABLE` at startup, so a store an older Zed wrote and a newer one never opened has only the original five; there is a fallback projection).
- `internal/sources/registry.go` — harness entry, `dbParse`/`dbParseFrom` like opencode's, which is all discovery, load and incremental indexing needed.
- `internal/sources/skip.go` — the two-tool gate.
- `cmd/deja/main.go` — `deja sources` row and its missing-tool note. `cmd/deja/handoff.go` — paste-only, for Roo's reason: the agent lives in the editor, there is no CLI to hand a prompt to.
- `docs/registry/zed.md`, `registry.json`, the registry index, the generated matrix (`go run ./scripts/genmatrix`), and the harness count in `README.md`, `npm/README.md`, `docs/index.html`. `llms-install.md` said "the seventeen harnesses deja knows" about a one-command-setup list; since zed has no install target it is now "the seventeen harnesses deja installs into" rather than gaining a row it cannot honour.
- Hermetic env lists in `internal/sources/registry_test.go`, `cmd/deja/coverage_extra_test.go` and `internal/index/coverage_extra_test.go` — `DEJA_ZED_*` plus `XDG_DATA_HOME`/`FLATPAK_XDG_DATA_HOME`, without which an index run reads the contributor's own threads.

No MCP install target, so `mcp`/`auto`/`skill`/`command` are false with gap entries. Zed does run MCP servers and does read rules files, so `mcp` and `skill` are `todo`, not dead ends; `auto` is a fact about the harness — no lifecycle hook exists to inject recall into.

## Fixture

`fixtures/registry/zed/zed.sql` carries both encodings: the `json` row is readable SQL, the `zstd` row is a real frame as a hex blob. That blob is the one thing here that is opaque in review, which the registry README's "stored as SQL so changes remain reviewable" is right to want. So the SQL states the exact plaintext it decompresses to and `TestRegistryFixtureZstdBlobMatchesItsDocumentedPlaintext` pins the two together — the comment cannot drift from the bytes.

The conformance case builds the fixture on **stdin** rather than as an argument: the file opens with a `--` comment and the sqlite3 CLI reads a leading `--` as an option.

## Checks

```
go test ./... -race -count=1     all packages pass except one noted below
go vet ./...                     clean
gofmt -s                         clean
```

Per-package coverage against the CI floors, measured the way CONTRIBUTING asks:

| package | before | after | floor |
| --- | --- | --- | --- |
| `internal/sources` | 89.8% | 89.9% | 87% |
| `internal/index` | 90.6% | 90.6% | 90% |
| `cmd/deja` | 89.4% | 89.4% | 88% |

`zed.go` is 87.5–100% per function; the 87.5% is `zedRows`, whose remaining branches are sqlite3 subprocess failures, and `TestParseZedDBSurvivesASqlite3ThatDoesNotAnswerInRows` reaches most of them with a stub on PATH. Note the issue asked for 95% on the touched packages, which is above every floor those packages actually stand at — I held the floors and did not lower one; say the word if 95% is the real bar and I'll push `internal/sources` up.

`zedDataDir` takes the platform as an argument instead of reading `runtime.GOOS`, so the macOS and Windows layouts are asserted on every machine rather than only where CI happens to run.

Two things to be straight about:

- **`TestOpenClawPluginRecallsForThePrompt` fails for me, and it also fails on unmodified `main`** — `install_openclaw_plugin_test.go:154: driving the plugin: signal: killed`, the node subprocess dying on a loaded machine. Not from this change; I did not touch it.
- **I could not verify against a populated store.** My install has `threads.db` with the schema above and zero rows, so the schema, the paths and the compression are read off a real install, while the two document shapes come from Zed's source at the commit named above. Every fixture is synthetic, per CONTRIBUTING, and no real session data is in this PR. If you want a round-trip against a real thread before merging, that is a fair thing to ask for.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Zed's built-in agent is the eighteenth harness deja indexes. Its threads are one SQLite store under Zed's *data* directory — `~/Library/Application Support/Zed/threads/threads.db` on macOS, not the `~/.config/zed` that holds only settings — and every thread body inside it is a zstd frame, so reading them needs the `zstd` CLI beside `sqlite3`. With either tool missing, `deja sources` says which one rather than reporting an empty history. Both thread document generations parse, because Zed rewrites a thread in the current shape only when that thread is next saved, so a store mixes them indefinitely. (#183)
+
 
 ## [0.17.2] - 2026-08-16
 
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Install also writes user-level guidance for the harnesses it detects: Claude Cod
 
 ## What you get
 
-**Solve it in Codex. Claude remembers.** Seventeen coding agents write every conversation
+**Solve it in Codex. Claude remembers.** Eighteen coding agents write every conversation
 to local files, and deja turns those files into one memory layer all of them read.
 
 | | |
@@ -206,7 +206,7 @@ The server exposes `recall`, `recall_context`, `blame`, `fix`, `how` and `rememb
 ## Supported harnesses
 
 <!-- matrix:start -->
-Claude Code &middot; Cline &middot; Codex CLI &middot; opencode &middot; aider &middot; Gemini CLI &middot; Cursor &middot; Antigravity &middot; Grok Build &middot; Hermes &middot; Goose &middot; Qwen Code &middot; Kimi Code &middot; pi &middot; OpenClaw &middot; Copilot CLI &middot; Roo Code.
+Claude Code &middot; Cline &middot; Codex CLI &middot; opencode &middot; aider &middot; Gemini CLI &middot; Cursor &middot; Antigravity &middot; Grok Build &middot; Hermes &middot; Goose &middot; Qwen Code &middot; Kimi Code &middot; pi &middot; OpenClaw &middot; Copilot CLI &middot; Roo Code &middot; Zed.
 
 <details>
 <summary>What each one supports</summary>
@@ -230,6 +230,7 @@ Claude Code &middot; Cline &middot; Codex CLI &middot; opencode &middot; aider &
 | OpenClaw | ✅ | ✅ | ✅ | ✅ | — | paste | — |
 | Copilot CLI | ✅ | ✕ | ✅ | ✅ | ✅ | ✅ | — |
 | Roo Code | ✅ | ⚠ | ✅ | ✅ | ✕ | paste | — |
+| Zed | — | ✕ | — | — | ✕ | paste | sqlite3 + zstd |
 
 ✅ works &middot; — possible, not built yet &middot; ✕ the harness has no such mechanism &middot; ⚠ blocked by an upstream bug &middot; ? not investigated
 
--- a/cmd/deja/coverage_extra_test.go
+++ b/cmd/deja/coverage_extra_test.go
@@ -54,6 +54,13 @@ func hermeticEnv(t *testing.T) string {
 	t.Setenv("DEJA_GROK_ROOT", filepath.Join(tmp, "grok"))
 	t.Setenv("DEJA_QWEN_ROOT", filepath.Join(tmp, "qwen"))
 	t.Setenv("DEJA_COPILOT_ROOT", filepath.Join(tmp, "copilot"))
+	// Zed resolves its store through the platform data directory, which on
+	// macOS sits under the home directory but on Linux follows XDG_DATA_HOME.
+	// Both are pinned so a contributor's own threads never reach a golden.
+	t.Setenv("DEJA_ZED_ROOT", filepath.Join(tmp, "zed"))
+	t.Setenv("DEJA_ZED_DB", "")
+	t.Setenv("XDG_DATA_HOME", "")
+	t.Setenv("FLATPAK_XDG_DATA_HOME", "")
 	t.Setenv("CLAUDE_CONFIG_DIR", "")
 	t.Setenv("CODEX_HOME", "")
 	t.Setenv("GEMINI_CLI_HOME", "")
--- a/cmd/deja/handoff.go
+++ b/cmd/deja/handoff.go
@@ -259,7 +259,9 @@ var handoffAlias = map[string]string{"agy": "antigravity"}
 
 // handoffPasteOnly mirrors the registry's `handoff: paste` entries; the
 // capability drift test keeps the two in sync.
-var handoffPasteOnly = map[string]bool{"openclaw": true, "hermes": true, "roo": true}
+// Zed is paste-only for the same reason Roo is: the agent lives in the editor,
+// so there is no CLI invocation to hand a prompt to.
+var handoffPasteOnly = map[string]bool{"openclaw": true, "hermes": true, "roo": true, "zed": true}
 
 func handoffTargets() []string {
 	return []string{"claude", "codex", "opencode", "cursor", "copilot", "gemini", "qwen", "antigravity", "aider", "pi", "grok", "cline", "goose", "kimi"}
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -1855,6 +1855,7 @@ func printSources(dir string) {
 		{"roo", strings.Join(sources.RooRoots(), string(os.PathListSeparator)), sources.RooRoots(), sources.LoadRoo},
 		{"pi", sources.PiRoot(), []string{sources.PiRoot()}, sources.LoadPi},
 		{"openclaw", sources.OpenClawRoot(), []string{sources.OpenClawRoot()}, sources.LoadOpenClaw},
+		{"zed", sources.ZedDB(), []string{sources.ZedDB()}, sources.LoadZed},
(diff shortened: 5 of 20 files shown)


vshulcz:
This is excellent work — the description alone did more than most PRs do, and it held up under checking.

I ran it against a **real 30-thread Zed store** rather than only the fixture, and verified your claims rather than taking them:

- **The timestamp bug is real.** I swapped `zedSinceWhere` for the naive text comparison and `TestParseZedDBSinceKeepsARowNewerByAFractionOfASecond` fails exactly as you said — a row newer by a fraction of a second is dropped from every incremental run afterwards. Good catch, and it would never have surfaced in use.
- **The message filtering is exact.** That store holds 1670 text segments across 3059 `ToolUse`, 751 `RedactedThinking` and 15 `Mention`; deja indexed 1669 messages. Tool calls and thinking are dropped, mentions kept, exactly as documented.
- **The store shape matches**, including the compression: 30 threads, all `data_type='zstd'`, all documents `version 0.3.0` with externally tagged `{"User":…}` / `{"Agent":…}` and bare `"Resume"`.
- Titles, start times, the incremental second pass and search all behave on real data: 30 sessions, no empty titles, no missing timestamps, a re-run adds nothing.

**On the zstd dependency: yes, take it.** The alternatives you rejected were rejected for the right reasons, and the gate is what makes it acceptable — `deja sources` naming which of the two CLIs is missing is the #794 lesson applied one layer down, rather than reporting an empty history. I would not have wanted a vendored decoder.

## One bug the fixture could not show, pushed to your branch

That real store has only the **original five columns**. `folder_paths` arrives by `ALTER TABLE`, so a store an older Zed wrote and a newer one never opened does not have it — which is the fallback you already handle for the projection. But the project comes only from that column, so all thirty threads landed in the unknown-project marker:

```
30  zed  project='-'
```

That is worse than a missing label. A session with no project is invisible to auto-recall, which ranks within the project the user is working in — so on such a store the whole integration indexes fine and is then unreachable by the thing that recalls without being asked. (It is also the shape that, until yesterday, matched every hyphenated project on the machine.)

The path is in the document regardless: `initial_project_snapshot.worktree_snapshots[].worktree_path`, present in all thirty. With it the same store reads as

```
17  d3fvxl        7  marketplace-price-tracker        2  metrics_course
 2  image2video-admin-ui        2  vk-bot-detection
```

`830150c` on your branch, with a test carrying a document that has the snapshot. Your fixture's thread document has none, so the fallback had nothing to find — which is exactly why this survived review until a real store ran through it.

## Two things I checked and deliberately left alone

- `zstdDecode` and `zedRows` shell out without a context timeout, and read into an unbounded buffer. So do `opencode.go` and `cursor.go`. That is a family-wide question, not something to hold your PR for.
- Per-frame decompression costs a process per changed thread. Your reasoning — a corrupt frame losing its own thread rather than the store — is the right trade for a memory tool, and incremental runs only touch what moved.

Thank you for this. The verification you did up front, and writing down what you rejected and why, is what made it reviewable at this size.

vshulcz:
Merged — Zed is the eighteenth harness deja reads.

Thanks again. Two things about this PR I'd point at for anyone doing the next one: you verified the on-disk shape on a real install before writing code and said which version you read it from, and you wrote down what you rejected and why. At +1467 lines that is the difference between a reviewable change and a leap of faith.

The zstd gate went in as you built it. `deja sources` naming which of the two CLIs is missing is the right answer to a dependency deja cannot vendor.

css521:
Thank you — and thank you for running it against a real store instead of taking the description at its word.

`830150c` is the part I want to name, because the lesson in it is sharper than "I missed a column". I did handle the missing column: there is a fallback projection, and a test for it. What I got wrong is that the test asserted `Project == "-"` as the expected result. So I pinned the degraded path as correct rather than asking whether the integration still worked once it degraded — and ranking within the project is exactly what stops working. My own test was standing guard over the bug. My fixture had no `initial_project_snapshot`, so I never learned the field was there to fall back to.

Next time I'll remember both halves of that: a fallback needs the question "does the feature still hold after this degrades?", not only an assertion that the degrade happened; and a fixture I wrote myself can only ever confirm my own assumptions about the format.

Glad the zstd gate reads right to you. Thanks for the review — the real-store numbers were worth more than another round of mocked assertions.

Issue #1119, opened 2026-08-07
windows: internal/index tests using unix path fixtures fail (silently unvalidated until timeout fix)


vshulcz:
Raising the CI test timeout from 6m to 12m let `test / windows-latest` finally run `internal/index` to completion — it had been dying in `cmd/deja` first, so internal/index was effectively unvalidated on windows for a while.

Nine tests fail on windows (pass on macOS + ubuntu). They seed transcripts with unix-style `"cwd":"/proj"` and flat/artificial store layouts, and deja derives session project + attribution from file paths via `filepath.Rel`/`Separator` and a lexicographic `s.Path < held.Path` comparison — OS-dependent. Predominantly test-portability, but one is unproven:

- TestHarnessSharedCounts / TestDateTokensMakeSessionsFindableByMonth / TestBoundarySurvivesIncrementalIngest / TestUnrelatedPushDoesNotWipeOtherBoundaries / TestExportWatermarksAreScopedPerPeer / TestExportKeepsMessagesSharingTheWatermarkTimestamp / TestImportedRecordsSurviveRebuildAndUpdate / TestRecentCarriesTouched
- TestSyncImportErrorKeepsThePathOnOneLine — windows yields a raw `CreateFile` error instead of the worded `no such directory`.

**Unproven, needs a windows repro:** `TestHarnessSharedCounts` gets `session count = 2, want 1` — two files with the same content `sessionId` should collapse to one manifest key (`claude:<id>`), and the id override is OS-independent in the code, yet windows doubles the count. Could be a real windows path-handling bug in the loader, or another fixture artifact. Do not dismiss without a windows run.

Skipped on windows for now (this issue) so windows CI validates the rest of internal/index; fix with windows CI to verify each.

vshulcz:
Windows CI on main is green, and `internal/index` ran to completion to get there.

What was in the way was not the nine fixtures. `cmd/deja` was hitting the
twelve-minute alarm again, and the dump named the test still running after 9m42s:

```
panic: test timed out after 12m0s
	running tests:
		TestUsageDocumentsSearchFlags (9m42s)
```

That test swapped `os.Stdout` for a pipe to capture `printUsage()`, which only
formats a 4 KB string. Reading the string directly instead (#1365) removed the
hang, and with it the reason the rest of the package went unvalidated.

From the run on main after that landed:

```
ok  github.com/vshulcz/deja-vu/cmd/deja        220.114s
[...]

PR #1286, opened 2026-08-17
test(windows): let windows CI say which of these still fail

vshulcz:
Working #1119, and using the windows machine we have.

Nine `internal/index` tests are skipped on windows. Eight are described as fixture portability; the ninth is not:

> **Unproven, needs a windows repro:** `TestHarnessSharedCounts` gets `session count = 2, want 1` — two files with the same content `sessionId` should collapse to one manifest key, and the id override is OS-independent in the code, yet windows doubles the count. Could be a real windows path-handling bug in the loader, or another fixture artefact. Do not dismiss without a windows run.

I read the path: the id starts as the filename base and is replaced by the `sessionId` in the content, which is the same on every OS, and the manifest key is `harness:id`. `attributeSession` compares `s.Path < held.Path`, which differs by separator but still yields one winner and one key. I cannot see where a second row comes from, and I cannot run windows to find out.

CI can. The skips are removed so `test / windows-latest` reports what actually fails instead of reporting nothing and calling it green.

**This PR is expected to go red**, and that is the point. I will read the failures, fix what is fixture and diagnose what is not, and land the result here — the final state of this branch will be green or precisely skipped with a reason per test, not a blanket.

Local runs are unchanged: `internal/index` passes on macOS as before.

### DIFF
--- a/internal/index/import_before_index_test.go
+++ b/internal/index/import_before_index_test.go
@@ -17,7 +17,6 @@ import (
 // A local session and the imported records must both be reachable after a
 // plain import-then-index.
 func TestImportBeforeFirstIndexKeepsLocalSessions(t *testing.T) {
-	skipWindowsPortability(t)
 	tmp := t.TempDir()
 	claudeRoot := filepath.Join(tmp, "claude")
 	proj := filepath.Join(claudeRoot, "-tmp-local")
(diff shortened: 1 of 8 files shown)


vshulcz:
Superseded by #1292. This branch did its job: with the skips off, windows came back with ten failures in `internal/index`, all one root cause — `t.Setenv("HOME")` isolates nothing there, so every test was sharing the package's home and reading each other's notes. The fix and the unskips live in #1292.

Issue #719, opened 2026-08-01
show suggests a longer prefix when the ids are identical and only --harness can help


vshulcz:
Two sessions from different harnesses can carry the same id — deja's own #698 is about this, and it happens naturally when a transcript is copied between tools:

```
$ deja show abc12345
deja: 2 sessions start with "abc12345" — showing the most recent; use a longer prefix for another
# qwen · proj/z · abc12345
```

There is no longer prefix: the two ids are the same string. The advice cannot be followed, and the thing that does work — `--harness claude` — is not mentioned.

```
$ deja show abc12345 --harness claude
# claude · proj/p · abc12345
```

PR #1305, opened 2026-08-17
fix(mcp): the server took the blocking index lock it had just avoided

vshulcz:
Night hunt, iteration 14, from the loose end left by iteration 13 — a lock-wait notice appeared on a path designed never to wait, and reading the code twice did not find it. Instrumenting `LockWaitNotice` did, first try:

```
index.lockDir → index.Ensure → main.findByPrefix → main.recallTextResult
```

`recallTextResult` opens with `EnsureForSearchStale`, which refuses to wait: it serves the snapshot on disk and hands the rebuild to a detached warmup, because rebuilding inline blows the client's tool timeout. Then, to print what the best session concluded, it read that session through `findByPrefix` — the CLI helper, which opens with a blocking `index.Ensure`. So the careful part was undone a few hundred lines below it. Waiting is the good case: if the lock happened to be free, the server ran the whole rebuild on the thread answering the client.

Three call sites, not one — `recall`, `recall_context`, and `resources/read`. The reviewer found the third; my first guard read only `mcp.go` and could not see it.

Also resolves by identity rather than by bare id. `FindByPrefix` matches the id across every harness and answers with the newest match, so the "full story" printed for a hit could be a different session sharing its id — the case #719 already documents. The hit knows its own harness, and `attachAnswers` in the same file was already doing it this way.

The guard is on the source, because the behavioural version needs a second process holding the lock at the right instant. It scans every `mcp*.go`, refuses to pass if it scanned fewer than two, checks that `findByPrefix` still exists before trusting a green result, and matches the bare name so `findByPrefixHarness` — which calls it and blocks identically — cannot slip through. All three ways of reintroducing the bug fail it.

Left for #1306, which the same review turned up: `resources/list`, `fix` and `how` still call the blocking `index.Ensure` and can rebuild inline, and two of them skip the "deja is indexing" guard so a cold index answers with a raw errno.

### DIFF
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -517,6 +517,25 @@ func recallText(dir, q, harness string, limit, budget int) (string, error) {
 	return text, err
 }
 
+// wholeSessionForMCP reads one session for a caller that must not wait.
+//
+// `findByPrefix` is the CLI helper and opens with a blocking `index.Ensure` —
+// right for a command someone typed, wrong here. Every caller below has already
+// been through `EnsureForSearchStale`, which deliberately refuses to wait: it
+// serves the snapshot on disk and hands the rebuild to a detached warmup,
+// because rebuilding inline blows the client's tool timeout. Calling the
+// blocking version afterwards undid exactly that — and if the lock happened to
+// be free it would run the whole rebuild on the server's thread rather than
+// merely wait for it.
+//
+// By identity rather than by prefix, too: `FindByPrefix` matches the id across
+// every harness and answers with the newest match, so the "full story" printed
+// for a hit could be a different session that shares its id — a real case, not
+// a theoretical one (#719). The hit knows its own harness; use it.
+func wholeSessionForMCP(dir string, s model.Session) (model.Session, bool, error) {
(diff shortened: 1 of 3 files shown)


vshulcz:
Follow-up measurement, and it is stronger than what this PR claimed.

The claim here was hedged — that the wait depends on timing — because the probe I used to test it passed `index --force`, which is not a flag. It failed instantly, so the "rebuild running in the background" was never running, and the measurement showed no blocking. Found while stress-testing concurrency two iterations later, when the same bad flag showed up in a different script.

With a real rebuild (`index --rebuild`), on the 16000-session store:

```
before this fix:  mcp recall during a rebuild   1.763s   (+ "another deja is building the index — waiting")
after this fix:   mcp recall during a rebuild   0.173s   (no wait)
baseline, idle:                                 0.235s
```

So it was a 10x spike on the agent path for the whole duration of any rebuild, not a timing-dependent maybe. The after-figure being below the idle baseline is expected: it serves the snapshot on disk instead of re-checking freshness.

Worth knowing for #1306, where the same call shape is still present in `resources/list`, `fix`, `how` and `blame` — those numbers should be assumed to be the before-figure, not the hedge.

Issue #1309, opened 2026-08-17
upgrade day: the first agent call rebuilds the index inline, with nothing said


vshulcz:
A version bump makes the previous index unreadable, so the first call after an upgrade rebuilds it. That part is unavoidable — there is no stale answer to serve when the format changed. What is avoidable is doing it inside the agent's tool call without saying so.

16000-session store, index written by the previous version:

```
first agent recall after the upgrade   2.86s
the one after that                     0.22s
```

The answer that comes back is correct and carries no sign that anything unusual happened. On a store several times this size that is tens of seconds inside one tool call, against MCP clients that have timeouts — so the failure mode on upgrade day is not "slow once", it is "the memory tool appears broken once", on exactly the day a user has reason to look at it.

deja already has the sentence for this. `buildingNowForAgent` answers "deja is indexing this machine's history … ask again then", and every surface uses it. It does not fire here because it opens with `index.HasManifest(dir)` — and on upgrade day the manifest exists, it is simply of a version this build cannot read. The one state the message was written for is the one its condition excludes.

Two things needed together, which is why this is not a patch:

- The condition wants to be "is there a usable index" rather than "is there a file" — a manifest whose version this build cannot read is not one it can answer from.
- It has to be consulted *before* `EnsureForSearchStale`, not after. Today the rebuild happens first and the guard runs when there is nothing left to report — the same ordering trap as #1306, where adding the guard alone produces a branch that cannot be reached.

The CLI path is fine and worth keeping as the model: it prints `indexing sessions into …` and the person waits knowing why. `doctor` is better still — `format written by an older deja — this build cannot read it; the next session rebuilds it, or run deja index now`. Only the agent is left guessing.

Found by the night hunt, iteration 27.

Issue #1049, opened 2026-08-06
imported note loses its lifecycle state on any full rebuild


vshulcz:
`sync import` records the state of a promoted note that arrived from another machine (#975): `orig_id`, `lifecycle`, `lifecycle_note`, `lifecycle_at` on the manifest row. Neither survives a full rebuild, and a row written before #984 never gets them at all.

Store C, imported by this build, before and after `deja index --rebuild`:

```
== before rebuild
[deja] imported:work/alpha · today · imported-…5e925396d6 — 3 matches
  tried and rejected (2026-08-06): we no longer invalidate on write (from claude:sess-alpha, 2026-08-06)
  [rejected] we no longer invalidate on write (from claude:sess-alpha, 2026-08-06)
deja: indexing sessions into /tmp/audit-N/n17c/C/idx ...
== after rebuild
[deja] imported:work/alpha · today · imported-…5e925396d6 — 3 matches
  [rejected] we no longer invalidate on write (from claude:sess-alpha, 2026-08-06)
== last --json after rebuild
{"id":"imported-315e925396d6","harness":"deja","project":"imported:work/alpha", ... }
```

`metaForSession` builds the row from scratch on a rebuild and does not copy the four fields, so any full rebuild — `index --rebuild`, a corrupt bucket, a format bump — drops the state. The batch is already deduped, so re-importing it does not bring it back: `deja: imported 0 records`.

Same store built by a binary from before #984 (8f7c2b5), then read by this build:

```
== search 'no longer invalidate'
[deja] imported:work/alpha · today · imported-…5e925396d6 — 3 matches
  [rejected] we no longer invalidate on write (from claude:sess-alpha, 2026-08-06)
== deja index (new)
deja: index is up to date (2 sessions)
== re-import same batch (new)
deja: imported 0 records
== deja index --rebuild
== search after rebuild
[deja] imported:work/alpha · today · imported-…5e925396d6 — 3 matches
  [rejected] we no longer invalidate on write (from claude:sess-alpha, 2026-08-06)
== doctor
  status   built (size=2.3 KB, updated=2026-08-06 17:39)
  freshness up to date
```

The same batch imported by this build prints `tried and rejected (2026-08-06): …` and carries the fields in `last --json`. #984 changed what goes into a manifest row without bumping the index format, and nothing re-derives the row, so a decision retracted on the other machine reads as accepted here forever — the case #975 fixed.

[...]

Issue #855, opened 2026-08-01
forget is the one command that does not accept the id printed on screen


vshulcz:
After #853 every command resolves the elided id a result line prints. `forget` does not:

```
$ deja anemometer
[claude] f5/proj · Jul 1 · a1b2c3d4e…d6e7f8a9b0 — 1 matches

$ deja show   'a1b2c3d4e…d6e7f8a9b0'  → # claude · f5/proj · a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a9b0
$ deja ctx    'a1b2c3d4e…d6e7f8a9b0'  → # deja context: claude · f5/proj · a1b2c3…
$ deja resume 'a1b2c3d4e…d6e7f8a9b0'  → cd '/tmp/f5/proj' && claude --resume a1b2c3…
$ deja promote 'a1b2c3d4e…d6e7f8a9b0' --state accepted --note x
                                        → promoted claude:a1b2c3d4e5f6… as accepted

$ deja forget --session 'a1b2c3d4e…d6e7f8a9b0' --dry-run
dry run — nothing was changed
would drop: 0 session(s), 0 message(s)
$ deja forget --session a1b2c3d4 --dry-run
would drop: 1 session(s), 1 message(s)
$ deja forget --unforget 'a1b2c3d4e…d6e7f8a9b0'
no tombstone matches "a1b2c3d4e…d6e7f8a9b0"
```

`forget` matches through its own selector in `internal/index/privacy.go` rather than the shared id predicate, so it missed the widening. Of all the commands to silently match nothing, this is the worst one: the reader copies an id off the screen, sees "nothing was changed", and has no reason to think the id was the problem rather than the session already being gone.

Issue #853, opened 2026-08-01
the id printed in a search hit cannot be pasted back: the ellipsis is part of it


vshulcz:
A long session id is elided in the middle of a result line, and copying exactly what is on screen into any command that takes an id fails:

```
$ deja anemometer
[claude] t14b/proj  · Jul 1 · a1b2c3d4e…d6e7f8a9b0 — 1 matches
  the anemometer mast bracket weld cracked

$ deja show 'a1b2c3d4e…d6e7f8a9b0'
deja: no session matches "a1b2c3d4e…d6e7f8a9b0"
```

#707 recognised this shape — "what a reader copies out of a search is a head and a tail rather than a prefix" — and added a substring fallback to `FindByPrefix`. The fallback cannot help here, because the string the reader copies contains `…`, which appears in no id: neither the prefix match nor the substring match can ever hit.

Short ids paste fine; this only bites the harnesses whose ids are long (the fixture above uses a 40-character one, which is what Codex thread ids look like).

PR #1316, opened 2026-08-18
fix(index): an id a session was synced under resolved to nothing

vshulcz:
Night hunt, iteration 40. Sync rewrites a session's id on import and keeps the old one in `OrigID`, deliberately — #1049 added it because rebuilding the row from scratch "dropped what only the import knew … the id it had on the machine it came from". Nothing then looked at it.

So the id a user reads on one machine does not work on the other:

```
deja show faraway1      deja: no session matches "faraway1"
deja resume faraway1    deja: no session matches "faraway1"
deja ctx faraway1       deja: no session matches "faraway1"
deja last --json        {"id": "imported-bfce745b3871", "orig_id": "faraway1"}
```

The last line is the same command telling the reader that id belongs to a session it holds.

`resume` turns out to have had the right answer written for it all along — `session imported-…ce745b3871 was synced from another machine — resume it there` — which could never fire, because the lookup never resolved the id. Same shape as #1309: the sentence exists, the predicate excludes the case.

The fallback runs after the exact-id pass and before the loose substring pass, so a local id still wins outright, and typing an imported id in full is not beaten by a local id that merely contains it.

`PrefixMatches` and `PrefixHarnesses` learned the same form. That is not tidiness: their own comment (#853) says the count and the resolver have to agree, "or a reader is told an id matches nothing and then watches it open" — resolving without counting is that failure with the sign flipped, and it would have silently swallowed the cross-harness case where two peers each had `s1`.

Left alone deliberately: `Forget`, `Unforget` and `FindByIdentity` still match the local id only. Destructive and exact-identity selectors are strict on purpose (#855), so `deja forget --session <original-id>` still reports nothing while `show` now works. Worth revisiting, but widening a destructive selector is not a change to make in the same breath as fixing a read.

Three mutations fail the test: dropping the fallback from the resolver, dropping it from the count, and — via the added case — a local session whose id contains the imported one. The fixture does a real export/import round trip and asserts import actually renamed the session before testing anything else, since staging that row by hand would only assert my own idea of what import does.

### DIFF
--- a/internal/index/find_orig_id_test.go
+++ b/internal/index/find_orig_id_test.go
@@ -1,0 +1,108 @@
+package index
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// Import rewrites a session's id, and OrigID keeps the one it had on the machine
+// it came from (#1049). Nothing looked at it, so a reader who saw an id on one
+// machine and typed it on another was told no session matches — about a session
+// deja holds and prints that very id for under `--json`. resume even has a
+// sentence written for the case ("synced from another machine — resume it
+// there") that could never fire, because the lookup never resolved the id.
+//
+// Done as a real round trip rather than a hand-staged manifest: the id is only
+// rewritten by import, so staging the row by hand would assert against my own
+// idea of what import does.
+func TestFindByPrefixResolvesTheIdTheSessionCameWith(t *testing.T) {
+	const origID = "faraway1"
+	line := `{"type":"user","sessionId":"` + origID + `","cwd":"/w","timestamp":"2026-09-01T09:00:00Z","message":{"role":"user","content":"the thrumbleknot migration"}}` + "\n"
+
+	// Machine A: index it, export it.
+	dirA, tmpA := syncPeerIndex(t, line)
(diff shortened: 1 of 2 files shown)


vshulcz:
Checked the other half of the asymmetry this creates, since it is worth having on the record rather than left to a reader to discover.

`deja forget --session <original-id>` still matches nothing after this change, and says so plainly:

```
nothing matched session "faraway1" — no session was dropped
```

That is honest — it names what did not happen, so nobody walks away believing the session was dropped — and the local id works: `forget --session imported-bfce745b3871` reports `sessions dropped: 1`. So the strictness of destructive selectors (#855) holds and nothing is silently missed.

What is now odd is the pair: `deja show faraway1` opens it and `deja forget --session faraway1` says nothing matched, on the same machine in the same minute.

A small improvement that stays within the strict rule, for whoever wants it: when the selector matches an `OrigID` but not an `ID`, have that same line name the local id — "nothing matched session \"faraway1\"; it was synced from another machine and is indexed here as imported-bfce745b3871". No widening of what forget accepts, just pointing at the id that works. Left out of this PR because it is new wording rather than a repair.

PR #1425, opened 2026-08-19
feat(zed): deja install zed

vshulcz:
Zed's agent panel reads MCP servers from `context_servers` in `~/.config/zed/settings.json`, but `deja install` had no target for it — Zed was the one harness where you had to hand-edit JSON to get recall.

That file is JSONC: comments and trailing commas, which `encoding/json` cannot read and re-marshalling would erase. So the installer edits the text in place, splicing the entry in and cutting it back out, and leaves every byte the reader wrote alone.

Stacked on #1424; review that one first.

Seven mutations run against the scanners — all caught, including the two that only knew `//` and lost their place inside a `/* */` comment (that was a real bug the reviewer found, fixed here).


vshulcz:
Closed by GitHub when the base branch merged. Continues as #1428.

PR #1436, opened 2026-08-19
fix(recall): say nothing when the working directory is not a project

vshulcz:
An agent started from a home directory gets project candidates `[shulcz, Users/shulcz]`. That scope holds everything ever launched from there and does not hold the repository being worked on, so ranking "this project's sessions" ranks unrelated work against each other and picks the best of a catch-all.

Measured on a real store, same prompt, same index, only the binary differs:

```
before  deja-vu — you have been here: "Ночной loop deja-vu, продолжение. Итерация 4 (на…"
after   (nothing)
```

The injected session was about a different piece of work entirely. The answer the prompt was actually about lived in a project the scope never covered.

Silence is both the right answer and the cheap one: a wrong injection costs tokens twice — once to carry it, and again when the agent learns to stop reading them.

The rule is deliberately narrow: the home directory itself and a filesystem root. A directory that is not a git repository is still somebody's project and is left alone.

Benchmarks, before → after: real 11/12 precision 1.00 → unchanged, negative false_fires 0 → unchanged, marathon and fresh unchanged, `bench recall` 1.00/1.00 unchanged, `bench context` 294 tokens at coverage 1.00 unchanged.

`bucket_scope` still reports 1 false fire, and that is honest: that arm models the other half of the problem — a scope with a legitimate project name that nonetheless holds unrelated work sharing ordinary words with the question. A cwd rule cannot answer that one; term selection or a stronger bar has to.

5 mutations, all caught. Two were not, at first: removing the gate entirely changed nothing until the fixture gained a session recorded *in* the home directory — without it the end-to-end test could not fail — and an `if cwd == ""` branch turned out to be unreachable, since `filepath.Clean("")` is `"."` and `"."` is its own parent. The branch is gone rather than left as code no test could distinguish.

### DIFF
--- a/cmd/deja/hook_prompt.go
+++ b/cmd/deja/hook_prompt.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
@@ -19,6 +20,7 @@ import (
 	"github.com/vshulcz/deja-vu/internal/index"
 	"github.com/vshulcz/deja-vu/internal/prompt"
 	"github.com/vshulcz/deja-vu/internal/search"
+	"github.com/vshulcz/deja-vu/internal/sources"
 	"github.com/vshulcz/deja-vu/internal/usage"
 
 	"github.com/vshulcz/deja-vu/internal/atomicfile"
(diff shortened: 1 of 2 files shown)


vshulcz:
Closing this: the evidence in the description does not hold.

The "before" output I quoted — a July 21 session recalled from a home-directory scope — turns out to be *relevant*. That session contains the instruction for the very thing the prompt was about:

> **1. awesome-claude-code** (самый ценный, 50k★) — если ещё не отправил: Открой: .../issues/new?template=recommend-resource.yml

So the home-directory scope produced a good recall in exactly the case I used to argue it produces bad ones. The separate wrong injection I saw in a live session (an unrelated code review) also came from that scope — which means the scope yields both, and I have no measurement of the ratio.

Silencing it outright would have suppressed the answer I was looking for. That is a worse failure than the one it fixes, and the premise "a home directory tells you nothing about what is being worked on" is now unproven rather than measured.

Reopening this needs a number: of the injections made from a catch-all scope, what share are useful. Until then the change is not justified.

PR #1448, opened 2026-08-19
fix(recall): stop guessing which words are specific by counting letters

vshulcz:
Two fixes to what recall counts as specific, and the corpus that made them measurable.

**The corpus.** `bench prompt` had about thirty sessions. At that size a word living in one of them scores idf 2.89 and a word living in ten scores 1.19; on a real store of ~1500 sessions the same words score 4.13 and 0.4. Measured on the small corpus, raising the informative floor from 2.0 to 3.0 took the benchmark from 11/12 to 0/12 and changed nothing at all on a real store — so nothing about rarity could be judged here. This adds 300 short sessions of ordinary working talk, each drawing two phrases from a pool of twenty, which puts a working word in about one session in twenty. `branch` really sits in 94 of ~1500.

**`hasIdentifierTerm` counted letters.** Any word of six letters was treated as naming something, so `paste the output of that command` and `walk me through the trace` recalled. The index has already answered the question that proxy is guessing at, so `recallWorthShowing` now asks it: a session is worth showing when it matched a term the index finds rare. The length rule stays in `promptTermsWorthAsking`, which runs before the store is opened and has nothing better.

**A session could answer half a question.** `did the kestrel timeout ever delay an escrow release` won on `kestrel` and `timeout` while `escrow` and `release` were decided in a different session entirely — the shape of a question that joins work nobody joined. The ranking now reports how many identifying words this project has an answer for anywhere, and a session is only worth showing when it accounts for all of them.

**The stem fold was decided by the whole store.** Whether `write` also tries `writes` depended on the exact form existing *anywhere* in the index. Put `write` in an unrelated project and `how often do we write parquet?` stops matching the session that says `parquet writes are batched per hour` — a question missing the one session that answers it because of work in another repository. The fold now looks at the sessions being ranked.

```
                     before    after
real_questions        11/12     12/12
negative_controls       0 f       0 f     (against a corpus that now holds working noise)
absent_subject          3 f       0 f
off_topic               3 f       1 f
shown_line            14/14     15/15
haystack (correct)      3/3       3/3
russian, marathon, fresh, bucket — unchanged
bench recall @5/@10   1.00/1.00 unchanged
bench context         294 median tokens, coverage 1.00, unchanged
live (45 real prompts, frozen index)   42 fired, 88% — identical to main
```

Live is unchanged either way: ordinary Russian words score 3.5 on a real store, so they clear the rarity bar there regardless. The gain here is that the benchmark now reads the defect instead of hiding it, and two proxies are gone.

Three unit fixtures needed a background before rarity could mean anything in them: in `TestHookPromptSkipsMarathonSessions`, `quetzalcoatl` sat in both sessions of a three-document index, giving it idf 0 — the most specific word imaginable, scored as filler. With a hundred ordinary sessions behind it the same word scores 3.53 and the working phrases 2.24.

On a real store the coverage rule blocks nothing: traced over 42 live recall decisions, the winning session accounted for every identifying word every time. It costs no recall there and removes a third of the false fires here.

Mutations, all caught: restoring the length rule gives 3 false negatives and 3 on `absent_subject`; restoring the `matched >= 2` fallback gives 4; restoring the store-wide fold drops `haystack` to 2/3; dropping the coverage rule puts `off_topic` back to 3 false; shrinking the fixture background to 3 sessions makes the marathon test fail again. The first draft of the stem-fold test passed under its own mutant and was rewritten to assert the match count rather than the order.

### DIFF
--- a/cmd/deja/bench_prompt.go
+++ b/cmd/deja/bench_prompt.go
@@ -306,13 +306,13 @@ func measurePrompt(seed int64) (promptReport, error) {
 // opening line came from the top of a long transcript does not, and that line
 // is the whole frame an agent reads before deciding to ignore the rest.
 func shownLineCarriesATerm(dir, project string, terms []string) bool {
-	ranked, matched, strong, err := index.ProjectRelevant(dir, []string{project}, terms, 8)
+	ranked, matched, strong, covered, err := index.ProjectRelevant(dir, []string{project}, terms, 8)
 	if err != nil {
 		return false
 	}
 	var keep []model.Session
 	for i, s := range ranked {
-		if !recallWorthShowing(terms, matched[i], strong[i]) {
+		if !recallWorthShowing(terms, matched[i], strong[i], covered) {
 			continue
 		}
 		keep = append(keep, s)
@@ -349,7 +349,7 @@ func promptBenchProbe(dir, project, chainID string, terms []string) (fired, corr
 	if !promptTermsWorthAsking(terms) {
 		return false, false
 	}
-	ranked, matched, strong, err := index.ProjectRelevant(dir, []string{project}, terms, 8)
+	ranked, matched, strong, covered, err := index.ProjectRelevant(dir, []string{project}, terms, 8)
 	if err != nil {
 		return false, false
 	}
@@ -357,7 +357,7 @@ func promptBenchProbe(dir, project, chainID string, terms []string) (fired, corr
 		// The same bar the hook applies, from the same function — kept in one
 		// place because the two drifted: this one asked whether the query held
 		// an identifier, the hook asked whether the session did.
-		if !recallWorthShowing(terms, matched[i], strong[i]) {
+		if !recallWorthShowing(terms, matched[i], strong[i], covered) {
 			continue
 		}
 		if len(s.Messages) > dejaVuMaxMessages {
--- a/cmd/deja/bench_prompt.go
+++ b/cmd/deja/bench_prompt.go
@@ -209,7 +209,7 @@ func measurePrompt(seed int64) (promptReport, error) {
 		}
 		// Filler that only exists to fill the bucket has no question of its
 		// own; it is asked about through the bucket-answer chain below.
-		if chain.Kind == "bucket" || chain.Kind == "haystack-noise" {
+		if chain.Kind == "bucket" || chain.Kind == "haystack-noise" || chain.Kind == "background" {
 			continue
 		}
 		terms := prompt.Terms(chain.Question)
--- a/cmd/deja/bench_prompt_test.go
+++ b/cmd/deja/bench_prompt_test.go
@@ -29,7 +29,7 @@ func TestPromptBenchCorpusIsMeasurable(t *testing.T) {
 		// asked about through the bucket-answer chain, whose question is the
 		// one that must go unanswered. A chain nobody asks about still has to
 		// earn its place, and this one earns it by being the wrong answer.
-		if c.Kind == "bucket" || c.Kind == "haystack-noise" {
+		if c.Kind == "bucket" || c.Kind == "haystack-noise" || c.Kind == "background" {
 			continue
 		}
 		if topics[c.Topic] {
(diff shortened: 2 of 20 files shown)


vshulcz:
Second commit fixes a mistake in the first: the background stepped by 13 over 26 phrases, and gcd(13,26)=13 reaches two of them. The rest stayed rare, so the benchmark scored ordinary words as words that identify something.

With the stride fixed, printing the candidates names both false paths in the gate:

```
length >= 6 in hasIdentifierTerm:  negatives 0 -> 3 false, absent_subject 0 -> 3
matched >= 2 fallback:             negatives 0 -> 4 false
```

Dropping both (`recallWorthShowing` = `strong >= 1`, length kept only in `promptTermsWorthAsking`, which runs before the store is opened) gives real 11/12 -> 12/12, negatives 0, absent_subject 3 -> 0, shown 14/14 -> 15/15, and 42/45 fired at 88% live — identical to main.

Still not in this branch, because three unit tests go silent under it. Their fixtures are three documents wide: in `TestHookPromptSkipsMarathonSessions`, `quetzalcoatl` sits in both sessions, so its idf is 0 and nothing there can be rare. They fail on a correct fact, and the promise they encode is real, so they need a background the way the benchmark corpus just got one — not a weaker assertion.

vshulcz:
Measured the other side of these changes on a wider live sample — 113 real prompts against a frozen index, six per session file rather than one, because one-per-file was only reaching 45 cases and flattering the baseline.

```
                 main     branch
fired           110/113   94/113
first line carries a term
                 80/110    71/94
                    73%       76%
```

16 recalls go quiet. I read all 16; every one was an unrelated neighbour:

```
"давай посмотрим логи на проде"        -> "1. Стратегический рефрейм всей категории"
"какая у нас зависимость от яндекса"   -> "не, пры желательно красиво расписывать"
"какую usb железяку надо купить"       -> "делай задачи в next.md"
"python3 audio-shrink.py --codec mp3"  -> "On quality: it was a wash, not a clear Exa win"
"sing box ни в коем случае не вырубай" -> "продолжай DDD-рефактор"
"отревьювь .../pull/879"               -> a review of a different PR
```

No answer was lost — 15% fewer injections, all of them ones an agent read for nothing.

vshulcz:
Do not merge yet. The live comparisons behind these commits were measured against an index that was not actually frozen — the hook rebuilds in place and `swapIndexDir` replaces the directory, so a `chmod -R a-w` copy does not survive, and my own loop transcripts were being indexed into the store I was measuring against.

Rebuilt the instrument: sources snapshotted (my own worktree transcripts excluded), every harness root pinned at a nonexistent path, verified the manifest mtime no longer moves across hook runs. On that index:

```
        fired      block opens on a term used   headline agrees
main    109/129    100/109 = 92%                63/100 = 63%
branch   30/129     30/30  = 100%               30/30  = 100%
```

Quality is perfect and coverage collapsed. Reading the losses, roughly half are real:

```
"проверь в deja, обсуждали ли мы connection pool exhausted"
   main recalled the session about connection pool exhausted; the branch is silent
"кстати реально давайте выбор агенту по вырезкам из сессии"  -> the same discussion, silenced
"я хочу форкнуть текущую сессию и запустить поиск багов в loop" -> the same prompt, silenced
```

The other half are the neighbours these changes were built to remove ("продолжай DDD-рефактор" under a question about database backups).

So the gates are too strict as a set. Next step is to measure each rule's share of the drop and keep only the ones that earn it. The benchmark work and the byte-vs-letter fix stand on their own; the coverage and scope rules are what need re-deciding.

vshulcz:
Superseded in part by #1449, which carries the three fixes that hold up on a real store: the per-project stem fold, the headline naming terms the block carries, and the one-word bar counting letters instead of bytes.

What stays here and why. Bisecting the branch on a single prompt found the cause of the coverage collapse:

```
"проверь в deja, обсуждали ли мы connection pool exhausted"
main       FIRED   the session that discussed it
45018d9c   silent  <- the `strong >= 1` bar
```

On a store of ~1500 sessions that session matches with `strong = 0` — those words are not rare enough to clear the bar. The benchmark cannot see it with any amount of background. Measured separately, the coverage and scope rules cost another 29 recalls of 112 and buy nothing (block quality 89% vs 88%).

So the corpus here still shows something real — with ordinary working noise present, the current bar fires on 5 of 10 negative controls — but the fix it suggested is wrong. Leaving this draft open with the corpus and the failing guard rather than weakening the assertion.

vshulcz:
Split: the corpus half is #1535 on current main, and the rest is #1534 to re-measure. The gate this branch rewrites has since moved to `internal/search/worth.go` and lost the third parameter — wiring it on the old corpus fired 2 of 12 negative controls, which is the same too-small-to-judge problem the corpus commits fix. Keeping this open until #1535 lands, then closing it in favour of #1534.

vshulcz:
Superseded: the corpus is on main (#1535) and the remaining rules are #1534, to be re-measured against the code as it is now.

PR #1450, opened 2026-08-20
wip(recall): rarity is measured in whole sessions, and that inverts it

vshulcz:
Draft — three unit tests fail and the question behind them is worth deciding before this goes further.

Rarity is measured in whole sessions: a term's document frequency is the number of sessions holding it. On a real store that inverts the signal. Eleven sessions of twenty to sixty thousand messages hold 99% of everything ever said, and every one of them mentions the ordinary words, so:

```
term          df/sessions   idf
pgbouncer          16      1.31
branch             13      1.50
omoda               5      2.35
suite               7      2.06
```

`pgbouncer` reads as commoner than `branch`. The most specific word in the store ranks below the most ordinary one — the signal is not weak, it is upside down. Every threshold on it is therefore mis-set, which is what a night of measurements kept running into: a rule that reads 12/12 on the benchmark cut live recall from 109 blocks to 30.

Counting messages instead brings back the failure session-counting exists to avoid — one marathon repeating a word three hundred times would make it common. Capping each session's contribution keeps both. Measured on a real store, a cap of 25 reproduces what treating each 200-message episode as a document gives:

```
term          cap 1   cap 5   cap 25   200-msg episodes
pgbouncer      1.31    2.16     3.40      3.37
omoda          2.35    3.03     4.09      4.42
singbox        2.53    2.85     4.17      4.70
branch         1.50    1.85     2.18      1.27
suite          2.06    2.27     2.53      2.07
```

Subjects land above 3, working words below 2.5 — the separation falls on the strong floor that was already in the code. The threshold was never the problem; the unit was.

Live, against a frozen index of 129 real prompts:

```
                                    main    branch
fired                              109/129  126/129
block opens on a term the product used  92%      99%
```

Both up. This is the only change measured tonight that improves coverage and quality at once. Mutation: setting the cap to 1 reproduces main exactly — 109/129 and 92%, to the number.

`bench prompt` every arm unchanged, `bench recall` 1.00/1.00, `bench context` unchanged. No manifest change and no reindex: the message count per session is already stored, and a manifest written before it degrades to one document per session, which is today's behaviour.

**What is unresolved.** Three tests fail, all on fixtures where one marathon session repeats a single line:

```
TestHookPromptSkipsMarathonSessions          quetzalcoatl in 310 of 311 messages
TestHookPromptNamesTheEarlierAttempt
TestToolHookCommandLineCarriesWhatHappenedLastTime
```

With the cap, a word saturating one session gets df 26 of a 27-document corpus and reads as filler. On a real store that shape does not appear — a subject word does not fill twenty-five messages unless the session really is about it. But whether "the topic of one long session" should count as rare is a judgement about what recall is for, not a bug to be patched around, and the fixtures encode the opposite answer today.

Numbers are here; the call is yours.

### DIFF
--- a/cmd/deja/hook_earlier_attempt_test.go
+++ b/cmd/deja/hook_earlier_attempt_test.go
@@ -26,6 +26,7 @@ func TestHookPromptNamesTheEarlierAttempt(t *testing.T) {
 		`{"type":"user","sessionId":"newpass","timestamp":"2026-07-20T12:00:00Z","message":{"role":"user","content":"how many retries for the payment webhook"}}`,
 		`{"type":"assistant","sessionId":"newpass","timestamp":"2026-07-20T12:01:00Z","message":{"role":"assistant","content":[{"type":"text","text":"we cut the payment webhook retries to 3"}]}}`,
 	})
+	writeOrdinaryBackground(t, 100)
 	if err := index.Ensure(index.DefaultDir(), "", true, nil); err != nil {
 		t.Fatal(err)
 	}
--- a/cmd/deja/hook_prompt_test.go
+++ b/cmd/deja/hook_prompt_test.go
@@ -257,12 +257,22 @@ func TestHookPromptSkipsMarathonSessions(t *testing.T) {
 	old := time.Now().Add(-72 * time.Hour).UTC().Format(time.RFC3339)
 	var lines []string
 	for i := 0; i < dejaVuMaxMessages+10; i++ {
-		lines = append(lines, `{"type":"user","sessionId":"hay","timestamp":"`+old+`","message":{"role":"user","content":"quetzalcoatl stampede msg `+fmt.Sprint(i)+`"}}`)
+		// A long session says its subject some of the time and ordinary work
+		// the rest of it. Repeating the subject in every one of 310 messages
+		// made it the commonest word in the fixture's whole corpus, which is a
+		// shape a real store does not have — measured across 129 live prompts,
+		// nothing was lost to counting rarity in messages.
+		text := "pushed the branch and ran the suite again msg " + fmt.Sprint(i)
+		if i%40 == 0 {
+			text = "quetzalcoatl stampede msg " + fmt.Sprint(i)
+		}
+		lines = append(lines, `{"type":"user","sessionId":"hay","timestamp":"`+old+`","message":{"role":"user","content":"`+text+`"}}`)
 	}
 	writeClaudeFixture(t, filepath.Join(claudeRoot, "-tmp-hay", "hay.jsonl"), "hay", lines)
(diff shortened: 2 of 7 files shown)


vshulcz:
The open question is answered by measurement rather than by opinion: capping silences nothing.

Compared the two builds over the same 129 live prompts and listed every case where main recalls and this branch does not:

```
lost: 0
```

So the shape the three failing tests encode — a subject word filling every message of a long session — does not occur on a real store. Their fixtures were two and three documents wide, where no measure of rarity means anything at all; the marathon one repeated `quetzalcoatl` in 310 of 311 messages, making the most specific word in that corpus its commonest. They now sit behind a hundred ordinary sessions and say their subject some of the time, and all three pass.

Mutation on that: shrinking the background from 100 sessions to 3 fails them again.

Full state — `go test`, `go vet`, `gofmt`, `golangci-lint` (0 issues) clean; `bench prompt` every arm at baseline, `bench recall` 1.00/1.00, `bench context` unchanged; live 109 → 126 fired of 129, first line carries a term the product used 92% → 99%, nothing lost.

vshulcz:
Not mergeable — the number in the description belongs to a binary that is not this branch.

The 126/129 fired at 99% was measured with a build made before the denominator was corrected. That build had `totalDocs = sessions * cap`, which counts a one-message session as twenty-five documents and inflates every idf. Replacing it with the honest `sum(min(messages, cap))` — what is committed here — measures:

```
                                     fired      block opens on a term used
main                                109/129     100/109 = 92%
this branch, as committed           113/129      79/113 = 70%
the earlier build, wrong denominator 126/129     125/126 = 99%
```

So the gain came from raising rarity across the board, not from counting document frequency in messages. As committed this is worse than main on block quality, and it goes back to draft.

What the cap sweep says, all against the same frozen index:

```
cap    10     18     23     25     28     35     50
       68%    65%    68%    70%    70%    73%    76%
```

No plateau anywhere near the value picked to match the episode simulation, and every point below main. The idf ordering the simulation showed is still real — measured offline, `pgbouncer` scores 1.31 against `branch` 1.50 when documents are whole sessions, and 3.37 against 1.27 when they are 200-message episodes — but capping per session is not the way to get it, and the thresholds it has to cross are fixed while the scale moves under them.

Leaving the branch up for the measurements. The finding stands; this particular fix does not.

vshulcz:
Померил на реальном сторе, как договаривались, и вопрос из описания снимается сам: если брать не одну единицу, а более редкий из двух вердиктов, три фикстуры проходят без правок. #1536 — 43/59 против 41/59 на живых вопросах, ни одного потерянного ответа, бенчи без изменений.

vshulcz:
Закрыт в пользу #1536: более редкий из двух вердиктов снимает вопрос про фикстуры, живьём 43/59 против 41/59.

PR #1453, opened 2026-08-20
fix(hook): a rare word earns the memory, not a pointer to it

vshulcz:
Whether a question gets the memory or a pointer to it depends on something the asker cannot see.

The payload rule asks for two informative terms. The bar just above it already says a single rare word is a real match — "pgbouncer answers a question, problem does not" — and lets such a question through. The payload rule ignored that, so a question whose whole subject is one word landed on the pointer path, and one padded with an ordinary word that happened to appear in the same session got the digest.

Measured over sixty subjects each store really holds, asked four ways:

```
"напомни, что мы решали про X"    38/60 = 63%   content
"почему мы выбрали X"             16/60 = 27%
"что там было с X в итоге"         0/60 =  0%   every one a pointer
"what did we decide about X"       0/60 =  0%   every one a pointer
```

All four reduce to two terms and clear the gate that opens the store. What decides is the second word: `решали` appears in those sessions, `итоге` and `decide` do not. The crisper the question, the surer the answer is a pointer.

This lets a strong term earn the digest. The "you have been here" line stays at two terms — saying it on one word teaches the reader to ignore it, which the test above the new one guards.

```
                              content with the subject      injected per prompt
store with marathon sessions      33% -> 51%                  615 -> 788 chars
store of small sessions            0% -> 64%                  264 -> 655 chars
```

On a store without marathon sessions the old rule returned content for a direct subject question **zero times in 78**.

Ordinary traffic pays nothing. Over 129 real prompts on one store and 41 on the other, fired count, block quality and injected size are identical before and after — those prompts already carry two informative terms, so the new clause never fires there.

```
marathon profile  112/129 fired, 88% opening on a subject term, 855 chars/prompt — before and after
small profile      17/41  fired, 59%,                           418 chars/prompt — before and after
```

`bench prompt` unchanged on every arm, `bench recall` 1.00/1.00, `bench context` unchanged.

Mutation: dropping the new clause fails `TestOneRareWordStillGetsTheMemory`. Two earlier drafts of that test passed under their own mutant and were rewritten — the first because its fixture already had two matching terms, the second because a pointer names the topic in full, so checking for the words in the block did not separate a memory from a pointer to one. It asserts the quoted transcript now.

### DIFF
--- a/cmd/deja/hook_prompt.go
+++ b/cmd/deja/hook_prompt.go
@@ -156,6 +156,11 @@ func runHookPromptMode(dir string, stdin io.Reader, stdout io.Writer, plain bool
 	}
 	ss := make([]model.Session, 0, 2)
 	confident := false
+	// worthDigest is the payload decision, kept apart from the announcement.
+	// A word rare enough to identify something is a real match — the bar above
+	// says so and lets such a question through — but saying "you have been here"
+	// on one word teaches the reader to ignore the line, so that stays at two.
+	worthDigest := false
 	seen := alreadyInjected(dir, input.SessionID)
 	pol := policy.Load()
 	for i, s := range ranked {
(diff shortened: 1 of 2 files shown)


vshulcz:
Measured the other side of this change: what the converted answers actually contain.

```
changed answers over 160 probes:
  pointer -> memory carrying the subject: 29
  pointer -> block without the subject:   10
  other transitions:                       0
```

Three to one. The ten are all the same shape:

```
what did we decide about drop-non-owner  ->  "- User: Decide the integration approach given…"
what did we decide about mm_status       ->  "- User: 1. **Decide**: User settings or project…"
```

The block comes from a session that does hold the subject; the line quoted from it is the one carrying `decide`. That is the line-selection defect, not a wrong session — it was there before and this change makes it visible more often, since those questions used to get a pointer instead.

What it points at: the ranking knows which term made the match strong and does not pass that down, so the digest weighs every query word alike. Four cheaper signals for the same job were measured and rejected — rarity within the session, frequency within the session, position, and mention count — each worse than the current rule.

PR #1455, opened 2026-08-20
Add omp (Oh My Pi) harness

hanchang:
## Summary

Adds **omp (Oh My Pi)** as a first-class harness. omp is the Oh My Pi coding agent ([`can1357/oh-my-pi`](https://github.com/can1357/oh-my-pi)), a pi-lineage agent that writes pi-format JSONL transcripts under `~/.omp/agent/sessions/<encoded-project>/<ISO-timestamp>_<uuid>.jsonl`.

Because omp shares the pi transcript shape, the parser reuses the existing `parsePiShaped` reader (the same one `pi` and `openclaw` use). The one difference: omp's project directories use Claude Code's single-dash encoding (e.g. `-Code-pleasure-course`), which decodes lossily, so the session header's `cwd` is promoted to the project key (`useHeaderCwd=true`) — the same approach `openclaw` takes.

## What's included

- **Parser** — `internal/sources/omp.go`: `OmpRoot` (env `DEJA_OMP_ROOT`), `OmpSessionFiles`, `LoadOmp`, `ParseOmpFile`, `ParseOmpFileFromOffset` (offset-incremental, append-only JSONL).
- **Registry** — `omp` entry in `internal/sources/registry.go` (jsonl, offset-incremental).
- **Install** — `deja install omp` writes `~/.omp/agent/mcp.json` (omp has an MCP client with the standard `mcpServers` shape).
- **Resume / handoff** — `omp --resume <id>` and `omp <prompt>`.
- **Docs** — `docs/registry/omp.md`, a synthetic conformance fixture, `registry.json` entry, and regenerated pages/matrix/sitemap.

## Capabilities

| Capability | Status |
| --- | --- |
| MCP recall | ✅ |
| Auto-recall | — (todo: omp has a `--hook`/`--extension` API, but deja writes no omp extension yet) |
| Skill | — (todo: omp reads `RULES.md`/skills) |
| Command | — (todo: follows the skill gap) |
| Resume | ✅ |
| Handoff | ✅ |

## Verification

- `go test ./...` passes (full suite).
- `go vet ./...` clean; `gofmt -s` clean.
- `internal/sources` coverage 89.7% (above the 87% floor); `omp.go` is 100% covered.
- New tests cover discovery, full parse (roles, thinking/toolCall skipping, tool-output mapping, header-cwd project attribution), offset-incremental parse, and empty/metadata-only files.

## Notes

- omp's `toolResult` messages map to `RoleToolOutput` (via the shared parser), so omp feeds friction and `--role tool` like the other harnesses that carry tool output.
- The `ARCHITECTURE.md` "Add a new harness" section describes the older hardcoded dispatch; the index is now fully registry-driven (`sources.Registry()`), so a new harness is a parser + one registry entry. I left that doc untouched except the harness-count word, but happy to update it if you'd like.


vshulcz:
Reviewed against a live omp, not only the diff. Installed `omp` 17.4.1 from `omp.sh/install`, pointed it at a local model through `~/.omp/agent/models.yml` (an OpenAI-compatible route), and drove real sessions through it.

**What checks out**

- The MCP path is right. Upstream `docs/mcp-config.md` names `~/.omp/agent/mcp.json` as the primary user-scope file, and `deja install omp` writes exactly there. Verified by running rather than reading: after the install, omp listed the tools itself — `mcp__deja_recall`, `mcp__deja_recall_context`, `mcp__deja_remember`, `mcp__deja_blame`, `mcp__deja_fix`, `mcp__deja_how`.
- The parser reads a real omp 17.4.1 transcript correctly: `user` → `tool-output` → `assistant`, the project comes from the header `cwd`, and `thinking` / `toolCall` blocks stay out of the transcript.
- `omp --resume <id>` works space-separated and with an id prefix, which is what `resumeCommand` emits.
- The empty `title` in my session is omp's own doing — it writes a padded placeholder line (`{"type":"title","title":"","pad":"   …"}`) and rewrites it later. Not a parser bug; I checked before saying so.

**Notes (not blockers)**

1. **Named profiles.** `omp --profile <name>` / `OMP_PROFILE` relocates the whole user scope: `~/.omp/profiles/<name>/agent/sessions` and `~/.omp/profiles/<name>/agent/mcp.json` (upstream `docs/config-usage.md`: "every OMP-native user-level path … normally resolves to `~/.omp/profiles/<name>/agent/…`", and it says that covers sessions). Today `OmpRoot()` and `deja install omp` see only the default profile, so a user with a profile gets an empty history and a config nobody reads.
2. **XDG.** Same paragraph: an existing `$XDG_DATA_HOME/omp` or `$XDG_STATE_HOME/omp` relocates the state that holds sessions.

Both are discovery failures, which is the worst class for deja: empty with no explanation. I am taking them in a separate PR rather than sending them back to you.

**What I did on your behalf**

While this sat, the nineteenth harness landed in main (DeepSeek Harness, #1518 / #1519) and the branch went to conflict on the generated pages and the harness count. I rebased it, regenerated `docs/registry/*.html`, the sitemap and the matrix, and moved the count to `twenty` in all seven places `TestReadmeSpellsTheHarnessCountTheRegistryHas` guards. Your commits are untouched.

Pushing to your fork is blocked on my side, so the rebased branch goes up on our repo and I merge from there — your commits keep their authorship.

vshulcz:
Merged as #1520 — your commits, rebased. Thank you.

Two things in this PR were better than they had to be. Reusing `parsePiShaped` with `useHeaderCwd=true` instead of writing a fourth pi-shaped reader is the right call, and it is the one a contributor usually gets wrong by copying the file next door. And the test that asserts `thinking` and `toolCall` blocks are *skipped* while `toolResult` becomes `RoleToolOutput` pins the part of a pi transcript that quietly rots — I ran omp 17.4.1 against a local model and the roles came out exactly as your test says.

You also flagged the stale "Add a new harness" section in ARCHITECTURE.md and left it alone rather than fixing it silently in a harness PR. That is the right instinct; it is now on my list.

The two discovery gaps I raised (named profiles, XDG) are mine — they follow from upstream paths neither of us would hit on a default install, and they land in a separate PR.

PR #1456, opened 2026-08-20
fix(recall): open the block on the line that answers

vshulcz:
Builds on #1454, which added the arm this fixes.

The block opens with what the user said, and that slot was filled by the best-matching user line whatever it carried. A session that says an ordinary word of the question in one place and answers the question in another led with the ordinary line:

```
what did we decide about mm_status      ->  "- User: 1. **Decide**: User settings or project settings?"
what did we decide about drop-non-owner ->  "- User: Decide the integration approach given…"
```

Two changes, both small:

- the ranking now hands the digest the query's terms most-identifying first, using the idf it already computed and used to throw away. Without that, `matchedLines` weighs `mm_status` and `decide` alike;
- when a line the agent wrote is worth more than the best user line, the block leads with it rather than framing it with a line that only echoes a word.

```
                       before   after
decoy_line              0/2      2/2
real_questions         11/12    11/12
shown_line             14/14    14/14
haystack                3/3      3/3
russian_questions       3/3      3/3
negative_controls    0 false   0 false
absent_subject, off_topic, bucket, marathon, fresh — unchanged
bench recall          1.00/1.00 unchanged
bench context         unchanged
```

Live, 120 direct subject questions against a frozen index:

```
quoted answer carrying the subject   81/120 -> 83/120
quoted but missing it                   20  -> 18
pointer only                            19  -> 19
```

Ordinary traffic is unaffected on both store profiles — one with marathon sessions, one without:

```
marathon profile  112/129 fired, 88% -> 89% opening on a subject term, 855 -> 852 chars per prompt
small profile      17/41  fired, 59% -> 59%,                            418 -> 413 chars per prompt
```

Slightly smaller blocks, because a line that carried nothing is no longer printed.

The measurement caught two of my own mistakes before the fix went in: the first version of the arm rebuilt the block from the subject alone, so it read 2/2 while the block opened on the decoy; and ordering the terms alone changed nothing, because the user slot is printed first regardless of what it is worth. Both are in the commit messages of #1454 and here.

### DIFF
--- a/cmd/deja/bench_prompt.go
+++ b/cmd/deja/bench_prompt.go
@@ -144,6 +144,11 @@ type promptReport struct {
 	// and read as worthless. Correct here means the first line the agent sees
 	// carries a word from the question.
 	Shown promptArmReport `json:"shown_line"`
+	// The decoy arm: the session that answers also says an ordinary word of the
+	// question, three times, about something else. Correct means the block opens
+	// on the line that carries the subject rather than the one that carries the
+	// ordinary word.
+	Decoy promptArmReport `json:"decoy_line"`
 	// Questions whose subject the project has never held, asked with words it
 	// has. Every fire is a false one: there is nothing to answer with.
 	//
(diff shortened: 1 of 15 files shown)


vshulcz:
Rebuilt on main as #1458 — the arm commit here conflicted once #1454 landed as a squash.

PR #1462, opened 2026-08-20
fix(prompt): keep short Russian subjects

vshulcz:
Russian keeps short subjects — кеш, хук, бот, сеть, порт, диск — and the floor for Cyrillic terms stood at five letters, so none of them could become a search term. #1460 did the same for English acronyms; this is the other side.

Lowering the floor was tried yesterday and measured worse: on a store of ordinary short sessions, blocks opening on a subject word fell from 88% to 53%. Rather than accept that, I printed what the lower floor actually adds — nine distinct words over forty-one live questions:

```
3x шину   3x про   1x сижу   1x маке   1x вот
1x раз    1x еще   1x чтоб   1x язык
```

Six of the nine are closed-class words missing from the list this file already keeps. Naming them leaves five, all of them things a question can be about.

```
                fired    block opens on a subject term    chars/prompt
small profile   17/41    88% -> 88%                       409 -> 409
marathon       112/129   90% -> 92%                       876 -> 874
```

The regression is gone and the marathon profile gains two points.

```
bench prompt   russian 4/4, real 12/13, shown_line 15/15 -> 16/16,
               negative controls 0 false fires, haystack 3/3, decoy 2/2 — rest unchanged
bench recall   1.00/1.00,  bench context unchanged
```

Mutations, both caught: putting the floor back at five fails the new test on the first subject; removing the six named words fails it on the filler line.

### DIFF
--- a/cmd/deja/hook_prompt.go
+++ b/cmd/deja/hook_prompt.go
@@ -322,7 +322,7 @@ func hasIdentifierTerm(terms []string) bool {
 		// something the way a five-letter Latin word does — so the bar is one
 		// letter lower there.
 		n := utf8.RuneCountInString(t)
-		if (n >= 5 || (n >= 4 && hasCyrillic(t))) && !soleWorkingWord[t] {
+		if n >= 3 && !soleWorkingWord[t] {
 			return true
 		}
 		for _, r := range t {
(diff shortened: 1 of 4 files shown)


vshulcz:
Two more commits: the reproduction from #1461 is now on main, and with it the case still failed — the extractor keeps `кеш` but the bar in front of the store refused it, because that one measures four letters for Cyrillic and five otherwise.

Lowering it to three was tried yesterday together with the unfiltered extractor and measured badly. With the six function words named it holds:

```
                       fired    block opens on a subject term   chars/prompt
small profile          17/41    88% -> 88%                      409 -> 409
marathon profile      112/129   92% -> 92%                      874 -> 874
```

Nothing lost on either store, and the benchmark closes:

```
russian_questions  4/5 -> 5/5
real_questions    12/13 -> 13/13
shown_line        16/16 -> 18/18
negative_controls  0 false fires — after naming line, file, code, read, open,
                   which started firing the moment they could reach the store
```

The bar still pays for itself: on a prompt that reduces to one ordinary word the hook takes 5.0ms with the store shut against 69.0ms with it open.

`TestTheSingleTermBarCountsLettersNotBytes` was calling the bar directly with words the extractor would never produce, so it asserted a contract nothing has. It goes through `prompt.Terms` now and states the four facts that matter: filler alone stays out, ordinary work stays out, and `кеш` and `ttl` get in.

Mutation: putting the bar back at five takes russian to 4/5 and real to 12/13.

PR #1464, opened 2026-08-20
fix(digest): decision markers for Russian too

vshulcz:
Follows #1463, which taught per-prompt recall to prefer a line that concluded something. The markers it uses are an English list, so every line of a Russian session reads as a passing mention — and half the sessions on the store these numbers come from are Russian.

A Russian chain that says its subject three times in passing before concluding anything reproduces it:

```
decision_line  4/5
```

The arm is new: the existing Russian arm asks whether the block carries a word of the *question*, and both a mention and a conclusion carry that. This one asks whether it carries what was *concluded*.

With the markers:

```
decision_line   4/5 -> 5/5
russian 5/5, real 13/13, shown_line 18/18, negative controls 0 false,
```

Live, on questions whose subject is a short word: 76/120 -> 78/120, misses 20 -> 18. Longer subjects unchanged at 85/120. Ordinary traffic on both store profiles: 112/129 fired at 92% -> 91% and 17/41 at 88% -> 88%, blocks two and three characters smaller. The one case is a single block out of 112 and sits inside the noise this metric has shown all along; the arms it was added for move in the other direction.

Mutation: removing `решили` and `в итоге` fails the new Russian test.

Reproducing it took two attempts, both recorded: the first fixture had one mention and one conclusion, and the block holds two lines, so both fitted and nothing could go wrong. Three mentions against two slots is what makes the choice visible.

### DIFF
--- a/internal/bench/prompt.go
+++ b/internal/bench/prompt.go
@@ -24,6 +24,10 @@ type PromptChain struct {
 	Project  string
 	Topic    string
 	Question string
+	// Fact is what the session concluded, so an arm can ask whether the block
+	// carries the conclusion rather than merely the subject: a passing mention
+	// says the subject too.
+	Fact     string
 	Sessions []model.Session
 	Negative bool
 	// Kind names what this chain is here to measure: "" for the plain case,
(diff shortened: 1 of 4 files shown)


vshulcz:
Rebuilt on main as a fresh branch — #1463 landed as a squash and the shared commit conflicted.

PR #1465, opened 2026-08-20
fix(digest): decision markers for Russian too

vshulcz:
Follows #1463, which taught per-prompt recall to prefer a line that concluded something. The markers it uses are an English list, so every line of a Russian session reads as a passing mention — and half the sessions on the store these numbers come from are Russian.

A Russian chain that says its subject three times in passing before concluding anything reproduces it:

```
decision_line  4/5
```

The arm is new: the existing Russian arm asks whether the block carries a word of the *question*, and both a mention and a conclusion carry that. This one asks whether it carries what was *concluded*.

```
decision_line   4/5 -> 5/5
russian 5/5, real 13/13, shown_line 18/18, negative controls 0 false,
```

Live, on questions whose subject is a short word: 76/120 -> 78/120, misses 20 -> 18. Longer subjects unchanged at 85/120. Ordinary traffic on both store profiles: 112/129 fired at 92% -> 91% and 17/41 at 88% -> 88%, blocks two and three characters smaller.

Mutation: removing `решили` and `в итоге` fails the new Russian test.

Reproducing it took two attempts, both worth recording: the first fixture had one mention and one conclusion, and the block holds two lines, so both fitted and nothing could go wrong. Three mentions against two slots is what makes the choice visible.

### DIFF
--- a/cmd/deja/bench_prompt.go
+++ b/cmd/deja/bench_prompt.go
@@ -245,7 +245,7 @@ func measurePrompt(seed int64) (promptReport, error) {
 		case "russian":
 			arm = &report.Russian
 			report.Decision.Cases++
-			if blockCarries(indexDir, scope, terms, chain.Fact) {
+			if blockCarries(indexDir, scope, terms, chain.Fact, chain.Topic) {
 				report.Decision.Fired++
 				report.Decision.Correct++
 			}
@@ -404,7 +404,7 @@ func firstShownLineCarries(dir, project string, terms []string, topic string) bo
 
 // blockCarries builds the block the hook would inject and reports whether it
 // holds a distinctive word of what the session concluded.
-func blockCarries(dir, project string, terms []string, fact string) bool {
+func blockCarries(dir, project string, terms []string, fact, topic string) bool {
 	ranked, matched, strong, idfOf, err := index.ProjectRelevant(dir, []string{project}, terms, 8)
 	if err != nil || len(ranked) == 0 {
 		return false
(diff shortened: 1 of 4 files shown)


vshulcz:
Correction: the arm this description quotes was missing from the branch. The commit that carried it was staged with `git add -A internal` and the arm lives in `cmd/`, so only the marker list went in — a description promising `decision_line 4/5 -> 5/5` against a diff that had no such arm.

Added now as its own commit. Verified on the branch: `decision_line` 5/5 with the markers, 4/5 without them, and the rest of the report unchanged — real 13/13, russian 5/5, shown_line 18/18, negative controls 0 false fires, haystack 3/3, decoy 2/2.

vshulcz:
Second correction, and it changes the numbers.

The arm was blind. It asked whether the block held any long word of the fact — and the subject is one of those words, present in every session that mentions the thing. Emptying the conclusion of its content entirely still read 5/5.

Skipping the subject makes it honest, and the markers earn less than the description claimed:

```
                    blind arm     honest arm
without the markers    4/5            2/5
with them              5/5            4/5
```

Still a real gain — from two of five to four — but not the clean sweep. The remaining case is one where the conclusion is quoted but its distinctive words fall outside the two lines the block holds.

Everything else on the branch is unchanged: real 13/13, russian 5/5, shown_line 18/18, negative controls 0 false fires, haystack 3/3, decoy 2/2, `bench recall` 1.00/1.00.

Three arms tonight have turned out to measure something other than what they were named for, this one included. The pattern that catches it is breaking the fixture on purpose and checking the arm actually falls — which is what I should have done the first time.

PR #1466, opened 2026-08-20
fix(recall): lead with the session that settled it

vshulcz:
Ranking counts how often and how rarely the query's words appear, so a session that mentioned the subject sixty times outranks the one that answered it in a sentence. An arm reproduces it:

```
concluded_session  0/2
```

Two sessions of one project hold the subject; only one concluded anything about it. The fix leads with that one.

Getting there took three attempts, and the first two are worth the space:

```
promote if any line of the session concludes         live 78 -> 70,  9 answers lost
promote if the lines the digest picks conclude       live 78 -> 70,  the same 9
promote if the FIRST shown line concludes            live 78 -> 81,  none lost
```

All nine losses showed the same block — "мы такого не решали, я такого не делал". The session had risen on a line the reader never saw: the digest holds two lines and the marker sat in the second.

Then the benchmark caught something the live numbers could not. `decision_line` fell 4/5 -> 3/5, and tracing which session got promoted showed why:

```
LEAD prompt-ru-01-session -> prompt-ru-04-session (terms [шардирование итоге чему])
```

`итоге` is how a Russian question asks for the outcome — and it is also one of the decision markers. A line matched on that phrase then counted as a conclusion because of it, and the session promoted was answering a different question. Two changes close it: the markers ignore any that the asker used, and "в итоге" joins the filler list where it belongs.

```
concluded_session  0/2 -> 2/2
decision_line      4/5   unchanged
real 13/13, shown_line 18/18, russian 5/5, negative controls 0 false,
```

Live: 78/120 -> 81/120 on questions whose subject is a short word, nothing lost. Ordinary traffic unchanged on both store profiles — 112/129 at 91% and 17/41 at 88%, blocks one character smaller.

Mutation: disabling the reorder returns the arm to 0/2.

### DIFF
--- a/cmd/deja/bench_prompt.go
+++ b/cmd/deja/bench_prompt.go
@@ -437,6 +437,8 @@ func blockCarries(dir, project string, terms []string, fact, topic string) bool 
 	if len(keep) == 0 {
 		return false
 	}
+	// The same ordering the hook applies, from the same function.
+	keep = search.LeadWithConclusion(keep, terms)
 	block := search.AutoRecallDigestFor(keep[:1], 2000, terms)
 	// Skip the subject itself. It appears in every session that mentions the
 	// thing, so checking for it asks whether the block is about the subject —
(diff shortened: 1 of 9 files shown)


vshulcz:
Rebuilt on main as a fresh branch — #1465 landed as a squash and the shared commits conflicted.

Issue #1534, opened 2026-08-22
The bigger prompt corpus shows 5 false fires the small one hid


vshulcz:
`bench prompt` used to run on thirty chains of invented vocabulary, where every word was rare and the recall gate looked perfect. With 300 background sessions of ordinary working talk behind it (#1448's corpus commits, now on main), the same corpus reports:

```
real questions     13/13   (was 11/12)
negative controls   5/12 false fires   (was 0)
```

The suite records 5 so it cannot grow quietly, but the target is 0.

First measurement, on this corpus: replacing `HasIdentifierTerm`'s word-length proxy with the rarity the index already knows (a lone match trusted when its term is `strong`) takes it 5 → 4 with real questions unchanged at 13/13. The remaining four need the other two rules from #1448 re-measured on current code — a session accounting for every identifying word the project has an answer for, and deciding the stem fold per project rather than per store.

Blocked on nothing; it is measurement work, not a decision.

PR #1536, opened 2026-08-22
Rarity: the rarer of the two verdicts wins

vshulcz:
Answers the open question in #1450 without anyone having to judge the fixtures.

Neither unit is right alone. Counting whole sessions calls a subject word common, because the eleven marathon sessions that hold 99% of the store all mention it — `pgbouncer` scores 1.31 against `branch` at 1.50. Counting messages capped per session fixes that, but then a word saturating one long session reads as filler, which is exactly what the name of the thing you have worked on all month looks like.

So compute both and take the rarer verdict. A term has to read as ordinary under **both** before it is treated as filler.

Live, 59 real questions from this machine's own history, each binary against its own untouched copy of a frozen index (the hook writes a seen-file beside it — sharing one directory lets the first run decide what the second is allowed to say, which is how an earlier reading of this got a phantom regression):

```
             answered  silent  pointer-only
main            41       17         1
capped only     43       16         0
rarer-of-two    43       16         0
```

Two questions gained, none lost.

The fixtures are the guard, and they are untouched: with the cap alone, `TestHookPromptSkipsMarathonSessions`, `TestHookPromptNamesTheEarlierAttempt`, `TestHookPromptDoesNotReadASessionItAlreadyShowed` and `TestToolHookCommandLineCarriesWhatHappenedLastTime` all fail — that is the saturating-subject case failing loudly. With the rarer-of-two rule the whole suite is green as it stands, so #1450's "three tests fail and the question is worth deciding" no longer needs deciding: the rule keeps what they assert.

`bench prompt` 13/13 real questions and 5 false fires, identical to main (the bench corpus's sessions are short, so the cap never binds there — that arm is #1534's business). `bench recall` 1.00/1.00, `bench context` 294 median tokens, coverage 1.00 — both unchanged. No manifest change and no reindex: the per-session message count is already stored, and a manifest written before it degrades to one document per session, which is today's behaviour.

### DIFF
--- a/internal/index/atomicity_test.go
+++ b/internal/index/atomicity_test.go
@@ -1,6 +1,7 @@
 package index
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


vshulcz:
Added the direct guards: `TestRarityTakesTheRarerOfTheTwoVerdicts` (a word one long session is about stays rare) and `TestRarityIsNotInvertedByMarathonSessions` (a subject mentioned in more sessions than an ordinary word does not end up the commoner of the two — the pgbouncer/branch shape from the report). Mutations: the cap alone and taking the commoner verdict both go red; session-counting alone no longer compiles, and removing the cap stays green, which is honest — with the rarer-of-two rule the cap's exact value is not load-bearing for either property.

Issue #1411, opened 2026-08-19
the brief's own wire line does not fit a 60-column pane


vshulcz:
#604 budgeted the brief's lines against the terminal so a 60-column split pane stops wrapping them. The lines carrying transcript text were budgeted; deja's own copy was not, and one line of it does not fit:

```
COLUMNS=60
wire       no agent is wired to this yet — `deja install --auto`     64 columns
```

Four columns over, so it wraps to a second line with `--auto` alone on it. This is the line a fresh install shows — the store has sessions, nothing is wired yet — so it is on the first screen a new user sees, which is the screen #604 was about.

Measured on a store with three sessions and no hooks installed, at COLUMNS=60, 80 and 120. Every other line of the brief fits 60 today: the header is 39, `covering` 36, `more` 44, and the `recent` lines are budgeted (they overflow only for CJK titles, which #1401 fixes).

The fix is a shorter sentence, which is your wording rather than a bug, so I have not touched it. Two shapes that fit, for what they are worth:

```
wire       nothing wired yet — `deja install --auto`             52
wire       no agent wired yet — `deja install --auto`            53
```

Found by the night hunt, iteration 100.

vshulcz:
Holding this one for a first-time contributor — it will not be picked up from the inside. If you want it, say so here and it is yours; ask anything in the thread.

Exberg:
Hi — I'd like to take this one as my first contribution here. Plan: shorten the wire line so it fits 60 columns (the `no agent wired yet — \`deja install --auto\`` shape, 53 columns), and add/extend the brief-width test to pin it. Happy to adjust wording if you prefer the other variant.

vshulcz:
Fixed by #1554: the wire line is 53 columns now, with a test that fails at 64 if the wording comes back.

Issue #604, opened 2026-07-31
The brief assumes a wide terminal


vshulcz:
From the overnight sweep, section E7.

Nothing in deja reads the terminal width — `grep -rn "COLUMNS|TIOCGWINSZ|GetSize"` finds nothing — so the brief truncates session titles to a fixed character count that assumes a wide window. Measured with a pty at several widths:

| columns | result |
|---|---|
| 100 | fits |
| 80 | fits |
| 60 | `recent` and `asked` lines wrap mid-word |
| 40 | every line wraps, the screen is unreadable |

A 60-column pane is not exotic — it is what a split editor gives you — and this is the first screen a new user sees.

The fix is not free: the binary has no dependencies, so terminal width means a `TIOCGWINSZ` ioctl with separate unix and windows files, which is the pattern `lock_unix.go`/`lock_windows.go` already uses. Worth doing, but it is a contained piece of work rather than a one-line change, and it should budget every column-aligned line — the brief, search hits, `files`, `restore` — not just titles.

Checked at the same time and **not** broken: `NO_COLOR` is respected, and colour is already off when the output is piped.

vshulcz:
Half of this shipped in #1367: `terminalWidth()` exists as `width_unix.go` / `width_windows.go` (the `lock_unix.go` pattern this issue pointed at), `briefWidth()` prefers `COLUMNS`, then the terminal, then 80, and the brief, the status line and the logo block all budget against it.

Still fixed-width, so a 40-column pane still breaks them: search hit headers, `deja files` and `deja restore`. Titles are also elided to 60 runes at index time, which bounds what those surfaces can do even once they read the width — a narrow terminal needs the cut at print time, not in the manifest.

PR #1554, opened 2026-08-23
fix(brief): fit the wire line into a 60-column pane

Exberg:
## Summary

The brief's wire line ran to 64 columns and wrapped `--auto` alone onto a second line of the first screen a fresh install shows (#1411). #604 budgeted the transcript lines against a 60-column pane, but deja's own copy was not budgeted.

The fix is the shorter wording suggested in the issue — "no agent wired yet — `deja install --auto`", 53 columns. No other line of the brief changes.

## Test plan

- Added `TestBriefWireLineFitsSixtyColumns`: renders the brief on an unwired store at COLUMNS=60, 80 and 120 and asserts the wire line stays within 60 columns.
- Sabotage check: with the old wording restored, the new test fails at every width with `wire line is 64 columns`; with the fix it passes.
- Updated `first_run_wiring_test.go` to match the new wording (the "goes quiet once wired" test asserted against the old string).
- `go vet ./...` clean; `go test -race ./...` passes in full locally (macOS, go 1.25.0); cmd/deja coverage 89.9% (floor 88).

## Notes

- First-time contributor; I commented on the issue before starting, per the thread where it was held for a first-time contributor.
- Chose the "no agent wired yet" shape (53 columns) over "nothing wired yet" (52) since it keeps saying what is missing. Happy to switch if you prefer the other.

### DIFF
--- a/cmd/deja/brief.go
+++ b/cmd/deja/brief.go
@@ -88,7 +88,10 @@ func runBrief(dir string, w io.Writer) error {
 	// header because that is the only line a first-time reader is certain to
 	// read.
 	if nothingWired() {
-		fmt.Fprintf(w, "wire       %sno agent is wired to this yet%s — `deja install --auto`\n", bold, reset)
+		// 53 columns: every line of the brief fits a 60-column pane (#604),
+		// and the longer wording left `--auto` wrapped alone on its own line
+		// of the first screen a fresh install shows (#1411).
+		fmt.Fprintf(w, "wire       %sno agent wired yet%s — `deja install --auto`\n", bold, reset)
 	}
 
 	recalls, bytes, _ := usage.TodayWithInjections(dir)
--- a/cmd/deja/brief_width_test.go
+++ b/cmd/deja/brief_width_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


vshulcz:
Read the whole diff and checked it by running rather than by eye.

**The test is real.** Restored the old wording on top of your branch and it fails at all three widths with `wire line is 64 columns, wraps on a 60-column pane` — so the case would catch this coming back, which is the part that usually goes missing in a one-line copy fix.

**One thing I changed, in a commit on your branch.** `TestTheBriefStopsSayingItOnceWired` asserted on the sentence, and the sentence is exactly what this PR rewords — the next rewording would leave that guard passing while guarding nothing. It now matches `deja install --auto`, which stays put whatever the copy says. Sabotage-checked that too: forcing the line to print on a wired machine fails it.

Not from this PR, so no action needed here: `go test ./...` on this branch still fails `TestParseZedDBSurvivesASqlite3ThatDoesNotAnswerInRows` under Go 1.27. That was fixed on main in #1568 and goes away on merge.

The wording is one of the two shapes #1411 proposed, so it is the owner's copy rather than a new invention. Good find and a clean fix.

vshulcz:
Thanks — the sabotage check is what makes this one worth merging as-is. Plenty of width fixes land with a test that passes either way; yours fails at 64 columns the moment the wording comes back, and you said so in the description instead of leaving it to be discovered.

PR #1557, opened 2026-08-23
fix(dsh): prefer the deja the user installed over the copy npm brought

vshulcz:
The plugin resolved its binary as: DEJA_BIN, then the platform package npm installed beside it, then PATH. That order pins a DSH user to whatever deja shipped with the plugin release they installed, while their own `deja update` or `brew upgrade` changes nothing.

Now: DEJA_BIN, then PATH, then the npm copy as the fallback for someone who never installed deja. Each candidate is asked for `version` before it is used, so a name on PATH that does not run falls through instead of silently answering nothing.

Checked against dsh 0.1.1-rc.2 with the mock model: three tools registered, and the runtime-context message carries the recall.

### DIFF
--- a/packages/dsh-deja/index.js
+++ b/packages/dsh-deja/index.js
@@ -26,16 +26,32 @@ try {
 const PLATFORM = process.platform === "win32" ? "windows" : process.platform;
 const ARCH = process.arch === "x64" ? "amd64" : process.arch;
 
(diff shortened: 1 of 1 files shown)


vshulcz:
Superseded by the move: the fix shipped in vshulcz/dsh-deja and is in dsh-deja@0.19.0. #1558 removes the copy this PR edited.

PR #1569, opened 2026-08-23
fix: align daily recall counts with weekly usage

### LINKED ISSUE #1403: brief counts the same events as recalls today and not this week
`deja brief` prints two lines that cannot both be true. The same events are recalls in the first and not recalls in the second, so the day can hold more of them than the week containing it.

`usage.TodayWithInjections` counts `hook`, `deja-vu` and `tool` events as recalls. `usage.Week` counts the same three as injections and leaves them out of its recall figure. Five hook injections and nothing else, measured:

```
today: recalls=5 bytes=2000 injected=2000
week : recalls=0 bytes=0    injected=5
```

Which is printed as:

```
today      1 session · 5 recalls served (2.0 kB)
this week  1 session · 0 recalls
```

The statusline reads the same two figures the same way, and `deja stats` splits them deliberately — "N recalls by your agents ... (plus M auto-injections)" — so the split itself is intended somewhere. What is not intended is one screen using both definitions without saying so.

It also breaks the suppression #842 added: `weekEchoesToday` compares `weekRecalls == recalls`, two counts that mean different things, so the week line survives the check precisely when it is about to print a smaller number than the line above it.

Two ways out, and the choice is a product one rather than a fix:

1. **Auto-injections are recalls everywhere.** `Week` counts them the way `Today` does. The week number grows for everyone, and "recall" means "memory reached an agent", however it got there. `stats` keeps its separate line for the breakdown.
2. **Auto-injections are not recalls anywhere.** `Today` stops counting them, and the statusline's headline number drops for anyone relying on hooks rather than the MCP tool. Closer to "a recall is something an agent asked for".

Either way both screens should say the same thing. I have not touched it — it changes a number people read every day.

Found by the night hunt, iteration 88.

mikemikimike:
Fixes #1403

## Summary
- Align daily and statusline recall counts with the weekly demand-side definition.
- Exclude empty results and automatic hook/deja-vu/tool injections from the recall headline.
- Keep injected bytes available as a separate statusline metric.
- Add regression coverage for empty results, injections, rotation, and statusline output.

## Compatibility
This changes the displayed daily/statusline recall count for users with automatic injections or empty results. Those events remain recorded and injections remain reported separately.

## Tests
- `go test ./internal/usage/... ./cmd/deja/... -run 'Test(StatuslineCountsRecalls|TodayExcludesEmptyRecallsAndInjections|RecordAndToday|RotateKeepsRecentWindow|Brief)' -count=1`
- `go vet ./...`
- `go build ./cmd/deja`

The repository-wide `go test ./... -race -count=1` could not start because this Windows environment has `CGO_ENABLED=0`. A non-race repository-wide run was attempted but is blocked by pre-existing environment-sensitive failures in embed fixtures, the SQLite helper path, and unrelated CLI tests that read host agent data.

### DIFF
--- a/cmd/deja/brief.go
+++ b/cmd/deja/brief.go
@@ -91,7 +91,7 @@ func runBrief(dir string, w io.Writer) error {
 		fmt.Fprintf(w, "wire       %sno agent is wired to this yet%s — `deja install --auto`\n", bold, reset)
 	}
 
-	recalls, bytes := usage.TodayDemand(dir)
+	recalls, bytes, _ := usage.TodayDemand(dir)
 	weekRecalls, _, _, _ := usage.Week(dir)
 	dejaVu := usage.DejaVuWeek(dir)
 	// Only when the week has nothing at all to report. Recalls and déjà vu
--- a/cmd/deja/brief.go
+++ b/cmd/deja/brief.go
@@ -91,7 +91,7 @@ func runBrief(dir string, w io.Writer) error {
 		fmt.Fprintf(w, "wire       %sno agent is wired to this yet%s — `deja install --auto`\n", bold, reset)
 	}
 
-	recalls, bytes, _ := usage.TodayWithInjections(dir)
+	recalls, bytes := usage.TodayDemand(dir)
 	weekRecalls, _, _, _ := usage.Week(dir)
 	dejaVu := usage.DejaVuWeek(dir)
 	// Only when the week has nothing at all to report. Recalls and déjà vu
--- a/cmd/deja/statusline.go
+++ b/cmd/deja/statusline.go
@@ -73,7 +73,8 @@ func runStatusline(dir string, stdin io.Reader, stdout io.Writer) error {
 		fmt.Fprint(stdout, withFileMemory(dir, in, line))
 		return nil
 	}
-	recalls, bytes, injected := usage.TodayWithInjections(dir)
+	recalls, bytes := usage.TodayDemand(dir)
+	_, _, injected := usage.TodayWithInjections(dir)
 	if recalls == 0 {
 		if wr, wb, _, _ := usage.Week(dir); wr > 0 {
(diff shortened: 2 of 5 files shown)


vshulcz:
Read every file and reproduced the issue's case before and after: five hook injections and nothing else, at `DEJA_INDEX_DIR` pointed at a scratch store.

Before: `deja · 5 recalls · 2.0 KB ctx today · 2.0 KB injected`. After your change: `deja · no recalls yet today · 0 B injected`.

The count is right — that is option 2 from #1403, and `Week` now agrees with it. The last four words were not: that branch prints a hardcoded `0 B injected`, and after this change it is exactly the branch an auto-recall user lands in every day. The line went from an inconsistent number to a false one, which is the same class of thing the issue was opened about.

Fixed in a commit on your branch, along with one more thing the diff introduced: the statusline read and parsed the usage log twice — `TodayDemand` and then `TodayWithInjections` — on a path that renders on every prompt, and two passes can also straddle a write and report two numbers that were never true together. `TodayDemand` now returns injections from the same pass, so there is one read, and the zero-recall branch says what did arrive: `deja · no agent recalls today · 2.0 KB injected`. Test added for that case; your `TestTodayExcludesEmptyRecallsAndInjections` grew the third return value.

One left deliberately, because it is a copy decision rather than a bug: `serviceReceipt` in `cmd/deja/hook_context.go` still counts injections, so the receipt that rides into every session can say `today: 5 recalls` while the statusline says none. Filed separately rather than settled here — that string goes to the model, and the wording is the owner's call.

Everything else holds: `Week`'s definition matches your `TodayDemand` kind for kind, including the `Empty` skip; `weekEchoesToday` now compares two counts that mean the same thing, which was the second half of #1403; and `usage.Today` has no callers outside tests, so nothing else silently kept the old meaning.

vshulcz:
Thanks. Two things stood out: you took the product decision the issue left open, picked one, and said in the description that the displayed number changes for people with injections — that is the part someone has to own, and it is easier to review than a fix that quietly changes what a daily counter means. And the test you added covers the empty-result case as well as the injection one, which is the half of the definition that is easy to forget.

PR #1573, opened 2026-08-23
feat(kimi): a Kimi Code plugin that recalls what the other agents wrote

### LINKED ISSUE #1572: Kimi Code has a plugin marketplace and deja is not in it
`MoonshotAI/kimi-code` ships a plugin marketplace inside the repository (`plugins/marketplace.json`), and its `/plugins` browser is in front of every Kimi Code user. The `curated` tier already carries third parties — obra/superpowers, Vercel, Google Chrome, Tencent CloudBase. deja indexes `~/.kimi-code` and installs into it already; what is missing is a package in Kimi's own idiom.

Three open threads there are asking for exactly this: #1283 (27 comments), #1478, #2613.

What a Kimi plugin can carry (`docs/en/customization/plugins.md`):

- `mcpServers` — the same schema as `mcp.json`, so `deja mcp` with its tools
- `skills` + `sessionStart.skill` — the `deja-history` skill, loaded into the main agent at session start
- `commands` — markdown files become slash commands
- `hooks` — `UserPromptSubmit` is the one event whose stdout Kimi appends to the turn, which is the same injection point `deja install kimi-auto` uses today

The part that needs care is duplication. `deja install kimi` writes `~/.kimi-code/mcp.json` and `kimi-auto` adds a marked `[[hooks]]` block to `config.toml`. Kimi registers a plugin's MCP server under its own namespace (`plugin-<id>:<name>` — `packages/agent-core-v2/src/app/plugin/manager.ts`), so it does *not* collide with the installer's entry: both would run, every tool would be listed twice, and recall would be appended twice. That was a real defect for opencode, dsh and Zed (#1567, #1570, #1571).

Note which repository: `MoonshotAI/kimi-cli` is the old project, `MoonshotAI/kimi-code` is the live one.

vshulcz:
Closes #1572.

`extensions/kimi` — a Kimi Code plugin: `deja mcp` as an MCP server, the shared `deja-history` skill loaded at session start, `/deja:recall`, and recall on every prompt through a `UserPromptSubmit` hook.

What a reviewer would otherwise have to reconstruct:

- **Both entry points are `.mjs`, and the manifest declares `command: "node"`.** That exact spelling is the one form Kimi rewrites to its own runtime when it ships as a native binary (`withPluginMcpRuntime`), so the server starts on a machine with no node on PATH. Nothing imports a `.js` file, so the plugin does not depend on the `package.json` — which is why the release zip excludes it.
- **Standing down is the whole duplication story.** Kimi namespaces a plugin's MCP server as `plugin-deja:deja`, so it does not collide with the `deja` entry `deja install kimi` writes into `mcp.json` — both would run. The hook and the server both check what the installer wrote (the marker comment in `config.toml`, the server in `mcp.json`) and stay out of the way when it is there. Whichever order you install in, exactly one of each runs.
- **Installing from the repository URL cannot work** — Kimi looks for `kimi.plugin.json` at the repository root or in a single top-level directory. Hence `kimi-deja.zip`, packed from `extensions/kimi` and attached at release, which is the same route Tencent CloudBase takes.

Verified against a real Kimi Code 0.28.1, not only by reading the docs — with a spy binary on `DEJA_BIN` so each call is attributable:

```
plugin only:   argv: mcp            argv: hook-prompt --plain     (plugin's copies ran)
both installs: installer spy: mcp, hook-prompt --plain
               plugin spy:    (empty)                             (plugin stood down)
```

The second run is the counter-case: the same two files, and the only difference is whether `config.toml`/`mcp.json` carry the installer's wiring. The last check ran against the exact zip that ships.

Tests: `extensions/kimi/test/pure.test.mjs` covers the stand-down matrix and binary resolution, and runs the real hook script the way Kimi runs it. On the Go side `TestKimiPluginManifest` pins the manifest against the loader's rules, `TestKimiPluginKnowsTheInstallersMarker` fails if either side edits the marker alone, and the bundled skill is pinned to the installer's text. New `kimi-deja` CI job alongside the other three.

The marketplace PR is not part of this — it needs the release asset to exist first, and it is a submission outside our own repositories.

### DIFF
--- a/.github/workflows/extensions.yml
+++ b/.github/workflows/extensions.yml
@@ -120,6 +120,44 @@ jobs:
             exit 1
           }
 
+  kimi:
+    name: kimi-deja
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    defaults:
+      run:
+        working-directory: extensions/kimi
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
+        with:
+          node-version: 22
+
+      - name: The plugin has to parse
+        run: node --check lib.mjs && node --check hooks/recall.mjs && node --check bin/deja-mcp.mjs
+
+      - run: npm test
+
+      - name: The manifest has to stay loadable by Kimi
+        run: |
+          node -e '
+            const fs = require("node:fs");
+            const m = JSON.parse(fs.readFileSync("kimi.plugin.json", "utf8"));
+            let bad = 0;
+            // Kimi ignores what it does not know, and reports it as a
+            // diagnostic nobody reads: a typo here is a capability that
+            // silently never arrives.
+            const known = new Set(["name", "version", "description", "keywords", "author", "homepage", "license", "interface", "skills", "agents", "commands", "sessionStart", "skillInstructions", "systemPrompt", "systemPromptPath", "mcpServers", "hooks"]);
+            for (const key of Object.keys(m)) {
+              if (!known.has(key)) { console.error("unknown manifest field: " + key); bad++; }
+            }
+            for (const rel of [m.skills, m.commands, ...(m.hooks ?? []).map((h) => h.command.split(" ").pop()), ...Object.values(m.mcpServers ?? {}).flatMap((s) => s.args ?? [])]) {
+              if (typeof rel !== "string" || !rel.startsWith("./")) { console.error("path must start with ./ : " + rel); bad++; continue; }
+              if (!fs.existsSync(rel)) { console.error("manifest points outside the plugin: " + rel); bad++; }
+            }
+            process.exit(bad ? 1 : 0);
+          '
+
   metadata:
     name: packages point at this repository
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,6 +50,21 @@ jobs:
           go run ./scripts/mcpb -version "${GITHUB_REF_NAME#v}" -in dist -out dist/mcpb
           go run ./scripts/mcpb -registry -version "${GITHUB_REF_NAME#v}" -out dist/mcpb
 
+      # Kimi Code installs a plugin from a local path, a zip URL or a GitHub
+      # repository — and the repository form looks for the manifest at the root
+      # or in one top-level directory, which a monorepo does not have. The zip
+      # is how extensions/kimi reaches a Kimi user.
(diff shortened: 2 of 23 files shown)


vshulcz:
Full audit of the install path, run against Kimi Code 0.28.1 with a spy binary on `DEJA_BIN` and on the installer's own entries, so every call is attributable to the half that made it.

**`deja install --auto` does not change.** No installer code was touched (`git diff origin/main...HEAD -- cmd/deja/install*.go` is empty): kimi still gets `mcp.json` plus the marked `[[hooks]]` block, and that stays the front door. The plugin is the other route in.

**Every install order, live:**

| state | plugin runs | installer runs |
|---|---|---|
| plugin only | `mcp`, `hook-prompt --plain` | — |
| plugin + `deja install kimi` | `hook-prompt --plain` | `mcp` |
| plugin + `deja install kimi-auto` | — | `mcp`, `hook-prompt --plain` |
| after `deja uninstall kimi-auto` | `mcp`, `hook-prompt --plain` | — |

Exactly one of each, in every order, with no reinstall in between — the check runs when the hook fires and when the server starts, not once at install time.

**Two defects the audit found, both fixed here:**

1. `deja doctor` called a plugin-only machine `not wired` / `stale`, which sends someone to run an install that only pushes the plugin aside. It now reports `plugin` for both lines. Disabled plugins and records pointing at a deleted managed copy do not count — tests for both.

2. Standing down on the marker alone was wrong. An older deja wired kimi through `SessionStart`, whose output Kimi runs and then throws away; that block still carries the marker, so the plugin would have stood down for wiring that recalls nothing — no recall from either half. It now stands down only for a block that still calls `hook-prompt`, which is the same rule doctor uses for `stale`. Verified live: with a stale block present the plugin recalls.

**Old installs are detectable.** The marker string has not changed since it was introduced (#362), so anything `deja install kimi-auto` has ever written is recognised.

**Duplicate skill text is not a risk.** Kimi discovers user skills from `~/.agents/skills` — where the installer writes the shared skill — and keys its catalog by normalised name, so the plugin's `deja-history` and the installer's collapse to one entry; `TestBundledSkillsMatchInstaller` pins the two texts to be identical. Commands cannot collide either: Kimi namespaces a plugin's command as `/deja:recall`.

**With deja missing entirely:** the hook exits 0 with no output (fail-open, verified), and the MCP launcher prints the install line and hands back the exit code rather than reporting an empty success.

**What the plugin cannot see, now documented in its README:** an MCP entry added by hand under a name other than `deja`, and a hook someone wrote themselves that calls deja without the marker. Windows is reasoned, not exercised: hooks run through `shell: true`, and `command: "node"` is the form Kimi rewrites to its own runtime.

PR #1590, opened 2026-08-23
fix(stats): pluralise singular impact counts

### LINKED ISSUE #1586: the impact screen says "1 recalls returned matches"
`deja stats --impact` prints two counts with a hardcoded plural, and one is exactly the state a new user is in. Reproduced with a usage log holding one recall and one injection:

```
deja impact — measured on this machine, nothing modeled
  recalls served     1 agent-initiated recalls returned matches
  memory at start    1 session starts began with project memory
```

`cmd/deja/stats_impact.go:55-56`:

```go
fmt.Fprintf(w, "  recalls served     %d agent-initiated recalls returned matches\n", r.Recalls)
fmt.Fprintf(w, "  memory at start    %d session starts began with project memory\n", r.Injections)
```

Two nouns and two verbs to agree: `1 agent-initiated recall returned matches`, `1 session start began with project memory`. Package `main` already has `pluralS` (`cmd/deja/brief.go`), so the counts need no new helper, but the verbs do need a decision — either a small `wasWere`-style helper or a reworded line that reads the same at any count. Say which you chose and why.

Same class as #1578, different screen. #1578 fixed the incremental index line; this one was found while reviewing that fix, and the sweep behind it found no other reachable case: `hook_plan.go:311` looks like the same bug but is guarded by `index.FrictionMinSessions`, so it never prints `1 sessions`.

A test belongs with it. `cmd/deja/stats_impact_test.go` already renders the screen from a synthetic usage log, so asserting the singular is a few lines beside what is there.

QinXi-ai:
## Summary

- make the impact report use singular nouns for one recall and one session start
- reuse the existing plural helper and add regression coverage for both singular lines

Closes #1586

## Test plan

- `go test ./cmd/deja -run 'TestStatsImpactCountsAndArithmetic|TestPrintImpactSingularGrammar'`
- `go vet ./...`
- `go build ./cmd/deja`
- `gofmt -s -d cmd/deja/stats_impact.go cmd/deja/stats_impact_test.go`
- `git diff --check`

## Notes

The race-enabled suite could not run on this Windows host because `gcc` is unavailable. A broader non-race Windows run also encountered existing executable-bit/worktree-path failures and an `os/exec.LookPath` timeout; the affected package tests above pass.

### DIFF
--- a/cmd/deja/stats_impact.go
+++ b/cmd/deja/stats_impact.go
@@ -52,8 +52,8 @@ func printImpact(w io.Writer, r usage.ImpactReport, credits int, jsonOut bool) e
 		return nil
 	}
 	fmt.Fprintln(w, "deja impact — measured on this machine, nothing modeled")
-	fmt.Fprintf(w, "  recalls served     %d agent-initiated recalls returned matches\n", r.Recalls)
-	fmt.Fprintf(w, "  memory at start    %d session starts began with project memory\n", r.Injections)
+	fmt.Fprintf(w, "  recalls served     %d agent-initiated recall%s returned matches\n", r.Recalls, pluralS(r.Recalls))
+	fmt.Fprintf(w, "  memory at start    %d session start%s began with project memory\n", r.Injections, pluralS(r.Injections))
 	if r.RawBytes > 0 && r.ServedBytes > 0 {
 		// The frame, the header and the session lines cost more than the text
 		// they wrap when sessions are short — which is exactly the state a new
--- a/cmd/deja/stats_impact_test.go
+++ b/cmd/deja/stats_impact_test.go
@@ -24,8 +24,8 @@ func TestStatsImpactCountsAndArithmetic(t *testing.T) {
 	}
 	got := out.String()
 	for _, want := range []string{
-		"2 agent-initiated recalls",
-		"1 session starts began with project memory",
+		"2 agent-initiated recalls returned matches",
+		"1 session start began with project memory",
 		"1 session recalled 2+ times",
 		"1 prompts matched work",
 		"50× less",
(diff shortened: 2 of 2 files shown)


vshulcz:
Read the whole diff and ran it. The change is correct and the test is real: `TestPrintImpactSingularGrammar` fails on `main` and passes with your fix, and you spotted that the existing test's `1 session starts began with project memory` assertion had to move with it.

I am going to merge #1675 instead, and I would rather say plainly why than leave you guessing.

The screen has a third line with the same bug — `déjà vu moments %d prompts matched work you had already done` at `stats_impact.go:80`, guarded only by `> 0`, so it is reachable at one. The issue said the sweep found no other reachable case; the issue was wrong, and that PR checked rather than trusted it. It also noticed that `TestStatsImpactCountsAndArithmetic` was asserting `1 prompts matched work` — the suite was holding the bug in place — where this PR keeps that assertion as it was. So one of the three lines, and the test that encodes it, would still be wrong on the same screen for the same user.

None of that is a criticism of the fix you wrote: you were first, you were right, and you did the part the issue asked for. It is a difference in how far each of you went past the issue text, and that is the thing worth naming rather than hiding behind "duplicate".

Two things I would rather you spend the next hour on than a rebase here:

- #1712 — `doctor --deep` says "index matches sources" while an altered record makes a session unreachable. It is a real correctness gap with a clear reproduction, not cosmetics.
- #1661 — `deja help` is a fixed 148-column page, so 12 of its 84 lines wrap on a default 80-column terminal.

Both are unclaimed. If either looks interesting, say so on the issue and it is yours. And if you would rather take the plural-direction coverage I mention in #1675 — no test asserts the plural of `session starts` or `prompts`, which I confirmed by mutation — say the word and I will leave it for you instead of writing it myself.

vshulcz:
Closing against #1675, for the reasons above — not because anything here is wrong. Thanks for finding it and for moving the stale assertion with the fix; the two issues I named are yours if you want them.

PR #1669, opened 2026-08-24
fix(install): write a config back with the line endings it already had

### LINKED ISSUE #1668: install rewrites a CRLF config as LF, or worse leaves it half and half
A Windows user's agent configs are usually CRLF. `deja install` does not keep them that way: four JSON configs are rewritten LF-only, and four TOML/YAML configs end up with mixed endings — deja's appended block is LF inside a file that is otherwise CRLF.

## Reproduction

Nine configs written with CRLF, then one `deja install --auto`:

```
$ for f in ...; do printf '%s' "$content" | unix2dos > $f; done
$ deja install --auto; echo "exit=$?"
exit=0

                                     before        after
  .gemini/settings.json              CR=6 LF=6     CR=0  LF=17   all CRLF lost
  .cursor/mcp.json                   CR=5 LF=5     CR=0  LF=13   all CRLF lost
  .claude/settings.json              CR=3 LF=3     CR=0  LF=63   all CRLF lost
  .qwen/settings.json                CR=3 LF=3     CR=0  LF=24   all CRLF lost
  .config/opencode/opencode.json     CR=3 LF=3     CR=0  LF=12   all CRLF lost
  .codex/config.toml                 CR=4 LF=4     CR=4  LF=9    mixed
  .kimi-code/config.toml             CR=1 LF=1     CR=1  LF=7    mixed
  .grok/config.toml                  CR=2 LF=2     CR=2  LF=6    mixed
  .hermes/config.yaml                CR=1 LF=1     CR=1  LF=12   mixed
```

(CR and LF are byte counts: `tr -cd '\r' < $f | wc -c`.)

The install itself succeeds — this is not a failure, it is a file the user did not ask deja to reformat.

## Impact

- The JSON five: a one-entry addition shows up as a whole-file rewrite in whatever tracks that dotfile.
- The TOML/YAML four are the worse half: one file, two conventions. Editors that pick an ending per file now guess from a majority that deja changed, and the next hand-edit flips the rest of the file.

## The repo already does this correctly in two places

- `cmd/deja/install_goose.go:77` converts its output back to CRLF when the file it read was CRLF.
- `cmd/deja/guidance.go:391` picks its newline from the existing file: `if strings.Contains(old, "\r\n") { newline = "\r\n" }`.

Everything else goes through `writeIfChanged(path, old, next)` (cmd/deja/install.go:735), which has both the old bytes and the new ones in hand and does not look at the endings.

## Expected

A file that was CRLF is written back CRLF. A file that was LF stays LF. A file deja creates itself keeps today's LF.

One caveat for whoever fixes it: the conversion has to happen *before* the `bytes.Equal(old, next)` comparison at the top of `writeIfChanged`, or every repeat install rewrites the file and reports it changed — the idempotence checked in the two iterations before this one would break.

Related: #1577 (Windows is a shipped platform), #1665 (the same JSON re-serialisation rewriting key order).

vshulcz:
A CRLF config came back LF-only from the five JSON targets, and half and half from the TOML/YAML ones — deja's appended block in LF inside a file that was CRLF throughout. Every writer already hands `writeIfChanged` both the old bytes and the new ones, so that is where the endings are restored: `matchLineEndings` converts the new content to CRLF when the file it is replacing was CRLF throughout. A mixed file is left alone — converting it whole would rewrite lines deja never touched.

The conversion happens **before** the `bytes.Equal(old, next)` comparison, or a repeat install would rewrite every CRLF config and report it changed.

That exposed a second defect the CRLF loss was hiding, fixed here too (see the issue comment for the transcript): `installKimiAuto`, `installHermesMCP` and `setHermesPluginEnabled` splice by counting `\n`, so on a CRLF file they do not recognise their own block from last time — kimi grows a blank line per install and hermes appends `mcp_servers:` and `plugins.enabled` again. They now work on LF text via `lfText` and let `writeIfChanged` put the endings back. Today this is invisible because the first install flattens the file to LF; preserving the endings makes it reachable, so the two changes have to land together.

Fail-before tests: `TestInstallKeepsCRLFConfigs` (four configs, all CRLF after install), `TestInstallCRLFStaysUnchangedOnRepeat` (six configs, byte-identical on the second run — this is the one that caught the duplication) and `TestInstallLeavesLFConfigsAlone` (no CR appears in a file that had none).

Closes #1668.

### DIFF
--- a/cmd/deja/install.go
+++ b/cmd/deja/install.go
@@ -732,7 +732,47 @@ func mentionsDeja(b []byte) bool {
 	return false
 }
 
+// lfText is the text of a config normalised to LF, for the writers that splice
+// blocks by counting newlines. They then work in one convention and
+// writeIfChanged puts the file's own endings back (#1668).
+func lfText(b []byte) string {
+	return strings.ReplaceAll(string(b), "\r\n", "\n")
+}
+
+// matchLineEndings writes back the line endings the file already had. A
+// Windows user's configs are CRLF, and install rewrote the JSON ones LF-only
+// while leaving the TOML and YAML ones half and half — deja's appended block in
+// LF inside a CRLF file (#1668). install_goose.go and guidance.go already did
+// this for their own two files; every other writer comes through here.
+//
+// Only a file whose endings are *all* CRLF is treated as a CRLF file. A mixed
+// one is left alone: converting it whole would rewrite lines deja never touched,
+// which is the thing this exists to avoid.
+func matchLineEndings(old, next []byte) []byte {
+	if len(old) == 0 || !bytes.Contains(old, []byte("\r\n")) {
+		return next
+	}
+	if bytes.Count(old, []byte("\n")) != bytes.Count(old, []byte("\r\n")) {
+		return next
+	}
+	// Only the newlines that are not already CRLF, so a writer that converted
+	// its own output first is not given a second carriage return.
+	var b bytes.Buffer
+	b.Grow(len(next) + bytes.Count(next, []byte("\n")))
+	for i := 0; i < len(next); i++ {
+		if next[i] == '\n' && (i == 0 || next[i-1] != '\r') {
+			b.WriteByte('\r')
+		}
+		b.WriteByte(next[i])
+	}
+	return b.Bytes()
+}
+
 func writeIfChanged(path string, old, next []byte) (string, error) {
+	// Before the comparison, not after: a CRLF config converted afterwards
+	// would differ from `old` on every run, so each repeat install would
+	// rewrite the file and report it changed.
+	next = matchLineEndings(old, next)
 	if bytes.Equal(old, next) {
 		return "unchanged", nil
 	}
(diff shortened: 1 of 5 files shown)


vshulcz:
Review caught a fourth writer with the same shape: `dshPatchWith(string(old), …)` (cmd/deja/install_deepseek.go:305) splits on `\n` without normalising, so the user's lines keep a trailing `\r` while deja's block is LF.

Checked before changing it — the outcome is not currently observable, because `matchLineEndings` converts whatever mix comes out into uniform CRLF, and the `# deja mcp:start` marker does not span a line ending, so a second install still matches its own block:

```
$ printf '# dsh patch\r\n- insert:\r\n    - id: mine\r\n      name: my-plugin\r\n' > ~/.dsh/cordis.patch.yml
$ deja install deepseek >/dev/null; deja install deepseek >/dev/null
second run: identical
marker count: 1
lines with CR not at end: 0
```

Fixed anyway, for the same reason the other three were: the splicer should not depend on a later conversion to clean up after it. `.dsh/cordis.patch.yml` is now in the repeat-install test alongside the other six.

PR #1675, opened 2026-08-24
stats: make the impact screen agree with a count of one

### LINKED ISSUE #1586: the impact screen says "1 recalls returned matches"

MsfPablo:
Closes #1586.

`deja stats --impact` hardcoded plural nouns, so a usage log holding one recall and one injection — the state a new user is in — rendered:

```
  recalls served     1 agent-initiated recalls returned matches
  memory at start    1 session starts began with project memory
```

## The judgement call the issue asked about

No verb helper, and no rewording. Both verbs on these lines are past tense, and English past tense does not inflect for number: `returned` and `began` are already correct at one and at many. Only the nouns disagreed, so `pluralS` on `recall`, `session start` and `prompt` is the whole fix.

That is why I did not add a `wasWere`-style helper — there is nothing for it to do here. If a line ever needs `is`/`are`, that is the point to add one, against a real case.

## One line beyond the issue text

The issue names two lines and reports the sweep found no other reachable case. There is a third on the same screen:

```go
// stats_impact.go:80
fmt.Fprintf(w, "  déjà vu moments    %d prompts matched work you had already done\n", r.DejaVuMoments)
```

It is guarded only by `> 0`, so it is reachable at one, and `TestStatsImpactCountsAndArithmetic` was already asserting `"1 prompts matched work"` — the existing test encoded the bug rather than catching it. Fixing two of the three would have left the same screen wrong for the same user in the same session, so I included it. Happy to split it into its own issue and PR if you would rather keep this one to the two lines you filed.

## Tests

- `TestStatsImpactAgreesWithCount` (new, table-driven, stdlib) — renders the screen at one and at many, asserts the singular forms appear and that the hardcoded plurals (`1 agent-initiated recalls`, `1 session starts`, `1 prompts`) do not. Both subtests fail on `main` before the change.
- `TestStatsImpactCountsAndArithmetic` — updated the two assertions that spelled the bug; its JSON assertions are untouched, since no report field changed.

Hermetic: index path is `t.TempDir()`, no HOME or `DEJA_*` read on this path, no platform-specific behavior.

## Verification

```text

ok  github.com/vshulcz/deja-vu/cmd/deja  69.788s  coverage: 89.8% of statements
total: (statements) 89.9%
```

`cmd/deja` floor in CI is 88; this stays above it. No documentation mentions these strings.

### DIFF
--- a/cmd/deja/stats_impact.go
+++ b/cmd/deja/stats_impact.go
@@ -52,8 +52,8 @@ func printImpact(w io.Writer, r usage.ImpactReport, credits int, jsonOut bool) e
 		return nil
 	}
 	fmt.Fprintln(w, "deja impact — measured on this machine, nothing modeled")
-	fmt.Fprintf(w, "  recalls served     %d agent-initiated recalls returned matches\n", r.Recalls)
-	fmt.Fprintf(w, "  memory at start    %d session starts began with project memory\n", r.Injections)
+	fmt.Fprintf(w, "  recalls served     %d agent-initiated recall%s returned matches\n", r.Recalls, pluralS(r.Recalls))
+	fmt.Fprintf(w, "  memory at start    %d session start%s began with project memory\n", r.Injections, pluralS(r.Injections))
 	if r.RawBytes > 0 && r.ServedBytes > 0 {
 		// The frame, the header and the session lines cost more than the text
 		// they wrap when sessions are short — which is exactly the state a new
@@ -77,7 +77,7 @@ func printImpact(w io.Writer, r usage.ImpactReport, credits int, jsonOut bool) e
 		fmt.Fprintf(w, "  knowledge re-used  %d session%s recalled 2+ times — fixes that keep paying\n", r.ReusedTwice, pluralS(r.ReusedTwice))
 	}
 	if r.DejaVuMoments > 0 {
-		fmt.Fprintf(w, "  déjà vu moments    %d prompts matched work you had already done\n", r.DejaVuMoments)
+		fmt.Fprintf(w, "  déjà vu moments    %d prompt%s matched work you had already done\n", r.DejaVuMoments, pluralS(r.DejaVuMoments))
 	}
 	served := r.Recalls + r.Injections
 	switch {
--- a/cmd/deja/stats_impact_test.go
+++ b/cmd/deja/stats_impact_test.go
@@ -25,9 +25,9 @@ func TestStatsImpactCountsAndArithmetic(t *testing.T) {
 	got := out.String()
 	for _, want := range []string{
 		"2 agent-initiated recalls",
-		"1 session starts began with project memory",
+		"1 session start began with project memory",
 		"1 session recalled 2+ times",
-		"1 prompts matched work",
+		"1 prompt matched work",
 		"50× less",
 	} {
 		if !strings.Contains(got, want) {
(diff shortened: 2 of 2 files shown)


vshulcz:
Read the whole diff and ran it here, on macOS.

**Verified, not just read.** Your `TestStatsImpactAgreesWithCount` on this branch's test file against `main`'s `stats_impact.go` fails on both subtests, printing `1 session starts` and `1 prompts`; with your change it passes. That is the property a regression test has to have and most do not.

The third line was the right call to include. I read the whole of `printImpact` looking for a fourth: `knowledge re-used` was already on `pluralS`, and `credited aloud` has no countable noun, so after this the screen has no hardcoded plural left. Your reading of the existing test is also correct — it asserted `1 prompts matched work`, so the suite was holding the bug in place rather than catching it.

And the verb answer is right. `returned` and `began` are past tense and do not inflect for number, so a `wasWere` helper would have had nothing to do. Adding the helper against a real case rather than a hypothetical one is the better order.

**One note, not a blocker.** The invariant only holds in the singular direction. Your `plural` subtest varies `recalls` but leaves injections and déjà vu moments at one, so no test anywhere asserts the plural of `session starts` or `prompts`. I checked by mutation rather than by eye: dropping `pluralS` from the `memory at start` line — hardcoding it the other way, to singular — leaves `go test ./cmd/deja -run TestStatsImpact` green. Nothing to change here; I will add the other direction in a follow-up rather than send you back for it.

The rest holds up: hermetic through `t.TempDir()`, no `DEJA_*` or `HOME` on this path, JSON assertions untouched because no report field changed.

CI needed a maintainer to approve the runs — first pull request from a fork. Approved, they are running now.

vshulcz:
Thank you — the part I want to name is not the fix, it is that you ran the existing test and read what it was asserting. `1 prompts matched work` was sitting in the suite as a passing assertion, and finding a third broken line that the issue explicitly said did not exist is the kind of thing a reviewer usually has to catch after the fact. You also answered the verb question with a reason instead of adding a helper for symmetry.

All checks green. Merging.

vshulcz:
The plural-direction coverage is in as #1735 — both mutations fail now. Nothing needed from you.

PR #1687, opened 2026-08-24
fix(install): name the config a permission error stopped, not deja's scratch file

### LINKED ISSUE #1686: a read-only config directory is reported as a permission error on a temp file the user cannot find
With `~/.codex` read-only, install refuses the target and names a file that does not exist and never will:

```
$ chmod 555 ~/.codex
$ deja install --all
…
deja: install finished what it could; 2 targets refused: claude-code: mkdir ~/.claude/skills: permission denied; codex: open ~/.codex/.deja-tmp-4168817699: permission denied — check those paths' permissions and run it again
$ echo $?
1
```

`~/.codex/.deja-tmp-4168817699` is the scratch file `writeIfChanged` creates to write atomically (cmd/deja/install.go:855). The reader cannot look at it, cannot chmod it, and if they search for it they find nothing. What they need is the config deja was trying to write — `~/.codex/config.toml`.

The claude-code half of the same line shows what it should look like: `mkdir ~/.claude/skills: permission denied` names a real path.

This is the shape #865 fixed for `update` — "names a temporary file and a package manager it could have named exactly".

## Control — everything else about the read-only case is right

Loud, per-target, and it keeps going rather than stopping at the first refusal (#902):

```
$ deja install --all      # .claude and .codex read-only, .gemini and .cursor writable
cursor: created ~/.cursor/mcp.json
gemini: created ~/.gemini/settings.json
deja: install finished what it could; 2 targets refused: … 
$ echo $?
1
```

An unreadable config is refused with its own name, correctly:

```
$ chmod 000 ~/.gemini/settings.json && deja install gemini
deja: install finished what it could; 1 target refused: gemini: open ~/.gemini/settings.json: permission denied — check that path's permissions and run it again
```

And a read-only *file* in a writable directory installs, preserving both the mode and the user's content — the atomic rename does not need to open the old file for writing:

```
$ chmod 444 ~/.codex/config.toml && deja install codex && ls -l ~/.codex/config.toml
codex: updated ~/.codex/config.toml
-r--r--r--  … config.toml
```

## Also

`cmd/deja/update.go:500` stages through `.deja-update-*` and returns that error raw as well. Same one-line shape; not fixed here because it is a different command with its own reproduction.

vshulcz:
With `~/.codex` read-only and no config in it yet, install refused the target and named `~/.codex/.deja-tmp-4168817699` — the scratch file `writeIfChanged` creates to write atomically. The reader cannot look at it, chmod it, or find it. The claude-code half of the same line already named a real path.

`os.CreateTemp`'s `*PathError` now has its `Path` rewritten to the destination before it is returned. Rewriting in place rather than wrapping keeps it a `*PathError`, so `errors.Is(err, fs.ErrPermission)` still holds and `refusalRemedy` still ends the run with the permissions sentence — that is asserted in the test, because the obvious alternative (wrapping in a new error) would quietly cost it.

Fail-before tests: `TestWriteIfChangedReportsTheDestination` on the helper directly, and `TestInstallNamesTheConfigNotTheTempFile` end to end. The second one needed care — with a config already present the *backup* step fails first and names a real path, so the fixture has to leave the read-only directory empty, which is the original reproduction. Both skip as root.

Closes #1686.

Everything else about the read-only case was already right and is quoted in the issue: exit 1, one line per refused target, the run continues to the writable harnesses, an unreadable config names itself, and a read-only *file* in a writable directory installs with its mode and content preserved.

`cmd/deja/update.go:500` stages through `.deja-update-*` and returns that error raw too. Same one line, different command, its own reproduction — left alone.

### DIFF
--- a/cmd/deja/install.go
+++ b/cmd/deja/install.go
@@ -860,8 +860,14 @@ func writeIfChanged(path string, old, next []byte) (string, error) {
 		// chmod-ed or found (#1686, the shape of #865). Rewriting the path in
 		// place keeps the error a *PathError, so errors.Is still sees the
 		// permission underneath and the remedy still names permissions.
+		//
+		// Only for a permission denial: the destination is the right thing to
+		// name when the directory refuses us, and the wrong thing when the
+		// failure is about the scratch file itself — "config.toml: no such
+		// file or directory" would send the reader after a directory that is
+		// the actual problem.
 		var pe *os.PathError
-		if errors.As(terr, &pe) {
+		if errors.Is(terr, fs.ErrPermission) && errors.As(terr, &pe) {
 			pe.Path = path
 		}
 		return "", terr
--- a/cmd/deja/install.go
+++ b/cmd/deja/install.go
@@ -854,6 +854,16 @@ func writeIfChanged(path string, old, next []byte) (string, error) {
 	}
 	tmp, terr := os.CreateTemp(filepath.Dir(path), ".deja-tmp-")
 	if terr != nil {
+		// The scratch file is deja's business; the reader's is the config it
+		// could not write. A read-only ~/.codex was reported as a permission
+		// error on ~/.codex/.deja-tmp-4168817699, which cannot be looked at,
+		// chmod-ed or found (#1686, the shape of #865). Rewriting the path in
+		// place keeps the error a *PathError, so errors.Is still sees the
+		// permission underneath and the remedy still names permissions.
+		var pe *os.PathError
+		if errors.As(terr, &pe) {
+			pe.Path = path
+		}
 		return "", terr
 	}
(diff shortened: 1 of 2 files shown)


vshulcz:
Review found two things, both applied.

The path rewrite now happens only when `errors.Is(terr, fs.ErrPermission)`. Naming the destination is right when the directory refuses us and wrong when the failure is about the scratch file itself — `"config.toml: no such file or directory"` would send the reader after a directory that is the actual problem.

Both tests now skip on Windows as well as root, matching the house pattern in `readonly_search_test.go` and `sources_denied_test.go`: a 0555 directory does not deny writes there, so the tests would have failed on the Windows matrix instead of skipping. Verified with `-count=2`.

Issue #1683, opened 2026-08-24
uninstall --all keeps the shared skill file, because it asks a record the same run is about to empty


vshulcz:
`deja uninstall --all` leaves `~/.agents/skills/deja-history/SKILL.md` on disk with deja's guidance in it, and says so plainly:

```
$ deja install --all >/dev/null
$ deja uninstall --all | grep '\.agents'
codex: guidance kept ~/.agents/skills/deja-history/SKILL.md
codex-auto: guidance kept ~/.agents/skills/deja-history/SKILL.md
cursor: guidance kept ~/.agents/skills/deja-history/SKILL.md
…
$ grep -c deja ~/.agents/skills/deja-history/SKILL.md
5
```

The file did not exist before the install:

```
$ ls ~/.agents
ls: ~/.agents: No such file or directory
```

## Cause

`sharedSkillStillWanted` (cmd/deja/guidance.go:247) asks whether any *other* harness still reads the shared skill, and answers from `readWiringState().Targets`:

```go
for _, t := range targets {
    other := guidanceHarness(t)
    if other != leaving && sharedSkillHarnesses[other] {
        return true
    }
}
```

During `uninstall --all` that record still lists every harness — `recordWiring` runs deferred, at the end of the run — so for each target in turn there is always another shared reader "still wanting" the file, and it is kept every time. The guard is right for one harness leaving and wrong for the run that removes them all, because it consults a record this very run is about to empty.

## Expected

Nothing deja wrote survives a full uninstall. Per-harness uninstall keeps the shared file, which is the behaviour the guard exists for (#840's neighbourhood) and must not change.

## Impact

Cosmetic on its own — but the file tells an agent to call deja, and deja is gone. The next session on codex, cursor, gemini, kimi, deepseek or openclaw reads an instruction to use a binary that is no longer installed.

vshulcz:
Fixed by #1685, on main since the collector merge in #1912.

PR #1694, opened 2026-08-24
fix(install): ignore a relative XDG_CONFIG_HOME, as the spec requires

### LINKED ISSUE #1693: a relative XDG_CONFIG_HOME is honoured, so wiring.json lands wherever the command was run
`XDG_CONFIG_HOME` set to a relative path makes deja write its wiring record into whatever directory the command was run from:

```
$ cd /tmp/somewhere
$ XDG_CONFIG_HOME=relcfg deja install claude-code
claude-code: guidance created ~/.claude/skills/deja-history/SKILL.md
claude-code: created ~/.claude.json
$ echo $?
0
$ find /tmp/somewhere -name wiring.json
/tmp/somewhere/relcfg/deja/wiring.json
```

The XDG Base Directory specification is explicit about this:

> All paths set in these environment variables must be absolute. If an implementation encounters a relative path in any of these variables it should consider the path invalid and ignore it.

deja honours it instead, relative to the process's working directory. The record then lives somewhere no later run will look: `deja update`'s wiring refresh, and the shared-skill guard fixed in #1683, both read `wiringStatePath()` and would find nothing — or find a different file, depending on where the user happened to be standing.

Same family as #1690: state written relative to the working directory, silently, exit 0.

## Control — a space in the path is handled correctly

```
$ XDG_CONFIG_HOME="/tmp/my cfg" deja install claude-code
claude-code: created ~/.claude.json
$ find "/tmp/my cfg" -name wiring.json
/tmp/my cfg/deja/wiring.json
```

So this is about the relative path specifically, not about path handling in general.

## Cause

Five places read the variable with the same shape, none checking that it is absolute:

- cmd/deja/wiring_state.go:38
- cmd/deja/install_aider.go:24
- cmd/deja/install_sync_timer.go:102
- cmd/deja/install_goose.go:119
- cmd/deja/guidance.go:119

```go
base := os.Getenv("XDG_CONFIG_HOME")
if base == "" {
	base = filepath.Join(homeDir(), ".config")
}
```

Nineteen `Getenv("XDG_…")` reads exist in total; the others (`XDG_DATA_HOME` in `internal/sources/opencode.go`, `XDG_CACHE_HOME`) deliberately mirror each harness's own path logic and are a separate question — upstream may honour a relative value, and deja has to find the files where the harness actually put them.

vshulcz:
`XDG_CONFIG_HOME=relcfg deja install claude-code` wrote `relcfg/deja/wiring.json` into the directory the command ran from. The XDG spec says a relative path in one of these variables is invalid and must be ignored; deja followed it. The record then sits where no later run looks — `deja update`'s wiring refresh and the shared-skill guard from #1683 both read `wiringStatePath()`.

New `xdgConfigHome()` applies the spec rule, and the two paths that are **deja's own** use it: the wiring record and the sync timer's systemd unit (a unit file in the working directory is one systemd will never see).

Deliberately not changed: `opencodeConfigHome`, `gooseConfigHome` and `aiderContextPath` read the same variable to mirror *the harness's* path logic. If opencode or goose honours a relative value, deja has to look where the harness actually put the file, so the spec rule is the wrong rule there. Worth a look of its own — it needs reading each upstream, not a sweep.

Fail-before tests: `TestRelativeXDGConfigHomeIsIgnored` (nothing created in the working directory, record at the spec's fallback, `wiringStatePath()` inside the home) and `TestAbsoluteXDGConfigHomeIsHonoured`, which keeps the control that a real absolute path — spaces and all — is still used.

Closes #1693.

### DIFF
--- a/cmd/deja/install_sync_timer.go
+++ b/cmd/deja/install_sync_timer.go
@@ -99,11 +99,10 @@ func syncAutoPlist(exe string) string {
 }
 
 func syncAutoUnitDir() string {
-	base := os.Getenv("XDG_CONFIG_HOME")
-	if base == "" {
-		base = filepath.Join(homeDir(), ".config")
-	}
-	return filepath.Join(base, "systemd", "user")
+	// deja's own unit file, so the spec's rule applies: a relative
+	// XDG_CONFIG_HOME is ignored rather than followed into the working
+	// directory, where systemd would never look for it (#1693).
+	return filepath.Join(xdgConfigHome(), "systemd", "user")
 }
 
 func installSyncTimerSystemd(exe string, uninstall bool) (installResult, error) {
--- a/cmd/deja/wiring_state.go
+++ b/cmd/deja/wiring_state.go
@@ -35,11 +35,24 @@ type wiringState struct {
 }
 
 func wiringStatePath() string {
-	base := os.Getenv("XDG_CONFIG_HOME")
-	if base == "" {
-		base = filepath.Join(homeDir(), ".config")
+	return filepath.Join(xdgConfigHome(), "deja", "wiring.json")
+}
+
+// xdgConfigHome is the base for the files that are deja's own — the wiring
+// record and the sync timer's unit — not for the ones that mirror a harness's
+// path logic.
+//
+// The XDG spec: "All paths set in these environment variables must be
+// absolute. If an implementation encounters a relative path in any of these
+// variables it should consider the path invalid and ignore it." deja honoured
+// a relative value instead, so `XDG_CONFIG_HOME=relcfg deja install` wrote
+// wiring.json into whatever directory the command ran from, where no later run
+// would look for it (#1693).
+func xdgConfigHome() string {
+	if base := os.Getenv("XDG_CONFIG_HOME"); filepath.IsAbs(base) {
+		return base
 	}
-	return filepath.Join(base, "deja", "wiring.json")
+	return filepath.Join(homeDir(), ".config")
 }
 
 func readWiringState() wiringState {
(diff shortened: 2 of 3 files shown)


vshulcz:
Review raised the upgrade path for someone whose `XDG_CONFIG_HOME` is relative today. Checked both halves; neither needs migration, and one of them cannot have it.

**The wiring record.** There is no single old location to migrate from. With a relative value the path was resolved against the working directory of each run, so a user who installed from two directories has two `wiring.json` files and neither is authoritative. What is lost is a record deja can rebuild: `recordWiring` rewrites it on the next `install`, and the readers degrade the way they are written to — `sharedSkillStillWanted` returns true on an empty record (keeps the shared file rather than removing one another harness still reads), and `refreshWiringAfterUpgrade` returns early and refreshes nothing. That is the same state as a machine that has never installed, which is a state the code already handles.

**The sync timer.** The unit was never live. `installSyncTimerSystemd` enables by name — `systemctl --user enable --now deja-sync.timer` — and systemd resolves that name in its own search path, never in a relative directory. So with a relative `XDG_CONFIG_HOME` the write landed somewhere systemd cannot see and `enable` failed, which the code reports as `wrote %s but systemctl refused it`. Nobody has a working timer installed that way for this change to orphan; what they have is a file in a directory that was never a unit directory.

**On the asymmetry** — whether opencode, goose and aider themselves ignore a relative `XDG_CONFIG_HOME` — that is the open question, and it is why those three are untouched here rather than swept along. deja has to find the file where the harness actually put it, which means reading each upstream's path logic, not applying the spec on their behalf. Queued as its own item.

PR #1696, opened 2026-08-24
fix(install): stop writing a stray comma into an opencode.jsonc that uses trailing commas

### LINKED ISSUE #1695: install puts a stray comma inside opencode.jsonc, leaving a config that no longer parses
`deja install opencode` writes a stray comma into `opencode.jsonc` when the file uses trailing commas, leaving a config that no longer parses. Exit 0, reported as `updated`.

## Reproduction

```
$ cat ~/.config/opencode/opencode.jsonc
{
  // my opencode config
  "theme": "system",
  "mcp": {
    "mine": {
      "type": "local",
      "command": ["my-server"],
    },
  },
}

$ deja install opencode
opencode: updated ~/.config/opencode/opencode.jsonc
$ echo $?
0

$ cat ~/.config/opencode/opencode.jsonc
{
  // my opencode config
  "theme": "system",
  "mcp": {
    "mine": {,                                    <-- here
      "type": "local",
      "command": ["my-server"],
    },
    "deja": {"type":"local","command":["/usr/local/bin/deja","mcp"]}
  },
}
```

`"mine": {,` is not legal in JSON or JSONC. Stripping comments and trailing commas the way a JSONC reader does:

```
Expecting property name enclosed in double quotes: line 5 column 14
```

opencode reads this file at startup. Nothing is deleted, but the user's config is broken and deja said it succeeded.

## Cause

cmd/deja/install.go, `updateOpencodeJSONC` — the loop that puts a comma on the previous entry before appending deja's:

```go
for i := len(body) - 1; i >= 0; i-- {
    trim := strings.TrimSpace(body[i])
    if trim != "" && !strings.HasPrefix(trim, "//") && !strings.HasSuffix(trim, ",") {
        body[i] += ","
        break
    }
}
```

It walks backwards until it finds a line that does *not* already end with a comma. With trailing commas in use — which is the whole point of a `.jsonc` file — every line of the last entry ends with one, so the scan runs past the entry entirely and lands on `"mine": {`, the line that *opens* the nested object.

The assumption is that the last meaningful line is the end of the last entry. It only holds when the file has no trailing comma.

## Expected

Look at the last non-empty, non-comment line only. If it already ends with `,`, append deja's entry and add nothing; otherwise add the comma there and stop. A `.jsonc` written with trailing commas is exactly the file this path exists to handle.

## Related

#1664 asks whether the other five JSON targets should accept comments at all; this one already accepts them and damages the file. The Zed writer (#1683's neighbourhood, `install_zed.go`) does the same job by locating the entry rather than by counting line endings, and iteration 56 confirmed it survives comments and trailing commas untouched.

vshulcz:
`deja install opencode` turned a valid `opencode.jsonc` into one that does not parse, and exited 0 saying `updated`:

```
    "mine": {,
      "type": "local",
```

The comma that joins deja's entry to the previous one was placed by walking backwards to the first line that does not already end in a comma. In a file written with trailing commas — which is what `.jsonc` is for — every line of the last entry ends in one, so the scan ran past the entry and landed on the line that *opens* it.

It now looks at the last line carrying content and stops there, adding a comma only if that line needs one. Lines ending in `{` or `[` never do.

Fail-before test: `TestOpencodeJSONCKeepsTrailingCommasValid` over four shapes — trailing commas, no trailing comma, an empty `mcp` block, and a comment as the last line in the block — each checked for parsing after comments and trailing commas are stripped, for deja's entry being present, for no `{,`, and for the user's own server surviving. Two of the four failed before the change.

Closes #1695.

Verified by hand as well: install, uninstall, install again over the same file leaves it byte-identical to the single-install result, comment and trailing commas intact.

### DIFF
--- a/cmd/deja/install.go
+++ b/cmd/deja/install.go
@@ -1404,6 +1404,76 @@ func updateOpencodeJSON(old []byte, exe string, uninstall bool) ([]byte, error) 
 	return append(next, '\n'), nil
 }
 
+// jsoncLastCodeLine finds the last line of a .jsonc block that a parser would
+// read as code, and where that code ends on it. It returns -1 when the block
+// holds nothing but comments and blank lines.
+//
+// The comma that joins deja's entry to the previous one has to land at the end
+// of the code, not the end of the line. Appended after a trailing // comment it
+// is stripped with the comment and the two entries lose their separator;
+// appended inside a /* … */ block it either vanishes with the block or is left
+// behind as a comma of its own (#1695).
+func jsoncLastCodeLine(body []string) (idx, end int, code string) {
+	idx = -1
+	inBlock := false
+	for i, line := range body {
+		var c string
+		c, inBlock, end = jsoncCodeOf(line, inBlock)
+		if c != "" {
+			idx, code = i, c
+		}
+	}
+	if idx < 0 {
+		return -1, 0, ""
+	}
+	_, _, end = jsoncCodeOf(body[idx], jsoncInBlockAt(body, idx))
+	return idx, end, code
+}
+
+// jsoncInBlockAt reports whether line i starts inside a /* … */ block.
+func jsoncInBlockAt(body []string, i int) bool {
+	inBlock := false
+	for _, line := range body[:i] {
+		_, inBlock, _ = jsoncCodeOf(line, inBlock)
+	}
+	return inBlock
+}
+
+// jsoncCodeOf returns the code a parser reads on one line, whether the line
+// leaves a /* … */ block open, and the offset where that code ends.
+func jsoncCodeOf(line string, inBlock bool) (code string, stillInBlock bool, end int) {
+	var b strings.Builder
+	inString, escaped := false, false
+	end = 0
+	for i := 0; i < len(line); i++ {
+		c := line[i]
+		if inBlock {
+			if c == '*' && i+1 < len(line) && line[i+1] == '/' {
+				inBlock, i = false, i+1
+			}
+			continue
+		}
+		switch {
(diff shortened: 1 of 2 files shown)


vshulcz:
Review found two more shapes that still came out invalid, and both were real. The first fix only knew about lines that are *entirely* a `//` comment.

**A trailing comment on the last entry.** The reviewer's fixture put the comment on the block's closing line, which is outside the body and so was already handled; moved inside, it breaks:

```
  "mcp": {
    "mine": {"type": "local"} // my server
  }
```
became `"mine": {"type": "local"} // my server,` — the comma goes with the comment when a reader strips it, and the two entries lose their separator.

**A block comment as the last thing in the block.** The comma landed on an interior line of the comment (`spanning,`), which either vanishes with the block or is left behind as a comma of its own.

The comma now goes at the end of the *code* on the line rather than the end of the line. `jsoncCodeOf` walks a line tracking string literals and block-comment state, so `"url": "https://example.com"` is not mistaken for a comment, and `jsoncLastCodeLine` carries that state across the block to find the last line a parser would actually read.

The test now covers seven shapes, three of which fail on the parent commit: trailing commas, no trailing comma, empty block, `//` comment last, block comment last, trailing comment on the last entry, and a `//` inside a string value.

One thing worth flagging: the test's own JSONC stripper had to be rewritten as a proper state machine. The first version was two regexes, and it reported both new shapes as broken when the output was fine — the `//` regex was eating the rest of `"https://example.com/my-server"`. It is written out independently rather than calling the production helper, so a bug in that helper cannot hide itself.

PR #1698, opened 2026-08-24
fix(install): refuse a goose extensions: list rather than writing a mapping into it

### LINKED ISSUE #1697: install goose writes its extension into an extensions: list, leaving YAML nothing can parse
`deja install goose` inserts its extension under `extensions:` without looking at what is already there. When the value is a YAML sequence, the result is a mapping and a sequence under one key — a file nothing can parse:

```
$ cat ~/.config/goose/config.yaml
extensions:
  - developer
  - fetch

$ deja install goose; echo $?
0

$ cat ~/.config/goose/config.yaml
extensions:
  deja:
    enabled: true
    type: stdio
    …
  - developer
  - fetch

$ python3 -c 'import yaml;yaml.safe_load(open("...config.yaml"))'
while parsing a block mapping … expected <block end>, but found '-'
```

A list is not a config goose can read either — its schema keys extensions by name — so the file was already wrong. But it was *valid YAML* that a person could open and fix, and after deja it is not, with exit 0 and `updated` on the terminal.

The repository already holds itself to the opposite rule, in the writer next door (`updateOpencodeJSONC`, cmd/deja/install.go):

> An "mcp" key exists but brace-counting could not bound it … Adding a fresh block here would leave the file with two "mcp" keys, which drops the user's other servers. **Refuse rather than corrupt a config that cannot be** [read].

## Cause

cmd/deja/install_goose.go — the insert finds the key and writes immediately after it:

```go
if i := strings.Index("\n"+next, "\nextensions:\n"); i >= 0 {
    at := i + len("\nextensions:\n") - 1
    next = next[:at] + entry + next[at:]
}
```

## Control — every shape that is a mapping is handled correctly

An `extensions:` block written by hand keeps its entries, its comments and its order, deja's entry joins them, a second install is byte-identical, and uninstall puts it back exactly:

```
extensions: ['deja', 'developer', 'fetch']     # after install
second install: identical
extensions: ['developer', 'fetch']             # after uninstall
```

An empty block, an inline `extensions: {}`, a config with no `extensions:` key, and an empty file all come out valid with deja's entry in place.

vshulcz:
`deja install goose` found `extensions:` and inserted its mapping entry straight after it, whatever the value was. Where the value is a sequence the result is a mapping and a sequence under one key — YAML nothing can parse — written with exit 0 and reported as `updated`. A config that was merely wrong for goose came out unreadable by anything.

It now refuses, naming the file and why, and leaves it byte-identical. That is the rule the opencode writer already states for itself: refuse rather than corrupt a config that cannot be read.

`yamlBlockIsSequence` looks at the first line of the block that is neither blank nor a comment; an empty block and a mapping block are both fine and take the existing path.

Fail-before tests: `TestInstallGooseRefusesAnExtensionsList` (refuses, names `extensions`, file unchanged) and `TestInstallGooseKeepsAHandWrittenBlock` as the control — a hand-written mapping keeps its entries and comments, deja's entry joins them, a second install is byte-identical, and uninstall puts the block back exactly.

Closes #1697.

The test reads the block without a YAML library: deja has no YAML parser, and adding one for a test would check the library rather than this code. Worth knowing that my first version of that reader was wrong — it counted `- "mcp"` inside deja's own `args:` as a sequence item and failed the control; it now only looks at the entry indent level.

### DIFF
--- a/cmd/deja/goose_yaml_test.go
+++ b/cmd/deja/goose_yaml_test.go
@@ -39,6 +39,36 @@ func TestInstallGooseRefusesAnExtensionsList(t *testing.T) {
 	}
 }
 
+// An inline value is not a block: the insert missed it and appended a second
+// `extensions:` key, and a parser takes the last of two — deja's — so the
+// user's extensions vanished without a word (#1697).
+func TestInstallGooseRefusesAnInlineExtensions(t *testing.T) {
+	hermeticEnv(t)
+	if os.Getenv("DEJA_GOOSE_ROOT") == "" {
+		t.Skip("no goose root in this environment")
+	}
+	for _, inline := range []string{"extensions: [a, b]\n", "extensions: {a: {enabled: true}}\n"} {
+		cfg := filepath.Join(gooseConfigDir(), "config.yaml")
+		if err := os.MkdirAll(filepath.Dir(cfg), 0o755); err != nil {
+			t.Fatal(err)
+		}
+		if err := os.WriteFile(cfg, []byte(inline), 0o644); err != nil {
+			t.Fatal(err)
+		}
+		if _, err := captureRun(t, "install", "goose"); err == nil {
+			after, _ := os.ReadFile(cfg)
+			t.Errorf("install accepted %q:\n%s", inline, after)
+		}
+		after, err := os.ReadFile(cfg)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if string(after) != inline {
+			t.Errorf("install changed a config it could not edit:\n%s", after)
+		}
+	}
+}
+
 // The control: a mapping — the shape goose actually uses — still gets deja's
 // entry, keeps the user's, and survives a second install and an uninstall.
 func TestInstallGooseKeepsAHandWrittenBlock(t *testing.T) {
--- a/cmd/deja/goose_yaml_test.go
+++ b/cmd/deja/goose_yaml_test.go
@@ -1,0 +1,123 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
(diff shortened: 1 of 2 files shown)


vshulcz:
Review found two more shapes. One is a worse bug than the one this PR started with; the other turned out to be an invalid fixture.

**Inline value — fixed, and it was silent content loss.** `extensions: [a, b]` and `extensions: {a: {enabled: true}}` are not followed by a block, so the insert missed them and appended a *second* `extensions:` key:

```
extensions: [a, b]
extensions:
  deja: …
```

A parser takes the last of two duplicate keys — deja's — so `[a, b]` was gone, with exit 0 and `updated` on the terminal. Now refused, file untouched. Covered by `TestInstallGooseRefusesAnInlineExtensions`.

**Tab-indented block — not a defect.** The reviewer is right that `dropEmptyYAMLKey` only counts spaces, and that a tab-indented block goes down a different path. But YAML forbids tabs as indentation outright, so

```
extensions:
	developer:
		enabled: true
```

was never a file goose or anything else could read — the parser fails on the user's own line 2, before deja's entry. I had a fix and a test for it and took both back out: it would have been machinery for an input that cannot exist, and the diff is more honest without it.

**Nested `extensions:` under a profile** — checked, comes out valid: deja writes a top-level key, which is where goose reads, and leaves the nested one alone.

Fail-before, with the two guards neutralised rather than the files reverted (reverting removes the helpers and the package stops building, which is how I first got a misleading clean run — that plus a stale test cache): `TestInstallGooseRefusesAnExtensionsList` and `TestInstallGooseRefusesAnInlineExtensions` both fail, the control passes.

PR #1702, opened 2026-08-24
fix(install): refuse a cordis.patch.yml where one of deja's markers is missing

### LINKED ISSUE #1701: deleting one of deja's two markers in cordis.patch.yml costs the user their own patch, or installs deja twice
deja's block in `~/.dsh/cordis.patch.yml` is delimited by two comments, and the opening one invites editing:

```
…
# deja mcp:end
```

`stripDSHBlock` (cmd/deja/install_deepseek.go:222) starts skipping at the first marker and stops at the second. Neither half of that is checked, so a user who deletes one marker loses either their own work or their sanity.

## 1. End marker gone: everything below it is deleted

```
$ sed -i '' '/deja mcp:end/d' ~/.dsh/cordis.patch.yml     # the user tidied one line
$ cat >> ~/.dsh/cordis.patch.yml <<'YAML'

# my own patch
- insert:
    - id: my-plugin
      name: "my-plugin.js"
YAML

$ deja install deepseek        # or any `deja update`
$ grep -c my-plugin ~/.dsh/cordis.patch.yml
0
```

The skip runs to end of file, so the user's patch — written *after* deja's block, nothing to do with deja — is gone. Exit 0.

## 2. Start marker gone: the entry is installed twice

```
$ sed -i '' '/deja mcp:start/d' ~/.dsh/cordis.patch.yml
$ deja install deepseek
$ grep -c 'id: mcp-deja' ~/.dsh/cordis.patch.yml
2
```

The orphaned copy is invisible to every later install and uninstall, and dsh loads the MCP client twice. Same shape as #1699 for kimi and what `install_zed.go` repairs for Zed.

## Control — an unedited file is handled correctly

User content before deja's block survives a reinstall and the file stays a two-item list with one marker pair; user content after it survives too; and an edit made *inside* deja's block is overwritten, which is right — the block says it is generated.

## Expected

A file with one marker and not the other cannot be edited safely, and deja should say so and change nothing, as `updateOpencodeJSONC` does when it cannot bound its block and as `installGoose` now does for a shape it does not recognise (#1697).

vshulcz:
deja's block in `cordis.patch.yml` is delimited by two comments, and its first line says "safe to delete". Delete one of them and:

- with the **end** marker gone, the skip ran to end of file and took a patch the user had written below deja's block — silently, exit 0;
- with the **start** marker gone, the orphaned block stayed and a second one was appended, so dsh loaded the MCP client twice and no later install or uninstall could find the orphan.

`dshPatchWith` now returns an error when one marker is present without the other: the file is left byte-identical and the message says which marker is missing. Same rule as `updateOpencodeJSONC` states for itself and as `installGoose` took on in #1697.

Fail-before test: `TestDSHPatchRefusesAHalfMarkedBlock`, with `TestDSHPatchKeepsTheUsersPatch` as the control — both markers present, and a file deja never touched, are edited exactly as before.

Closes #1701.

The function grew an error return, so its two existing tests now check it; nothing else calls it.

Verified end to end: with the end marker removed and a user patch below it, install refuses, the patch is still there and the file is unchanged; with the start marker removed, there is still exactly one `id: mcp-deja`.

### DIFF
--- a/cmd/deja/dsh_patch_test.go
+++ b/cmd/deja/dsh_patch_test.go
@@ -1,8 +1,12 @@
 package main
 
 import (
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/sources"
 )
 
 const dshDejaBlock = dshBlockStart + `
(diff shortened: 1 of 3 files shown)


vshulcz:
Review found five things. Three were real and are fixed; one is a broader question I have queued rather than answered here; one falls out of the fixes.

**Uninstall had no guard at all** — and it is the worse half of the original bug. `deja uninstall deepseek` on a file whose end marker is gone deleted the user's patch, exactly when someone with a hand-edited file most needs it left alone. Both directions now go through the same check.

**Markers in the wrong order passed.** `hasStart != hasEnd` is satisfied when both exist, so an end marker above a start marker slipped through and the skip ran to end of file — the user's patch gone again, exit 0. The check now compares positions.

**`strings.Contains` matched a marker inside a value.** That direction is a false refusal rather than damage, but it would have blocked a perfectly good file. Markers are now matched as whole lines, the way `stripDSHBlock` reads them.

All three are one helper, `checkDSHMarkers`, called from both paths.

**Partial install** — plugins written before the patch file is touched, so a refusal leaves `command.js` and `auto.js` on disk with nothing pointing at them. Real, and not specific to dsh: several targets write more than one thing. Queued as its own item rather than answered inside a marker fix.

**Test gap for the wrong order** — covered now, along with a case asserting a marker quoted inside a value is *not* treated as one.

Fail-before, with both checks neutralised in place: `TestDSHPatchRefusesAHalfMarkedBlock` and `TestUninstallDeepSeekRefusesAHalfMarkedBlock` fail; the controls pass.

After, by hand: uninstall on a half-marked file refuses and `my-plugin` is still there; install with the markers reversed refuses and `my-plugin` is still there.

PR #1704, opened 2026-08-24
fix(install): say when it replaced a skill file it has no record of writing

### LINKED ISSUE #1703: install replaces a SKILL.md it has no record of and calls it "guidance updated"
A `SKILL.md` at deja's path that deja never wrote is replaced on install, and the terminal says only `guidance updated`:

```
$ cat ~/.claude/skills/deja-history/SKILL.md
---
name: deja-history
description: MY OWN skill, nothing to do with deja
---

I wrote this by hand and I want it kept.

$ deja install claude-code
claude-code: guidance updated ~/.claude/skills/deja-history/SKILL.md
$ grep -c "I wrote this by hand" ~/.claude/skills/deja-history/SKILL.md
0
```

A `.bak` is written next to it, so the text is recoverable. Nothing says so.

## Why this is a reporting bug and not an overwrite bug

`skillWasEdited` (cmd/deja/skill_marks.go:70) treats a file with no recorded mark as deja's, and the comment there says why:

> A file with no mark is treated as deja's. That is the state every existing install is in, and calling all of them edited would refuse to update a skill on every machine that has one — the protection starts from the next write.

That trade-off is right, and refusing unmarked files would freeze the skill everywhere it was installed before marks existed. deja genuinely cannot tell its own pre-marks copy from a stranger's file at the same path.

What it can do is say what it did. Replacing content deja has no mark for is exactly the case where the reader should hear about the backup, in the same voice as the recorded-edit case already speaks:

```
skill: kept your edited ~/.claude/skills/deja-history/SKILL.md — `deja install --force` to take deja's version
```

## Expected

When deja replaces a non-empty `SKILL.md` it has no mark for, and the content differs from what it is writing, the line names the replacement and the backup rather than calling it an update.

## Control

- The recorded-and-edited case already keeps the file and says so.
- A fresh install with no file writes it silently, which is right.
- A file deja wrote and nobody touched is byte-identical, so nothing is reported at all.

vshulcz:
A `SKILL.md` at deja's path that deja never wrote is replaced on install, and the only line about it was `guidance updated`. A `.bak` is taken, so nothing is lost — but nothing said so either.

The overwrite itself stays. `skill_marks.go` explains why and the reasoning holds: deja cannot tell its own pre-marks copy from a stranger's file at the same path, and treating every unmarked file as a stranger's would freeze the skill on every machine installed before marks existed. What it can do is report it, in the same voice the recorded-edit case already uses:

```
skill: replaced ~/.claude/skills/deja-history/SKILL.md, which deja has no record of writing — your copy is at ~/.claude/skills/deja-history/SKILL.md.bak
```

Only when the file was non-empty, differs from what deja is writing, and has no mark. A second install says nothing — by then the mark exists.

Fail-before test: `TestInstallReportsReplacingAnUnmarkedSkill`, which also checks the backup holds the user's original. `TestInstallDoesNotCryReplacedForItsOwnSkill` is the control across a first and second install.

Closes #1703.

The repo caught one of my own mistakes on the way: `TestDocCommentsNameWhatTheyDocument` failed because I had inserted the new helper between `skillWasEdited`'s doc comment and its declaration. Good test.

### DIFF
--- a/cmd/deja/guidance.go
+++ b/cmd/deja/guidance.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -296,9 +297,19 @@ func writeGuidanceFile(path string, old, next []byte) (string, error) {
 		fmt.Printf("skill: kept your edited %s — `deja install --force` to take deja's version\n", shortHome(path))
 		return "kept", nil
 	}
+	// deja cannot tell its own pre-marks copy from a file someone else wrote at
+	// the same path, and treating every unmarked file as a stranger's would
+	// freeze the skill on every machine installed before marks existed
+	// (skill_marks.go). What it can do is say what it replaced, rather than
+	// calling it an update and leaving the backup unmentioned (#1703).
+	replacing := len(old) > 0 && !bytes.Equal(old, next) && !skillIsMarked(path)
 	action, err := writeIfChanged(path, old, next)
 	if err != nil {
 		return action, err
+	}
+	if replacing && action != "unchanged" {
+		fmt.Printf("skill: replaced %s, which deja has no record of writing — your copy is at %s\n",
+			shortHome(path), shortHome(path+".bak"))
 	}
 	rememberSkill(path, next)
 	return action, nil
--- a/cmd/deja/guidance.go
+++ b/cmd/deja/guidance.go
@@ -303,6 +303,16 @@ func writeGuidanceFile(path string, old, next []byte) (string, error) {
 	// (skill_marks.go). What it can do is say what it replaced, rather than
 	// calling it an update and leaving the backup unmentioned (#1703).
 	replacing := len(old) > 0 && !bytes.Equal(old, next) && !skillIsMarked(path)
+	if replacing {
+		// backupOnce keeps the first .bak it ever made and skips the rest, so
+		// the promise below would have named a copy of some older file while
+		// the one being destroyed went unsaved. This copy is of the content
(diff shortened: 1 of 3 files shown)


vshulcz:
Review caught a false promise in the line I had just added, and it is worse than the thing this PR set out to fix.

`backupOnce` keeps the first `.bak` it ever wrote and skips the rest. So on a machine whose mark record was lost — a wiped cache — the second replacement printed "your copy is at …SKILL.md.bak" while that file still held the *first* file. The one being destroyed went nowhere:

```
$ printf 'A: my first file\n' > ~/.claude/skills/deja-history/SKILL.md
$ deja install claude-code                       # .bak = A
$ rm ~/.cache/deja/index.db.skills               # marks lost
$ printf 'C: my second file\n' > ~/.claude/skills/deja-history/SKILL.md
$ deja install claude-code
skill: replaced …SKILL.md, which deja has no record of writing — your copy is at …SKILL.md.bak
$ cat ~/.claude/skills/deja-history/SKILL.md.bak
A: my first file
```

C was gone, and the message made the loss look handled — which is a worse failure than the silence I was fixing.

The replacing branch now writes the backup itself, from the content it is about to replace, before `writeIfChanged` runs. An older `.bak` is deja's earlier rescue of a file the user has evidently already dealt with; keeping it while losing the live one is the wrong trade.

Fail-before test: `TestReplacedSkillBackupHoldsWhatWasReplaced`, which loses the marks between two installs and asserts the backup holds the second file. Verified by hand too — `.bak` now reads `C: my second file`.

PR #1706, opened 2026-08-24
fix(install): refuse a guidance file whose block deja cannot bound, and repair a duplicated one

### LINKED ISSUE #1705: a missing guidance marker makes install duplicate the block, and the next uninstall deletes the whole file
deja's guidance sits between two HTML comments in a file the user also writes in. If one marker goes missing, a later install adds a *second* block, and the uninstall after that deletes the file — the user's text with it.

## Reproduction

```
$ deja install grok                       # writes ~/.grok/GROK.md with the block
$ sed -i '' '/guidance:end/d' ~/.grok/GROK.md      # the user tidies one line
$ printf 'MY OWN text below\n' >> ~/.grok/GROK.md

$ deja install grok                       # a reinstall, or any `deja update`
$ grep -n 'guidance:\|MY OWN' ~/.grok/GROK.md
1:<!-- deja guidance:start -->
10:MY OWN text below
12:<!-- deja guidance:start -->
21:<!-- deja guidance:end -->

$ deja uninstall grok
$ cat ~/.grok/GROK.md
cat: ~/.grok/GROK.md: No such file or directory
```

The file is gone. `MY OWN text below` is gone with it, and there is no `.bak` for this path.

## Cause

`guidanceMarkerLines` (cmd/deja/guidance.go:432) looks for the first start marker and then the first end marker after it. With the end marker missing, the install finds no pair, so `updateGuidanceBlock` appends a whole new block — the file now has two starts and one end.

The uninstall then cuts from the *first* start to the *only* end, which spans the orphaned block, the user's line and deja's new block. What is left is empty, and `writeIfChanged`'s uninstall branch deletes an empty result on purpose (#840: "Nothing is left once deja's block comes out, which means deja wrote the whole file"). That reasoning is right in general and wrong here: the file is empty only because the cut swallowed someone else's text.

## Two more shapes, same cause

- **End marker without a start**: same duplication, one start and two ends.
- **Two complete blocks** (a merge, a copied file, an older deja): the install removes only the first and appends a new one, so the file keeps two blocks for ever. Nothing repairs it.

## Control — the ordinary case is right

Text above *and* below an intact block survives a reinstall, and the file keeps exactly one start and one end.

## Expected

deja should not edit a guidance file whose markers it cannot pair, and should say so — the rule #1701 settled for `cordis.patch.yml`. A file with two complete blocks should come out with one, as `install_zed.go` repairs a config carrying deja's old id alongside the current one.

vshulcz:
Delete deja's `guidance:end` marker from `~/.grok/GROK.md`, write a line of your own below it, and the next install appends a *whole second block*: two starts, one end. The uninstall after that cuts from the first start to the only end — across your line and both blocks — and what is left is empty, so the file is **deleted**.

```
$ deja uninstall grok
$ cat ~/.grok/GROK.md
cat: ~/.grok/GROK.md: No such file or directory
```

`updateGuidanceBlock` now refuses when a start marker has no end after it, in both directions, leaving the file byte-identical. And it removes *every* complete block rather than the first, so a file that somehow carries two comes out with one instead of keeping both for ever.

A lone **end** marker is deliberately still accepted: that is what an inline start looks like to a line scanner, and `TestGuidanceBlockPreservesInlineMarkers` documents that a marker written inside a sentence is prose, not deja's. Appending below such a file costs nobody anything. My first version refused it and broke that test — the narrower rule is the right one.

Fail-before tests: `TestGuidanceBlockRefusesUnpairedMarkers` and `TestGuidanceBlockRepairsADuplicate`, with `TestGuidanceBlockKeepsTextAround` as the control — text above and below an intact block survives an install and an uninstall.

Closes #1705.

Verified by hand: install and uninstall both refuse on the half-marked file, `MY OWN text below` is still there, and a file with two blocks comes out with one.

### DIFF
--- a/cmd/deja/guidance.go
+++ b/cmd/deja/guidance.go
@@ -226,7 +226,10 @@ func dropRetiredGuidance(harness string) error {
 			}
 			continue
 		}
-		next := updateGuidanceBlock(string(old), true)
+		next, uerr := updateGuidanceBlock(string(old), true)
+		if uerr != nil {
+			return uerr
+		}
 		// A file that held nothing but our block was ours to begin with, so
 		// leaving it behind empty would be litter rather than someone's content.
 		if strings.TrimSpace(next) == "" {
@@ -385,7 +388,11 @@ func installGuidance(harness string, uninstall bool) (installResult, error) {
 			}
 		}
 	} else {
-		next = []byte(updateGuidanceBlock(string(old), uninstall))
+		updated, uerr := updateGuidanceBlock(string(old), uninstall)
+		if uerr != nil {
+			return installResult{}, uerr
+		}
+		next = []byte(updated)
 	}
 	a, err := writeGuidanceFile(path, old, next)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


vshulcz:
Three findings. One is real and fixed; the other two describe behaviour I think should stay, and I want to say why rather than change it quietly.

**The grok twin — fixed.** `grok` writes `GROK.md` and then `AGENTS.md` from one install. With the twin's markers unpaired, the primary was already rewritten before the refusal. Reproduced with a *stale* primary (an up-to-date one is "unchanged" and hides it):

```
$ deja install grok
… 1 target refused: grok: deja's guidance block has no end marker …
  GROK.md unchanged: NO — the primary was written while the twin refused
```

The twin's markers are now checked before the primary is touched. `TestGuidanceGrokTwinIsCheckedFirst` covers it; after the fix the primary is byte-identical.

**Whitespace around a marker — left as is.** `checkGuidanceMarkers` and `guidanceMarkerLines` both compare the whole line minus its line ending, so a marker with a trailing space or a leading indent is not deja's marker to either of them. They agree, which is what matters: a line deja does not recognise is prose, deja appends its own block below, and nothing is cut. That is the same rule `TestGuidanceBlockPreservesInlineMarkers` already documents for a marker written inside a sentence.

Making the match tolerant would have deja start *claiming* markers it has always ignored, and cutting from a line someone wrote as text — the failure this PR exists to prevent. If the two ever disagreed that would be a bug; they don't.

**Accumulation** follows from the same reading and is bounded by the user: deja adds one block per install, and now removes every block it recognises, so a file cannot grow copies on its own.

Issue #876, opened 2026-08-02
doctor reports a hook as wired when the binary it names is gone


vshulcz:
Install from one location, move (or reinstall elsewhere) the binary, and every config deja wrote still names the old absolute path:

```
$ ~/oldbin/deja install claude-auto
$ rm -rf ~/oldbin
$ deja doctor
Hooks:
  precompact   wired       ~/.claude/settings.json      ← wired to a binary that is not there
```

The harness gets `no such file or directory` on every session start. The repair from #773 exists and works — but it runs from the hook path, which is the one path a dead binary cannot reach; `last` and `doctor` do not repair, so nothing surfaces it:

```
before hook-context: 3 references to the old path
after  last:         3
after  doctor:       3
after  hook-context: 0    ← only reachable if the binary still runs
```

After, doctor names it once for every harness at once, from the wiring record it already keeps:

```
  wiring       stale       configs name /old/path/deja, which is not there — `deja install claude-auto` rewrites them for this binary
```

PR #1709, opened 2026-08-24
fix(doctor): say when the wiring runs a binary that is no longer there

### LINKED ISSUE #1708: the binary moves, every hook exits 127, and doctor still reports the machine as wired
Move the deja binary after installing — a `go install` into a different GOBIN, a brew prefix change, a renamed home — and every hook fails on every prompt while `deja doctor` reports the machine as wired.

```
$ deja install claude-auto          # wires ~/bin/deja into ~/.claude/settings.json
$ mv ~/bin/deja ~/bin/deja-moved

$ ~/bin/deja hook-prompt --plain; echo "exit=$?"
sh: ~/bin/deja: No such file or directory
exit=127

$ deja doctor
MCP wiring:
  claude-code  wired          guidance written     ~/.claude.json
Hooks:
  precompact   wired       ~/.claude/settings.json
$ echo $?
0
```

Both rows are true about the file on disk and useless to the reader: the command in them names a path that no longer exists. Every session-start, pre-tool and pre-compact hook now exits 127, and the only tool whose job is to explain why memory stopped arriving says everything is fine.

`codexHasSeenItsHook` already names this shape in its own comment — "scripted runs end up with no memory while every file on disk looks correctly installed". This is the same failure through a different door.

## Running deja again does not repair it

```
$ ~/bin/deja-moved index
$ grep -o '"command": "[^"]*"' ~/.claude/settings.json | head -1
"command": "~/bin/deja hook-tool-after"
```

The wiring still names the old path.

## What deja already knows

`~/.config/deja/wiring.json` records the executable it wired (`Exe`, cmd/deja/wiring_state.go), so doctor can check that path exists without parsing each harness's config format.

## Expected

`deja doctor` says the wiring points at a binary that is gone, and what to run. Whether deja should *re-point* the wiring by itself when it notices is a separate decision — `refreshWiringAfterUpgrade` does that after an upgrade, and extending it to "the recorded exe has vanished" changes when deja writes to a user's config unasked.

vshulcz:
Move the deja binary after installing and every hook exits 127 on every prompt, while `deja doctor` reports `wired` for each config and exits 0. The rows are true about the files and useless to the reader: the command in them names a path that is gone.

doctor now checks the executable the wiring recorded — one check, in the one place every harness has in common, rather than parsing each config format:

```
MCP wiring:
  binary       the wiring runs ~/bin/deja, which is not there — `deja install --all` re-points it at this binary
  claude-code  wired          guidance written     ~/.claude.json
```

Fail-before test: `TestDoctorNamesAVanishedBinary`, with two controls — a binary that is where the wiring says, and a machine that never installed, both silent.

Closes #1708.

Not done here, and worth a decision of its own: deja does **not** re-point the wiring when it notices. `refreshWiringAfterUpgrade` does that after an upgrade; extending it to "the recorded exe has vanished" means deja writes into a user's config unasked, on a machine where the thing they last did was move a file. The message says what to run instead.

### DIFF
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -807,6 +807,7 @@ func doctorPolicy(w io.Writer, dir string) {
 
 func doctorMCP(w io.Writer) {
 	fmt.Fprintln(w, "MCP wiring:")
+	doctorWiredExe(w)
 	for _, c := range doctorMCPConfigs() {
 		status := "config missing"
 		if doctorExists(c.path) {
--- a/cmd/deja/doctor_auto.go
+++ b/cmd/deja/doctor_auto.go
@@ -188,3 +188,22 @@ func codexHookTrustSection(cfg string) string {
 	}
 	return rest
 }
+
+// doctorWiredExe reports a wiring that points at a binary which is no longer
+// there. Every hook the harnesses run names the executable that installed
+// them, so a deja that moved — a `go install` into a different GOBIN, a brew
+// prefix change, a renamed home — leaves every hook exiting 127 on every
+// prompt while each config file still reads correctly (#1708). The recorded
+// path is checked rather than each harness's own format, which is the one
+// thing all of them have in common.
+func doctorWiredExe(w io.Writer) {
+	st := readWiringState()
+	if st.Exe == "" || len(st.Targets) == 0 {
+		return
+	}
+	if _, err := os.Stat(st.Exe); err == nil {
+		return
+	}
+	fmt.Fprintf(w, "  %-12s the wiring runs %s, which is not there — `deja install --all` re-points it at this binary\n",
+		"binary", shortHome(st.Exe))
+}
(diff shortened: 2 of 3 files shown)


vshulcz:
Closing unmerged: the check already exists.

`doctorWiringExe` (cmd/deja/doctor.go:196, from #876) does exactly this, and its comment describes the same scenario — "Move the binary and those configs keep naming the old one … doctor happily printed \"wired\" for a hook that cannot run". Its message is better than the one I added, because it names the exact install command for the recorded targets:

```
  wiring       stale       configs name ~/bin/deja, which is not there — `deja install claude-auto` rewrites them for this binary
```

That line was in my own reproduction all along, on line 69 of doctor's output. I missed it because the grep I used to read the output ended in `head -5` and the harness store rows came first — the same class of measurement error as reading an exit code through a pipe, which has caught me twice tonight already.

Nothing here to merge. #1708 closed as already fixed.

PR #1711, opened 2026-08-24
fix(doctor): emit the stale_stores the JSON schema documents, and name the real policy file

### LINKED ISSUE #1710: doctor --json omits stale_stores the docs promise, and the docs name a policy file deja does not write
`docs/json-output.md` is the contract for anyone parsing `deja doctor --json`. Two places where it and the output disagree.

## 1. `stale_stores` is documented but never emitted on a healthy machine

The example shows it:

```json
"index": {
  "state": "ok",
  "path": "/home/user/.cache/deja/index.db",
  "stale_stores": 0
}
```

The field carries `omitempty` (cmd/deja/doctor_report.go:47), so the value the example shows is exactly the one that is never written:

```
$ deja index && deja doctor --json | jq .index
{
  "state": "ok",
  "path": "~/.cache/deja/index.db"
}
```

A script written against the example does `report["index"]["stale_stores"]` and raises on every machine whose index is fresh — the common case. The document is explicit about what may be absent — "`embed`, `ingest_health` and `deep` (present only under `--deep`) are omitted when unavailable; `policy` is always present" — and `stale_stores` is not on that list.

## 2. The policy path in the example is a file deja does not write

```json
"policy": {
  "state": "active",
  "path": "/home/user/.config/deja/policy.toml",
```

deja writes and reads `policy.json` (`internal/policy.Path`, policy.go:45), and that is what the output carries:

```
$ deja doctor --json | jq -r .policy.path
~/.config/deja/policy.json
```

The text report says `policy.json` too. Only this example says otherwise, and it is the one place a reader goes to learn the filename.

## Control — the rest of the shape matches

On a fresh machine and on an indexed one, the top-level keys are exactly `schema_version, stores, index, mcp, sqlite3, version, policy`, with `embed` and `ingest_health` absent as documented. Store, index, version and policy `state` values seen in both runs are all from the documented sets, and `stores[0]` carries `name/state/paths/files/indexed_sessions` as described.

vshulcz:
`docs/json-output.md` is the contract for anyone parsing `deja doctor --json`, and it disagreed with the output in two places.

`index.stale_stores` is in the documented example as `0`, and `0` was the one value never emitted: the field carried `omitempty`. A script written against the example raised on every machine whose index is fresh. The document names the three keys that may be absent — `embed`, `ingest_health`, `deep` — and this is not one of them, so the code was wrong, not the doc.

The index now has its own type. `doctorComponent` was shared with `sqlite3`, and simply dropping `omitempty` put a meaningless `stale_stores` in the sqlite3 block — caught by the golden, which is what it is for.

The policy example named `~/.config/deja/policy.toml`. deja reads and writes `policy.json` (`internal/policy.Path`), the text report says so, and this example was the one place a reader would go to learn the filename.

Fail-before tests: `TestDoctorJSONCarriesEveryDocumentedField` over the seven always-present keys plus `index.stale_stores`, and `TestDocsNameThePolicyFileDejaWrites`, which fails while the docs name a file deja does not write.

Closes #1710.

The golden gained the field; the rest of the shape was already right on both a fresh machine and an indexed one, which is in the issue as the control.

### DIFF
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -1019,7 +1019,7 @@ func doctorTOMLWired(path string) bool {
 // index this build cannot read without shipping a manifest writer.
 var indexFormatDirection = index.FormatDirection
 
-func doctorIndex(w io.Writer, idx doctorComponent, dir string) {
+func doctorIndex(w io.Writer, idx doctorIndexReport, dir string) {
 	fmt.Fprintln(w, "Index:")
 	loc := idx.Path
 	if loc == "" {
--- a/cmd/deja/doctor_building_test.go
+++ b/cmd/deja/doctor_building_test.go
@@ -29,7 +29,7 @@ func TestDoctorSaysWhenTheIndexIsBeingBuilt(t *testing.T) {
 	}
 
 	var out bytes.Buffer
-	doctorIndex(&out, doctorComponent{State: "missing", Path: dir}, dir)
+	doctorIndex(&out, doctorIndexReport{State: "missing", Path: dir}, dir)
 	got := out.String()
 	if !strings.Contains(got, "building now (reading sessions 42%)") {
 		t.Errorf("doctor does not mention the build in flight:\n%s", got)
@@ -43,7 +43,7 @@ func TestDoctorSaysWhenTheIndexIsBeingBuilt(t *testing.T) {
 		t.Fatal(err)
 	}
 	out.Reset()
-	doctorIndex(&out, doctorComponent{State: "missing", Path: dir}, dir)
+	doctorIndex(&out, doctorIndexReport{State: "missing", Path: dir}, dir)
 	if got := out.String(); !strings.Contains(got, "not built (run `deja warmup`)") {
 		t.Errorf("an index that is really missing lost its advice:\n%s", got)
 	}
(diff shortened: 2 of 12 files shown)


vshulcz:
The schema question is the right one to raise. Reading the policy the document sets for itself:

> **Within a `schema_version`**, changes are additive only: new optional fields may appear, but existing field names, types, and meanings stay the same.
> **Bumping `schema_version`** signals a breaking change (field removal, rename, or type change).

A field starting to appear is the additive case, not the breaking one. Nothing is removed, renamed or retyped, and no consumer can have depended on `stale_stores` being *absent* — absence is what the document said would not happen. Someone who coded to the document gets what it promised; someone who coded to the observed output reads it with a default and is unaffected.

The other reading — that presence itself is a contract — would mean the correct fix is to change the document instead, and that lands worse: it would publish "the field is there when non-zero", which is a shape scripts have to special-case for the *healthy* machine. So: no bump, and 2 stays 2.

**The `embed` block has the same defect and I have not touched it.** `model` and `dim` both carry `omitempty` (doctor_report.go:112-113) while the example shows them present. That is the same mismatch in the same document, but it needs an embedding-enabled machine to reproduce properly, and it is a different key — queued for its own iteration with its own before and after rather than folded in here.

**The relative path in `TestDocsNameThePolicyFileDejaWrites`** is how the repo's other doc-reading tests do it (`TestDocCommentsNameWhatTheyDocument` reads `../../cmd/deja/...`), and `go test` sets the working directory to the package directory by definition. Left as is for consistency.

PR #1713, opened 2026-08-24
fix(doctor): have --deep say what it compared instead of promising no memory was lost

### LINKED ISSUE #1712: doctor --deep says "index matches sources" while an altered record makes a session unreachable
Alter one indexed record in place, keeping its length, and `deja doctor --deep` reports the index healthy while the session becomes unreachable by any query.

## Reproduction

Three sessions indexed, then five bytes of one record's text overwritten — same length, so no header or size changes:

```
$ python3 - ~/.cache/deja/index.db/records.bin <<'PY'
import sys
p=sys.argv[1]; b=bytearray(open(p,'rb').read())
i=b.find(b"retry loop keeps firing")
b[i:i+5]=b"XXXXX"
open(p,'wb').write(bytes(b))
PY

$ deja doctor --deep
Deep verification:
  checked  3 source files, 3 indexed sessions, 3 sampled files re-parsed, 10 postings resolved
  status   index matches sources — no memory lost
$ echo $?
0
```

It does not match, and memory is lost:

```
$ grep -o "retry loop keeps firing 0" ~/.claude/projects/p/*0.jsonl
retry loop keeps firing 0                      # the source

$ deja "retry loop keeps firing 0" | grep -c 444444440
0                                              # the session is not found

$ deja XXXXX
deja: no matches in 3 indexed sessions         # nor by what the index now holds
```

The postings still key on `retry`, so the record is reachable by neither the old word nor the new one.

## Why it is missed

`DeepVerify` (internal/index/deepverify.go:90) compares **per-session message counts**:

> Parse drift: deterministically sample in-sync source files …, full-parse them fresh, and compare per-session message counts with the index.

A same-length edit leaves the count identical. The other guards are structural too: the manifest carries `RecordsSize`, not a checksum, which is why truncation *is* caught and this is not.

## What is right today (the control)

`--deep` is good at everything it does check, each with exit 1 and a named drift kind:

```
truncated records.bin   integrity damaged — records or postings are missing
flipped length prefix   drift [torn-log] records.bin unreadable: record length 4294967253 exceeds cap
garbage in a bucket     drift [unreadable-postings] token catalog unreadable: bad bucket magic
healthy index           status index matches sources — no memory lost, exit 0
```

## The defect

Not that counts are what it compares — that is a reasonable, cheap check. It is that the status line claims more than the check can support. "index matches sources — no memory lost" is what a reader takes to the bank when deciding whether to rebuild.

Two ways out, and the second is a design call:

1. Say what was compared: counts and structure, not content.
2. Store a content hash per session (or for `records.bin`) at build time and verify it here. That catches this class outright, and needs care to avoid false alarms on the normalisation the count check already has to model — `stripSelfRecall`, dedup, clipping.

vshulcz:
`--deep` compares per-session message counts and the structure around them. A same-length edit inside a record leaves the count identical, so the check passes — while the session becomes unreachable by any query, because the postings still key on the old word and the record no longer holds it. The line it printed was:

```
  status   index matches sources — no memory lost
```

That is the sentence a reader takes to the bank when deciding whether to rebuild, and this pass cannot support it. It now says what it checked:

```
  status   every session's message count matches its source, and the records and postings read cleanly
```

Nothing about the checking changes. `--deep` is good at what it does check, and the issue quotes all four cases: truncation, a corrupted length prefix, a garbage bucket, and a healthy index.

Fail-before test: `TestDeepStatusSaysWhatItCompared`, with `TestDeepStatusStillLeadsWithFindings` as the control that a report with findings is untouched. The existing `TestDoctorDeepCleanAndDrift` asserted the old sentence and now asserts the new one.

Closes #1712 — the reporting half.

The larger option is in the issue and is yours to call: store a content hash per session at build time and verify it here, which catches this class outright. It needs care against false alarms on the normalisation the count check already models (`stripSelfRecall`, dedup, clipping), so I have not reached for it inside a wording fix.

### DIFF
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -140,7 +140,12 @@ func doctorDeep(w io.Writer, r index.DeepReport) {
 		fmt.Fprintf(w, "  stale    %s changed since last pass — `deja index` will absorb them\n", doctorCount(len(r.Stale), "source"))
 	}
 	if r.Clean() {
-		fmt.Fprintln(w, "  status   index matches sources — no memory lost")
+		// What this pass actually compares is message counts per session, plus
+		// the structure around them: sizes, magic numbers, postings that
+		// resolve. It cannot see a same-length edit inside a record, which
+		// leaves the count identical and the session unreachable — so the line
+		// says what was checked rather than promising nothing was lost (#1712).
+		fmt.Fprintln(w, "  status   every session's message count matches its source, and the records and postings read cleanly")
 		return
 	}
 	for _, f := range r.Findings {
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -140,12 +140,26 @@ func doctorDeep(w io.Writer, r index.DeepReport) {
 		fmt.Fprintf(w, "  stale    %s changed since last pass — `deja index` will absorb them\n", doctorCount(len(r.Stale), "source"))
 	}
 	if r.Clean() {
-		// What this pass actually compares is message counts per session, plus
-		// the structure around them: sizes, magic numbers, postings that
-		// resolve. It cannot see a same-length edit inside a record, which
-		// leaves the count identical and the session unreachable — so the line
-		// says what was checked rather than promising nothing was lost (#1712).
-		fmt.Fprintln(w, "  status   every session's message count matches its source, and the records and postings read cleanly")
+		// What this pass compares is message counts per session, plus the
+		// structure around them: sizes, magic numbers, postings that resolve.
+		// It cannot see a same-length edit inside a record, which leaves the
+		// count identical and the session unreachable — so the line says what
+		// was checked rather than promising nothing was lost (#1712).
+		//
+		// And only what was actually checked: nothing is sampled when every
+		// source is stale, or when the sampled tokens carry no postings, and a
+		// clean report then means "found nothing wrong", not "compared and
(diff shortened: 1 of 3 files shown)


vshulcz:
Both red findings are right, and they are the same defect I was fixing, one level down: the sentence claimed comparisons that may not have happened.

Nothing is sampled when every source is stale — `samplePaths(inSync, …)` gets an empty list and the count loop never runs — and `SampledPostings` stays 0 when every sampled token carries no postings. `Clean()` is `len(Findings) == 0`, so both states report clean. My replacement sentence then asserted a count comparison and a postings read that had not occurred.

The status is now conditional on what the run actually did:

```
  status   every sampled session's message count matches its source, and the sampled postings resolve
  status   the sampled postings resolve; no source was in sync to re-parse, so no message count was compared
  status   every sampled session's message count matches its source; no sampled token carried postings
  status   nothing to compare — no source was in sync to re-parse and no sampled token carried postings
```

It also says "sampled" where it means sampled: `deepParseSample` is 5 files and `deepPostingSample` is 32 tokens, so "the records and postings read cleanly" was overstating that too.

`TestDeepStatusDoesNotClaimAnUnsampledComparison` covers the three zero cases and all three fail against the single unconditional sentence — which answers the 🟡 and ❓ as well: the test now exercises the paths where `Clean()` is true and nothing was compared, rather than only the wording of the happy one.

PR #1745, opened 2026-08-24
fix(index): follow a store directory that is a symlink instead of calling it empty

### LINKED ISSUE #1744: a store directory that is a symlink indexes nothing, and every surface calls it found and empty
Symlinking a store — `~/.claude` into a dotfiles repo, onto an external disk, into a synced folder — takes it out of the index entirely, with no error anywhere.

Repro. One claude store, two paths to it:

```
$ ln -s /tmp/store/real /tmp/store/claude-link

$ DEJA_CLAUDE_ROOT=/tmp/store/claude-link deja index
deja: index is up to date (0 sessions)
$ DEJA_CLAUDE_ROOT=/tmp/store/claude-link deja search quibbleknot
(nothing)

$ DEJA_CLAUDE_ROOT=/tmp/store/real deja index          # the control: same directory, real path
$ DEJA_CLAUDE_ROOT=/tmp/store/real deja search quibbleknot
[claude] proj · Aug 7 · real-a — 1 matches
  quibbleknot lives in the real store
```

What the diagnostics say about the symlinked one:

```
$ deja doctor  | grep claude
  claude       found     /tmp/store/claude-link  (0 files)
$ deja sources | grep claude
claude	/tmp/store/claude-link	sessions=0 messages=0 size=0 B redacted=0
```

"found … (0 files)" is the wrong sentence: the store is there and full, and deja never looked inside it.

Cause: `walkFiles` (`internal/sources/util.go:271`) calls `filepath.WalkDir(root, …)`, which lstats the root — a symlink is not a directory, so the walk visits that one entry and descends nowhere. `cursor.go:85` and `gemini.go:47` walk their own roots the same way.

Not the same thing, and working as intended: a symlink *inside* a store is skipped (`d.Type().IsRegular()`), which is what keeps a FIFO from hanging the build and a link from walking out of the store. That case is at least visible — doctor counts it as "1 not recognised here". A symlink loop inside a store is survived cleanly: the walk finishes and the row still reports the real files.

Expected: a store root that is a symlink is followed, since it is the path the user configured. Everything inside keeps today's rules.

vshulcz:
Closes #1744.

`filepath.WalkDir` lstats its root, so a store kept behind a symlink — dotfiles repo, external disk, synced folder — was visited as one symlink entry and never descended into. deja indexed nothing and said `found … (0 files)`.

Only the root is resolved. A link *inside* a store is still skipped: that is what keeps a FIFO from hanging the build and a link from walking out of the store, and `TestLinksInsideAStoreAreStillSkipped` pins it, loop included.

```
before  deja: index is up to date (0 sessions)   ·  doctor: claude found /…/claude-link (0 files)
after   deja: claude: 1 session, 1 message       ·  doctor: claude found /…/claude-link (1 file, 1 indexed session)
```

`cursor.go` and `gemini.go` walk their own roots and get the same treatment.

Tests: `TestASymlinkedStoreRootIsWalked` (fails before) and `TestLinksInsideAStoreAreStillSkipped`.

### DIFF
--- a/internal/sources/cursor.go
+++ b/internal/sources/cursor.go
@@ -82,7 +82,8 @@ func CursorDBs() []string {
 func CursorTranscripts() []string {
 	root := filepath.Join(CursorCLIRoot(), "projects")
 	var out []string
-	_ = filepath.WalkDir(resolvedRoot(root), func(p string, d os.DirEntry, err error) error {
+	walked, under := walkRoot(root)
+	_ = filepath.WalkDir(walked, func(p string, d os.DirEntry, err error) error {
 		// Regular files only: a FIFO matching the glob would block the
 		// parser's Open forever (same hang walkFiles guards against).
 		if err != nil || !d.Type().IsRegular() {
@@ -95,7 +96,7 @@ func CursorTranscripts() []string {
 			return nil
 		}
 		if cursorPathHasDir(p, "agent-transcripts") {
-			out = append(out, p)
+			out = append(out, under(p))
 		}
 		return nil
 	})
--- a/internal/sources/cursor.go
+++ b/internal/sources/cursor.go
@@ -82,7 +82,7 @@ func CursorDBs() []string {
 func CursorTranscripts() []string {
 	root := filepath.Join(CursorCLIRoot(), "projects")
 	var out []string
-	_ = filepath.WalkDir(root, func(p string, d os.DirEntry, err error) error {
+	_ = filepath.WalkDir(resolvedRoot(root), func(p string, d os.DirEntry, err error) error {
 		// Regular files only: a FIFO matching the glob would block the
 		// parser's Open forever (same hang walkFiles guards against).
 		if err != nil || !d.Type().IsRegular() {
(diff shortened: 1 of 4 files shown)


vshulcz:
Caught in my own follow-up before the reviewer reported: walking the target and reporting **its** paths broke the incremental pass. The kind predicates test `strings.HasPrefix(p, ClaudeRoot())` against the configured root, so resolved paths stopped matching and the second run indexed the store back down to nothing:

```
deja: incremental index changed_files=1 removed_files=0 sessions=0
$ deja search quibbleknot
(nothing)
```

Fixed by walking the target and mapping each path back under the configured root, which is the spelling every downstream consumer keys on. Now:

```
deja: updated 1 file (1 new message)
[claude] proj · Aug 7 · real-a — 3 matches
```

`TestWalkReportsPathsUnderTheConfiguredRoot` pins it.

vshulcz:
Reviewer flagged cursor and gemini as returning resolved paths that their kind predicates would then not match — right, and now fixed the same way: both use the shared `walkRoot` and report paths under the configured root.

While writing the test for them I found my first one was vacuous. Their walk starts at a path *built under* the configured root (`<root>/projects`, `<root>/tmp/<hash>/chats`), so a link higher up was never broken for them — the path resolves on the way in and the walk descends normally. The case that does differ is a link **at** that path: `~/.cursor/projects` pointing elsewhere, a `chats` directory moved off the disk. `TestCursorAndGeminiFollowASymlinkedWalkRoot` is against that, and it fails without the change.

PR #1746, opened 2026-08-24
mcp: answer ping and resources/templates/list

### LINKED ISSUE #1720: the MCP server answers ping with "method not found" — a host that pings sees a dead server
`ping` is part of the MCP spec at the version we claim (`2024-11-05`, `cmd/deja/mcp.go:27`) and both sides must answer it promptly with an empty result. A host that uses it for keepalive is entitled to treat an error as a stale connection and drop the server.

Repro against the real binary over stdio:

```
$ printf '%s\n' \
  '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"probe","version":"1"}}}' \
  '{"jsonrpc":"2.0","id":2,"method":"ping"}' \
  '{"jsonrpc":"2.0","id":3,"method":"resources/templates/list","params":{}}' | deja mcp
initialize                 ok ['capabilities', 'instructions', 'protocolVersion', 'serverInfo']
ping                       ERROR {'code': -32601, 'message': 'method not found'}
resources/templates/list   ERROR {'code': -32601, 'message': 'method not found'}
```

Control that this is not a blanket failure — the same session answers the rest correctly:

```
tools/list                 ok tools=6
tools/call recall          ok content[0]=<deja-recall> Recalled history from prior sessions…
resources/list             ok ['resources']
notifications/initialized  (no response — correct, it is a notification)
prompts/list               ERROR method not found  (correct: we do not declare a prompts capability)
```

Expected: `ping` returns `{"result":{}}`. `resources/templates/list` returns `{"resources_templates":[]}` — we declare a `resources` capability at `cmd/deja/mcp.go:102`, so clients ask for its templates, and an error there is noise at best.

Cause: `handleMCP` (`cmd/deja/mcp.go:96`) has no case for either, so both fall through to the `-32601` default.

MsfPablo:
Closes #1720.

`handleMCP` had no case for either method, so both fell through to the `-32601` default:

```
ping                       ERROR {'code': -32601, 'message': 'method not found'}
resources/templates/list   ERROR {'code': -32601, 'message': 'method not found'}
```

`ping` is in the spec at the version we claim (`2024-11-05`) and both sides must answer it with an empty result — a host using it for keepalive is entitled to read the error as a stale connection and drop the server. We declare a `resources` capability, so clients ask for its templates as a matter of course; deja exposes concrete session resources and no URI templates, and the empty list is the shape that says so.

## One deviation from the issue text

The issue writes the expected payload as `{"resources_templates":[]}`. I used **`resourceTemplates`**, the key the MCP schema defines and what clients actually read — `resources_templates` would parse as a result with no template list at all, leaving the client no better off than with the error. I took that for a slip in the probe script rather than an intended shape; say the word if you meant otherwise.

## Tests

- `TestMCPPingAndResourceTemplates` (new) — drives both methods through `driveMCP`, i.e. `serveMCP` over the real stdio framing, asserting neither carries an `error`, that `ping` returns an empty object, and that `resourceTemplates` is present and empty. Fails on `main` with the exact `-32601` from the issue.
- `TestMCPUndeclaredMethodStillErrors` (new) — the control: `prompts/list` must still return `-32601`, since we declare no prompts capability. Without it, a fix that turned the default branch into a catch-all would pass the test above.

Hermetic via the existing `hermeticEnv(t)`; neither method touches the index, so no store seeding is needed.

## Verification

```text

ok  github.com/vshulcz/deja-vu/cmd/deja  65.377s  coverage: 89.8% of statements
total: (statements) 89.9%
```

## Notes for reviewers

I did not touch the `initialize` capabilities block. Answering `ping` needs no declaration, and `resources/templates/list` is covered by the `resources` capability we already send.

The branch is based on `884ac6b` rather than current `main`: my token lacks the `workflow` scope, so I cannot push a branch that carries the newer `.github/workflows/hol-plugin-scanner.yml`. The change itself touches only `cmd/deja/mcp.go` and `cmd/deja/mcp_contract_test.go`, so it should rebase cleanly — tell me if you'd rather I find another way to get it onto the current tip.

### DIFF
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -142,6 +142,18 @@ func handleMCP(dir string, req rpcRequest) (any, int, string) {
 				"inputSchema": map[string]any{"type": "object", "properties": map[string]any{"text": map[string]any{"type": "string", "description": "A durable fact, decision, or conclusion to remember."}, "project": map[string]any{"type": "string", "description": "Optional project name; defaults to notes."}, "tags": map[string]any{"type": "array", "items": map[string]any{"type": "string"}, "description": "Optional navigation tags, searchable as #tag."}}, "required": []string{"text"}},
 			},
 		}}, 0, ""
+	case "ping":
+		// Part of the spec at the version we claim, and both sides must
+		// answer it with an empty result. A host that pings for keepalive
+		// is entitled to read an error as a stale connection and drop the
+		// server, so this cannot fall through to -32601.
+		return map[string]any{}, 0, ""
+	case "resources/templates/list":
+		// We declare a resources capability, so clients ask for its
+		// templates. deja exposes concrete session resources and no URI
+		// templates, and the empty list is the shape that says so —
+		// an error here reads as a broken capability.
+		return map[string]any{"resourceTemplates": []map[string]any{}}, 0, ""
 	case "resources/list":
 		return mcpResourcesList(dir)
 	case "resources/read":
--- a/cmd/deja/mcp_contract_test.go
+++ b/cmd/deja/mcp_contract_test.go
@@ -314,3 +314,65 @@ func TestRecallReturnsTheDecisionNotOnlyTheQuestion(t *testing.T) {
 		t.Fatalf("recall dropped the question:\n%s", got)
 	}
 }
+
+// TestMCPPingAndResourceTemplates covers #1720: neither `ping` nor
+// `resources/templates/list` had a case in handleMCP, so both fell through
+// to -32601. A host that pings for keepalive reads that error as a stale
+// connection and drops the server, and we declare a resources capability,
+// so clients ask for its templates as a matter of course.
+func TestMCPPingAndResourceTemplates(t *testing.T) {
+	hermeticEnv(t)
+
+	resp := driveMCP(t,
+		`{"jsonrpc":"2.0","id":1,"method":"ping"}`,
+		`{"jsonrpc":"2.0","id":2,"method":"resources/templates/list","params":{}}`,
+	)
+	if len(resp) != 2 {
+		t.Fatalf("got %d responses, want 2: %#v", len(resp), resp)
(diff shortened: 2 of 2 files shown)


vshulcz:
Read the whole diff and ran it. Reproduced the failure on `main` — `response 1 carried an error: map[code:-32601 message:method not found]` — and confirmed both methods answer with your change.

**You were right to overrule the issue.** `resourceTemplates` is the key in the MCP schema; `resources_templates` would have parsed as a result with no template list, so a client would have been no better off than with the error, only quieter about it. Changing what the issue asked for and saying why is the correct move, not a liberty.

Checked the two things a reader would want checked and could not take on trust:

- We do declare the capability — `mcp.go:102` sends `"resources": {}` in `initialize` — so clients are entitled to ask for its templates, and an error there really does read as a broken capability rather than an unsupported one.
- `mcpProtocolVersion` is `2024-11-05`, the version whose spec makes `ping` mandatory on both sides. The claim in your description holds at the version we actually advertise, which is the part that would have been easy to get wrong.

`TestMCPUndeclaredMethodStillErrors` is the part I would have asked for if it were missing. A fix in this shape is one careless edit away from becoming a catch-all default, and that control is what makes the next person's mistake visible.

**The branch base is mine to fix, not yours.** You are on `884ac6b` because your token has no `workflow` scope and current `main` carries `.github/workflows/hol-plugin-scanner.yml`. That is a permissions wall on our side of the fence, not something for you to work around — GitHub reports the merge as clean, so I will take it from here. If it ever does conflict, I will rebase it myself rather than send it back.

vshulcz:
Thank you — the part worth naming is that you checked the key against the schema instead of against the issue, and then said so. A silent `resources_templates` would have shipped a result no client reads and closed the issue anyway.

Merging.

PR #1748, opened 2026-08-24
warmup: name the phase before it is written

### LINKED ISSUE #1731: during a rebuild an agent is told "deja is indexing this machine's history ()" — the phase renders as an empty parenthesis
`buildingNowForAgent` interpolates `warmupStatus.progress()` (`cmd/deja/mcp.go:1130`), and `progress()` returns `s.Phase` unguarded (`cmd/deja/warmup_status.go:126`). A status file that exists but has not had its phase written yet renders as an empty parenthetical, so the sentence a model is handed is:

```
deja is indexing this machine's history (). Recall comes online in a few seconds; ask again then.
```

Repro — three MCP clients on one index at once (two recalling, one calling `remember`, which triggers the refresh). Of 22 recall answers:

```
   11 x "deja is indexing this machine's history ()."
   11 x "deja is indexing this machine's history (finding transcripts…)."
```

The second half is the control: the same code path, the same run, renders correctly as soon as the phase is on disk — so the empty form is a window, not a missing feature.

The human-facing sentence already handles it: `warmupStatus.line()` (`cmd/deja/warmup_status.go:149`) substitutes `starting` when `Phase` is empty. Only the agent-facing one is missing the guard.

Expected: the same substitution, so the sentence reads `(starting)` — or no parenthetical at all — but never an empty one.

Impact: cosmetic in the sense that no answer is wrong, and not cosmetic in that it is the first thing an agent reads about deja on a cold machine, in a sentence deja wrote about itself.

Found while checking two clients on one index (BUG-HUNT 77). Everything else there held: 20 concurrent `remember` writes all landed with no torn lines, the usage log stayed valid JSONL, and doctor reported the index intact.

MsfPablo:
Closes #1731.

A status file can exist with its phase not yet written. `progress()` returned `s.Phase` unguarded, and every caller wraps that fragment in a parenthetical, so the sentence handed to a model was:

```
deja is indexing this machine's history (). Recall comes online in a few seconds; ask again then.
```

`line()` already substituted `starting` for an empty phase. This moves that substitution into a `phaseName()` helper and routes both fragments through it, so the guard cannot be present on one sentence and missing from the other again — which is exactly how this bug arose.

## Two things wider than the issue text

**It is not only the agent sentence.** `progress()` has five callers, all of which wrap it: `buildingNowForAgent` (`mcp.go:1022`), the statusline (`statusline.go:48`), the still-building recall notice (`warmup_status.go:217`), the MCP instructions block, and `doctor`'s `status building now (%s)` row. Guarding at the one call site the issue names would have left the other four rendering `()`. Fixing `progress()` itself fixes all five, and is why the diff is smaller than the blast radius.

**There is a second shape.** The issue reproduces the empty phase with no total. When the total *is* known and the phase is not yet written, the old code returned `" 25%"`, rendering as `"( 25%)"` — a leading space inside the parenthesis. Same root cause, same fix; the test covers both.

I kept `starting`, the word `line()` already used, rather than dropping the parenthetical — it matches the human-facing sentence and reads as a phase name next to `finding transcripts…`.

## Tests

- `TestWarmupProgressNamesAnUnwrittenPhase` (new, table-driven) — all four combinations of phase written/unwritten × total known/unknown. Two subtests fail on `main`: `progress() = "", want "starting"` and `progress() = " 25%", want "starting 25%"`.
- `TestWarmupSentencesNeverShowAnEmptyParenthetical` (new) — asserts the property on the rendered output of both `progress()` and `line()` rather than on the fragment, so a future caller cannot reintroduce `()` by formatting around the guard.
- `TestWarmupProgressFragment` and `TestWarmupStatusLineReadsLikeASentence` — unchanged, both still pass.

Pure value formatting, no I/O, so no environment setup is needed.

## Verification

```text

ok  github.com/vshulcz/deja-vu/cmd/deja  91.708s  coverage: 89.8% of statements
total: (statements) 89.9%
```

## Note for reviewers

Based on `884ac6b` rather than current `main` for the same reason as #1746: my token lacks the `workflow` scope and cannot push a branch carrying the newer `.github/workflows/hol-plugin-scanner.yml`. This touches only `cmd/deja/warmup_status.go` and its test, so it should rebase cleanly.

### DIFF
--- a/cmd/deja/warmup_status.go
+++ b/cmd/deja/warmup_status.go
@@ -121,17 +121,29 @@ func readWarmupStatus(dir string) *warmupStatus {
 	return &st
 }
 
+// phaseName is the phase to show. A status file can exist with its phase not
+// yet written, and every caller wraps this fragment in a parenthetical or a
+// sentence, so an empty phase renders as "()" — a sentence deja wrote about
+// itself, with nothing in it. Both fragments substitute the same word rather
+// than each guarding separately, so they cannot drift apart again.
+func (s *warmupStatus) phaseName() string {
+	if s.Phase == "" {
+		return "starting"
+	}
+	return s.Phase
+}
+
 // progress is the bare "phase 42%" fragment, for callers that supply their
 // own sentence.
 func (s *warmupStatus) progress() string {
 	if s.Total <= 0 {
-		return s.Phase
+		return s.phaseName()
 	}
 	p := 100 * s.Done / s.Total
 	if p > 100 {
 		p = 100
 	}
-	return fmt.Sprintf("%s %d%%", s.Phase, p)
+	return fmt.Sprintf("%s %d%%", s.phaseName(), p)
 }
 
 // line is the one-sentence status a host can show its user. It names what is
@@ -146,11 +158,7 @@ func (s *warmupStatus) line() string {
 		}
 		pct = fmt.Sprintf(" %d%%", p)
 	}
-	phase := s.Phase
-	if phase == "" {
-		phase = "starting"
-	}
-	return fmt.Sprintf("deja: indexing your history (%s%s) — recall comes online in a few seconds", phase, pct)
+	return fmt.Sprintf("deja: indexing your history (%s%s) — recall comes online in a few seconds", s.phaseName(), pct)
 }
 
 // publishBuildProgress installs the file sink for the work that happens before
--- a/cmd/deja/warmup_status_test.go
+++ b/cmd/deja/warmup_status_test.go
@@ -249,3 +249,40 @@ func TestOpencodePluginIndexesBeforeCompaction(t *testing.T) {
 		t.Fatalf("compaction hook does not index:\n%s", src)
 	}
 }
+
+// TestWarmupProgressNamesAnUnwrittenPhase covers #1731: a status file that
+// exists but has not had its phase written yet left progress() returning "",
+// and every caller wraps that fragment in a parenthetical — so the sentence
+// handed to an agent was "deja is indexing this machine's history ()."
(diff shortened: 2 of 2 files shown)


vshulcz:
Read the whole diff and ran it. Both new subtests fail on `main` exactly as you describe — `progress() = "", want "starting"` and the ` 25%` shape — and pass with the change.

**The blast-radius claim checks out, and it is the reason to merge this rather than the one-line version.** I traced `progress()` myself rather than taking the count: `doctor.go:1039`, `mcp_instructions.go:19`, `mcp.go:1022`, `statusline.go:48` and `warmup_status.go:225`. Every one of them wraps the fragment in a parenthetical or a sentence, so guarding at the call site the issue named would have left four screens still rendering `()` — including `deja doctor`, which is where someone looks precisely when something seems wrong.

The second shape is the better catch. `" 25%"` inside a parenthetical is the kind of thing that survives review because it looks like a spacing nit rather than the same bug with a total attached.

`phaseName()` on the receiver is the right shape: the two fragments cannot drift apart again, which is exactly how the bug arose — `line()` had the guard, `progress()` did not.

One note, not a blocker: `TestWarmupSentencesNeverShowAnEmptyParenthetical` asserts on rendered output rather than on the fragment, so it survives a future caller reintroducing `()` by formatting around the guard. That is the assertion I would have asked for, and it is already here.

vshulcz:
Thank you — the thing I want to name is that you followed `progress()` to its five callers before choosing where to fix it. The issue pointed at one sentence; fixing that one would have left `deja doctor` printing `()` at exactly the moment a user is trying to work out whether deja is broken.

Merging.

PR #1749, opened 2026-08-24
fix(doctor): name the transcript it may not open instead of blaming the harness format

### LINKED ISSUE #1747: a session file the user cannot read is reported as a format change, and doctor never names it
`chmod 000` on one transcript. Indexing does the right thing — the readable session is indexed and the run says one path was skipped — and then points at a `doctor` that says something else entirely:

```
$ deja index
deja: claude: 1 path could not be read — `deja doctor` names it

$ deja doctor | grep -E 'claude|warning'
  claude       unreadable /…/claude  (2 files, 1 indexed session)
  warning      claude store cannot be read — its format may have changed; please report it

$ deja doctor | grep -c locked.jsonl
0
```

Two things are wrong. The advice is for a different fault — nothing changed format, a file is mode 000 — and it asks the user to open a report about their own permissions. And `deja index` promises doctor names the path; it names none, in the text form or the JSON one (`{"name":"claude","state":"unreadable","paths":["/…/claude"],"files":2,"indexed_sessions":1}`).

deja already has the right answer for this: the `denied` state, whose warning reads `permission denied on <path>; check its permissions (on macOS, also Full Disk Access for your terminal)` (`cmd/deja/doctor.go:350`), and which distinguishes a wholly unreadable store from a partly readable one (#802, #816). It is only reached for a directory. A *file* that cannot be opened falls into `unreadable` at `cmd/deja/doctor_report.go:396`, which is the format-drift state.

Control from the same run: with the file readable, the store reports `ok` and both sessions are indexed; making the *directory* unreadable does produce the `denied` warning with its path. So it is the file case that lands in the wrong bucket.

Expected: a file deja is not allowed to open is `denied`, named, with the permission advice — and `Partial` set, since the rest of the store did index.

vshulcz:
Closes #1747.

A transcript at mode 000 landed in the `unreadable` state, whose warning is for a harness that changed its format and asks the user to report it. deja already has the right state for this — `denied`, which names the path and gives the permission advice — but only a *directory* reached it.

```
before  claude       unreadable /…/claude  (2 files, 1 indexed session)
        warning      claude store cannot be read — its format may have changed; please report it
        (and `deja index` had just said "`deja doctor` names it" — it named nothing)

after   claude       denied    /…/claude  (2 files, partly unreadable, 1 indexed session)
        warning      claude store is only partly readable — some sessions are missing from recall — permission denied on /…/proj/locked.jsonl; check its permissions (on macOS, also Full Disk Access for your terminal)
```

`Partial` is set when other files did index, which is what makes the sentence say sessions are missing rather than that the store is gone. The JSON form carries `denied` and `partial` too.

Test: `TestAnUnreadableFileIsDeniedAndNamed` — skips under root and on a filesystem that ignores the mode.

### DIFF
--- a/cmd/deja/doctor_denied_file_test.go
+++ b/cmd/deja/doctor_denied_file_test.go
@@ -58,3 +58,51 @@ func TestAnUnreadableFileIsDeniedAndNamed(t *testing.T) {
 	}
 	_ = tmp
 }
+
+// "Partly readable" has to mean something read. A store whose files are all
+// locked is not partly anything.
+func TestAWhollyLockedStoreIsNotCalledPartlyReadable(t *testing.T) {
+	if os.Getuid() == 0 {
+		t.Skip("root reads everything")
+	}
+	hermeticEnv(t)
+	proj := filepath.Join(os.Getenv("DEJA_CLAUDE_ROOT"), "-proj")
+	if err := os.MkdirAll(proj, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	var locked []string
+	for _, name := range []string{"a.jsonl", "b.jsonl"} {
+		p := filepath.Join(proj, name)
+		body := `{"type":"user","sessionId":"` + name + `","cwd":"/w","timestamp":"2026-08-08T01:00:00Z","message":{"role":"user","content":"flumberjack"}}` + "\n"
+		if err := os.WriteFile(p, []byte(body), 0o644); err != nil {
+			t.Fatal(err)
+		}
+		if err := os.Chmod(p, 0o000); err != nil {
+			t.Skipf("cannot lock a file here: %v", err)
+		}
+		locked = append(locked, p)
+	}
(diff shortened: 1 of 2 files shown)


vshulcz:
Followed my own check on the `Partial` flag before the review landed: `len(check.files) > 1` was wrong for a store whose files are *all* locked — two locked files reported "partly unreadable — some sessions are missing from recall", when nothing is readable at all.

It now probes: another file that opens is what makes a store partly readable, bounded to 64 attempts so a large store does not turn doctor into an audit.

```
one file, locked          claude denied /…/claude2  (1 file, cannot be read)
two files, both locked    claude denied /…/claude3  (2 files, cannot be read)
one locked, one readable  claude denied /…/claude   (2 files, partly unreadable, 1 indexed session)
```

`TestAWhollyLockedStoreIsNotCalledPartlyReadable` pins it.

PR #1750, opened 2026-08-24
mcp: refuse a session uri that carries no id

### LINKED ISSUE #1728: resources/read with an empty session id serves an arbitrary session
`deja://session/` and `deja://session/claude:` — a URI with no id at all — both return a full session digest. `FindByPrefix` matches on `strings.HasPrefix(meta.ID, p)` (`internal/index/retrieval.go:1409`), and every id has the empty string as a prefix, so the first match wins and the agent is handed a session it never asked for. The reply even echoes the requested URI back, so nothing says which session it actually is.

Repro over stdio, against an index of 12 sessions:

```
  real uri           ok bytes=5977 mime=text/markdown uri='deja://session/claude:sess-11'
  empty ref          ok bytes=5977 mime=text/markdown uri='deja://session/'
  prefix only        ok bytes=5977 mime=text/markdown uri='deja://session/claude:'
```

Control — a wrong id in the same session is refused, so the surface is not simply permissive:

```
  unknown id         ERROR -32602 no session matches "nosuchid"
  wrong scheme       ERROR -32602 unknown resource uri
  traversal          ERROR -32602 no session matches "../../etc/passwd"
  id with slash      ERROR -32602 no session matches "a/b"
```

Expected: an empty id is refused the same way an unknown one is.

Impact: a whole transcript reaches the model on a malformed URI, and `deja log` records it as a resource served on purpose. `resources/list` only ever emits full URIs, so nothing legitimate depends on the empty form.

MsfPablo:
Closes #1728.

`FindByPrefix` matches on `strings.HasPrefix(meta.ID, p)`, and every id has the empty string as a prefix. So `deja://session/` and `deja://session/claude:` — URIs carrying no id at all — returned a full session digest: a whole transcript the agent never asked for, echoed back under the URI it did send, and recorded in `deja log` as a resource served on purpose.

An empty ref is now refused with the same `-32602 no session matches ""` an unknown id already gets.

## Why the guard is here and not in FindByPrefix

`FindByPrefix` has five other callers (`hook_context.go`, `view.go`, two in `main.go`, and the resources path), and they pass an id they already hold or a CLI argument. Prefix-matching the empty string is only dangerous where the input is attacker-shaped, which is this boundary — the URI arrives from the client. Guarding here fixes the reported hole without changing a shared lookup's semantics for callers that are fine.

If you'd rather `FindByPrefix` refuse an empty prefix outright as a matter of principle, that's a one-liner too and arguably the better invariant — I didn't want to change the shared function's contract uninvited.

## Tests

`TestMCPResourceReadRefusesAnEmptySessionRef` (new), driving real stdio through `driveMCP`:

- `deja://session/` → refused with `-32602`. Fails on `main` (served instead of refused).
- `deja://session/claude:` → refused with `-32602`. Fails on `main`.
- **Control**: a full URI taken from `resources/list` is still served and still contains the session text — so the guard cannot be satisfied by refusing everything.

The existing `TestMCPResourcesAndPagination` read path is unchanged and still passes.

## Verification

```text

ok  github.com/vshulcz/deja-vu/cmd/deja  78.624s  coverage: 89.8% of statements
total: (statements) 89.9%
```

## Note for reviewers

Based on `884ac6b` rather than current `main`, same as #1746 and #1748 — my token lacks the `workflow` scope and cannot push a branch carrying the newer `.github/workflows/hol-plugin-scanner.yml`. This touches only `cmd/deja/mcp_resources.go` and `cmd/deja/mcp_contract_test.go`.

### DIFF
--- a/cmd/deja/mcp_contract_test.go
+++ b/cmd/deja/mcp_contract_test.go
@@ -376,3 +376,52 @@ func TestMCPUndeclaredMethodStillErrors(t *testing.T) {
 		t.Errorf("prompts/list error code = %v, want -32601", e["code"])
 	}
 }
+
+// TestMCPResourceReadRefusesAnEmptySessionRef covers #1728: FindByPrefix
+// matches on strings.HasPrefix, and every id has "" as a prefix, so a URI
+// carrying no id at all returned a full session digest — a whole transcript
+// the agent never asked for, echoed back under the URI it did send.
+func TestMCPResourceReadRefusesAnEmptySessionRef(t *testing.T) {
+	tmp := hermeticEnv(t)
+	claude := filepath.Join(tmp, "claude")
+	t.Setenv("DEJA_CLAUDE_ROOT", claude)
+	seedClaude(t, claude, "app", "sess-alpha", "the frobnicator crash in parser.go", "fixed the frobnicator")
+	seedClaude(t, claude, "app", "sess-beta", "another frobnicator regression today", "frobnicator again")
+
+	refused := []struct {
+		name string
+		uri  string
+	}{
+		{name: "no id at all", uri: "deja://session/"},
+		{name: "harness prefix only", uri: "deja://session/claude:"},
+	}
+	for _, tt := range refused {
+		t.Run(tt.name, func(t *testing.T) {
+			resp := driveMCP(t, `{"jsonrpc":"2.0","id":1,"method":"resources/read","params":{"uri":"`+tt.uri+`"}}`)
+			e, ok := resp[0]["error"].(map[string]any)
+			if !ok {
+				t.Fatalf("%s was served instead of refused: %#v", tt.uri, resp[0])
+			}
+			if code, _ := e["code"].(float64); code != -32602 {
+				t.Errorf("error code = %v, want -32602", e["code"])
+			}
+		})
+	}
+
+	// Control: a real URI from resources/list must still be served, so the
+	// guard cannot be satisfied by refusing everything.
+	t.Run("a full uri is still served", func(t *testing.T) {
+		list := driveMCP(t, `{"jsonrpc":"2.0","id":2,"method":"resources/list","params":{}}`)
+		resources := list[0]["result"].(map[string]any)["resources"].([]any)
+		uri := resources[0].(map[string]any)["uri"].(string)
+
+		read := driveMCP(t, `{"jsonrpc":"2.0","id":3,"method":"resources/read","params":{"uri":"`+uri+`"}}`)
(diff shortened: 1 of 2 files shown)


vshulcz:
Read the whole diff and ran it here. Confirmed both ways: on `main` both subtests fail with the URI *served* rather than refused, and the control — a full URI taken from `resources/list` — still returns the session text with your guard in place, so the fix cannot be satisfied by refusing everything.

**On the question you asked.** I agree with where you put it, and I want to give you the reason rather than just the verdict.

`index.FindByPrefix` has five other callers and I traced them: `hook_context.go:950` and `view.go:159` pass `s.ID` from a session they already loaded, `main.go:3092` passes `m.ID` from a stored mark, and `main.go:1501` is the CLI path where the commands reject an empty argument before it gets there. None of them can be handed an empty string by anyone outside the process. The MCP URI is the only input that arrives from a client, so that is where the boundary is, and a boundary check reads better at the boundary.

That said, your instinct about the invariant is right too, and it goes one layer further than you saw: `internal/search/search.go:1235` has the same `strings.HasPrefix` shape and is the fallback `findByPrefix` uses when the index is unavailable. So there are two functions that treat "" as "match anything". That is a footgun waiting for the next caller, but it is a change to a shared contract with its own blast radius and its own test — not something to smuggle into a security fix. I will do it separately; this PR is the right size.

One note, not a blocker: the error message reuses `no session matches %q`, which renders as `no session matches ""`. That is deliberate on your part and I would keep it — an empty ref genuinely matches no session, and saying anything more specific would tell a client something about our lookup internals that it has no use for.

vshulcz:
Thank you — and specifically for asking about `FindByPrefix` instead of either changing it quietly or ignoring it. That question is what sent me looking, and it turned up a second function with the same shape.

Merging.

vshulcz:
Rebased this onto current main and pushed it to your branch — the conflict was mine, not yours: #1746 landed first and both PRs append to the end of `mcp_contract_test.go`. Resolution was keep-both, your test is unchanged, and the commit is still yours. Suite green, checks clean.

vshulcz:
Done as #1774, and your instinct was better than my first answer to it.

`PrefixMatches` has answered 0 for an empty prefix all along — the comment above it says the count and the resolver have to agree, or someone is told an id matches nothing and then watches it open (#853). So this was that same invariant broken, with the sign flipped, sitting in the file the whole time.

Both copies are guarded now: `internal/search.FindByPrefix` is the fallback the CLI uses when the index is unavailable, so guarding only the index one would have made the behaviour depend on whether a rebuild was running.

PR #1752, opened 2026-08-24
fix(index): tell a run that waited for another build where the index stands

### LINKED ISSUE #1751: an index run that waited for another one ends with no outcome line at all
Two `deja index` runs at once: one builds, the others wait. The waiters print the wait and then nothing — no summary, no session count, exit 0.

```
$ deja index &  deja index &  deja index &
  run 0 rc=0: deja: incremental index changed_files=100 removed_files=0 sessions=100
  run 1 rc=0: deja: another deja is building the index — waiting for it to finish
  run 2 rc=0: deja: another deja is building the index — waiting for it to finish
```

The control is the same command a second later, when nothing is in flight:

```
$ deja index
deja: index is up to date (760 sessions)
```

So a run that waited leaves "waiting for it to finish" as the last thing on screen, and the user is not told that it did finish or what the index now holds.

Cause: the freshness check that prints that line runs *before* the lock (`cmd/deja/main.go:273`). A waiter fails it, enters the build path, blocks inside `index.Ensure`, and by the time it holds the lock the index is current — so `Ensure` returns at its own early return (`internal/index/ingest.go:187`) having printed nothing, and the command has no closing line of its own.

Everything else about concurrent runs is sound and was checked in the same session: four cold runs at once produce one build and three waits, all exit 0 and the index is complete (600 files, 600 indexed sessions); eight runs over an index 100 files behind produce one incremental pass and seven waits, with all 100 late sessions present afterwards.

vshulcz:
Closes #1751.

A second `deja index` fails the pre-lock freshness check, enters the build path, blocks inside `Ensure`, and by the time it holds the lock the index is current — so `Ensure` returns silently and the command had no closing line. "waiting for it to finish" was the last thing on screen.

The build's progress writer is counted now; a build that printed nothing ends with the same line the command prints when nothing was in flight.

```
before  run 1: deja: another deja is building the index — waiting for it to finish
                (end of output)

after   run 1: deja: another deja is building the index — waiting for it to finish
               deja: index is up to date (960 sessions)
        run 2: deja: incremental index changed_files=100 removed_files=0 sessions=100
                (unchanged — a build that reported its work adds nothing)
```

Test: `TestAnIndexRunThatBuiltNothingStillSaysWhereItStands` over `countingWriter` and `indexQuietOutcome`.

### DIFF
--- a/cmd/deja/index_waiter_test.go
+++ b/cmd/deja/index_waiter_test.go
@@ -1,0 +1,32 @@
+package main
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+)
+
+// A run that waits behind another build finishes with the index current and
+// nothing printed by the build itself — so the command owes the closing line
+// the same run would print when nothing was in flight (#1751).
+func TestAnIndexRunThatBuiltNothingStillSaysWhereItStands(t *testing.T) {
+	var out bytes.Buffer
+	w := &countingWriter{w: &out}
+	if _, err := w.Write([]byte("deja: incremental index changed_files=1\n")); err != nil {
+		t.Fatal(err)
+	}
+	if w.n == 0 {
+		t.Error("the writer did not count what the build printed")
+	}
+
+	quiet := &countingWriter{w: &bytes.Buffer{}}
+	if quiet.n != 0 {
+		t.Error("a build that printed nothing counted output")
+	}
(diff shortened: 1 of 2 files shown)


vshulcz:
Reviewer was right that the freshness bool was being discarded: a store that changed again while the other build was finishing would have been called "up to date". It reports what is on disk now — `another deja finished the build (N sessions indexed); newer sessions are not in it yet` when the index is behind — and `TestAnIndexRunThatBuiltNothingStillSaysWhereItStands` covers both.

Issue #696, opened 2026-08-01
sessions stamped in the future are counted as today's work


vshulcz:
A store with four sessions — one today, one yesterday, one dated six days ahead, one dated next year:

```
deja-vu dev · 4 sessions across 1 agent
today      3 sessions
this week  4 sessions · 0 recalls
recent     [claude] proj/p · Sep 5 2027 · work in future_year
           [claude] proj/p · Aug 7 · work in future_week
           [claude] proj/p · today · work in today
```

"today 3" counts one session from today and two that have not happened. The counters test `Updated.After(midnight)` with no upper bound, so anything ahead of now is today, this week, and at the top of "recent" forever.

This is not hypothetical: a machine with a wrong clock, a transcript written in a different timezone, or a harness stamping in UTC while the reader is behind it all produce timestamps in the future. On a 10k-session store the same bug reported "today 2222 sessions".

The first screen is the first thing anyone sees, and its top line is a count that can only be wrong upward.

PR #1754, opened 2026-08-24
feat(recall): say when a session's date is later than this machine's clock

### LINKED ISSUE #1753: a session dated in the future outranks today's work in recall, for as long as the index holds it
A transcript whose timestamp is ahead of the clock — a machine with a bad clock, a harness writing local time as UTC, a file copied from a device set wrong — becomes the newest session deja knows and stays there.

Six sessions, one query:

```
$ deja search blorptastic --all
[claude] proj · Jan 2 2031  · fut-1 — 1 matches
[claude] proj · Jan 1 10000 · max-1 — 1 matches
[claude] proj · Aug 24      · now-1 — 1 matches      ← today's work, third
[claude] proj · Jan 1 1970  · epo-1 — 1 matches
[claude] proj · Jul 20 1969 · pre-1 — 1 matches
[claude] proj · -           · nots-1 — 1 matches
```

Recall hands an agent the same order:

```
1. [claude] proj · fut-1 · updated 2031-01-02 (Jan 2 2031)
2. [claude] proj · max-1 · updated 10000-01-01 (Jan 1 10000)
3. [claude] proj · now-1 · updated 2026-08-24 (today)
```

`deja last` too. Nothing anywhere says the date is impossible, and `doctor` reports the store as plain `found (6 files, 6 indexed sessions)`.

Working correctly, and worth keeping as the control: the epoch, a pre-epoch date and a missing timestamp are all handled — indexed, ranked last, and printed as `Jan 1 1970`, `Jul 20 1969` and `-` rather than as errors.

deja already treats "ahead of the clock" as a signal not to trust, in the one place it was previously bitten: `readWarmupStatus` calls a status stamped in the future stale rather than fresh forever (#889, `cmd/deja/warmup_status.go`). Session recency has no such guard.

Expected: a timestamp implausibly far ahead of now does not count as "newest". What to put in its place is a judgement call — the file's own mtime is the obvious substitute, since the indexer already has it — and so is the tolerance, which has to be generous enough for ordinary timezone and clock jitter rather than fighting it.

Impact: one bad transcript sits at the top of every recall page and every `deja last`, ahead of what the user actually did today, and the "updated 10000-01-01" line goes to the model as fact.

vshulcz:
Closes #1753.

A transcript stamped ahead of the clock leads every recall page as the newest thing deja holds, with `updated 2031-01-02` printed as fact. `deja brief` already counts these (`ahead  N sessions stamped later than this machine's clock`), so deja knew and the page an agent reads did not.

The date is left exactly as the transcript wrote it. What is added is the caveat, in the same place the listing already carries `[earlier attempt …]` and `[this session abandoned one approach partway]`:

```
1. [claude] proj · fut-1 · 2 matches · updated 2031-01-03 (Jan 3 2031)
[stamped later than this machine's clock — its date cannot place it against the others]
3. [claude] proj · now-1 · 1 matches · updated 2026-08-24 (today)
```

Tolerance is 48h, the reading `readWarmupStatus` already takes — zone and clock jitter is not a wrong clock.

My first attempt was to damp the timestamp to the file's mtime at ingest. It worked and it was wrong: it edits what the transcript says and destroys the signal brief reports, and `TestBriefNamesSessionsAheadOfTheClock` and `TestHandoffReceiptOnAFutureSession` both failed — which is what they are for. Backed out; the correction is on the issue.

Test: `TestRecallMarksASessionStampedAheadOfTheClock`, which also pins that only the session ahead gets the marker.

### DIFF
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -704,6 +704,11 @@ func wholeSessionForMCP(dir string, s model.Session) (model.Session, bool, error
 // needs precisely when the answer was too big to fit.
 const recallCountLineReserve = 160
 
+// futureStampTolerance is how far ahead of the clock a transcript may be
+// stamped before the date stops being usable as "when". Ordinary zone and clock
+// jitter lands inside a day; the same reading readWarmupStatus takes.
+const futureStampTolerance = 48 * time.Hour
+
 func recallTextResult(dir, q, harness string, limit, offset, budget int) (string, int, int64, []string, error) {
 	if limit <= 0 {
 		limit = 5
@@ -829,6 +834,15 @@ func recallTextResult(dir, q, harness string, limit, offset, budget int) (string
 		}
 		if h.Superseded != "" {
 			fmt.Fprintf(&hb, "[earlier attempt — a newer session in this project covers the same ground, updated %s]\n", h.Superseded)
+		}
+		// `deja brief` counts these — a wrong clock, a transcript copied from a
+		// machine set wrong, a harness writing local time as UTC — so deja knew
+		// and the page that an agent reads did not. The date stays as the
+		// transcript wrote it; what is added is that it cannot be trusted for
+		// "newest", which is the one thing the top of a recall page implies
+		// (#1753).
+		if h.Session.Updated.After(time.Now().Add(futureStampTolerance)) {
+			fmt.Fprintln(&hb, "[stamped later than this machine's clock — its date cannot place it against the others]")
 		}
 		if h.Tier != search.TierExact {
 			fmt.Fprintf(&hb, "[%s]\n", h.Tier)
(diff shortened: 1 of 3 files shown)


vshulcz:
Second correction, from my own follow-up: the marker carried a 48h tolerance of its own while `brief` counts anything after now. Two surfaces would then disagree about the same session — brief saying one is ahead, recall marking none.

Tried moving brief to the tolerance instead; `TestOverviewExcludesSessionsFromTheFuture` says no, in as many words: "A session three hours from now is still ahead of the clock, even though it falls on today's date" (#696). So the pinned decision stands and the marker matches it. Both now go through `index.StampedAhead`, and `TestBriefAndRecallAgreeOnWhatIsAheadOfTheClock` pins the rule itself.

Issue #967, opened 2026-08-04
unforget reports success for an imported session that cannot come back


vshulcz:
An imported session lives only in the index — the transcript is on the machine it came from. `forget` drops it and the rebuild has nothing to re-read, so it is gone for good. `--unforget` says otherwise:

```
$ deja forget --session imported-c5b8a55a4da2
sessions dropped: 1
$ deja forget --unforget claude:imported-c5b8a55a4da2
restored 1 session and rebuilt the index

$ deja last
[claude · tmp/l23 · 2026-07-11 · loc23] local work …          ← the imported one is not there
$ deja "PEER SECRET"
deja: no matches in 1 indexed session
```

A local session in the same store round-trips correctly, which is the contrast:

```
$ deja forget --session loc23    → sessions dropped: 1
$ deja forget --unforget claude:loc23 → restored 1 session and rebuilt the index
$ deja last → [claude · tmp/l23 · 2026-07-11 · loc23] local work …    ← back
```

Two honest options: `forget` warns that dropping an imported session is permanent because deja holds the only copy, or `--unforget` reports that the tombstone was lifted but the records are gone and names `sync import` as the way back. Today it lifts the tombstone and claims a restore that did not happen.

vshulcz:
Verified fixed on current main. unforget reports 'gone' for an imported session that cannot come back (main.go, restoredSessions).

PR #1756, opened 2026-08-24
fix(forget): say a deleted transcript is gone from this machine, not that it came from another

### LINKED ISSUE #1755: unforget blames a machine you never had: a locally deleted transcript is reported as one that came from another machine
Forget a session, delete its transcript, then lift the tombstone. deja says the records came from another machine and points at `sync import`:

```
$ deja forget --session ts-01
tombstones added: 1
$ rm ~/…/claude/proj/ts-01.jsonl
$ deja index
deja: incremental index changed_files=0 removed_files=2 sessions=0
$ deja forget --unforget claude:ts-01
1 of them came from another machine and deja held the only copy — the tombstone is lifted, but the records are gone; `deja sync import` brings them back
```

That transcript was written on this machine and deleted from it. `deja sync import` will not bring it back, and there is no other machine to import from — the honest sentence is that the transcript is no longer on disk.

The branch is right about *what* happened (nothing was restored) and wrong about *why*: it assumes any lifted-but-absent session is an imported one. That case is real and is what #967 fixed (`cmd/deja/main.go:2438`), but it is not the only one, and the index knows which is which — imported sessions are counted separately by `index.ImportedSessionCounts`, and a local session's `Path` is a file that either exists or does not.

Control from the same run, all correct: the tombstone survives the file's deletion and a full rebuild; `forget --list` still names the vanished session; a file that comes back while its session is still forgotten stays out of the index (`changed_files=1 … sessions=0`, 0 hits); unforgetting the other one restores it and rebuilds.

Expected: two sentences, one for a session deja only ever held in the index ("`deja sync import` brings them back") and one for a transcript that is simply no longer on this machine.

vshulcz:
Closes #1755.

Lifting a tombstone restores nothing in two different situations, and they had one sentence between them. A session deja only ever held in its own index really did come from another machine, and `sync import` brings it back (#967). A transcript this machine wrote and the user then deleted is simply not on disk — and was told to import it from a machine they never had.

```
before  1 of them came from another machine and deja held the only copy … `deja sync import` brings them back
after   claude:ts-09 is no longer on this machine — the tombstone is lifted, but the transcript it named is gone, so there is nothing to index
```

Imported sessions carry the `imported-` id prefix `sync.go` mints, which is the same marker `ImportedSessionCounts` reads. `restoredSessions` now returns which keys did not come back, so both sentences name them instead of counting them; a prefix selector, which finds rows by manifest rather than by key, keeps a count-only line.

Test: `TestUnforgetSaysWhyNothingCameBack` over the imported case, the local case, and both at once.

### DIFF
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -3214,11 +3214,6 @@ func dayWord(n int) string {
 	return fmt.Sprintf("%d days", n)
 }
 
-// unforgetGoneLines says why a lifted tombstone restored nothing. Two causes
-// with two answers: a session deja only ever held in its own index came from
-// another machine and sync import brings it back (#967), while a transcript
-// this machine wrote and then deleted is simply not on disk — telling that
-// reader to import from another machine names one they never had (#1755).
 // pluralIsLifted and pluralTranscript keep the two sentences readable for one
 // key and for several; joinCapped names up to five.
 func pluralIsLifted(n int) string {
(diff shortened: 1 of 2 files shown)


vshulcz:
Reviewer's 🔴 was right — `joinCapped` names several keys and the sentence said "is no longer". Both sentences now read for one key and for many ("claude:ts-01, claude:ts-02 are no longer on this machine — the tombstones are lifted, but the transcripts they named are gone"), pinned in the test rather than left to substring containment.

PR #1759, opened 2026-08-24
fix(index): say when only half a store could be read, in the run and in doctor --json

### LINKED ISSUE #1758: a store whose sqlite half is unreadable reports success: the index run counts what it got, doctor --json calls it ok
Cursor keeps two stores: CLI transcripts as JSONL and the IDE sessions in SQLite. With `sqlite3` off the PATH, the IDE half is unreadable — and the run that indexes it says only what it managed to read.

```
$ deja index                      # sqlite3 present
deja: cursor: 2 sessions, 4 messages

$ PATH= deja index                # same store, no sqlite3
deja: cursor: 1 session, 2 messages
```

Half the store is gone from recall and nothing on that screen says so. When the *whole* store is unreadable the run does say it, which is the control:

```
deja: opencode: skipped — sqlite3 CLI not found
deja: zed: skipped — sqlite3 and zstd CLIs not found
deja: cursor: skipped — sqlite3 CLI not found      # when cursor has only the IDE half
```

Cause: the skip line is printed only for a store that yielded nothing (`internal/index/ingest.go:558`, "if len(r.ss) == 0"), while `sources.SkipReason` would answer for this one too.

`doctor`'s text row is honest — `cursor found … (1 CLI transcript, 1 store IDE 12.0 KB, sqlite3 CLI missing — IDE sessions unavailable, 1 indexed session)`. Its JSON form is not:

```
{"name": "cursor", "state": "ok", "paths": [...], "files": 2, "indexed_sessions": 1}
```

No `needs-sqlite3`, no note at all — the two forms disagreeing about one store, which is what #999 fixed for the states that existed then, and what I fixed for zed's whole-store case in #1738. The state depends on which file is newest: with the CLI transcript newest, doctor inspects that, it parses, and the store is `ok`.

Everything else in this area is right and worth keeping as the control: `opencode`, `zed`, `grok`, `goose` and `hermes` each report correctly with sqlite3 missing, doctor gives `needs-sqlite3` with a warning naming the fix (`install it, then run deja index`), and with sqlite3 present all five index normally.

vshulcz:
Closes #1758.

Cursor keeps CLI transcripts as JSONL and its IDE sessions in SQLite. Without the `sqlite3` CLI the IDE half is unreadable, and the index run reported only what it managed to read — the skip line was printed for a store that yielded nothing at all.

```
before  deja: cursor: 1 session, 2 messages
after   deja: cursor: 1 session, 2 messages — part of this store could not be read: sqlite3 CLI not found
        (with sqlite3: deja: cursor: 2 sessions, 4 messages — unchanged)
        (a wholly unreadable store still reads: deja: cursor: skipped — sqlite3 CLI not found)
```

Second half of the same fault: doctor's text row has always carried "sqlite3 CLI missing — IDE sessions unavailable" in its detail while the JSON form called the store `ok`, because the state comes from whichever file is newest and the CLI transcript parses. Now both say `needs-sqlite3`, with `partial` set:

```
{"name": "cursor", "state": "needs-sqlite3", …, "files": 2, "indexed_sessions": 1, "partial": true}
```

Tests: `TestNarrationNamesAHalfReadStore` and `TestBothFormsAgreeWhenHalfAStoreNeedsSqlite`.

### DIFF
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -358,11 +358,17 @@ func printDoctorStoreWarnings(w io.Writer, stores []doctorStore) {
 				what = "store is only partly readable — some sessions are missing from recall"
 			}
 			fmt.Fprintf(w, "  warning      %s %s — permission denied on %s; check its permissions (on macOS, also Full Disk Access for your terminal)\n", store.Name, what, store.Denied)
-		case "needs-sqlite3":
+		case "needs-sqlite3", "needs-zstd":
 			// Not a format change: the parser could not run at all. Saying so
 			// points at installing one package instead of at a bug report
-			// against the harness (#792).
-			fmt.Fprintf(w, "  warning      %s store needs the sqlite3 CLI — install it, then run `deja index`\n", store.Name)
+			// against the harness (#792). Which package depends on the store —
+			// zed and deepseek need zstd (#1758) — and a partly readable one
+			// says so rather than reading as a store that is entirely gone.
+			what := "store"
+			if store.Partial {
+				what = "store is only partly readable — part of it"
+			}
+			fmt.Fprintf(w, "  warning      %s %s needs %s — install it, then run `deja index`\n", store.Name, what, toolFromSkip(store.Skipped))
 		}
 	}
 }
@@ -490,7 +496,7 @@ func doctorHarnesses(w io.Writer, dir string) {
 			// `doctor --json`, which has said `denied` all along while this
 			// row, the one people read, said `found` (#993).
 			switch inspected[name] {
-			case "denied", "unreadable", "parsed-zero", "needs-sqlite3":
+			case "denied", "unreadable", "parsed-zero", "needs-sqlite3", "needs-zstd":
 				status = inspected[name]
 				if status == "denied" {
 					if detail != "" {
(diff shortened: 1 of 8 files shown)


vshulcz:
Reviewer's 🔴 and a false alarm I hit myself both land on the same place, and both are fixed:

- deepseek was outside the doctor branch (`storeNeedsSQLite3` is false for it), so the run said "part of this store could not be read: zstd CLI not found" while the JSON said `ok` — the disagreement this PR is about. The branch now covers any store with a skip reason, and the state names the tool: `needs-sqlite3` or `needs-zstd`, with the reason carried in a new `skipped` field. `docs/json-output.md` and the contract test list both.
- My own: `SkipReason("deepseek")` answered whenever zstd was missing and any session file existed, so a store of plain `session.jsonl` — which reads perfectly without zstd — was told part of it could not be read. It now requires a `.zstd`/`.zst` file to be there.

```
cursor    deja: cursor: 1 session, 2 messages — part of this store could not be read: sqlite3 CLI not found
          warning  cursor store is only partly readable — part of it needs the sqlite3 CLI — install it, then run `deja index`
zed       deja: zed: 1 session, 2 messages — part of this store could not be read: zstd CLI not found
          warning  zed store needs the zstd CLI — install it, then run `deja index`
deepseek  deja: deepseek: 1 session, 4 messages       (plain jsonl, no zstd needed — silent, correctly)
```

The unused buffer in the test is gone too.

Issue #1307, opened 2026-08-17
an exclusion set today does not cover what is already indexed, and sync export ships it


vshulcz:
`DEJA_EXCLUDE_PROJECTS` and `~/.config/deja/exclude` are described in `docs/SECURITY-MODEL.md` as privacy controls — "project patterns skipped at ingest". The "at ingest" is accurate and is the whole problem: a pattern set today does nothing about what was ingested yesterday, and nothing says so.

Reproduced end to end:

```
DEJA_EXCLUDE_PROJECTS=private deja index      → the private project is still findable
DEJA_EXCLUDE_PROJECTS=private deja sync export --full
                                              → exported 3 records
                                              → imported on the other machine, the private session is searchable there
DEJA_EXCLUDE_PROJECTS=private deja index --rebuild
                                              → gone from search
DEJA_EXCLUDE_PROJECTS=private deja sync export --full
                                              → exported 2 records
```

So the sequence a careful user would actually perform — realise a project should not be shared, set the exclusion, sync — is exactly the one that ships it. `deja index` returns silently, having applied nothing, and the export that follows carries the project to another machine where the receiving side has no exclusion of its own.

`forget` does not have this shape: a forgotten session stays gone across an export (checked in the same run).

Two ways out, and the choice is a product decision rather than a patch:

- **Apply the list where the data leaves**, not only at ingest. `sync export` reading the exclude patterns would close the case that matters most, and would match what "privacy control" leads a reader to expect.
- **Say something.** If exclusion stays ingest-time, then `deja index` noticing that the patterns changed since the last build and printing "run `deja index --rebuild` to apply them to sessions already indexed" turns a silent no-op into an instruction. `doctor` could carry the same line.

Not fixing it here because either option changes user-visible behaviour or wording rather than repairing a defect, and the first one needs a decision about whether an exclusion is retroactive at all.

Found by the night hunt, iteration 21. The measurement that matters: on the receiving machine the excluded project was searchable by its own text, with no exclusion configured there.

PR #1761, opened 2026-08-24
fix(index): re-read the stores that needed a CLI once that CLI is installed

### LINKED ISSUE #1760: install the missing CLI, run `deja index` as told, and the store is still not indexed
doctor's advice for a store that needs an external tool is exact:

```
warning      opencode store needs the sqlite3 CLI — install it, then run `deja index`
warning      zed store needs the zstd CLI — install it, then run `deja index`
```

Doing that leaves the store unindexed. The transcripts have not changed since the last run, so the incremental pass has nothing to revisit and reports success:

```
$ PATH= deja index
deja: opencode: skipped — sqlite3 CLI not found
$ deja index                    # sqlite3 now installed, exactly as advised
deja: index is up to date (0 sessions)
$ deja last | grep -c opencode
0
$ deja index --rebuild
deja: opencode: 1 session, 2 messages    ← only this works
```

Zed is the same with zstd, and it is worse there because the store is half-readable — the plain thread indexed, the compressed one did not, and "up to date (1 session)" is technically true of the index and false about the store:

```
$ PATH=/usr/bin deja index      # sqlite3 only
deja: zed: 1 session, 2 messages — part of this store could not be read: zstd CLI not found
$ deja index                    # zstd now installed
deja: index is up to date (1 session)
$ deja index --rebuild
deja: zed: 2 sessions, 4 messages
```

Expected: either the advice names `deja index --rebuild`, or — better — a store skipped for a missing tool is remembered as unread, so the next run re-reads it once the tool is there. The second is what the user means by "install it, then run deja index", and deja already knows which stores were skipped: it prints the reason on the run and doctor carries it in `skipped`.

Found while checking Zed's store with `zstd` off the PATH (BUG-HUNT 93). The rest of that check is sound: the plain thread indexes complete, the compressed one is skipped rather than half-parsed, the run says so, and doctor reports `needs-zstd` with the path.

vshulcz:
Closes #1760.

doctor says "install it, then run `deja index`". Doing exactly that left the store unindexed: the transcripts had not changed, so the incremental pass had nothing to revisit and the run reported success.

```
before  $ PATH= deja index      → deja: opencode: skipped — sqlite3 CLI not found
        $ deja index            → deja: index is up to date (0 sessions)
        $ deja last | grep -c opencode  → 0        (only --rebuild worked)

after   $ PATH= deja index      → deja: opencode: skipped — sqlite3 CLI not found
        $ deja index            → deja: opencode: 1 session, 2 messages
        $ deja index            → deja: index is up to date (1 session)
        $ PATH= deja index      → deja: opencode: skipped — sqlite3 CLI not found
```

The manifest records which external CLIs the build could use (`tool_fingerprint`, alongside the exclude fingerprint that solves the same shape of problem for #1307). A machine that has gained or lost one is not fresh, and the pass is forced full — an incremental one would walk the same unchanged files and skip the store again.

Additive: an index written by an older deja carries no fingerprint, and an empty one means "not recorded", so nothing is rebuilt for a field that was never written.

Zed is the same story with zstd, and the store is half-readable there: 1 session before, 2 after.

Test: `TestInstallingTheMissingToolMakesTheIndexStale`.

### DIFF
--- a/internal/index/index.go
+++ b/internal/index/index.go
@@ -282,6 +282,14 @@ type Manifest struct {
 	// it decodes with it empty, and an empty one means "not recorded", which
 	// says nothing rather than guessing.
 	ExcludeFingerprint string `json:"exclude_fingerprint,omitempty"`
+	// ToolFingerprint records which external CLIs were available when this
+	// index was built. A store deja could not read for want of sqlite3 or zstd
+	// is not stale by its file state — the transcripts did not change — so
+	// installing the tool and running `deja index`, which is what doctor tells
+	// the reader to do, reported "index is up to date" and left the store out
+	// of recall (#1760). Additive like the field above: empty means "not
+	// recorded", which is what an index from an older deja carries.
+	ToolFingerprint string `json:"tool_fingerprint,omitempty"`
 }
 
 // HarnessIngest is one harness's ingestion health from its last indexing pass.
@@ -311,8 +319,9 @@ type manifestCore struct {
 	RecordsSize      int64
 	BucketFiles      int
 	IngestHealth     map[string]HarnessIngest
-	// ExcludeFingerprint: see Manifest.
+	// ExcludeFingerprint, ToolFingerprint: see Manifest.
 	ExcludeFingerprint string
+	ToolFingerprint    string
 }
 
 type RedactionStats struct {
--- a/internal/index/ingest.go
+++ b/internal/index/ingest.go
@@ -184,6 +184,13 @@ func Ensure(dir string, harness string, force bool, progress io.Writer) error {
 	if !force && err == nil && notesZoneDrifted(m) {
 		force = true
 	}
+	// A store skipped for a missing CLI has unchanged files, so the
+	// incremental pass has nothing to revisit and the store stays out of the
+	// index. Installing the tool is a change to what deja can read, not to the
+	// transcripts, and only a full pass acts on it (#1760).
+	if !force && err == nil && toolsChanged(m) {
+		force = true
+	}
 	if !force && err == nil && manifestFresh(m, want, scope) && recordsIntact(dir, m) {
 		return nil
 	}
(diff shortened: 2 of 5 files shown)


vshulcz:
The PATH question was the right one to ask, and the first version did have the loop: a hook running with a minimal PATH would have flipped the fingerprint and forced a full rebuild, then the terminal would have flipped it back and rebuilt again.

Two changes make it stable:

- Only a tool **gained** counts. Losing one reads nothing new, and what is indexed stays valid — so the minimal-PATH run is not a change at all.
- A build never records less than the index already knew: `mergedToolFingerprint` keeps a tool marked available if it was there now *or* at the last build, so a hook's build cannot erase what a terminal build learned and make the next terminal run look like a gain.

Measured end to end, with a hook-like run in the middle:

```
PATH= deja index   → deja: opencode: skipped — sqlite3 CLI not found
deja index         → deja: opencode: 1 session, 2 messages
PATH= deja index   → deja: index is up to date (1 session)
deja index         → deja: index is up to date (1 session)
deja index         → deja: index is up to date (1 session)
```

`TestAToolLostDoesNotStartARebuildLoop` pins both halves. On the other questions: `SQLite3Available`/`ZstdAvailable` are `exec.LookPath`, no exec of the binary; all three manifest write sites carry the field, and `writeManifestOnly` passes it through with the rest of the core.

vshulcz:
Reviewer's 🔴 confirmed and fixed: `initEmptyIndex` set the exclude fingerprint and not this one, so an index created by `deja sync import` recorded no tools — and a blank fingerprint reads as "written by an older deja", which would have kept a skipped store out forever. `TestAnImportedIndexRecordsTheTools` pins it. The 🔵 on import carrying a stale fingerprint stands as described: the next Ensure notices and rebuilds, which is the intended path.

PR #1764, opened 2026-08-24
fix(index): say a store went away instead of greeting the reader as new

### LINKED ISSUE #1762: a store whose directory is not there drops its sessions from the index, and the run says the machine has no history
Index a store, then take its directory away — an unmounted disk, a synced folder that has not arrived yet, a relative `DEJA_*_ROOT` read from another working directory. An ordinary `deja index` prunes every session it had from that store:

```
$ DEJA_CLAUDE_ROOT=/Volumes/ext/claude deja index
$ deja search pluckwhistle | grep -c '^\[claude\]'
1

# disk unmounted, nothing else changed
$ DEJA_CLAUDE_ROOT=/Volumes/ext/claude deja index
deja: nothing to index yet — no agent history was found on this machine; `deja sources` shows where deja looked
$ deja search pluckwhistle | grep -c '^\[claude\]'
0
```

Two things are wrong. The memory is gone from recall while the transcripts are perfectly safe on a disk that is merely not mounted — and it comes back only because the *files* come back:

```
# plugged back in
$ deja index
deja: incremental index changed_files=1 removed_files=0 sessions=1
$ deja search pluckwhistle | grep -c '^\[claude\]'
1
```

And the sentence is wrong: "no agent history was found on this machine" is what deja says on a fresh install. Here it has an index full of that machine's history and has just deleted it.

deja already draws this distinction elsewhere — doctor separates a store whose disk went away from one that was deleted (#933, the `unplugged` state) — but the index prunes on file absence alone, so a root that is not there reads exactly like a user who deleted their transcripts. Those are opposite intentions: #1755/#1756 confirmed that a *deleted* transcript should leave the index and stay gone.

Expected: a store whose configured root is absent keeps what is already indexed, and the run says the store was not there rather than that the machine has none.

Found while checking `DEJA_*_ROOT` overrides (BUG-HUNT 94). The overrides themselves behave: an absolute root, a bare relative one, `./claude`, a nested relative path and a trailing slash all index the same store, and `~/nope` is not expanded (the shell's job) so it reads as missing.

vshulcz:
Closes #1762.

When a store disappears, the run explains the eviction and then contradicts itself:

```
before  deja: /…/ext4 is gone, and 3 indexed files with it — if that disk is simply not mounted, reconnect it and run `deja index`
        deja: incremental index changed_files=0 removed_files=3 sessions=0
        deja: nothing to index yet — no agent history was found on this machine; `deja sources` shows where deja looked

after   deja: /…/ext4 is gone, and 3 indexed files with it — if that disk is simply not mounted, reconnect it and run `deja index`
        deja: incremental index changed_files=0 removed_files=3 sessions=0
        deja: 3 indexed files went away with the store that held them, and nothing is left to index — no agent history was found on this machine; `deja sources` shows where deja looked
```

"Nothing to index **yet**" is the first-run sentence, and it was the last thing a reader saw after being told what they had just lost. A machine deja really has never seen still gets it unchanged.

The issue as filed claimed the eviction itself was the bug. It is not: `TestUnmountedVolumeKeepsWhatItAlreadyIndexed` pins "a store the user deleted is still a deletion", and keeping the records made it fail. The correction is on the issue.

Test: `TestTheEmptyIndexSentenceKnowsSomethingWasLost`.

### DIFF
--- a/cmd/deja/empty_after_eviction_test.go
+++ b/cmd/deja/empty_after_eviction_test.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/index"
+)
+
+// "nothing to index yet" is the sentence for a machine deja has never had
+// history from. Printed after a store went away it contradicts the line above
+// it, which has just said what was lost: the reader is told they are new here
+// and that N indexed files are gone, in that order (#1762).
+func TestTheEmptyIndexSentenceKnowsSomethingWasLost(t *testing.T) {
+	fresh := emptyIndexReason(index.BuildSummary{Initial: true}, 0)
+	if !strings.Contains(fresh, "nothing to index yet") {
+		t.Errorf("a first run says %q", fresh)
+	}
+
+	lost := emptyIndexReason(index.BuildSummary{}, 3)
+	if strings.Contains(lost, "yet") {
+		t.Errorf("a run that evicted a store still reads as a first run: %q", lost)
+	}
+	if !strings.Contains(lost, "3") {
+		t.Errorf("the sentence does not say how much went away: %q", lost)
+	}
+	if !strings.Contains(lost, "gone") && !strings.Contains(lost, "went away") {
+		t.Errorf("the sentence does not say what happened: %q", lost)
+	}
+
+	// One file reads as one file.
+	if one := emptyIndexReason(index.BuildSummary{}, 1); strings.Contains(one, "1 indexed files") {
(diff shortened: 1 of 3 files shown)


vshulcz:
One correction from my own follow-up before the review lands: the counter was incremented before the keep-back loop, so records kept because their volume is merely unmounted would have counted as gone. Moved after it — an unmounted disk keeps its records and its count stays zero. `TestUnmountedRecordsAreNotCountedAsGone` also pins that the counter clears on read, so a later command in the same process cannot inherit a spent count.

Issue #857, opened 2026-08-01
deja view writes a shareable page without the redaction floor the other outbound paths print


vshulcz:
A store holding one masked secret, through every path that puts deja's content in a file:

```
$ deja share g1
deja: 1 secrets masked in this share. pattern redaction is a floor — review before sending; rotate anything that leaked.
$ deja sync export dir
deja: records were redacted at index time (0 masked). pattern redaction is a floor — review the export before moving it; …
$ deja promote … --to file.md
deja: 1 secret masked in this file. pattern redaction is a floor — review before sending; …           (#848)
$ deja view --no-open
deja: view written to /tmp/g4/idx.view.html
```

The page itself is fine — the token is a `[redacted:…]` marker, because redaction happens at index time — but it is a self-contained HTML file whose whole purpose is to be looked at and passed around, and it is now the only outbound path that says nothing about the floor.

Printing on every run would be noise for local browsing; naming the count when the page actually carries masked content would not.

PR #1770, opened 2026-08-24
fix(view): redact the page it writes instead of trusting what the index kept

### LINKED ISSUE #1768: `deja view` writes secrets to disk that share and sync export strip — and reports them as "masked"
`share` and `sync export` redact on the way out. `view` does not: it writes whatever the index holds into an HTML page on disk, and then counts the `[redacted:…]` markers that were already there and calls them "masked in this page".

Repro. One store, five secret shapes, indexed with redaction off (an index from an older deja, or one built before a redaction pattern was fixed — #1766 was today):

```
$ DEJA_NO_REDACT=1 deja index
deja: claude: 1 session, 5 messages

$ deja share exp-1 | grep -cE '8f14e45…|ghp_…|BEGIN RSA PRIVATE|abcdefghijklmnop1234567890'
0
$ deja sync export ./out
deja: exported 5 records
deja: records were redacted at index time (0 masked)…
$ strings ./out/deja-sync-*.jsonl | grep -cE '8f14e45…|ghp_…|BEGIN RSA PRIVATE|…'
0

$ deja view --no-open
deja: view written to ~/.cache/deja/index.db.view.html
$ grep -cE '8f14e45…|ghp_…|BEGIN RSA PRIVATE|abcdefghijklmnop1234567890' ~/.cache/deja/index.db.view.html
4
```

All four are in the page, one of them inside a `"preview"` field:

```
"preview":"user: vexlorb plain json: {\"api_key\": \"8f14e45fceea167a5a36dedd4bea2543\"}\nuser: vexlorb bearer: Authorization: Bearer abcdefghijklmnop1234567890\nuser: vexlorb github: ghp_…
```

The second half is the reporting. Over an index deja *did* redact, the same command says:

```
deja: 6 secrets masked in this page. pattern redaction is a floor — review before sharing; rotate anything that leaked.
```

Nothing was masked by `view` — `cmd/deja/view.go:112` counts `strings.Count(string(body), redact.Marker)`, the markers ingest wrote. So the page that is safe gets the warning, and the page that carries four secrets in the clear says only "view written to …". The reassurance and the danger are the wrong way round.

Expected: `view` re-redacts like the other two ways out, and the count reports what this command masked.

`deja view` is the local HTML browser of your memory — a file on disk, meant to be opened and, per its own warning, sometimes shared.

vshulcz:
Closes #1768.

`share` and `sync export` redact on the way out; `view` wrote whatever the index held. On an index built by an older deja — or before a pattern was fixed, which #1766 did today — the page carried the secrets in the clear:

```
before  $ deja view --no-open
        deja: view written to ~/.cache/deja/index.db.view.html
        $ grep -cE '8f14e45…|ghp_…|BEGIN RSA PRIVATE|…' index.db.view.html
        4

after   deja: view written to …
        deja: 6 secrets masked in this page. pattern redaction is a floor — review before sharing; rotate anything that leaked.
        $ grep -cE …  →  0
```

Titles, previews, project names, recall digests and terms, and promoted notes all go through `redact.Text` before they reach the page. The count is the markers in the finished page, so it reports what was masked here and what ingest had already masked without claiming which — and a page with none stays quiet, which is what `TestViewNamesTheRedactionFloorWhenThePageHoldsAMaskedSecret` and #857 ask for.

Test: `TestTheViewRedactsWhatTheIndexKept` builds an index with `DEJA_NO_REDACT=1` and asserts the page holds neither the api key nor the token.

### DIFF
--- a/cmd/deja/view.go
+++ b/cmd/deja/view.go
@@ -208,6 +208,12 @@ func writeViewHTML(dir, out string) (string, int, error) {
 	for i := range notes {
 		notes[i].Title = redactMask(notes[i].Title)
 		notes[i].Text = redactMask(notes[i].Text)
+		// A note's project and tags are what the user typed, so they are text
+		// like the rest rather than structure deja minted.
+		notes[i].Project = redactMask(notes[i].Project)
+		for j := range notes[i].Tags {
+			notes[i].Tags[j] = redactMask(notes[i].Tags[j])
+		}
 	}
 	sj, err := json.Marshal(sessions)
 	if err != nil {
--- a/cmd/deja/view.go
+++ b/cmd/deja/view.go
@@ -98,7 +98,7 @@ func runView(dir string, args []string) error {
 	if out == "" {
 		out = dir + ".view.html"
 	}
-	path, err := writeViewHTML(dir, out)
+	path, masked, err := writeViewHTMLCounted(dir, out)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


vshulcz:
Two follow-ups of my own while the review runs. A note's `project` and `tags` are what the user typed rather than structure deja minted, so they go through the same pass now — session ids, harness names and the date range are left alone deliberately, since deja mints those. And the cost, measured on the 12000-session index from BUG-HUNT 78: `deja view` goes from 0.62s to 0.90s, for at most 200 previews and 100 recall digests per page.

vshulcz:
Reviewer's 🟡 and ❓ both land: the count read the JSON strings before rendering while the comment claimed the finished page. It counts `b.String()` now — what a reader will actually find in the file — and the pass-through wrapper is gone, with the test calling `writeViewHTML` directly.

PR #1773, opened 2026-08-24
fix(sync): give ssh a connect timeout so a sleeping machine costs seconds, not minutes

### LINKED ISSUE #1772: `deja sync ssh` to a machine that is asleep hangs for minutes with no output and no timeout
A laptop that is asleep, a VPN that is down, a firewall that drops packets: ssh sits in its connect until the OS gives up. deja passes no timeout and says nothing while it waits.

Measured against 192.0.2.1 (RFC 5737 test net, which blackholes):

```
$ time deja sync ssh 192.0.2.1
   (no output at all)
   446.7s
```

Over seven minutes on one host, and `deja sync` with no argument walks every machine deja knows, so the wait multiplies. The control is a host that fails fast — a name that does not resolve answers in 0.1s with a clear line:

```
$ deja sync ssh nosuchhost.invalid
deja: exported 10 records
deja: ssh nosuchhost.invalid: exit status 255: ssh: Could not resolve hostname nosuchhost.invalid…
```

Cause: `sshRunner` (`cmd/deja/sync_ssh.go:20`) runs `exec.Command("ssh", …)` with no `-o ConnectTimeout=…` and no `-o BatchMode=yes`, so a host that neither answers nor refuses is waited on for the system default, and a host that wants a password can block on a prompt that nothing is reading.

Expected: a connect timeout deja chooses (and passes to `scp` too), and a line saying which host it is waiting for before it starts.

Everything else in the round trip is sound and was checked in the same session: export on A and import on B carried all five sessions with their turns; a second import of the same directory added 0 records and said so; B's own export carried none of A's records, so nothing echoes back; a self-import is recognised ("10 records were already here word for word"); and a directory of damaged batches imports what it can while naming each file it could not read.

vshulcz:
Closes #1772.

A laptop that is asleep neither answers nor refuses, so ssh waits out the system default. Measured against 192.0.2.1 (RFC 5737, blackholed):

```
before  446.7s, no output at all
after    10.5s, "Connection to 192.0.2.1 port 22 timed out"
```

Every `ssh` and `scp` call carries `-o ConnectTimeout=10 -o BatchMode=yes` now. BatchMode is the other half: a host that wants a password was blocking on a prompt nothing is reading.

Test: `TestSSHCarriesAConnectTimeout`. The stubs in the existing ssh tests read the remote command by position, so they read it from the end of the argument list now — the calls kept their shape, which that test also pins.

The rest of the round trip was checked in the same session and is sound: export on A and import on B carry all sessions with their turns, a second import of the same directory adds 0, B's export carries none of A's records so nothing echoes back, a self-import is recognised word for word, and a directory of damaged batches imports what it can while naming each file it could not read.

### DIFF
--- a/cmd/deja/coverage_extra_test.go
+++ b/cmd/deja/coverage_extra_test.go
@@ -450,7 +450,7 @@ func TestSyncSSHErrorBranches(t *testing.T) {
 		t.Fatalf("sshCapture quote err=%v", err)
 	}
 	sshRunner = func(name string, args ...string) (string, error) {
-		if name == "ssh" && args[1] == "mktemp -d" {
+		if name == "ssh" && args[len(args)-1] == "mktemp -d" {
 			return "/tmp/remote", nil
 		}
 		return "boom", os.ErrPermission
@@ -671,7 +671,7 @@ func TestSyncSSHPushMoreErrorBranches(t *testing.T) {
 	old := sshRunner
 	defer func() { sshRunner = old }()
 	sshRunner = func(name string, args ...string) (string, error) {
-		if name == "ssh" && args[1] == "mktemp -d" {
+		if name == "ssh" && args[len(args)-1] == "mktemp -d" {
 			return "", errors.New("mktemp failed")
 		}
 		return "", nil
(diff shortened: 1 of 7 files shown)


vshulcz:
Reviewer's find is gone — the first stub assignment was dead code, overwritten two lines later. Also from my own pass: the cleanup call that removes the remote batch directory was the one site still running bare, so a host that went away between the transfer and the cleanup would have held the sync open; `TestEverySSHCallCarriesTheOptions` reads sync_ssh.go and fails on any `sshRunner("ssh"|"scp"` line without them. The stubs that read the host positionally now skip past the options.

PR #1776, opened 2026-08-24
feat(recall): say when a hit is the other machine's copy of a session this one holds

### LINKED ISSUE #1775: the same session from two machines is shown twice with contradictory conclusions and nothing says they are one session
Sync keeps both copies when a session id exists on two machines, which is right — nothing is overwritten. What is missing is the connection between them.

One session, `dup-2`, worked on twice: five hours ago on this machine, an hour ago on another, and the two disagree. After `sync import`, recall hands an agent both, in recency order, with no hint they are the same session:

```
$ deja search "the limit" --all
[claude] imported:proj · today  · imported-…b1285ecbfe — 2 matches     ("the limit is nine")
[claude] proj          · 1d ago · dup-2               — 2 matches     ("the limit is five")
```

`deja ctx "the limit"` picks the newer one and says nothing about the other. Nothing on any surface — search, recall, `ctx`, doctor — mentions that these are one session's two histories.

deja already knows it. The imported row keeps the id it had on the machine it came from (#1049):

```
claude:imported-d5b1285ecbfe   id=imported-d5b1285ecbfe   orig=dup-2   project=imported:proj
claude:dup-2                   id=dup-2                   orig=        project=proj
```

And the listing already has a vocabulary for exactly this kind of caveat — `[earlier attempt — a newer session in this project covers the same ground, updated …]`, `[this session abandoned one approach partway]`, and since #1754 `[stamped later than this machine's clock …]`.

Expected: when a recall page holds both copies, say so on the older one — the reader is looking at two answers to the same question and needs to know they came from one session on two machines.

Ranking itself is sound and was checked both ways: with the local copy 2 hours old and the imported one 60 days old the local one leads; with the local one 5 hours old and the imported one an hour old the imported one leads. Recency decides, and the imported copy is visibly marked `imported:proj`.

Everything else about the conflict is right: neither copy overwrites the other, both keep their turns, and re-importing changes nothing.

vshulcz:
Closes #1775.

Sync keeps both copies when a session id exists on two machines — neither overwrites the other, which is right — and the imported row keeps the id it had there (#1049). Nothing connected them, so an agent saw one session's two histories as two unrelated sessions, sometimes disagreeing:

```
1. [claude] imported:proj · imported-d5b1285ecbfe · updated 2026-08-25 (today)      "the limit is nine"
2. [claude] proj · dup-2 · updated 2026-08-24 (1d ago)                              "the limit is five"
```

Now each copy names the other, from its own side:

```
1. [claude] imported:proj · imported-d5b1285ecbfe · …
[another machine's copy of claude:dup-2, which this machine has too — the two may not say the same thing]
2. [claude] proj · dup-2 · …
[this machine's copy; claude:imported-d5b1285ecbfe is the same session as it arrived from another machine — …]
```

`index.TwinSessions` does the pairing by harness and origin id: the same id under another harness is another session, and an import whose origin is not on this machine has no counterpart to name. One manifest read per page.

The marker sits with the ones already there — `[earlier attempt …]`, `[this session abandoned one approach partway]`, `[stamped later than this machine's clock …]` — on the agent-facing page, which is the reader that cannot ask a human. `deja search` renders its own listing and is unchanged.

Ranking was checked both ways and is sound: recency decides which copy leads, in both directions.

Tests: `TestTwinsAcrossMachinesFindEachOther` and `TestRecallNamesTheOtherMachinesCopy`.

### DIFF
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -704,6 +704,18 @@ func wholeSessionForMCP(dir string, s model.Session) (model.Session, bool, error
 // needs precisely when the answer was too big to fit.
 const recallCountLineReserve = 160
 
+// twinSessionsFor pairs each session with the same session as it exists on
+// another machine, so a page holding both copies can say so. One manifest read
+// per page; empty when the index cannot be read, which costs a marker rather
+// than an answer.
+func twinSessionsFor(dir string) map[string]string {
+	metas, err := index.AllMeta(dir)
+	if err != nil {
+		return nil
+	}
+	return index.TwinSessions(metas)
+}
+
 func recallTextResult(dir, q, harness string, limit, offset, budget int) (string, int, int64, []string, error) {
 	if limit <= 0 {
 		limit = 5
@@ -765,6 +777,7 @@ func recallTextResult(dir, q, harness string, limit, offset, budget int) (string
 	// attempts and none of the 5 clean ones.
 	total := len(hits)
 	attachLifecycles(dir, hits)
+	twins := twinSessionsFor(dir)
 	demoted := demoteRejected(hits)
 	if offset > 0 {
 		if offset >= total {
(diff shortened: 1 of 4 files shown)


vshulcz:
Both 🔴 are real and are fixed — and the second one is a process note against myself: I answered them in a comment before the edits were actually committed, so the branch still carried the old code when the review read it. They are in now:

- `TwinSessions` returns every counterpart, sorted, so two machines holding one session are both named rather than the second import silently overwriting the first. `TestTwinsAcrossMachinesFindEachOther` covers the three-machine case.
- The copies are told apart by `OrigID` rather than by an `imported-` id prefix — the prefix is the convention sync mints, the field is the fact.

On the 🟡 about `AllMeta` per recall: measured on the 12000-session index from BUG-HUNT 78, recall is 0.60 s with this change (the read goes through `readManifestCached`). The marker is one line per hit and only for a hit that has a twin, which the test now pins by count.

PR #1778, opened 2026-08-24
test(kimi): compare the two manifests whole instead of five keys of them

### LINKED ISSUE #1777: the two Kimi manifests can advertise different things: everything but five keys is unchecked
The Kimi plugin ships as two manifests — `kimi.plugin.json` at the repo root, which a `/plugins install <github url>` reads, and `extensions/kimi/kimi.plugin.json`, which goes into the release zip. They are the same plugin with different path prefixes, and `TestKimiManifestsAgree` compares exactly five keys: `name`, `version`, `description`, `license`, `sessionStart`.

Everything else can drift unnoticed. Changing only the packaged copy:

```
$ python3 - <<'PY'    # keywords + interface.displayName, packaged manifest only
…
$ go test ./cmd/deja/ -run Kimi
ok      github.com/vshulcz/deja-vu/cmd/deja      0.338s
```

`interface` is what Kimi's `/plugins` browser shows — display name, short and long description, developer, website. Two front doors can advertise different things, and the test that exists says nothing.

The control: a change to `mcpServers` or to the hook list *is* caught, but by `TestKimiPluginManifest` pinning one specific hook rather than by any comparison of the two files — so the day a second hook is legitimately added and that expectation is updated, the root manifest can be left behind again.

Expected: compare the two manifests whole, modulo the path prefix that is the only intended difference (`./extensions/kimi/…` against `./…`).

Everything else in the packaging check is sound and was verified in the same session: `go run ./scripts/pinmanifests -check` says "scoop, winget and the codex plugin match the newest release" against the live release API, and all three are at 0.18.0; the Codex plugin's manifest points at files that all exist (skills, `.mcp.json`, hooks, both logos); `server.json` and `packaging/mcpb/manifest.json` carry placeholder versions because the registry and bundle workflows stamp the release version at publish time.

vshulcz:
Closes #1777.

The plugin ships as two manifests — the repo root one a `/plugins install <github url>` reads, and `extensions/kimi/kimi.plugin.json` in the release zip — and `TestKimiManifestsAgree` compared five keys of them. Everything else could drift: `interface` is what Kimi's `/plugins` browser shows, so the two front doors could advertise different things and the suite stayed green.

They are compared whole now, after rewriting the root manifest's `./extensions/kimi/` prefix to `./` — the one intended difference. A failure names the keys rather than printing two manifests.

Before and after, changing only the packaged copy:

```
keywords + "drifted"                 before: ok     after: FAIL
interface.displayName changed        before: ok     after: FAIL
homepage changed                     before: ok     after: FAIL
```

The rest of the packaging check is sound and was verified in the same session: `pinmanifests -check` says scoop, winget and the Codex plugin match the newest release, against the live API, all at 0.18.0; the Codex plugin's manifest points at files that all exist; `server.json` and the mcpb manifest carry placeholder versions because their workflows stamp the release version at publish time.

### DIFF
--- a/cmd/deja/kimi_plugin_test.go
+++ b/cmd/deja/kimi_plugin_test.go
@@ -2,9 +2,11 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 	"testing"
 )
@@ -102,12 +104,12 @@ func TestKimiManifestsAgree(t *testing.T) {
 	if plugin["version"] != kimiPluginVersion {
 		t.Fatalf("kimiPluginVersion is %q, the manifest says %v", kimiPluginVersion, plugin["version"])
 	}
-	for _, key := range []string{"name", "version", "description", "license", "sessionStart"} {
-		a, _ := json.Marshal(plugin[key])
-		b, _ := json.Marshal(root[key])
-		if string(a) != string(b) {
-			t.Fatalf("%s differs between the two manifests: %s vs %s", key, a, b)
-		}
+	// Whole, not key by key: the two differ only in where they point, and a
+	// list of keys to compare is a list of keys someone forgets to extend —
+	// `interface` is what Kimi's /plugins browser shows, and it went unchecked
+	// (#1777).
+	if diff := manifestDiffIgnoringPaths(plugin, root); diff != "" {
+		t.Fatalf("the two manifests disagree beyond their paths: %s", diff)
 	}
 	// The root manifest addresses the same files from one directory up.
 	if root["skills"] != "./extensions/kimi/skills/" || root["commands"] != "./extensions/kimi/commands/" {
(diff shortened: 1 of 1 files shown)


vshulcz:
Checked the two things that could make a whole-manifest comparison lie, before the review lands:

- **Formatting.** `encoding/json` sorts a map's keys and normalises numbers (`{"b":2,"a":1,"n":1.0}` and `{"a":1,"n":1,"b":2}` marshal identically), so the two files can be indented or ordered differently and still compare equal. Noted in the comment.
- **The prefix rewrite.** It is a string replacement, so it is only honest while `./extensions/kimi/` appears in paths and nowhere else — a description mentioning it would be rewritten too and a real difference would pass. Today it appears in exactly the four path fields; the test now fails if it turns up in `description`, `homepage`, `keywords`, `interface` or `author`.

vshulcz:
The 🔴 does not hold, and it is worth settling with the source rather than by assertion. `go doc encoding/json.Marshal` says it outright:

```
Map values encode as JSON objects. The map's key type must either be a
string, an integer type, or implement encoding.TextMarshaler. The map keys
are sorted and used as JSON object keys …
```

Randomised iteration order applies to `range` over a map, not to what `json.Marshal` emits. Measured on the nested shape the two 🟡 are about — an object inside an object inside an object, built from two differently ordered sources:

```
2000 marshals, identical every time: true (2000)
output: {"a":{"b":[3,4],"m":{"c":2,"q":1},"y":2},"k":"v","z":1}
```

So the string comparison and the `json.RawMessage` comparison are both order-robust, `interface` included. The comment stays as written.

PR #1779, opened 2026-08-24
hunt 2026-08-23: 56 iterations of the bug hunt, 72 merged fixes

### LINKED ISSUE #1716: the failure hook goes silent above 1 MiB of tool output, even when the error is the first line
The failure hook recognises a known error and names what fixed it last time — until the tool output crosses 1 MiB, at which point it goes silent even when the error is the very first line.

## Reproduction

An error deja has seen four times, with the command that followed it recorded:

```
$ deja fix "psql: could not connect to server: Connection refused"
psql: could not connect to server: Connection refused · 2026-08-04
  ran next: $ make db-up ENV=staging
```

Feed the same error to the hook as stderr, with padding after it:

```
$ deja hook-tool-after < payload-1000KiB.json
deja: this error came up here before (2026-08-04) — what followed it: make db-up ENV=staging

$ deja hook-tool-after < payload-1100KiB.json
                                     # nothing, exit 0
```

The error is line one in both. The difference is the size of what came after it.

## Cause

`readHookPayload` reads at most 1 MiB (`io.LimitReader(r, 1<<20)`, cmd/deja/hook_context.go:81). Past that the JSON is cut mid-string, the decode fails, and the hook sees an empty input — not a truncated one.

The bound itself is right: a hook must not read unbounded input. What is wrong is what happens at the edge. `hook_tool_after.go:37` says the scan is capped at 8000 bytes because

> An error names itself at the top or the bottom; the middle of a 10 MB build log is not worth the scan inside an action the user is waiting on.

That reasoning describes a 10 MB build log, and a 10 MB build log is exactly what never arrives: it is dropped whole, a megabyte before the scan gets a chance.

## Impact

A failing build or test run is where this hook earns its keep, and a verbose one is the common case — `make`, `cargo`, `pytest -v`, a docker build. Every one of those that exceeds a megabyte of output loses the recall, silently.

## Expected

Either the comment should say that output over 1 MiB is not read at all, or the hook should use what did arrive. The second is small: when the payload does not decode, scan the raw bytes for a friction line rather than discarding them. deja already takes that shape elsewhere — `hook-prompt` decodes rather than unmarshals precisely so that "a host that writes anything after the object keeps its recall".

## Control

Nothing else about the enormous-output cases is wrong: all of them exit 0 in 10–29 ms with nothing on stderr, whether the payload is 1 MB, 5 MB with no newlines at all, or 11.8 MB. The hook never stalls and never crashes; it just stops answering.

vshulcz:
The collector branch for the BUG-HUNT run of 2026-08-23/25. Draft on purpose: it is Vlad's call whether this lands as one squash or gets split, and nothing here has been reviewed by anyone but the hunt's own reviewer subagent.

Every commit arrived the same way: a reproduction and a snapshot before the change, an issue with the measurement, a failing test, a reviewer pass, and a small PR merged into this branch with green CI. The ledger in `deja-vu-loop/BUG-HUNT-LEDGER.md` carries a row per iteration with what was checked and what was found, including the checks that found nothing.

Some of what is in here:

- **Hooks.** The failure hook went silent above 1 MiB of tool output even when the error was its first line (#1716).
- **MCP.** `ping` answered "method not found", which a host may read as a dead server (#1720); `tools/call` errors handed back the Go decoder's text (#1723); a full recall page dropped the line telling the agent how to reach the rest (#1726); `deja://session/` with no id served an arbitrary session (#1728, #1729); every tool went blind for the length of a background refresh (#1733).
- **Index and stores.** doctor never named three stores it reads from (#1738); the aider parser kept bytes that are not UTF-8 (#1740); `recall_context` spent 22 s on one large session to return 8 KB (#1742); a symlinked store indexed nothing while every surface called it found and empty (#1744); a transcript the user cannot read was reported as a format change (#1747); installing a missing CLI and running `deja index` as advised left the store unindexed (#1760).
- **Privacy.** A secret in escaped JSON was not redacted (#1765); `deja view` wrote to disk what `share` and `sync export` strip (#1768).
- **Sync.** `deja sync ssh` waited 446 s on a machine that is asleep (#1772); the same session from two machines was shown twice with nothing connecting them (#1775).

Closes #1716, #1720, #1723, #1726, #1728, #1729, #1731, #1733, #1736, #1738, #1740, #1742, #1744, #1747, #1751, #1753, #1755, #1758, #1760, #1762, #1765, #1768, #1772, #1775, #1777.


vshulcz:
Superseded by #1912, which merged the collector into main with a merge commit — all 129 commits are in history and the 127 issues they fixed are closed. The collector branch stays open for the hunt still running against it.

PR #1781, opened 2026-08-24
fix(sync): stop calling a push with nothing to send an exchange

### LINKED ISSUE #1780: a machine deja never contacted is reported as "last exchange just now"
`deja sync ssh <host>` with nothing new to send never contacts the host — and records the exchange anyway. doctor then contradicts itself in one line:

```
$ deja sync ssh githbu.example        # a typo, on a machine with nothing to push
deja: exported 0 records
deja: nothing new to push

$ deja doctor | grep githbu
  githbu.example last exchange just now, nothing received yet — last attempt failed: ssh githbu.example: exit status 255: ssh: Could not resolve hostname githbu.example…
```

"Last exchange just now" about a machine that does not exist, in the same sentence that says the attempt failed. The first half is the one a reader believes.

The peers file shows where it comes from:

```
{"host": "githbu.example",
 "last_push": "2026-08-24T23:40:35.228755Z",   ← a push that never happened
 "last_pull": "0001-01-01T00:00:00Z",
 "last_error": "ssh githbu.example: … Could not resolve hostname …"}
```

Control — the same host with something to send records no timestamp, only the error, which is right:

```
peer: githbu.example push= 0001-01-01T00:00:00 err= ssh githbu.example: exit status 255: …
```

Cause: `syncSSHPush` returns `nil` when the export produced zero records (`cmd/deja/sync_ssh.go:154`), so `syncOneWay` calls `peers.Record(host, …, nil)` and stamps `LastPush`. `peers.Record` has the rule right in its own comment — "The failure is recorded, the timestamp is not: 'last exchanged tuesday' has to mean memory actually moved" — and a push of zero records over a connection that was never opened did not move memory either.

Follow-on, already in my queue and not part of this: there is no way to remove a peer. A typo'd hostname stays in `peers.json` and every bare `deja sync` retries it, which now costs the 10 s connect timeout from #1772.

vshulcz:
Closes #1780.

`deja sync ssh <host>` with nothing new to send never opens a connection, and recorded the exchange anyway — so doctor said a machine that does not exist was reached a moment ago, in the same line that reported the failure:

```
before  githbu.example last exchange just now, nothing received yet — last attempt failed: … Could not resolve hostname …
after   githbu.example never exchanged
        (and once there is something to send: never exchanged — last attempt failed: … Could not resolve hostname …)
```

`peers.Record` already had the rule in its own comment — "the timestamp is not: 'last exchanged tuesday' has to mean memory actually moved" — and a push of zero records over a connection that was never opened did not move memory. `syncSSHPush` returns `errNothingToSend` now; the host is still remembered as a machine deja knows, without a timestamp and without an error, and the caller still treats it as a quiet success so a machine with no new work is not counted as one that could not be reached.

Test: `TestNothingToSendIsNotAnExchange`, which also pins that a real exchange still stamps and a real failure still records.

### DIFF
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -2778,6 +2778,7 @@ Usage:
   deja sync import <dir>
   deja sync                       (exchange with every machine deja knows, both ways)
   deja sync ssh <host> [--pull] [--both] [--full]
+  deja sync forget <host>
   deja last [n] [--json] [--project name] [--harness name] [--from machine|local] [--since duration] [--role user|assistant|tool|files|command|edit]
   deja sources
   deja completion <bash|zsh|fish>
--- a/cmd/deja/sync.go
+++ b/cmd/deja/sync.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/vshulcz/deja-vu/internal/peers"
 	"io/fs"
 	"os"
 	"path/filepath"
@@ -49,6 +50,21 @@ func runSync(dir string, args []string) error {
 	switch args[0] {
 	case "ssh":
 		return runSyncSSH(dir, args[1:])
+	case "forget":
+		// A machine deja knows is retried by every bare `deja sync`, and a
+		// typo'd hostname is one of those forever — at the cost of the connect
+		// timeout each time. peers.Forget did the work already and nothing
+		// reached it (#1780).
+		host := args[1]
+		found, err := peers.Forget(host)
+		if err != nil {
+			return err
+		}
+		if !found {
+			return fmt.Errorf("deja does not know a machine called %q — `deja doctor` lists the ones it does", host)
+		}
+		fmt.Fprintf(os.Stdout, "deja: %s forgotten — bare `deja sync` will not try it again\n", host)
+		return nil
 	case "export":
 		full := false
 		rest := args[1:]
--- a/cmd/deja/sync_nothing_sent_test.go
+++ b/cmd/deja/sync_nothing_sent_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"errors"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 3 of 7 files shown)


vshulcz:
Two things found while checking my own change, both in now:

**A regression it would have introduced.** `peers.Record` with a zero time takes the success branch, so "nothing to send" would have cleared the error *and* overwritten a real `last_push` with the zero time — a machine synced an hour ago would read as never exchanged. `Record` now leaves an existing row alone when there is nothing to report, and `TestRecordingNothingKeepsWhatWasThere` pins it.

**A capability that was written and unreachable.** `peers.Forget` exists, is tested, and no command called it — so the queue item I opened this iteration from ("no way to remove a peer") was half wrong: the work was done, the door was missing. `deja sync forget <host>` is wired now, named in the usage summary and on the commands page:

```
$ deja sync forget githbu.example
deja: githbu.example forgotten — bare `deja sync` will not try it again
$ deja sync forget nosuch.example
deja: deja does not know a machine called "nosuch.example" — `deja doctor` lists the ones it does
```

Wording of both lines is yours to change; the mechanism is what the issue is about.

vshulcz:
On the 🟡 about clearing `LastError`: keeping it is the intended behaviour, and I think the right one — a push with nothing to send never opens a connection, so it learns nothing about the machine, and clearing the last failure would say the host is fine when nothing was checked. A real exchange is what clears it. The second 🟡 is fair either way, so that sequence is pinned now: `TestNothingToSendLeavesAnEarlierFailureStanding` covers error → nothing-to-send → success.

vshulcz:
CI caught what my machine could not: on Linux `peers.Path()` follows `XDG_CONFIG_HOME`, so setting only `HOME` left every test in the package sharing one real file, and my two new ones then read each other's rows. The package already had the answer — `t.Setenv("DEJA_PEERS_FILE", …)` — and both tests use it now. Reproduced the failure locally with `XDG_CONFIG_HOME=/tmp/fake-xdg go test ./...`, which is green both ways.

PR #1783, opened 2026-08-25
test(index): stand where the swap-window wait is instead of racing a goroutine against it

### LINKED ISSUE #1782: TestDamagedRereadsAPairThatDisagreed races a 5 ms goroutine against a 20 ms sleep, and CI loses sometimes
The test plants a truncated `records.bin`, starts a goroutine that restores the full file after 5 ms, and asserts `Damaged` comes back false — which only holds if the restore lands inside the 20 ms `swapWindowWait` that `Damaged` sleeps before its second read. On a loaded runner the goroutine can be starved past that, and the test fails on a store that is behaving exactly as designed.

It failed once on the ubuntu runner during BUG-HUNT 85, on a branch that touched only `cmd/deja`:

```
--- FAIL: TestDamagedRereadsAPairThatDisagreed (0.23s)
    swap_stat_test.go:86: a store that agrees with its manifest a moment later was called damaged
```

I could not reproduce it by repetition — 40 runs at GOMAXPROCS=1 under 24 CPU hogs stayed green — so here it is by construction instead, moving the restore to 25 ms, which is what starvation does to a 5 ms sleep:

```
$ sed -i '' 's/time.Sleep(5 \* time.Millisecond)/time.Sleep(25 * time.Millisecond)/' internal/index/swap_stat_test.go
$ go test ./internal/index/ -run TestDamagedRereadsAPairThatDisagreed
FAIL
```

Expected: the test synchronises with the re-read instead of racing it, so it proves what it is about — a store that agrees a moment later is not damaged — without depending on which goroutine the scheduler picks.

`TestRealDamageSurvivesTheReread` next to it has no such dependency and is unaffected.

vshulcz:
Closes #1782.

`TestDamagedRereadsAPairThatDisagreed` restored the truncated file from a goroutine sleeping 5 ms and relied on that landing inside the 20 ms `swapWindowWait` before `Damaged` looks again. On a loaded runner the goroutine can be starved past the window, and the test then fails on a store behaving exactly as designed — it did once on ubuntu during BUG-HUNT 85, on a branch that touched only `cmd/deja`.

The pause is a variable now (`waitOutSwapWindow`), so the test restores the file *inside* it and asserts the second read happened at all. Production behaviour is unchanged: the variable holds the same `time.Sleep(swapWindowWait)`.

```
before, with the restore moved to 25 ms (what starvation does to 5 ms):   FAIL
after,  same 25 ms delay inside the wait:                                 ok
```

`TestRealDamageSurvivesTheReread` beside it never depended on the scheduler and is untouched.

### DIFF
--- a/internal/index/manifest.go
+++ b/internal/index/manifest.go
@@ -481,7 +481,7 @@ func Damaged(dir string) bool {
 	if !RebuildInProgress(dir) {
 		return true
 	}
-	time.Sleep(swapWindowWait)
+	waitOutSwapWindow()
 	m2, err := readManifest(dir)
 	if err != nil {
 		// The manifest went away under us, which is the swap itself rather than
--- a/internal/index/swap_stat_test.go
+++ b/internal/index/swap_stat_test.go
@@ -3,6 +3,7 @@ package index
 import (
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


vshulcz:
Checked the two things a package-level seam can break, before the review lands: nothing in `internal/index` calls `t.Parallel`, so no other test runs while the variable is swapped, and the swap is restored by `defer` on every path including a failure. `go test ./internal/index/ -race` is green, and so is the timing-sensitive subset at `-race -count=8`. The retry loop calls the pause once per attempt exactly as it did before — only the call site's name changed.

vshulcz:
Second 🟡 taken: `statIndexFile`'s retry loop still slept directly while `openIndexFile`'s went through the variable — two loops, one window, two spellings. Both go through `waitOutSwapWindow` now, and a test reads `swap_window.go` and fails on any direct `time.Sleep(swapWindowWait)` outside the variable's own definition, so the next loop to grow one is caught rather than quietly racing the scheduler again. On the first 🟡: nothing in the package calls `t.Parallel`, the swap is restored by defer, and `-race` is green — worth remembering if parallel tests ever arrive here.

Issue #1306, opened 2026-08-17
three more MCP handlers block on the index lock and can rebuild inline


vshulcz:
Turned up by the review of #1305, which fixed the same shape in `recall`, `recall_context` and `resources/read`. These are still open:

- `cmd/deja/mcp_resources.go:24` — `resources/list` calls the blocking `index.Ensure`. On a stale store that is not merely a wait: it runs the whole rebuild inside the handler.
- `cmd/deja/mcp.go:279` (`fix`) and `cmd/deja/mcp.go:313` (`how`) — same call, on two read-only tools. Harder to justify than `remember`, which at least just wrote something and wants it searchable: both of these are described to the model as things to call *before* diagnosing an error or inventing a command, so they sit on the latency-critical path by design.
- `blame` reaches `index.EnsureForSearch` through `findBlameHits` (`cmd/deja/main.go:1727`), which also blocks and can rebuild inline.

Second, smaller thing in the same places: `resources/list`, `fix` and `how` skip `buildingNowForAgent`, so on a cold index they answer with a raw errno carrying an internal manifest path instead of "deja is indexing" — the state #972 fixed for the other tools.

Measured context for whether this matters: on a 16000-session store a full rebuild is 2.7s and an incremental one 0.65s, against a 0.31s steady-state recall. So the wait is between two and nine times the whole call, and the inline-rebuild case is worse than the wait.

Not fixed in #1305 because each needs a decision rather than a substitution — whether the tool can serve a stale answer at all, and what it says when it does. `recall` already had that answer (`EnsureForSearchStale` plus an honest note in the payload); these do not.

vshulcz:
Checked the second half — the claim that these answer with a raw errno on a cold index — and it does **not** reproduce, for a reason that is worth writing down because it ties the two halves together.

On a genuinely cold index every tool answers properly:

```
recall  → "No prior deja sessions matched …"
fix     → "No session on this machine ran a command after that error."
how     → "No command on this machine mentions …"
blame   → []
resources/list → 1 resource
```

They build the index synchronously first, which is the blocking behaviour this issue is about, and then answer normally.

The rebuild window does not surface it either. Hammering `fix` and `how` for the whole of a 2.7s rebuild on a 16000-session store got through exactly **one** round, and it answered correctly — because each call waited for the lock rather than observing the directory mid-swap.

[...]

PR #1785, opened 2026-08-25
fix(mcp): let blame answer from the snapshot while a rebuild runs

### LINKED ISSUE #1784: blame is the only read tool that goes silent during a refresh, because its path blocks on the index lock
Since #1734 the reading tools answer from the snapshot while a rebuild is in flight; `blame` still declines, because its path takes the blocking `index.EnsureForSearch` and would otherwise wait out the whole rebuild:

```
# a rebuild holds the index lock
  blame     0.01s  deja is indexing this machine's history. Recall comes online in a few seconds; ask again then.
  recall    0.01s  (index refresh running in the background — the very newest sessions may not appear yet) + results
```

Both tools want the same thing — what is on disk now — and the snapshot is readable throughout a swap, which is why recall reads it without waiting. The blocking call is what forces blame's hand: `findBlameHits` (`cmd/deja/main.go:1959`) calls `EnsureForSearch`, which takes `lockDir`.

Expected: the agent-facing blame reads the way recall does — `EnsureForSearchStale`, answer from the snapshot, and say a refresh is running. `deja blame` on the command line keeps the blocking call: someone typed it and is watching the progress, which is the distinction the code already draws for `#1306`.

Measured cost of not doing it: on a 12000-session store a refresh is seconds, and `blame` is the tool an agent calls *before editing a file* — the moment where "ask again then" means the edit happens without the history.

Follow-on from #1733, where the same reasoning was applied to the other five tools and blame and remember were left as the two that block.

vshulcz:
Closes #1784.

Since #1734 the reading tools answer from the snapshot during a refresh; blame still declined, because its path takes the blocking `EnsureForSearch`. blame is the tool an agent calls *before editing a file*, so "ask again then" meant the edit happened without the history.

```
before  blame   0.01s  deja is indexing this machine's history. Recall comes online in a few seconds; ask again then.
after   blame   0.43s  10 sessions, and the payload opens with
                       {"note":"index refresh running in the background — the very newest sessions may not appear yet"}
```

`findBlameHitsStale` is the agent-facing path — `EnsureForSearchStale`, hand the rebuild to the detached warmup, answer from disk. `deja blame` on the command line keeps the blocking call, which is the distinction #1306 drew: someone typed it and is watching the progress.

The note goes in the payload rather than in prose, because blame answers in JSON — recall's sentence in the shape this tool speaks.

Tests: `TestBlameReadsWithoutWaitingForARebuild` (which also fails if the tool goes back to declining) and `TestBlameSaysWhenItServedASnapshot`.

`remember` stays on the blocking path: it writes first and then indexes, which is a different question.

### DIFF
--- a/cmd/deja/blame_stale_test.go
+++ b/cmd/deja/blame_stale_test.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"encoding/json"
+	"github.com/vshulcz/deja-vu/internal/model"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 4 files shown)


vshulcz:
One of my own before the review lands: the note was inside the byte-budget loop, so on a full page a session could have been dropped to make room for a sentence about the index. The payload is trimmed on the hits alone now and the note added afterwards; `TestTheRefreshNoteDoesNotCostASession` builds a payload well over the budget and asserts the session count is the same with the note and without it. The count `deja log` records was already the session count rather than the row count, so it is unaffected.

vshulcz:
Both taken. The source scan asserted nothing if the blame case vanished — it fails now when the case is missing. And the note can put the payload over the budget: that is deliberate (about a hundred bytes, against dropping a whole session for it), so it is said in the comment and bounded by the test — the overshoot must be no more than the note's own length.

PR #1787, opened 2026-08-25
fix(search): run the co-occurrence rescue when the words never appear together

### LINKED ISSUE #1786: the co-occurrence rescue never runs for the case it was written for: two words that never appear together
The rescue exists for a query whose words are right but not together — "the corpus itself knows that this user's login lives next to jwks and rotation" (`internal/index/cooccur.go:11`). That is the state where the ANDed postings intersect to nothing. The chain does not call it there.

Fixture: 160 filler sessions, 10 that say "keytab renewal failed on the build agent", 12 that say "kerberos keytab rotation on the runner". The map learns `kerberos -> [keytab rotation runner]`, which is exactly right.

```
$ deja search "kerberos renewal"
deja: no matches in 182 indexed sessions — try fewer words or --re (query "kerberos renewal")
deja: on their own: "kerberos" in 12, "renewal" in 10 — no session has them together
```

And the rescue, asked directly, answers:

```
cooccurSearch -> 10 sessions, variants: map[kerberos:[keytab] renewal:[renewal]]
```

Ten sessions found by swapping `kerberos` for its proven neighbour `keytab`, and the search that a user runs returns none of them.

Cause: `searchDetailedOnce` has two zero-result branches. The one for "the postings did not intersect" (`internal/index/retrieval.go:148`) tries stem, fuzzy, error signatures, then relevance — no `cooccurSearch`. The one for "postings intersected but the records scan came back empty" (`:178`) is the only place the rescue is called, and that is a much rarer state than the one the feature describes.

Control from the same store, showing the map itself is fine:

```
tokens with neighbours: 8
  kerberos  -> [keytab rotation runner]
  keytab    -> [kerberos rotation runner agent build failed]
  renewal   -> [agent build failed keytab]
```

Expected: the empty-intersection branch tries the rescue too, in the same order as the other one — after stem and fuzzy, before falling back to relevance.

This also explains why the map's staleness (my queue item `[cooccur-staleness]`) was hard to demonstrate: the rescue rarely runs at all.

vshulcz:
Closes #1786.

The rescue exists for a query whose words are right but not together — "this user's login lives next to jwks and rotation". That is the state where the ANDed postings intersect to nothing, and the chain called the rescue only in the *other* zero-result branch, the one for postings that intersect but scan to nothing. So the case it was written for never reached it.

```
before  $ deja search "kerberos renewal"
        deja: no matches in 182 indexed sessions …
        deja: on their own: "kerberos" in 12, "renewal" in 10 — no session has them together
        (while cooccurSearch, asked directly, answered with 10 sessions)

after   deja: no exact match, trying word forms: kerberos -> keytab
        [claude] proj · 5d ago · y-160 — 6 matches · stemmed (kerberos->keytab)
          keytab renewal failed on the build agent 160
```

The call goes where the others already are — after stem and fuzzy, before the error-signature pass and the relevance fallback — and through `withRelevanceTail` like the sibling branch, so the tier and the narration are the same.

Test: `TestCooccurRescueRunsWhenThePostingsDoNotIntersect` builds a store that teaches the map `kerberos -> keytab`, asserts the map learned it and that the rescue itself finds the sessions, then asserts the search a user runs returns them; it also pins that a query matching directly is unaffected.

This is also why `[cooccur-staleness]` was hard to demonstrate — the map's staleness rarely mattered because the rescue rarely ran.

### DIFF
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -686,7 +686,11 @@ func cmdCtx(dir string, rest []string) error {
 	// Which rung answered, said out loud on stderr as the search screen says
 	// it — stdout stays the context block an agent parses. ctx served an
 	// answer to a word the caller never typed and said nothing about it.
-	if result.Stemmed {
+	if result.Neighbour {
+		printNeighbour(os.Stderr, result.Variants)
+		o.Stemmed = true
+		o.FuzzyVariants = result.Variants
+	} else if result.Stemmed {
 		printStemmed(os.Stderr, result.Variants)
 		o.Stemmed = true
 		o.FuzzyVariants = result.Variants
@@ -962,7 +966,11 @@ func runSearch(dir string, args []string, sourceInstance string) error {
 	ss, policyHidden := policyFilterSessionsCounted(policy.ActivationSearch, result.Sessions)
 	o.PolicyWithheld = policyHidden
 	o.Tier = result.Tier
-	if result.Stemmed {
+	if result.Neighbour {
+		printNeighbour(os.Stderr, result.Variants)
+		o.Stemmed = true
+		o.FuzzyVariants = result.Variants
+	} else if result.Stemmed {
 		printStemmed(os.Stderr, result.Variants)
 		o.Stemmed = true
 		o.FuzzyVariants = result.Variants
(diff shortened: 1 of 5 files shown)


vshulcz:
Three of the review's questions, measured before it landed:

- **Does it steal a pasted error from the error-signature pass?** No — a real Go panic with its goroutine trace answers the same on both binaries (9 matches, same session). The rescue only runs when it finds a neighbour that produces sessions, and a pasted stack trace's tokens have none.
- **What does it cost when a query genuinely has no answer?** 0.07 s on the 12000-session index from BUG-HUNT 78, the same as before to two decimals.
- **Is the fixture in the branch this is about?** It is now asserted rather than assumed: the test checks both tokens have postings before searching, so an empty-postings query cannot pass for the empty-intersection case.

vshulcz:
The 🔵 turned out to matter more than a nit, because this fix is what makes that path reachable: a co-occurrence result set `Stemmed: true`, so the search narrated it as `no exact match, trying word forms: kerberos -> keytab`. It is not a word form — the corpus says those two words keep company — and reading it as a spelling correction invites the user to think they typo'd. The result carries `Neighbour` now and says:

```
deja: no session has those words together, so deja tried one this corpus keeps beside "kerberos": keytab
```

The ❓ is also addressed: the test now proves the precondition rather than assuming it — it searches each token separately and fails if any session holds both, so the fixture cannot drift into the other branch.

Wording of the new sentence is yours to change.

PR #1789, opened 2026-08-25
fix(show): tell the two halves of the repeated hour apart

### LINKED ISSUE #1788: during the DST fall-back `deja show` prints two different turns as the same minute
An hour repeats when the clocks go back, and `deja show` renders both halves the same, so an hour of conversation reads as a duplicated message:

```
$ TZ=America/New_York deja show dst
# claude · proj · dst
2026-11-01 01:30 user:      ← 05:30Z, EDT
2026-11-01 01:30 assistant: ← 06:30Z, EST, an hour later
2026-11-01 02:30 user:
```

The control is the same session in UTC, where the three turns are an hour apart as they actually were:

```
$ TZ=UTC deja show dst
2026-11-01 05:30 user:
2026-11-01 06:30 assistant:
2026-11-01 07:30 user:
```

Both stamps are correct — both instants really are 01:30 local — and that is exactly why the reader cannot tell them apart. The order is right; what is missing is any sign that the second one is an hour later rather than a copy of the first.

Cause: `internal/search/search.go:1274` formats every stamp as `2006-01-02 15:04`, which carries no offset.

Expected: when a session repeats a local stamp, the repeated ones say which side of the change they are on — the offset (`01:30 -04:00` / `01:30 -05:00`) is enough. An ordinary session, where no stamp repeats, keeps the format it has: this happens twice a year and should not cost every other transcript a wider column.

Found in BUG-HUNT 88 and queued as `[show-repeated-hour]`.

vshulcz:
Closes #1788.

When the clocks go back an hour happens twice, and both halves rendered as the same minute — so an hour of conversation read as a duplicated message:

```
before  2026-11-01 01:30 user:        ← 05:30Z, EDT
        2026-11-01 01:30 assistant:   ← 06:30Z, EST, an hour later
        2026-11-01 02:30 user:

after   2026-11-01 01:30 -04:00 user:
        2026-11-01 01:30 -05:00 assistant:
        2026-11-01 02:30 user:
```

Only the repeated minutes carry the offset: this happens twice a year and must not cost every other transcript a wider column. The same session in UTC, where nothing repeats, is unchanged.

Test: `TestTheRepeatedHourSaysWhichSideItIsOn`, which also pins that an ordinary session gains nothing. It skips where tzdata is unavailable.

From BUG-HUNT 88, queued as `[show-repeated-hour]`.

### DIFF
--- a/internal/search/repeated_hour_test.go
+++ b/internal/search/repeated_hour_test.go
@@ -51,3 +51,28 @@ func TestTheRepeatedHourSaysWhichSideItIsOn(t *testing.T) {
 		t.Errorf("an ordinary session gained an offset:\n%s", pb.String())
 	}
 }
+
+// Two turns at the same instant are not a repeated hour — a burst of messages
+// in one minute is ordinary — and a minute that comes back three times still
+// says which is which.
+func TestRepeatedStampDetection(t *testing.T) {
+	base, err := time.Parse(time.RFC3339, "2026-11-01T05:30:00Z")
+	if err != nil {
+		t.Fatal(err)
+	}
+	same := repeatedStamps([]model.Message{{Time: base}, {Time: base}, {Time: base}})
+	if len(same) != 0 {
+		t.Errorf("messages at one instant were called a repeated hour: %v", same)
+	}
+	// The same wall-clock minute from three different instants: one entry, and
+	// every one of them gets the offset when it prints.
+	thrice := repeatedStamps([]model.Message{
+		{Time: base}, {Time: base.Add(time.Hour)}, {Time: base.Add(2 * time.Hour)},
+	})
+	if len(thrice) != 0 {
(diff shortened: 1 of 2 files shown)


vshulcz:
Measured the questions this raises before the review lands:

- **Cost.** `deja show` on the 30001-message session from BUG-HUNT 82: 1.34 s with the change, 1.35 s without.
- **Odd offsets.** The format carries them exactly — `+05:30` Kolkata, `+05:45` Kathmandu, `+13:45` Chatham, `+00:00` UTC.
- **Other renderers.** `PrintSession` is the only place in this package printing per-message stamps; `deja ctx` prints none, and `deja log`'s stamps are events rather than turns, where a repeated minute is two events and reads correctly.
- **Detection.** Pinned now: messages at one instant are not a repeated hour, a session with no stamps produces nothing, and a minute repeated from different instants marks every one of them.

Issue #1742, opened 2026-08-24
recall_context spends 22 s on one large session to return 8 KB — every message is redacted and sanitised, then thrown away


vshulcz:
One session of 30001 messages (279 MB on disk, a real shape for a long-running agent transcript). Reading it is fast; rendering the 8 KB digest is not:

```
search:               12.6 ms   sessions=1
wholeSessionForMCP:  189.9 ms   messages=30001
search.PrintContext:  22.55 s   bytes=8064
```

Over MCP that is the whole tool call, three runs in a row, nothing cached:

```
  recall            0.21s  bytes=1262
  recall_context   22.65s  bytes=8215
  run: 22.63s / 22.75s / 22.73s
```

MCP clients have timeouts — the repo already fixed one 2.86 s call for that reason (#1309). 22 s is a tool that fails.

Control: `recall` over the same index answers in 0.21 s, `deja index` of the same store takes 4.7 s, and the session-start hooks are 0.03–0.04 s. Nothing else on this store is slow.

Cause: `printContextChunks` (`internal/search/search.go:1372`) walks **every** message and builds its rendered chunk — `SafeText(contextText(m.Text, …))`, and `contextText` opens with `redact.SafeForDisplay` — before the window that selects ~8 KB is chosen at line 1398. So 279 MB is redacted and sanitised to produce 8064 bytes, and the other 99.997% is discarded.

The same function backs `deja ctx`, `deja show`'s context header and the `deja://session/…` resource read, so all of them pay it.

Expected: the cost is bounded by what can be shown, not by the size of the session.

Found while checking a single session of hundreds of megabytes (BUG-HUNT 82). The rest of that check held: indexing 279 MB takes 4.7 s at 638 MB peak RSS, the index stays 3.7 MB, search is 30 ms, and `deja show` on it prints the whole thing but says `30001 messages — --offset n --limit n reads a slice` first.

vshulcz:
Fixed by #1743, on main since the collector merge in #1912.

PR #1791, opened 2026-08-25
perf(ctx): render a digest's turns across cores

### LINKED ISSUE #1790: deja ctx spends 3.5 s of its 4.5 s rendering turns it will not print
`deja ctx` prints an ~8 KB window, but renders every kept turn of the session first — redaction, prose collapsing, terminal safety — and then throws away everything the window does not reach. On a 279 MB single-session corpus that is most of the command's runtime.

## Reproduction

One session, 30001 messages (the fixture from #1742's measurement):

```
$ time deja ctx resharding
# deja context: claude · proj · whopper · updated 2026-08-05

## assistant

grumbleflux fixed by resharding the whopper

real  4.51s
```

Three runs: 5.18s, 4.50s, 4.51s. Output is 123 bytes.

Where it goes — `DEJA_CPUPROFILE=/tmp/cpu.out deja ctx resharding`, `go tool pprof -top -cum`:

```
      flat  flat%   sum%        cum   cum%
         0     0%     0%      3.49s 79.86%  internal/search.PrintContext
         0     0%     0%      3.49s 79.86%  internal/search.printContextChunks
         0     0%     0%      2.41s 55.15%  internal/search.contextText
         0     0%     0%      2.38s 54.46%  internal/search.proseForSnippet
     1.58s 36.16% 36.16%      1.58s 36.16%  runtime.memmove
     0.50s 11.44% 47.60%      1.19s 27.23%  strings.Fields
         0     0% 47.60%      1.02s 23.34%  strings.Join
```

Index loading is 0.12s of the run; the rest is rendering. The control that this is rendering and not matching: the include callback (`PrintContext.func1`, which lowercases and counts query terms in every turn) is 0.51s — a seventh of what the render costs.

## Cause

`printContextChunks` (internal/search/search.go:1388) renders in one sequential pass:

```go
for _, m := range s.Messages {
    ...
    text := SafeText(contextText(m.Text, weight > 0))
```

and only afterwards does the sliding window pick which chunks fall inside the 8 KB budget.

## Not the fix

Rendering only the window was tried in #1742 and reverted: predicted sizes diverge from rendered ones (prose collapsing drops whole tool-dump lines), so the window moves and the output changes. Whatever lands here has to be byte-identical to today's output on every query.

vshulcz:
Closes #1790.

`deja ctx` renders every kept turn before the 8 KB window picks which ones print, and that render — redaction, prose collapsing, terminal safety — was 3.5 s of a 4.5 s run on the 30001-message session. Which turns are kept still happens in one ordered pass (the include callback carries `prevKept` from turn to turn); only the rendering is spread across cores, written back by index, so the digest is the same bytes in the same order however many cores run it. Sessions under 32 kept turns render in place.

Rendering only the window is the other option and was tried and reverted in #1742: collapsed sizes diverge from predicted ones and the window moves.

Measured on the same fixture, three runs each: 4.51 / 4.50 / 5.18 s before, 1.16 / 1.19 / 1.15 s after. Output byte-identical on `resharding`, `grumbleflux`, `shard`, `whopper`.

Tests: `TestContextIsTheSameBytesOnOneCoreAndMany` renders a mixed fixture (prose, tool dumps, a fenced block, an escape byte) on 1 core and 8 and compares; `TestShortSessionsRenderInPlaceAndMatch` pins the in-place path. Control: writing results at the wrong index fails the first test on all four queries.

### DIFF
--- a/internal/search/context_parallel_test.go
+++ b/internal/search/context_parallel_test.go
@@ -1,0 +1,81 @@
+package search
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/vshulcz/deja-vu/internal/model"
+)
+
+// mixedContextSession is the shape that broke rendering only the window in
+// #1742: prose next to tool dumps and numbered output, so collapsing changes
+// sizes, plus a fenced block and an escape byte that the renderers treat
+// differently from ordinary text.
+func mixedContextSession(turns int) model.Session {
+	base := time.Date(2026, 8, 25, 9, 0, 0, 0, time.UTC)
+	s := model.Session{Harness: "claude", Project: "proj", ID: "mixed", Updated: base}
+	for i := range turns {
+		role := "user"
+		if i%2 == 1 {
+			role = "assistant"
+		}
+		var text string
+		switch i % 5 {
+		case 0:
+			text = fmt.Sprintf("turn %d: the resharding question came back today", i)
+		case 1:
+			text = fmt.Sprintf("  %d | 0x%04x  some tool dump line\n  %d | 0x%04x  another\nwe decided to cap retries at 3", i, i, i+1, i+1)
+		case 2:
+			text = fmt.Sprintf("```go\nfunc turn%d() {}\n```", i)
+		case 3:
+			text = strings.Repeat(fmt.Sprintf("turn %d prose that runs long enough to matter. ", i), 40)
+		case 4:
+			text = fmt.Sprintf("turn %d\x1b[31m coloured \x1b[0m and \t tabbed\n\n\n  spaced   out  ", i)
+		}
+		s.Messages = append(s.Messages, model.Message{Role: role, Text: text, Time: base.Add(time.Duration(i) * time.Minute)})
+	}
+	return s
+}
+
+// The digest is the same bytes however many cores render it: turns are written
+// back by index, so parallel rendering cannot reorder or drop one (#1790).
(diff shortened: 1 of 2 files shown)


vshulcz:
Checks run while the review was in flight:

- `GOMAXPROCS=1`: output identical to the sequential build.
- `go test ./internal/search/ -race -count=2`: clean.
- Peak RSS on the 279 MB fixture: 1.007 GB before, 1.120 GB after (+11%). The workers hold a rendered copy of a large turn each, so the extra is bounded by workers × biggest turn. Trading that for 4.5 s → 1.2 s looks right, but it is a real number and worth a second opinion.

vshulcz:
Correcting a number I posted on this PR: I reported peak RSS going from 1.007 GB to 1.120 GB (+11%) and attributed it to the workers each holding a rendered copy. Re-measured properly — same 279 MB fixture, same query, worker count pinned at 1, 2, 4, 8 and 11, two runs each:

```
workers=1   4.98–5.74s   930–1080 MB
workers=2   2.87–3.45s   1028–1073 MB
workers=4   1.83–2.35s   1082–1088 MB
workers=8   1.41–1.98s   1084–1085 MB
workers=11  1.33–1.89s   1090–1097 MB
```

Peak RSS is flat across the range — the spread inside one worker count is as large as the spread across all of them, so the earlier +11% was allocator noise, not the workers. Time keeps improving to the core count.

That also answers the follow-up I had queued (would a bounded worker count keep most of the win at a smaller peak): no — capping workers costs the time and saves nothing.

Issue #1090, opened 2026-08-06
search, show and last print terminal escape sequences from transcript text


vshulcz:
`safeForStatusline` (#634) and `trimBriefTitle` strip control characters
because transcript text arrives verbatim and a terminal acts on it. The
plain-text renderers for search, show and last do not.

Fixture: one claude session whose only message contains
`\x1b[31m`, `\x1b[2K`, `\x1b[1A`, `\x1b]0;pwned\x07`. Indexed normally
(1 session, 1 message), then each surface run with NO_COLOR=1 and its stdout
scanned for those exact byte sequences:

    search      rc=0 bytes=102    OSC set title, BEL
    show        rc=0 bytes=119    SGR colour, erase line, cursor up, OSC set title, BEL
    share       rc=0 bytes=164    BEL
    last        rc=0 bytes=106    SGR colour, erase line, cursor up, OSC set title, BEL
    view        rc=0 bytes=53     clean
    search json rc=0 bytes=682    clean

So a session can move the reader's cursor, erase the line above, and set the
terminal window title when someone runs `deja last`. `--json` and the HTML
view are already clean; this is the text path only.

It matters most for imported memory: `deja sync import` brings transcripts
from another machine, and the trust policy exists because that boundary is
not trusted. A NUL, a bell and an ESC also reach stdout from an ordinary
corrupt transcript — same fixture family, `deja "control bytes"` prints
NUL/BEL/ESC verbatim.

PR #1793, opened 2026-08-25
fix(remember): bound and de-escape the project name the confirmation echoes back

### LINKED ISSUE #1792: deja remember echoes the project name raw: escape bytes recolour the line and 5000 characters print whole
`deja remember --project X` prints "deja: remembered under X" with X exactly as given. Escape bytes reach the terminal and there is no length bound, on the CLI and over MCP.

## Reproduction

Escapes and a carriage return, shown through `cat -v`:

```
$ deja remember "shard limit is 64" --project "$(printf 'red\033[31mALERT\033[0m\rrewound')" 2>&1 | tail -1 | cat -v
deja: remembered under red^[[31mALERT^[[0m^Mrewound
```

Without `cat -v` the line is recoloured and then rewound by the `\r`, so what stays on screen is not what was stored.

Length, same command with a 5000-character project:

```
$ deja remember "the shard limit is 64" --project "$(python3 -c "print('w'*5000)")" | wc -c
5114
```

Over MCP the same value comes back in the tool reply a model reads:

```
reply length 3044
'Remembered under proj\x1b[31mALERT\x1b[0m\rrewoundwwwwwwwwwwwwwwwwwwww…'
```

Control: the listing surfaces already bound this. `mcp_resources.go:19` sets `mcpResourceNameMax = 80` and passes both title and project through `safeForStatusline`, which maps control and format runes to spaces and cuts at the width; `promote.go:282` cuts a line at 80 runes with an ellipsis. The remember echoes are the surface that does neither.

## Cause

- `cmd/deja/remember.go:76` — `fmt.Fprintf(os.Stdout, "deja: remembered under %s%s\n", strings.TrimSpace(project), suffix)`
- `cmd/deja/remember.go:56` — the "already remembered under" line
- `cmd/deja/mcp.go:432` and `cmd/deja/mcp.go:449` — the two replies that name the project

Same class as #1090, where an escape byte in a project or id recoloured the `deja ctx` header. Impact is cosmetic-to-misleading rather than content loss: the confirmation of what was stored is the one line a user reads to check they got it right.

vshulcz:
Closes #1792.

`deja remember` and the MCP `remember` tool echoed the project name exactly as typed: an escape byte recoloured the confirmation and a `\r` rewound it, and a 5000-character name printed whole (3 KB of it into a model's context over MCP). Both now go through the bound the listing surfaces have used since #1090 — `safeForStatusline` at `mcpResourceNameMax` (80), then `neutralizeFrameMarkers`.

Only the echo changes; the note is still stored under the name as given, so nothing about lookup or dedup moves.

Tests fail before the fix: `TestRememberEchoDoesNotHandTheTerminalWhateverWasTyped`, `TestRememberEchoBoundsALongProject`, `TestMCPRememberReplyBoundsTheProject`.

Before / after on the CLI, through `cat -v`:

```
deja: remembered under red^[[31mALERT^[[0m^Mrewound
deja: remembered under red [31mALERT [0m rewound
```

Wording is yours: the cut name ends in `…` and keeps the first 80 columns. If you would rather the confirmation name no project at all when it was cut, say so and I will change it.

### DIFF
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -429,7 +429,7 @@ func callMCPTool(dir, name string, raw json.RawMessage) (string, error) {
 			// Not an error to the agent: the fact it wanted stored is stored.
 			// Saying so stops it retrying, and stops one fact costing a line of
 			// every later recall (#1736).
-			return fmt.Sprintf("Already remembered under %s.", strings.TrimSpace(a.Project)), nil
+			return fmt.Sprintf("Already remembered under %s.", projectForEcho(a.Project)), nil
 		case err != nil:
 			return "", notesWriteError(err)
 		}
@@ -446,7 +446,7 @@ func callMCPTool(dir, name string, raw json.RawMessage) (string, error) {
 		// The journal is where the user sees what the agent did with their
 		// store; a write belongs there at least as much as a read.
 		usage.RecordResult(dir, usage.KindRemember, len(a.Text), 1, false)
-		return fmt.Sprintf("Remembered under %s.", strings.TrimSpace(a.Project)), nil
+		return fmt.Sprintf("Remembered under %s.", projectForEcho(a.Project)), nil
 	default:
 		return "", fmt.Errorf("unknown tool %q", name)
 	}
--- a/cmd/deja/remember.go
+++ b/cmd/deja/remember.go
@@ -12,6 +12,16 @@ import (
 	"github.com/vshulcz/deja-vu/internal/sources"
 )
 
+// projectForEcho is how a project name is written back to whoever named it.
+// The value is theirs and reaches a terminal or a model unchanged otherwise:
+// an escape byte recoloured the confirmation and a carriage return rewound it,
+// and a 5000-character name printed whole. The listing surfaces have bounded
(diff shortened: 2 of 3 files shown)


vshulcz:
Follow-up in the same PR: a project of nothing but control bytes bounded down to an empty string, so the line read `deja: remembered under ` and the MCP reply `Remembered under .`. It now says "a name with no printable characters" — wording is yours to change. Test: `TestRememberEchoNamesAProjectThatBoundsToNothing`.

vshulcz:
Reviewer flagged the tombstone hint at `remember.go:85` and proposed routing the dayNote through `projectForEcho` too. Declining that fix: `internal/sources/notes.go:272` builds the note's id as `"deja-" + day + "-" + project` from the raw name, so a sanitised id names no session and the `deja forget --unforget deja:<id>` the line hands over stops working. The observation is real — reproduced, the raw id does carry an escape byte to the terminal — so it is filed as #1794 with the three options, rather than fixed here with a command that no longer runs.

Everything else in the review came back clean: the four tests fail on the unfixed code and pass on this branch, storage and lookup still use the raw name, CJK width / combining marks / the 80-column boundary behave, and `go test ./... -count=1` passes.

PR #1796, opened 2026-08-25
fix(mcp): refuse a batch as an invalid request instead of claiming it did not parse

### LINKED ISSUE #1795: a JSON-RPC batch is answered "parse error" and never runs; jsonrpc 1.0 is served as if it were 2.0
A JSON-RPC batch — an array of requests in one frame — is answered with `-32700 parse error` and a null id, and none of its requests run. The frame is valid JSON; nothing about it failed to parse.

## Reproduction

```
    '[{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}},{"jsonrpc":"2.0","id":3,"method":"ping","params":{}}]' \
    '{"jsonrpc":"1.0","id":4,"method":"ping","params":{}}' \
    '{"id":5,"method":"ping","params":{}}' | deja mcp
{"id":1,"jsonrpc":"2.0","result":{"capabilities":…}}
{"error":{"code":-32700,"message":"parse error"},"id":null,"jsonrpc":"2.0"}
{"id":4,"jsonrpc":"2.0","result":{}}
{"id":5,"jsonrpc":"2.0","result":{}}
```

Three separate things in that output:

1. The batch gets `-32700`, which says the frame was not JSON. It was. A client that reads the code cannot tell "deja does not do batches" from "your bytes were corrupt", and the null id means it cannot match the failure to either request it sent.
2. `"jsonrpc":"1.0"` is served as if it were 2.0.
3. A request with no `jsonrpc` member at all is served too.

Control that the frames themselves are fine: id 4 and id 5 both get results, and the same two requests sent one per line without the array wrapper both answer normally.

## Cause

`cmd/deja/mcp.go:56` unmarshals every frame into a struct; an array fails that unmarshal and falls into the `-32700` branch at line 57. Nothing anywhere looks at `req.JSONRPC`.

## Impact

No harness we install into sends a batch today, so nothing is broken in practice — this is about the answer being true. JSON-RPC 2.0 allows batching and the protocol version we advertise (2024-11-05) does not forbid it; 2025-06-18 removed it. Either we run the array, or we say plainly that batches are not supported — `-32600 invalid request` with a sentence — rather than claiming the frame did not parse.

vshulcz:
Closes #1795.

A JSON-RPC batch got `-32700 parse error` with a null id — the frame parsed fine, and the client could not tell "deja does not do batches" from "your bytes were corrupt", nor which request died. It is now `-32600` with a sentence, carrying the id of the first request in the array so the refusal can be matched to something the client sent. `{"jsonrpc":"1.0"}` was served as if it were 2.0 and is now refused the same way.

A frame that omits `jsonrpc` entirely is still served — clients in the wild omit it and the request is unambiguous. That leniency is now pinned by a test rather than being an accident.

Before / after on the same four frames:

```
{"error":{"code":-32700,"message":"parse error"},"id":null,"jsonrpc":"2.0"}
{"id":4,"jsonrpc":"2.0","result":{}}

{"error":{"code":-32600,"message":"batch requests are not supported — send one request per line"},"id":2,"jsonrpc":"2.0"}
{"error":{"code":-32600,"message":"unsupported jsonrpc version 1.0 — this server speaks 2.0"},"id":4,"jsonrpc":"2.0"}
```

Not doing: running the array. No harness we install into sends one, 2025-06-18 removed batching from MCP, and a half-supported batch path is worse than a clear refusal. If a client turns up that needs it, the refusal is the thing to change.

Tests fail before the fix: `TestABatchIsRefusedAsInvalidNotAsUnparseable`, `TestAWrongJSONRPCVersionIsRefused`. Control that corrupt bytes keep their parse error: `TestCorruptBytesStillParseError`. Leniency pinned by `TestAMissingJSONRPCVersionIsStillServed`.

The two refusal sentences are wording — yours to change.

### DIFF
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -41,18 +41,32 @@ func isNotification(id json.RawMessage) bool {
 	return len(id) == 0 || string(id) == "null"
 }
 
+// parseBatch reports whether a frame is an array of requests, and returns them.
+// A frame that only starts like one — a truncated `[` — is not a batch and
+// keeps its parse error.
+func parseBatch(frame string) ([]json.RawMessage, bool) {
+	if !strings.HasPrefix(frame, "[") {
+		return nil, false
+	}
+	var elems []json.RawMessage
+	if err := json.Unmarshal([]byte(frame), &elems); err != nil {
+		return nil, false
+	}
+	return elems, true
+}
+
 // batchID returns the id of the first request in a batch, so the refusal can
 // be matched to something the client sent rather than coming back as null.
 // A batch whose first element carries no id gets a null id, as before.
-func batchID(frame string) json.RawMessage {
-	var reqs []rpcRequest
-	if err := json.Unmarshal([]byte(frame), &reqs); err != nil || len(reqs) == 0 {
+func batchID(elems []json.RawMessage) json.RawMessage {
+	if len(elems) == 0 {
 		return nil
 	}
-	if isNotification(reqs[0].ID) {
+	var first rpcRequest
+	if err := json.Unmarshal(elems[0], &first); err != nil || isNotification(first.ID) {
 		return nil
 	}
-	return reqs[0].ID
+	return first.ID
 }
 
 const mcpMaxFrame = 10 * 1024 * 1024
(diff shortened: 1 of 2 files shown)


vshulcz:
Second commit, from probing the detection myself: `strings.HasPrefix(trimmed, "[")` alone called a truncated `[` a batch. Detection now decodes the array (`[]json.RawMessage`) and only a frame that really is one counts, so `[` and `"[not a batch"` keep `-32700` while `[]`, `[1,2,3]` and a real batch get the refusal. The id is taken from the first element whatever its type — a string id comes back as a string.

Pinned by `TestOnlyARealArrayCountsAsABatch` and `TestABatchRefusalCarriesAStringID`.

One deliberate divergence from JSON-RPC: a batch containing only notifications gets the refusal rather than silence. The spec says a notification gets no reply, but a client that sends a batch of them would otherwise learn nothing and keep sending them.

vshulcz:
Reviewer was right and I was wrong on the notification-only batch: I had argued the client learns more from a refusal, but the spec forbids answering a notification inside a batch or out of it, and deja already ignores a bare notification on its own line — the refusal was the inconsistent half.

Third commit: `batchReply` decides both whether an answer is owed and which id it carries. A batch of nothing but notifications gets silence; a mixed batch gets the refusal on the id of the first request that asked for one (previously it would have come back null if the first element happened to be a notification); an empty or malformed array is an invalid request like any other and still gets the refusal with a null id.

Test: `TestABatchOfNotificationsIsAnsweredWithSilence`, plus the notification case added to the table in `TestOnlyARealArrayCountsAsABatch`.

Issue #1726, opened 2026-08-24
a full recall page drops the line telling the agent how to get the rest


vshulcz:
`recall` ends with `N more match(es) — call recall again with offset=N`, and the header's arithmetic is written around it ("the agent asks for offset=served and the arithmetic has to hold", `cmd/deja/mcp.go:890`). The line is appended before the final `trimUTF8(out, budget)`, so on a full page it is the first thing cut — the agent is told 5 of 12 matched and never told how to reach the other 7.

Repro — 12 sessions matching `zqxwv`, each long enough to fill the page, recalled over MCP:

```
{"query": "zqxwv"}              bytes=4096  deja recall for "zqxwv" (5 of 12 m…  paging line: MISSING
{"query": "zqxwv", "limit": 2}  bytes=1812  deja recall for "zqxwv" (2 of 12 m…  paging line: 10 more match(es) — call recall again with offset=2.
{"query": "zqxwv", "limit": 1}  bytes=1032  deja recall for "zqxwv" (1 of 12 m…  paging line: 11 more match(es) — call recall again with offset=1.
```

The last two are the control: the same query, the same index, and the line is there whenever the payload is small. It disappears exactly when the budget is full — which is the case where paging is the point.

Expected: the page ends with the paging line at any budget; the excerpts are what should give way, not the instruction.

Impact: an agent that fills a page has no way to learn the offset to ask for, so `offset` is unreachable in the one situation it exists for.

Cause: `recallTextResult` (`cmd/deja/mcp.go:918`) trims the whole assembled string, paging line included.

Everything else at the cap is sound and was checked in the same session: the body stops at exactly 4096 bytes, never mid-rune (a Cyrillic + emoji payload came back with no replacement characters), `offset` pages correctly, and `offset=99` past the end answers `"zqxwv": 12 total, offset 99`.

vshulcz:
Fixed by #1727, on main since the collector merge in #1912.

PR #1798, opened 2026-08-25
fix(mcp): hold recall_context to the ~8KB it promises, and say when a digest is cut

### LINKED ISSUE #1797: recall_context has no byte cap: the documented ~8KB is 8221 on an ordinary session and 8335 when the project name is long
`recall_context` documents "~8KB" and applies no cap at all. `recall` subtracts the frame overhead from its 4096 so the framed reply fits (`cmd/deja/mcp.go:305`); the context path (`mcp.go:329`) passes no budget, so the reply is whatever `PrintContext` produced plus the header plus the frame.

## Reproduction

400-message session, mixed Cyrillic and ASCII, queried over MCP:

```
    '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"recall_context","arguments":{"query":"резервирование шардов"}}}' | deja mcp
bytes 8221 runes 5338
```

8221 against a documented 8192. That much is a rounding argument. This is not:

```
# same session, project directory name 180 chars, session id 120 chars
bytes 8335
```

The digest header is `# deja context: <harness> · <project> · <id> · updated <date>`, and none of those are bounded here, so the overshoot grows with names the user chose. There is no upper bound in the code — only the 8000-byte body budget inside `printContextChunks`, which the header, the tier prefix and the frame are all added on top of.

Control that the cap is not simply "8000 means 8000 of body": `recall` at the same layer is exact, because 4096 is passed as `4096-recallFrameOverhead` and the trim happens before the frame goes on.

## Impact

An agent's context window is the budget being spent, and the tool description is what an agent plans against. Over by 29 bytes on an ordinary session is cosmetic; over by 143 because a project has a long path is the same bug with a bigger number, and nothing stops it growing further.

## Cause

`cmd/deja/mcp.go:329` — `recallContextResult(dir, a.Query, a.Harness)` takes no budget, and `recallContextResult` (mcp.go:1108) returns `search.PrintContext`'s output with a tier prefix, unframed and untrimmed.

vshulcz:
Closes #1797.

`recall_context` passed no budget at all: `PrintContext` bounds the body at 8000, and the header, the tier prefix and the frame went on top. An ordinary session came back at 8221 bytes; a session in a project with a 180-character path came back at 8335, and nothing bounded that — the header carries the project name and the session id. `recall` at the same layer has been exact since it started passing `4096-recallFrameOverhead`; this is the same shape.

Measured on the same two fixtures: 8221 → 8192, 8335 → 8192.

The trim would otherwise end a digest mid-word, so it cuts at a line boundary when one is within 400 bytes and appends a line saying the digest was trimmed and how to get the rest. The marker is reserved before the trim, the way recall reserves its paging line (#1726), so the thing that explains the cut cannot itself be cut.

Tests: `TestRecallContextFitsTheBudgetItDocuments` (both fixtures; fails before at 8217/8507), `TestATrimmedDigestSaysItWasTrimmed` with a control that a short session gets no marker.

The marker's wording is yours. It is also half of the queued question about truncation markers on the recall side — this only covers the digest.

### DIFF
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -328,7 +328,7 @@ func callMCPTool(dir, name string, raw json.RawMessage) (string, error) {
 		}
 		text, sessions, raw, ids, err := recallContextResult(dir, a.Query, a.Harness)
 		if err == nil {
-			text = frameRecall(text)
+			text = frameRecall(fitContextDigest(text, contextMCPBudget-recallFrameOverhead))
 			usage.RecordServedSessions(dir, usage.KindContext, len(text), sessions, sessions == 0, raw, ids)
 			usage.SnapshotPolicy(dir, usage.KindContext, text, sessions, policy.Load().Describe(policy.ActivationMCP))
 		}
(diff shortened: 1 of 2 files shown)


vshulcz:
Reviewer's finding stands and is fixed: the backup to a line break could take the matched turn with it. `fitContextDigest` now takes the query and only backs up when the shorter body still carries a query word the longer one had — tidiness never costs the answer.

Their reproduction and mine differed on one point worth recording: a match that does not survive the trim itself is gone regardless, so the case that matters is a long last line whose match sits early in it. `TestTheBackupNeverDropsTheMatch` builds exactly that (a 500-byte final line, match at its start, budget cutting 200 bytes into it) and fails without the guard; its control checks the backup still happens when nothing is at stake.

PR #1800, opened 2026-08-25
fix(mcp): mark the excerpt the recall page budget cut, like every other cut line

### LINKED ISSUE #1799: at the recall byte cap the last excerpt is cut with no marker, while every other truncated line ends with …
When a recall page hits the byte cap, the last excerpt line is cut mid-word with nothing to say so, while every other truncated line in the same payload ends with `…`.

## Reproduction

12 sessions of repeated text, one query, over MCP:

```
    '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"recall","arguments":{"query":"grumblewidget retry budget"}}}' | deja mcp
bytes 4096
```

tail of the reply:

```
… grumblewidget retry budget 8 1 grumblewidget retry b …
- grumblewidget retry budget 8 2 grumblewidget retry budget 8 2 grumblewidget retry budget 8 2 grumblewi
8 more match(es) — call recall again with offset=4.
```

The line above ends in `…` because the excerpt itself was truncated. The last line ends in `grumblewi` — cut by the page budget, with no marker. An agent reading it has no way to tell it is a fragment, and the words after the cut are the ones it might have gone looking for.

Control that the marker convention exists and is applied everywhere else: the same payload has `…` on every excerpt that was shortened; only the final line, cut by `trimUTF8` at `cmd/deja/mcp.go:1037`, has none.

## Cause

`cmd/deja/mcp.go:1037` — `out = trimUTF8(out, budget-len(more))`. The trim is byte arithmetic and adds nothing.

The fix has to reserve the marker before trimming, the way the paging line is reserved (#1726) and the digest marker now is (#1797), so the thing that reports the cut cannot itself be cut.

vshulcz:
Closes #1799.

Every excerpt shortened on its own ends with `…`; the one the page budget cut ended mid-word with nothing, so the last line an agent reads was the only line it could not tell was a fragment. The marker is reserved before the trim, the way the paging line is (#1726) and the digest marker now is (#1797), so the thing that reports the cut cannot itself be cut. The page still comes back at exactly 4096.

```
- … grumblewidget retry budget 8 2 grumblewidget retry budget 8 2 grumblewi
- … grumblewidget retry budget 8 2 grumblewidget retry budget 8 2 grum …
```

A trim that lands exactly on a line break gets no marker inside the line — it already ends cleanly.

Tests: `TestTheLastExcerptOnAFullPageSaysItWasCut` (fails before), with `TestAShortPageIsNotMarkedAsCut` as the control that a page inside its budget is not marked.

### DIFF
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -1125,6 +1125,11 @@ func markCut(s string) string {
 	if strings.HasSuffix(s, "\n") {
 		return s
 	}
+	// A trim landing between an excerpt's own ellipsis and its newline already
+	// says the line was shortened; a second one reads as "… …".
+	if strings.HasSuffix(strings.TrimRight(s, " "), "…") {
+		return s + "\n"
+	}
 	return s + cutMarker
 }
 
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -1086,7 +1086,11 @@ func recallTextResult(dir, q, harness string, limit, offset, budget int) (string
 		if len(more) >= budget {
 			more = ""
 		}
-		out = trimUTF8(out, budget-len(more))
+		// Every excerpt shortened on its own ends with the marker; the one the
+		// page budget cut ended mid-word saying nothing, so the last line an
+		// agent reads was the one line it could not tell was a fragment
+		// (#1799). Reserved before the trim, like the paging line above.
+		out = markCut(trimUTF8(out, budget-len(more)-len(cutMarker)))
 	}
 	out += more
 	var raw int64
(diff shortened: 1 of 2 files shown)


vshulcz:
Found while probing the arithmetic myself, before the review landed: at tiny budgets the marker was added on top rather than reserved — budget 1 produced a 5-byte body, budget 60 produced 61. Second commit guards it: when there is no room for both the page and the marker, keep the bytes and drop the marker.

Measured across budgets 1, 5, 10, 60, 200, 1000 and 4096-recallFrameOverhead: body never exceeds the budget, the framed page is exactly 4096 at the real one, and `… …` never appears. Pinned by `TestTheCutMarkerNeverBreaksTheBudget`.

vshulcz:
Reviewer's latent case is real: a trim landing between an excerpt's own `…` and its newline would have produced `… …`. Third commit ends the line instead of adding a second marker, pinned by `TestACutOnAnEllipsisIsNotMarkedTwice` on markCut directly rather than hoping fixture text reaches that state.

PR #1803, opened 2026-08-25
feat(completion): support PowerShell

### LINKED ISSUE #1577: no shell completion on Windows, which is a shipped platform
`deja completion` covers bash, zsh and fish. Windows has no completion at all, and Windows is a shipped platform: the release carries `windows_amd64` and `windows_arm64` archives, `packaging/scoop/deja-vu.json` and three winget manifests, and `deja install` has a `cmd /c deja mcp` path written specifically for it.

```
$ deja completion powershell
deja: unknown shell "powershell"; want bash, zsh, or fish
```

So a scoop or winget user tabs `deja ins<TAB>` and gets nothing, while the same user on macOS gets the install targets, the harness names and the handoff targets.

The scripts are hand-written and live in `cmd/deja/completion.go`, one const per shell, with three placeholders substituted at print time so the lists are not duplicated per shell:

- `%INSTALL_TARGETS%` — `installTargetNames()`
- `%HARNESSES%` — `sources.HarnessNames()`
- `%HANDOFF_TARGETS%` — `handoffTargets()`

A PowerShell script registers an argument completer, roughly:

```powershell
Register-ArgumentCompleter -Native -CommandName deja -ScriptBlock { ... }
```

What makes this a good change rather than a copy-paste: the three lists must come from the same substitution, not from a fourth hand-written copy. The comment above them says why — the lists fell seven harnesses behind the last time they were duplicated.

Worth covering in `cmd/deja/completion_test.go` the way the others are: the script mentions every install target and harness name, and `deja completion powershell` exits zero.

One judgement call to make and state in the PR: whether `pwsh` should be accepted as an alias for the same script, or whether one name is enough.

jjj-n:
## Summary

- add native PowerShell completion for commands, subcommands, and common flags
- populate harness, install, and handoff candidates from the same runtime lists used by the existing shells
- accept `pwsh` as an alias and expose PowerShell in the existing completions, CLI help, and command guide

Closes #1577

## Test plan

- `go test ./cmd/deja -run 'TestCompletion|TestPowerShellCompletion' -count=1`
- `go vet ./...`
- `go build ./cmd/deja`
- sourced the generated script and verified `TabExpansion2` results in PowerShell 7 and Windows PowerShell 5.1

## Notes

`deja completion pwsh` emits the same script as `deja completion powershell`.

### DIFF
--- a/cmd/deja/completion.go
+++ b/cmd/deja/completion.go
@@ -11,11 +11,11 @@ import (
 // runCompletion writes a shell-specific script so the binary stays dependency-free.
 func runCompletion(args []string) error {
 	if len(args) != 1 {
-		return fmt.Errorf("completion needs bash, zsh, or fish")
+		return fmt.Errorf("completion needs bash, zsh, fish, or powershell")
 	}
 	script, ok := completionScripts[args[0]]
 	if !ok {
-		return fmt.Errorf("unknown shell %q; want bash, zsh, or fish", args[0])
+		return fmt.Errorf("unknown shell %q; want bash, zsh, fish, powershell, or pwsh", args[0])
 	}
 	// Every list a shell offers is substituted rather than duplicated per shell:
 	// three hand-maintained copies is how this one fell seven harnesses behind.
@@ -30,9 +30,11 @@ func runCompletion(args []string) error {
 }
 
 var completionScripts = map[string]string{
-	"bash": bashCompletion,
-	"zsh":  zshCompletion,
-	"fish": fishCompletion,
+	"bash":       bashCompletion,
+	"zsh":        zshCompletion,
+	"fish":       fishCompletion,
+	"powershell": powershellCompletion,
+	"pwsh":       powershellCompletion,
 }
 
 const bashCompletion = `# bash completion for deja
@@ -75,7 +77,7 @@ _deja_completion() {
             fi
             ;;
         completion)
-            COMPREPLY=( $(compgen -W "bash zsh fish" -- "$cur") )
+            COMPREPLY=( $(compgen -W "bash zsh fish powershell pwsh" -- "$cur") )
             ;;
         doctor)
             COMPREPLY=( $(compgen -W "--json --offline --deep" -- "$cur") )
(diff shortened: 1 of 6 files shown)


vshulcz:
Read the whole diff and ran it here, on macOS.

**Checked the thing that has bitten this file before.** The comment above the substitution says three hand-maintained copies is how the completion fell seven harnesses behind, so the question for a fourth shell is whether it joins the substitution or starts a fourth copy. It joins: `%HARNESSES%`, `%INSTALL_TARGETS%` and `%HANDOFF_TARGETS%` are all substituted in the emitted PowerShell, with nothing left unexpanded. And the command list matches bash exactly — 35 against 35, no difference in either direction — which I checked by extracting both from the generated scripts rather than by reading them side by side.

Adding `powershell` to `completionTestShells` is the right move for the same reason: the two list tests now fail for PowerShell too if a future harness stops reaching it, which is the failure that actually happened once.

**What I could not check.** You say you sourced the script and exercised `TabExpansion2` in PowerShell 7 and 5.1. There is no `pwsh` on this machine, so I have not run a single line of that script — I read it and I checked what the binary emits. The parsing logic around `$wordToComplete` and `$commandAst` is yours; I am taking it on your word that it behaves in a real shell, and saying so rather than implying I verified it.

Two details I did read closely, since they are where this style of completer usually goes wrong:

- `$completingCommand` handles the empty-word case (`elements.Count -le 1`) separately from the partial-word case, so `deja <TAB>` and `deja st<TAB>` both offer commands rather than the first one falling through to the flag list.
- `$previous` is computed differently depending on whether a word is in progress, which is what makes `--harness <TAB>` offer harnesses instead of the command's flag set.

`pwsh` as an alias for the same script is a nice touch and costs one map entry.

CI needed a maintainer to approve the runs — first pull request from this fork. Approved; everything is green.

vshulcz:
Thank you — the part worth naming is that you put PowerShell into the existing substitution instead of writing a fourth list. That file has one scar on it, and it is exactly the scar a new shell reopens.

All checks green. Merging.

PR #1805, opened 2026-08-25
fix(mcp): decide remember's index refresh with one lock attempt instead of a check and a wait

### LINKED ISSUE #1804: remember is the last MCP tool that can block on a rebuild, and only a race-prone guard stops it
Every MCP tool but one reads through the non-blocking `index.EnsureForSearchStale` and answers from the current snapshot while a refresh runs (#1733). `remember` still calls the blocking `index.EnsureForSearch` (`cmd/deja/mcp.go:497`), and what keeps it from waiting out a whole rebuild is a check one moment earlier:

```go
if line := buildingNowForBlockingTool(dir); line != "" {
    return "Saved. " + line, nil
}
if err := index.EnsureForSearch(dir, search.Options{All: true}, false, mcpProgress()); err != nil {
```

`buildingNowForBlockingTool` asks `index.RebuildInProgress`, which is `tryLockDir` — "is the lock held right now" (`internal/index/manifest.go:498`). The blocking call then takes `lockDir`, the same lock. Guard and call are one lock acquisition apart: a rebuild that starts in that window makes the tool wait for it inside the client's tool call. Iteration 78 measured >8 s against a held lock with the guard removed; #1733 carries that number.

## What the guard does cover, measured

Lock held for 12 s by another process, three frames in one MCP session:

```
id 2 "Saved. deja is indexing this machine's history. Recall comes online in a few seconds; ask ..."
id 3 "<deja-recall>\nRecalled history from prior sessions..."
elapsed 0.0s
```

So today the common case is answered honestly and fast — this is not a hang report. The complaint is that `remember` is the last tool whose correctness depends on a check that cannot be atomic with the thing it protects, while the fix is the same call every other tool already makes.

`blame` was the other one named in #1733 and now goes through `findBlameHitsStale`; `remember` is what is left.

## What to change

Give `remember` `EnsureForSearchStale` plus `requestWarmup`, keeping the same sentence when a refresh is in flight, so the tool cannot block at all rather than being protected from blocking by a race.

vshulcz:
Closes #1804.

`remember` was the last MCP tool that could wait out a rebuild inside a client's call. What stopped it was `RebuildInProgress` — a try-lock — followed by a blocking `EnsureForSearch` on the same lock: the same question asked twice, a lock acquisition apart, so a rebuild starting in that window was waited out.

`index.EnsureForSearchNoWait` takes the lock or reports that someone else holds it, in one attempt, and `EnsureForSearch`'s body moves behind `ensureLocked` so both share it. `remember` keeps the pre-check for the things a lock cannot tell it — an index written by another version, a directory it cannot write — which is what #1309's "don't rebuild inside the call" rests on.

Behaviour is unchanged where it was already right: with the lock held for 10 s a fresh note answers in under a second with "Saved. deja is indexing this machine's history…", and with the lock free the note is indexed inline and findable straight away (`TestMCPToolContract/remember_then_recall_round-trips`, `TestRememberSavesOnUpgradeDayWithoutRebuilding` unchanged).

Tests: `TestNoMCPToolTakesTheBlockingIndexPath` fails before (mcp.go had one `index.EnsureForSearch(`), `TestRememberAnswersWhileTheIndexIsLocked` holds the real lock and asserts the answer and the note on disk.

### DIFF
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -491,11 +491,27 @@ func callMCPTool(dir, name string, raw json.RawMessage) (string, error) {
 		// On upgrade day that meant rebuilding the whole index inside the call
 		// (#1309), so the agent is told instead — the detached warmup picks it
 		// up with everything else.
+		// The note is on disk either way; what is left is making it findable.
+		// On upgrade day that meant rebuilding the whole index inside the call
+		// (#1309), so the agent is told instead — the detached warmup picks it
+		// up with everything else.
 		if line := buildingNowForBlockingTool(dir); line != "" {
 			return "Saved. " + line, nil
 		}
-		if err := index.EnsureForSearch(dir, search.Options{All: true}, false, mcpProgress()); err != nil {
+		// The check above cannot cover a rebuild that starts after it: it and a
+		// blocking Ensure ask the same lock a moment apart, and what began in
+		// between was waited out inside the client's call (#1804). One attempt
+		// at the lock decides both.
+		busy, err := index.EnsureForSearchNoWait(dir, search.Options{All: true}, mcpProgress())
+		if err != nil {
 			return "", err
+		}
+		if busy {
+			requestWarmup(dir)
+			if line := buildingNowForBlockingTool(dir); line != "" {
+				return "Saved. " + line, nil
+			}
+			return "Saved. deja is refreshing its index; this note becomes findable when that finishes, in a few seconds.", nil
 		}
 		// The journal is where the user sees what the agent did with their
 		// store; a write belongs there at least as much as a read.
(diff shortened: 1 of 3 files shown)


vshulcz:
Second commit, found while checking the Windows and read-only paths myself: `tryLockDir` returns "no lock" for two different things — someone else holds it, and this machine cannot write the lock file at all (both platforms map a permission error to `ok=false, err=nil`). The first version read the second as busy, so `remember` on a read-only index would have said "deja is refreshing its index" every time, about a refresh that never comes. `EnsureForSearchNoWait` now separates them.

Test: `TestRememberOnAReadOnlyIndexDoesNotPromiseARefresh` (skipped when running as root, where the case cannot be built).

vshulcz:
Reviewer's one finding was a comment block pasted twice above the pre-check — removed in the third commit. Everything else came back clean on their side too: ensureLocked's body matches the original, the read-only case is distinguished from a held lock, Windows locking matches unix, and every remember branch leaves the index in a sane state.

PR #1807, opened 2026-08-25
fix(mcp): take the environment block out of recall's budget instead of adding it on top

### LINKED ISSUE #1806: the first recall of a session is 4505 bytes against a 4096 cap: the environment block is appended after the trim
`recall` trims its page to `4096-recallFrameOverhead` and then appends the environment block *after* the frame:

```go
text, sessions, raw, ids, err := recallTextResult(dir, a.Query, a.Harness, int(a.Limit), int(a.Offset), 4096-recallFrameOverhead)
if err == nil {
    text = frameRecall(text) + environmentOnce(dir)
```
(`cmd/deja/mcp.go:305`)

So the first recall of an MCP session — the one an agent plans the rest of its work against — is over the documented cap by whatever the block weighs.

## Reproduction

12 sessions, three recurring errors across them so the block has walls to name, one full recall page:

```
bytes 4505
env block: True
framed part: 4096 after frame: 409
```

The page itself is exact. The 409 bytes after `</deja-recall>` are the block.

Control: with the block suppressed (its 15-minute marker already stamped), the same query answers at 4096.

Same shape as #1797 on the context side, where the cap was applied to the body but the frame went on top. Here the frame is accounted for and the suffix is not.

## Impact

10% over on this fixture. The block is bounded — three walls, 96 bytes of error text each — so the overshoot is bounded too, but it lands on the first call of every session and nothing in the arithmetic knows about it.

vshulcz:
Closes #1806.

`recall` trimmed its page to `4096-recallFrameOverhead` and then appended the environment block after the frame, so the first recall of every MCP session came back at 4505 bytes on this fixture — 10% over the cap, on the call an agent plans the rest of its work against.

The block is computed first and its length comes out of the page budget, so the reply is exactly 4096 with the block included: 3687 of page, 409 of block. Later calls in the same session are unchanged — the block goes out once.

Tests: `TestTheFirstRecallOfASessionFitsItsBudget` (fails before at 4505). It resets both the per-process `sync.Once` and the 15-minute marker so it measures a session's first call however the package's tests are ordered — without that it passed alone and failed in the suite, which is the failure mode worth naming.

### DIFF
--- a/cmd/deja/environment.go
+++ b/cmd/deja/environment.go
@@ -85,9 +85,12 @@ func environmentBlock(dir, activation string) string {
 	return b.String()
 }
 
-// environmentServed is per process, which is what makes "once" countable on
+// environmentSpent is per process, which is what makes "once" countable on
 // this side: an MCP server lives for the whole session.
-var environmentServed sync.Once
+var (
+	environmentMu    sync.Mutex
+	environmentSpent bool
+)
 
 // environmentTTL is how long a delivered block suppresses the next one.
 //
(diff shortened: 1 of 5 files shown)


vshulcz:
Second commit, from checking the ordering myself: computing the block before the page meant a recall that then failed spent the session's one block — the `sync.Once` and the 15-minute marker were both consumed by the asking. `environmentOnce` now returns the block and a `deliver()` the caller runs once the reply carrying it is built, and `environmentServedRecently` no longer writes the stamp; `stampEnvironmentServed` does, called from the delivery and from the hook path that emits the block.

`TestAFailedRecallDoesNotSpendTheEnvironmentBlock` pins it. `TestEnvironmentBlockIsNotDeliveredTwice` changed with the contract — it asserted the check itself writes the stamp; it now asserts the opposite for the same guarantee, that both paths still send at most one block.

PR #1809, opened 2026-08-25
fix(sync): bound and de-escape peer names on the screens that report them

### LINKED ISSUE #1808: deja sync and doctor print peer host names raw: escape bytes rewrite the line and a 300-character host prints whole
Peer host names are printed exactly as the peers file holds them. A host with an escape byte in it recolours and rewinds the line; a 300-character host prints whole, three times over in one `deja sync` run.

## Reproduction

A peers file with two awkward hosts (RFC 5737 style: nothing here is a real machine):

```
$ python3 - <<'PY'
import json
esc = chr(27)
open("peers.json","w").write(json.dumps({"peers":[
  {"host":"build"+esc+"[31mALERT"+esc+"[0m\rrewound","last_push":"2026-08-01T10:00:00Z"},
  {"host":"w"*300,"last_push":"2026-08-02T10:00:00Z"}]}))
PY
$ DEJA_PEERS_FILE=peers.json deja sync 2>&1 | cat -v | head -3
deja: build^[[31mALERT^[[0m^Mrewound
deja: build^[[31mALERT^[[0m^Mrewound: ssh build^[[31mALERT^[[0m^Mrewound: exit status 255: hostname contains invalid characters:
deja: wwwwwwwwww… (300 characters, unbroken)
```

`deja doctor` prints the same value in its Sync section:

```
Sync:
  build^[[31mALERT^[[0m^Mrewound last exchange 23 days ago, nothing received yet — last attempt failed: ...
  wwwwwwww…(300)… last exchange 22 days ago, nothing received yet — ...
```

Without `cat -v` the `\r` rewinds each line, so what stays on screen is not the host deja actually tried.

Control that the surrounding machinery is not the problem: `peers.trimError` already collapses whitespace and cuts at 160 runes with an ellipsis, and `doctor --json` carries no escape byte or carriage return — the JSON encoder escapes them. The plain-text surfaces are the ones with nothing.

## Cause

- `cmd/deja/sync_ssh.go:135` and `:139` — the per-host progress and failure lines in `runSyncAll`
- `cmd/deja/sync.go:66` — the "forgotten" line
- `cmd/deja/doctor_peers.go:28` and `:54` — the doctor's Sync section, host and stored error

Same class as #1792 (`deja remember` echoing a project name raw) and #1090 before it. The listing surfaces have used `safeForStatusline` + a bound since then.

## Impact

A host name is something a user typed once, or that arrived with a shared config; it is not attacker-controlled in the usual case. The damage is a status screen that lies about which machine deja is talking to, which is exactly the screen someone reads when sync is misbehaving.

vshulcz:
Closes #1808.

`deja sync` and `deja doctor` printed a peer host exactly as the peers file holds it, so an escape byte recoloured and rewound the line a user reads when sync is misbehaving, and a 300-character host filled three lines of one run. Both now go through the bound the listing surfaces have used since #1090, and so does the stored `last_error`, which usually quotes the host back.

Before / after through `cat -v`:

```
deja: build^[[31mALERT^[[0m^Mrewound
deja: build [31mALERT [0m rewound
```

The name is bounded on screen and nowhere else: ssh is handed the host as written and `deja sync forget <host>` still matches on it in full — `TestForgetStillMatchesTheHostAsWritten` covers that, and `TestTheDoctorDoesNotHandTheTerminalAPeerName` fails before the fix.

`doctor --json` was already safe (the encoder escapes both), and `peers.trimError` was already rune-safe — the plain-text surfaces were the ones with nothing.

### DIFF
--- a/cmd/deja/doctor_peers.go
+++ b/cmd/deja/doctor_peers.go
@@ -25,7 +25,7 @@ func doctorPeers(w io.Writer, dir string, now time.Time) {
 	}
 	from := index.ImportedByMachine(dir)
 	for _, p := range list {
-		fmt.Fprintf(w, "  %-12s %s\n", p.Host, peerLine(p, from[p.Host], now))
+		fmt.Fprintf(w, "  %-12s %s\n", hostForEcho(p.Host), peerLine(p, from[p.Host], now))
 	}
 }
 
@@ -51,7 +51,9 @@ func peerLine(p peers.Peer, sessions int, now time.Time) string {
 		line += fmt.Sprintf(", %s from there", doctorCount(sessions, "session"))
 	}
 	if p.LastError != "" {
-		line += " — last attempt failed: " + p.LastError
+		// The stored error usually quotes the host back, so it carries whatever
+		// the name carried.
+		line += " — last attempt failed: " + safeForStatusline(p.LastError, 200)
 	}
 	return line
 }
--- a/cmd/deja/peers_echo.go
+++ b/cmd/deja/peers_echo.go
@@ -1,0 +1,18 @@
+package main
+
+// hostForEcho is how a peer's name is written to a screen. The name comes from
+// a config file — typed once, or shared with a team — and reached the terminal
+// exactly as stored: an escape byte recoloured the line a user reads when sync
+// is misbehaving, a carriage return rewound it, and a 300-character host filled
+// three lines of one `deja sync` run (#1808). The same bound the listing
+// surfaces have used since #1090.
+//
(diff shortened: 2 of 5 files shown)


vshulcz:
Second commit, from walking every host surface myself: sync_ssh.go:204 and :243 print the peer's own stdout raw, so a machine on the other end of the ssh could recolour and rewind this terminal. Same class, wider blast radius than the host name — de-escaped through SafeText and capped at 2000 bytes, with `TestRemoteOutputCannotRewriteTheLocalScreen` and a control that an ordinary report passes through untouched. The 'synced with %s, but could not record it' line took the host bound too.

PR #1811, opened 2026-08-25
fix(remember): clean and bound a tag where it is written, and where it is echoed

### LINKED ISSUE #1810: remember echoes tags raw and stores a tag with an escape byte in it
`deja remember --tag` echoes tags back exactly as typed, and `NormalizeTags` lets a control byte through into what is stored.

## Reproduction

```
$ deja remember "the shard limit is 64" --project proj \
    --tag ok --tag "$(printf 'red\033[31mALERT\033[0m\rrewound')" --tag "$(python3 -c "print('w'*400)")"
deja: remembered under proj #ok #red^[[31malert^[[0m^Mrewound #wwww…(400 characters)
```

The line is recoloured and then rewound by the `\r`, and the 400-character tag prints whole. The project name beside it is bounded — that was #1792; the tags next to it were not.

What is stored keeps the escape too:

```
stored tags: ['ok', 'red\x1b[31malert\x1b[0m\rrewound', 'wwwwwwwww…(400)']
```

Control that the rest of the pipeline is fine: `deja recall "shard limit"` prints the same note as `#ok #redalert rewound #wwww… …` — the snippet path de-escapes and truncates already. It is the write path and the confirmation line that do neither.

## Cause

- `cmd/deja/remember.go:91` — `suffix = " #" + strings.Join(norm, " #")`, printed raw.
- `internal/sources/notes.go:62` — `NormalizeTags` lowercases, trims `#`, drops empties and dupes, and caps the *count* at 8. Nothing bounds one tag's length and nothing removes control characters, so they reach `notes.jsonl` and every surface that reads it.

## Impact

Cosmetic on the confirmation line, the same class as #1792 and #1808. The stored value matters more: a tag is meant to be a short handle you can search for, and `#red\x1b[31malert` is neither short nor searchable as typed.

Bounding a tag's length changes what is written, which is a product call — flagged rather than assumed. Stripping control characters at the same place is not: a tag with an escape byte in it is not a tag anyone meant to write.

vshulcz:
Closes #1810.

Two halves of the same gap. `NormalizeTags` capped the *count* at 8 and nothing else: a tag carrying an escape byte went into `notes.jsonl` as typed, and a 400-character tag was stored whole. And `deja remember`'s confirmation printed the tag list raw beside a project name that has been bounded since #1792, so the line was recoloured and rewound.

Now: control and format runes are dropped where a tag is written, whitespace folds to `-`, one tag is cut at 64 bytes, and the echo takes the same 80-column bound as the project beside it.

```
deja: remembered under proj #ok #red^[[31malert^[[0m^Mrewound #wwww…(400)
deja: remembered under proj #ok #red-[31malert-[0m-rewound #wwww…(64, then …)
```

Stored: `['ok', 'red-[31malert-[0m-rewound', 'wwww…(64)']`.

Cutting a tag at 64 bytes changes what is written, so say if you want a different number or no bound at all — it is the one part of this that is a product call rather than a mechanism. Dropping control bytes is not: a tag with an escape sequence in it is not a handle anyone meant to file a note under.

Tests: `TestNormalizeTagsDropsControlBytes`, `TestNormalizeTagsBoundsOneTag` (with a control that an ordinary tag is untouched), `TestRememberEchoBoundsTheTagsToo`. Control that the rest was already right: `deja recall` renders the same note de-escaped and truncated, which is why this only shows on the write and confirm paths.

### DIFF
--- a/cmd/deja/remember.go
+++ b/cmd/deja/remember.go
@@ -88,7 +88,10 @@ func runRemember(dir string, args []string) error {
 	}
 	suffix := ""
 	if norm := sources.NormalizeTags(tags); len(norm) > 0 {
-		suffix = " #" + strings.Join(norm, " #")
+		// Cleaned at the source now, but the echo is the line a user checks
+		// what was filed against, and it is one line: bound it the way the
+		// project name beside it is bounded (#1810).
+		suffix = " " + safeForStatusline("#"+strings.Join(norm, " #"), mcpResourceNameMax)
 	}
 	fmt.Fprintf(os.Stdout, "deja: remembered under %s%s\n", projectForEcho(project), suffix)
 	return nil
--- a/cmd/deja/remember_echo_test.go
+++ b/cmd/deja/remember_echo_test.go
@@ -94,3 +94,34 @@ func TestRememberEchoNamesAProjectThatBoundsToNothing(t *testing.T) {
 		t.Errorf("the line does not say why the name is missing: %q", line)
 	}
 }
+
+// The project name has been bounded since #1792; the tags printed beside it
+// were not, so a tag with an escape byte in it still rewrote the line (#1810).
+func TestRememberEchoBoundsTheTagsToo(t *testing.T) {
+	t.Setenv("DEJA_NOTES_FILE", filepath.Join(t.TempDir(), "notes.jsonl"))
+	t.Setenv("DEJA_INDEX_DIR", filepath.Join(t.TempDir(), "index.db"))
+	out := captureStdout(t, func() {
+		err := runRemember(index.DefaultDir(), []string{
+			"--project", "proj",
+			"--tag", "ok",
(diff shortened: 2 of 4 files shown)


vshulcz:
Two things measured while the review runs:

- Multibyte tags survive the 64-byte cut as valid text: 32 Cyrillic runes, 21 CJK runes, `utf8.ValidString` true in both.
- Folding whitespace to `-` merges `retry budget` and `retry-budget` into one handle, and the dedupe then keeps one. That is deliberate and now pinned by `TestATagWithASpaceBecomesOneHandle`: tags are folded into `#tag` tokens in the indexed text, so a tag with a space in it never matched as one thing anyway.

Also checked the read path, which was the risk worth naming: `NormalizeTags` is called on the write side and in the promote dedupe comparison only — nothing on the search path normalises, so notes written before this change are still found by the tag they were stored with. The one visible effect on old data is that re-promoting a note whose stored tag was odd no longer compares equal and appends once more.

PR #1813, opened 2026-08-25
fix(remember): keep checking for duplicates past an oversized note

### LINKED ISSUE #1812: one oversized note turns off the duplicate check for every note written after it
`remember` refuses a note it already holds (#1736). One oversized note in the file turns that check off for every note after it.

## Reproduction

Over MCP, because argv cannot carry the payload. Two identical 5.5 MB notes:

```
id 2 'Remembered under proj.'
id 3 'Remembered under proj.'
notes.jsonl: 11040128 bytes, 2 lines
```

Both stored. That much is documented — `maxNoteLine` says a note longer than 4 MB "is written, and the duplicate check simply does not claim to cover it".

This is not. One 5.5 MB note, then the same small note twice:

```
id 2 'Remembered under proj.'      # the big one
id 3 'Remembered under proj.'      # small note
id 4 'Remembered under proj.'      # the same small note again — accepted
```

Control, the same three calls without the big note first:

```
id 2 'Remembered under proj.'
id 3 'Already remembered under proj.'
```

So the dedupe works until a long line exists, and then stops working for the whole store.

## Cause

`internal/sources/notes.go:127` — `noteAlreadyStored` scans with

```go
sc := bufio.NewScanner(f)
sc.Buffer(make([]byte, 0, 64*1024), maxNoteLine)
for sc.Scan() { ... }
```

`bufio.Scanner` stops at the first line longer than its buffer and reports `bufio.ErrTooLong` from `sc.Err()`, which nothing reads. The loop ends there, so every line after the oversized one is never compared — and notes.jsonl is append-only, so the oversized note is permanently in front of everything written later.

## Impact

Repeated facts are what #1736 was about: one fact re-remembered costs a line of every later recall. After one big note, nothing stops that happening again, and the store keeps growing with copies. The user sees "Remembered under proj." each time, so nothing signals it.

vshulcz:
Closes #1812.

`bufio.Scanner` stops at the first line longer than its buffer, and `sc.Err()` was never read — so one note past `maxNoteLine` ended the duplicate scan, and since notes.jsonl is append-only, every note written after it was compared against nothing. The user kept seeing "Remembered under proj." while the store filled with copies.

The scan now skips a line past the cap and carries on. A note that long is still not compared — that part is what the cap says — and it is still written whole.

```
id 2 'Remembered under proj.'      # 5.5 MB note
id 3 'Remembered under proj.'      # small note
id 4 'Remembered under proj.'      # the same small note — accepted (before)
id 4 'Already remembered under proj.'                          # (after)
```

Test: `TestOneHugeNoteDoesNotDisableTheDuplicateCheck`, with a control that the same duplicate is refused without the oversized note in front, and an assertion that the big note itself is still written whole.

Not fixed here: nothing bounds how big one note may be, which is the open product call in the queue (a single `remember` accepting 3 MB, iteration 79). This only stops one such note from disabling the check for everything else.

### DIFF
--- a/internal/sources/note_dedupe_test.go
+++ b/internal/sources/note_dedupe_test.go
@@ -52,3 +52,49 @@ func TestOneHugeNoteDoesNotDisableTheDuplicateCheck(t *testing.T) {
 		t.Errorf("the oversized note was not written whole: file is %d bytes", len(body))
 	}
 }
+
+// The cap is on a note's content: a line of exactly maxNoteLine bytes is still
+// compared, and one byte more is skipped. Counting the newline made the
+// boundary one byte tight.
+func TestTheLineCapCountsTheNoteNotItsNewline(t *testing.T) {
+	for _, tc := range []struct {
+		name    string
+		fill    int
+		compare bool
+	}{
+		{"one under", -1, true},
+		{"exactly the cap", 0, true},
+		{"one over", 1, false},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			path := filepath.Join(t.TempDir(), "notes.jsonl")
+			t.Setenv("DEJA_NOTES_FILE", path)
+			// A note whose JSON line lands exactly on the cap.
+			body := strings.Repeat("z", 64)
+			if err := AppendNote("proj", body, time.Now()); err != nil {
+				t.Fatal(err)
+			}
+			line, err := os.ReadFile(path)
+			if err != nil {
+				t.Fatal(err)
+			}
+			pad := maxNoteLine - len(strings.TrimRight(string(line), "\n")) + tc.fill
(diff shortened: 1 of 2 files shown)


vshulcz:
Second commit, from walking the reader's boundaries myself before the review lands: the new cap counted the terminating newline, so a line of exactly `maxNoteLine` bytes was skipped where the old scanner compared it. Table run over the parsing paths — no trailing newline, empty file, only newlines, an over-long final line, over-long then short, exactly the cap, one over — now reads:

```
exactly max   -> compared
max plus one  -> skipped
long then short -> long skipped, short compared
long last line  -> earlier lines compared, long one skipped
```

`TestTheLineCapCountsTheNoteNotItsNewline` pins all three boundary cases.

PR #1818, opened 2026-08-25
test(view): pin that a note's tag cannot close the page's script block

### LINKED ISSUE #1817: the view page's notes array has no test that a tag cannot close its script block
`deja view` writes a page carrying two arrays: sessions and promoted notes. Session text has been pinned against closing the page's own `<script>` block since the view landed (`cmd/deja/view_test.go`). The notes array has no such test, and a note's project, title, text and tags are all text the user typed — cleaned at write time only since #1811, so a store can still hold anything written before that.

## Measured

A promoted note stored the old way, with tags `['ok', 'red<ESC>[31malert<ESC>[0m', '</script><script>alert(1)</script>', 'w'*400]`, then `deja view`:

```
page bytes: 8464
raw escape byte present: False
literal </script> inside a script block: False
unescaped alert(1) present: True        # inside a JSON string, see below
400-char tag present whole: True
tags json: ["ok","red[31malert[0m","</script><script>alert(1)</script>","wwww…"]
```

So the injection is inert: `json.Marshal` escapes `<` and `>` to `<`/`>`, and `jsonForScript` rewrites `</` to `<\/` on top of that. Nothing to fix in the page.

What is missing is the test. Removing either layer alone still passes — the page survives on the other one — so a future change that drops both (switching to a different embedding, or marshalling with an encoder that has HTML escaping off) would go unnoticed on this array while the session array stays covered.

Two cosmetic leftovers, recorded rather than fixed: an escape byte in an old tag still reaches the page as `` (invisible in HTML, no terminal to recolour), and a 400-character tag renders whole and stretches its row.

vshulcz:
Closes #1817.

No behaviour change — a gap in what is pinned. The page's session text has been tested against closing its own `<script>` block since the view landed; the promoted-notes array beside it had no such test, and a note written before #1811 can still hold a tag with anything in it.

Measured first, on a page built from a note whose tags are `['ok', 'red<ESC>[31malert<ESC>[0m', '</script><script>alert(1)</script>', 'w'*400]`:

- no raw escape byte reaches the page
- no literal `</script>` inside the script block; `alert(1)` appears only inside a JSON string as `</script>…`
- the 400-character tag renders whole (cosmetic, and only for notes written before the write path was cleaned)

So nothing is broken today: `json.Marshal` escapes `<`/`>` and `jsonForScript` rewrites `</` to `<\/`. The test holds both layers. Control: removing either one alone still passes — the page survives on the other — and removing both fails the test, which is what makes it worth having.

### DIFF
--- a/cmd/deja/view_notes_escaping_test.go
+++ b/cmd/deja/view_notes_escaping_test.go
@@ -10,11 +10,29 @@ import (
 	"github.com/vshulcz/deja-vu/internal/index"
 )
 
-// The page carries promoted notes in their own array beside the sessions, and
-// a note's project, title, text and tags are all text the user typed. Session
-// text has been pinned against closing the page's script block since the view
-// landed; the notes array had no such test, and notes written before #1811 can
-// still hold a tag with anything in it.
+// notesPayload returns the page's notes array as it was written into the
+// script, before any JSON decoding — the notes are their own array beside the
+// sessions, and reading the whole page instead would be satisfied by the same
+// text appearing in a session preview.
+func notesPayload(t *testing.T, page string) string {
+	t.Helper()
+	const marker = ",N="
+	i := strings.Index(page, marker)
+	if i < 0 {
+		t.Fatalf("the view page no longer embeds a notes array; this test reads the wrong thing now")
+	}
+	rest := page[i+len(marker):]
+	if j := strings.Index(rest, "];"); j >= 0 {
+		return rest[:j+1]
+	}
+	return rest
+}
+
+// The page carries promoted notes in their own array, and a note's project,
+// title, text and tags are all text the user typed. Session text has been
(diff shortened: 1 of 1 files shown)


vshulcz:
Reviewer was right that the control was vacuous, and chasing it turned up a second reason: the note's tags also appear in the *session* preview, so reading the whole page could not tell the notes array apart from the sessions array either way.

Second commit reads the `N=` payload itself and asserts on it. Four mutations against the reworked test:

```
drop json.Marshal's HTML escaping only:  pass   (page still safe on the other layer)
drop jsonForScript's </ rewrite only:    pass   (page still safe on the other layer)
drop both:                               FAIL
notes never reach the page:              FAIL
```

The last one is the control the first version was missing: with `LoadPromotedNotes` stubbed out, the test now fails instead of passing on the template's own closing tag.

Issue #1794, opened 2026-08-25
the tombstone hint from remember prints a session id with escape bytes in it


vshulcz:
The tombstone hint from `deja remember` prints a session id built from the project name, and an escape byte in that name reaches the terminal:

```
$ deja remember "first thing" --project "$(printf 'p\033[31mX')"
$ deja forget --session "deja-2026-08-25-$(printf 'p\033[31mX')"
$ deja remember "second thing" --project "$(printf 'p\033[31mX')" 2>&1 | cat -v
deja: it is written but stays hidden — deja-2026-08-25-p^[[31mX was forgotten, and its tombstone lifts only through `deja forget --unforget deja:deja-2026-08-25-p^[[31mX`
deja: remembered under p [31mX
```

The confirmation on the last line is bounded since #1792; the hint above it is not.

The tension is why it was left out of that fix: the id is not decoration, it is the argument of a command the reader is meant to paste. `internal/sources/notes.go:272` builds it as `"deja-" + day + "-" + project` from the raw name, so a sanitised id names nothing and the paste fails. Bounding it the way the echo is bounded would print a command that does not work.

Options, none of them free:

- Print the id through `%q`. Control bytes become visible escapes, but the quoted form is Go's, not the shell's, so pasting still does not reproduce the bytes.
- Neutralise for display and drop the paste — say which note is hidden and point at `deja forget --list` instead of handing a command.
- Leave the raw id and accept that a name with escapes in it recolours one line of one rare path.

A project name with an escape byte in it is a user typing something odd, not an attack surface — impact is one recoloured line. Filing it so the choice is recorded rather than rediscovered.

Follow-on from #1792.

vshulcz:
Fixed in #2763. The id in the hint is quoted the way a shell pastes it: ANSI-C quoting when it carries anything a terminal acts on — the same predicate safeForStatusline strips by, so both halves of the line agree — single quotes for anything a shell acts on, bare otherwise, and the line names the shells that read the quoted form.

The same quoting covers forget --list's unforget line and promote's corrections line; the restored line is prose and is neutralised instead. The rest of the printed commands that still interpolate raw are #2768.

PR #1820, opened 2026-08-25
fix(sync): bound the peer's output where the failure is turned into an error

### LINKED ISSUE #1819: a failing sync puts the peer's raw output in the error: escape bytes reach the terminal and peers.json
#1809 bounded the peer's output on the lines that print it. The failure paths build it into an error instead, and that error is printed by `main` exactly as it is:

```go
return fmt.Errorf("remote import: %v: %s", err, out)   // cmd/deja/sync_ssh.go:198
return fmt.Errorf("remote export: %v: %s", err, out)   // :240
```

```go
fmt.Fprintln(os.Stderr, "deja:", rebuildWindowError(err))   // cmd/deja/main.go:40
```

## Measured

A stubbed remote that fails and writes `"deja: import failed\x1b[31m\rHACKED " + "padding "*200`:

```
error length: 1661
escape in error: true, carriage return: true
stored LastError: 163 bytes, escape=true cr=false
```

So on `deja sync ssh <host>` the whole 1661 bytes reach the terminal with the escape and the carriage return intact — the `\r` rewinds the line, which is the same effect #1808 was about, with the text coming from the machine on the other end. The value also lands in `peers.json` as `last_error`: `trimError` collapses whitespace (which is why the `\r` is gone there) and cuts at 160 runes, but an escape byte is not whitespace and survives.

Control: `deja sync` over all peers does not show this — `runSyncAll` runs the same text through `safeForStatusline` since #1809 — and `deja doctor` bounds the stored value on the way out. It is the single-host path and the stored value that have nothing.

## Cause and fix

Bound the remote's output where the error is built, not where it is printed: `remoteOutputForEcho` already exists for exactly this text (de-escape via `search.SafeText`, cap at 2000 bytes) and is what the printing paths use. Doing it at the source covers the terminal, the stored `last_error`, and anything else that later reads the error.

vshulcz:
Closes #1819.

#1809 bounded the peer's output where it is printed. The failure paths build it into an error instead — `fmt.Errorf("remote import: %v: %s", err, out)` — and `main` prints an error exactly as it is, so `deja sync ssh <host>` against a failing peer still put 1661 bytes of remote text on this terminal with the escape byte and the carriage return intact, and the same text landed in `peers.json` as `last_error` (`trimError` drops the `\r` as whitespace, but an escape byte is not whitespace).

Bounded at the source instead of at each printer: both error paths now wrap `remoteOutputForEcho(out)`, which is the same de-escape-and-cap the printing paths use. That covers the terminal, the stored value, and anything else that reads the error later.

Test: `TestAFailingRemoteCannotWriteOnThisTerminal` fails before — on both the error text and `peers.json` — with a control that the remote's own report ("import failed") survives the bound, so the diagnosis is not thrown away.

### DIFF
--- a/cmd/deja/sync_remote_error_test.go
+++ b/cmd/deja/sync_remote_error_test.go
@@ -1,0 +1,82 @@
+package main
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/vshulcz/deja-vu/internal/index"
+	"github.com/vshulcz/deja-vu/internal/peers"
+)
+
+// failingRemote points sshRunner at a peer that writes hostile output and then
+// fails, and returns the error the sync produced.
+func failingRemote(t *testing.T, out string) error {
+	t.Helper()
+	tmp := hermeticEnv(t)
+	t.Setenv("DEJA_PEERS_FILE", filepath.Join(tmp, "peers.json"))
+	root := filepath.Join(os.Getenv("DEJA_CLAUDE_ROOT"), "-proj")
+	if err := os.MkdirAll(root, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	rec := `{"type":"user","sessionId":"v1","cwd":"/proj","timestamp":"2026-08-22T01:00:00Z","message":{"role":"user","content":"the retry budget question"}}` + "\n"
+	if err := os.WriteFile(filepath.Join(root, "a.jsonl"), []byte(rec), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	dir := index.DefaultDir()
+	if err := index.Ensure(dir, "", true, nil); err != nil {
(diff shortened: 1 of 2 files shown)


vshulcz:
Reviewer's two leftovers are taken: the scp failures at sync_ssh.go:191 and :261 now bound their output the same way, with `TestAFailingScpCannotWriteOnThisTerminalEither` covering the first one.

One deliberate exception in that second sentence: the host stays as written, because the line hands over `deja sync ssh <host> --pull --full` to paste and a bounded name would name no machine. Same tension as the tombstone id in #1794, and noted in the code.

Their other findings match what I measured: SafeText maps the escape and the `\r` to spaces and leaves "import failed" intact; `strings.Contains(out, "exported 0 records")` at :248 reads the raw output, so the bound cannot break it; and the double bound is by design — the terminal gets the full error, `peers.json` keeps its 160-rune summary.

PR #1822, opened 2026-08-25
fix(doctor): line the Sync column up by what the terminal draws

### LINKED ISSUE #1821: the doctor's Sync column is padded by runes, so a long or CJK peer name breaks the alignment
The doctor's Sync section formats each peer as `"  %-12s %s"`. That column holds only for ASCII names of twelve characters or fewer, which is not what the field contains.

## Reproduction

Four peers — an ordinary name, a longer one, a CJK one, and a 300-character one (bounded to 80 columns since #1808):

```
Sync:
  build-server-in-the-other-office last exchange 1 day ago, nothing received yet
  laptop       last exchange 19 hours ago
  wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww… last exchange 3 days ago, nothing…
  数据分片服务器      last exchange 2 days ago, nothing received yet
```

Three different things go wrong:

- the 32-character name gets no padding at all, so its state starts at column 34 while `laptop`'s starts at 15
- `%-12s` pads by runes, so 数据分片服务器 (7 runes, 14 display columns) is padded as if it were 7 columns wide and its state starts two columns further right again
- an 80-column name pushes the state past the width of an ordinary terminal, where it wraps

Control: `deja doctor`'s other columns are fine, because what they hold is deja's own short labels — `doctor_auto.go` prints harness names in the same `%-12s` and they are all under twelve ASCII characters. The peers row is the one that puts a user-supplied value in a column.

Elsewhere in the repo the padding is measured, not assumed: `files.go:280` pads with `strings.Repeat(" ", max(0, col-termwidth.Columns(path)))`.

## Impact

Cosmetic — a first-screen report reads as broken when the columns do not line up, which is the screen someone opens when sync is already misbehaving.

vshulcz:
Closes #1821.

The Sync section's column was `%-12s`, which pads by runes: a 32-character name got no padding, a CJK name was padded as if each character were one column wide, and an 80-column name pushed the state off the screen.

It now pads by what the terminal draws (`termwidth.Columns`, the idiom `files.go:280` already uses), grows the column to the widest name up to 32 columns, and gives a name wider than that a line of its own with the state indented underneath — so nothing is truncated to buy the alignment.

```
  build-server-in-the-other-office last exchange 1 day ago, nothing received yet
  laptop       last exchange 19 hours ago
  数据分片服务器      last exchange 2 days ago, nothing received yet

  build-server-in-the-other-office last exchange 1 day ago, nothing received yet
  laptop                           last exchange 19 hours ago
  wwww…(80 columns)…w
                                   last exchange 3 days ago, nothing received yet
  数据分片服务器                   last exchange 2 days ago, nothing received yet
```

Test: `TestTheSyncSectionLinesUpWhateverThePeerIsCalled` measures where each state begins in terminal columns and fails before at `[35 15 84 22]`, with a control that every name is still on the screen.

### DIFF
--- a/cmd/deja/doctor_peers.go
+++ b/cmd/deja/doctor_peers.go
@@ -3,10 +3,12 @@ package main
 import (
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/vshulcz/deja-vu/internal/index"
 	"github.com/vshulcz/deja-vu/internal/peers"
+	"github.com/vshulcz/deja-vu/internal/termwidth"
 )
 
 // doctorPeers reports the machines this one exchanges memory with, and when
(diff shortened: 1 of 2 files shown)


vshulcz:
Rendered the awkward cases myself while the review runs:

```
no peers            peers        no other machines yet — …
one peer            laptop       last exchange 19 hours ago…
exactly 12          aaaaaaaaaaaa                     last exchange …
exactly 32          bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb last exchange …
33 columns          ccccccccccccccccccccccccccccccccc
                                                     last exchange …
emoji               🚀rocket     last exchange …            (aligned)
nothing printable   a name with no printable characters
                                 last exchange …
```

One case is still off by a column, and it is not this section's doing: a name carrying a combining mark. `termwidth.Columns` counts the mark as one column —

```
"écombining" (decomposed) -> 11 columns   ("écombining" precomposed -> 10)
"a​b"                -> 3 columns    (a zero-width space counted as 1)
"🚀rocket" -> 8, "数据分片" -> 8           (both right)
```

so the padding is computed from a width one too wide and the state starts a column early. That is the same for every caller of `Columns` — the brief, the files screen, the statusline — so it is filed separately rather than patched inside the doctor. Emoji and CJK are correct.

PR #1825, opened 2026-08-25
fix(termwidth): measure a name the same however it was typed

### LINKED ISSUE #1824: the same name measures 10 or 11 columns depending on its normal form, so NFD names misalign every column
`termwidth.Columns` counts a combining mark as a column of its own, so the same name measures differently depending on how it was typed or which filesystem produced it:

```
"écombining" decomposed (e + U+0301)  -> 11 columns
"écombining" precomposed (U+00E9)     -> 10 columns
"über-server" decomposed              -> 12 columns
"🚀rocket" -> 8, "数据分片" -> 8         (both correct)
```

Every column-aligned screen reads that number — the brief, the files list, the statusline, and the doctor's Sync section since #1822 — so a name in NFD is padded a column short and its column starts early. macOS hands paths back in NFD, and project names come from paths, so this is the ordinary case on a Mac rather than an exotic one.

## Not the whole story

`RuneColumns`'s own comment already says over-counting is deliberate: CJK combining marks (U+3099, U+302A) count two because the table finds them there, `が` measures four and draws two, and the reasoning is that over-counting shortens a line while under-counting runs it past the edge. That stays right — a lone combining mark does get a cell in some terminals, so counting Mn/Me as zero would be the unsafe direction.

What is wrong is narrower: the same text in two normal forms should measure the same. Composing before measuring fixes exactly that and nothing else:

```
decomposed é   raw=11 composed=10
precomposed é  raw=10 composed=10
decomposed ü   raw=12 composed=11
zero-width space  raw=3 composed=3    (unchanged, still over-counted)
CJK mark       raw=4 composed=4       (unchanged, still over-counted)
```

`nfcfold.Compose` already exists for the search path and returns the string untouched when it holds no combining mark, so the ordinary case pays a scan and no allocation.

vshulcz:
Closes #1824.

`termwidth.Columns` counted a combining mark as a column of its own, so the same name measured differently depending on how it was typed — and macOS hands paths back decomposed, so on a Mac that is the ordinary case:

```
"über-server" decomposed  -> 12 columns
"über-server" precomposed -> 11 columns
```

Every aligned screen reads that number — the brief, the files list, the statusline, the doctor's Sync column since #1822 — and padded the decomposed spelling a column short.

`Columns` now composes before measuring. That fixes exactly the normal-form difference and leaves the deliberate over-counting alone: a zero-width space still measures 1, `か` + U+3099 still measures 4, and a combining mark that composes with nothing still measures 1. `RuneColumns`'s own comment gives the reason — over-counting shortens a line, under-counting runs it past the edge — so this does not touch that.

Cost, 200k iterations on a 132-byte line:

```
BenchmarkColumnsPlain-11         226.3 ns/op    0 B/op   0 allocs/op
BenchmarkColumnsDecomposed-11    718.6 ns/op  480 B/op   3 allocs/op
```

Plain text — nearly everything these screens print — pays a scan and no allocation, because `nfcfold.Compose` returns the string untouched when it holds no combining mark.

Tests: `TestTheSameNameMeasuresTheSameInEitherNormalForm` fails before (11 vs 10, 12 vs 11, 6 vs 5), `TestOverCountingStaysWhereItWasDeliberate` pins what is intentionally unchanged, `TestPlainTextIsUnchanged` pins the ordinary cases.

### DIFF
--- a/internal/termwidth/cost_test.go
+++ b/internal/termwidth/cost_test.go
@@ -1,0 +1,24 @@
+package termwidth
+
+import (
+	"strings"
+	"testing"
+)
+
+// Columns runs on every aligned line the tool prints, so composing first has to
+// stay free for text that holds no combining mark — which is nearly all of it.
+func BenchmarkColumnsPlain(b *testing.B) {
+	s := strings.Repeat("build-server-in-the-other-office ", 4)
+	b.ReportAllocs()
+	for b.Loop() {
+		_ = Columns(s)
+	}
+}
+
+func BenchmarkColumnsDecomposed(b *testing.B) {
+	s := strings.Repeat("über-server ", 4)
+	b.ReportAllocs()
+	for b.Loop() {
+		_ = Columns(s)
+	}
+}
--- a/internal/termwidth/cut_pairs_test.go
+++ b/internal/termwidth/cut_pairs_test.go
@@ -1,0 +1,26 @@
+package termwidth
+
+import "testing"
+
+// Every caller measures with Columns and then cuts with Cut, so the two have to
+// agree about the same string: `if Columns(s) > max { Cut(s, max) }`.
+func TestCutAgreesWithColumns(t *testing.T) {
+	for _, s := range []string{
+		"über-server-in-the-other-office",
+		"über-server-in-the-other-office",
+		"naïve-project",
+		"数据分片服务器",
+		"🚀rocket-launcher",
+		"plain-ascii-name",
+	} {
+		for _, max := range []int{1, 4, 8, 12, 20} {
+			cut := Cut(s, max)
+			if got := Columns(cut); got > max {
+				t.Errorf("Cut(%q, %d) measures %d columns", s, max, got)
+			}
+			if Columns(s) <= max && cut != s {
+				t.Errorf("Cut(%q, %d) shortened a string that already fits", s, max)
+			}
+		}
+	}
+}
--- a/internal/termwidth/cut_pairs_test.go
+++ b/internal/termwidth/cut_pairs_test.go
@@ -1,6 +1,10 @@
 package termwidth
 
-import "testing"
+import (
+	"testing"
+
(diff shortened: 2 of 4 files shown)


vshulcz:
Checked the Columns/Cut pairing myself, since that is where a mismatch would hurt: every caller does `if Columns(s) > max { Cut(s, max) }`, so what has to hold is that a cut never measures wider than the budget, and that a string which already fits is left alone. Both hold for decomposed and precomposed spellings, CJK, emoji and plain ASCII, at budgets 1, 4, 8, 12 and 20 — `TestCutAgreesWithColumns`, second commit.

`Cut` still counts raw runes, so for a decomposed string it stops a column early rather than late. That is the same direction the package already chose for the zero-width cases: a short line is untidy, a long one runs past the edge.

vshulcz:
Reviewer's first finding is right, and chasing it turned up something worse than an off-by-one: `CutRight` could start a line with a mark whose base it had just cut away —

```
CutRight("и◌̆ндекс", 6) -> "◌̆ндекс"
```

and that mark draws on whatever precedes it, which on these screens is the ellipsis. Third commit composes inside `Cut` and `CutRight` too, so a mark and its base are one rune before anything is counted. `TestACutNeverStartsWithAnOrphanedMark` fails without it (the line above is its output).

One consequence worth naming: a cut now returns composed text, so for NFD input the bytes out differ from the bytes in. These are display paths and the same normalisation the measurement already does, but it is a change in what the function returns, not only in what it counts.

On the second finding — `hasCombining` only looking at U+0300–U+036F — that range does cover the Cyrillic and Greek marks that matter here (U+0306 and U+0301 are both inside it, and both compose correctly in the tests above). What it misses is the extended blocks (U+1AB0–, U+1DC0–) and script-specific marks like U+0483. Filed separately rather than widened here.

Also: the linter rejected a literal U+200B in the test; it is an escape now.

PR #1834, opened 2026-08-25
fix(sync): bound the ssh capture's error like the rest of the sync path

### LINKED ISSUE #1833: sshCapture builds its error from the raw host and the raw remote output
#1819 bounded the peer's output in the errors built by `syncSSHPush` and `syncSSHPull`, and #1820's review caught the two scp sites. `sshCapture` builds one too, from the raw host *and* the raw remote output, and it was missed:

```go
func sshCapture(host, cmd string) (string, error) {
	out, err := sshRunner("ssh", append(sshOpts(), host, cmd)...)
	s := strings.TrimSpace(out)
	if err != nil {
		return "", fmt.Errorf("ssh %s: %v: %s", host, err, s)   // cmd/deja/sync_ssh.go:286
	}
```

## Reproduction

A remote whose `mktemp -d` fails while writing `"mktemp: no space\x1b[31m\rHACKED " + "padding "*300`, against a peer named `peer\x1b[31mX`:

```
length=2459 escape=true cr=true
head="ssh peer\x1b[31mX: exit status 1: mktemp: no space\x1b[31m\rHACKED padding padding padding …"
```

`main` prints an error exactly as it is, so all 2459 bytes reach the terminal with the escape and the rewind intact — the same effect #1819 fixed one function over. This is the first thing `deja sync ssh <host> --pull` does, so it is the most likely remote failure of the three, not the least.

## Fix

The same one: `remoteOutputForEcho` for the output, `hostForEcho` for the name. The host is not part of a pasteable command in this sentence, so unlike the recovery line in #1820 there is nothing to keep raw.

Filed with the sweep the queue asked for. The rest of `fmt.Errorf` in cmd/deja that carries text deja did not write:

- `install_sync_timer.go:76` and `:134` — `launchctl`/`systemctl` output, from this machine's own service manager rather than a remote; recorded, not changed.
- `mcp.go:1322` — `want a number, got %s` echoes the raw JSON value back to the model; protocol text rather than a terminal, and bounded by the frame size.
- `update.go:80` — an HTTP status line from the release host.

vshulcz:
Closes #1833.

`sshCapture` was the one sync error #1819 and #1820 did not reach: it built its message from the raw host and the raw remote output, and `main` prints an error exactly as it is.

```
length=2459 escape=true cr=true
head="ssh peer\x1b[31mX: exit status 1: mktemp: no space\x1b[31m\rHACKED padding padding …"
```

It now uses the same two bounds as the rest of the file — `hostForEcho` for the name, `remoteOutputForEcho` for the output. Nothing here is a command to paste, so unlike the recovery sentence in #1820 the host has no reason to stay raw.

This is the first thing `deja sync ssh <host> --pull` does, so it is the most likely of the remote failures rather than the least.

Test: `TestAFailingSSHCaptureCannotWriteOnThisTerminal` fails before (2459 bytes, escape and rewind present), with a control that the remote's own report — "no space" — survives the bound.

The rest of the sweep the queue asked for, recorded in the issue rather than changed: `install_sync_timer.go` quotes launchctl/systemctl from this machine's own service manager, `mcp.go`'s `want a number, got %s` echoes a JSON value back to a model inside a size-bounded frame, and `update.go` quotes an HTTP status line.

### DIFF
--- a/cmd/deja/sync_remote_error_test.go
+++ b/cmd/deja/sync_remote_error_test.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


vshulcz:
Second commit: the two sibling errors in the same function — `unexpected output` and the TMPDIR one — also named the host raw. Their output was already safe (`%q` quotes an escape byte); the host was not. Bounded now, so every sentence sshCapture can produce is.

vshulcz:
Took the minor you flagged: `sshRunner` folds stderr into its error, and its own comment says what stderr holds — host-key notices and server banners, written by the machine at the other end. That is the root of the sync path, so every error built from it inherited the leak whatever the call site did.

Third commit bounds it. Control: without the bound the error is 3233 bytes with the escape and the rewind; with it, bounded and de-escaped, and "banner" survives.

The fixture is worth a note — the first version redirected only the last command in the shell string, so the banner went to stdout, the runner dropped it, and the test passed against unfixed code. It wraps the whole group now.

Agreed on the three you listed as deliberate, and on the recovery sentences keeping the host raw.

PR #1836, opened 2026-08-25
test(nfcfold): tie the composition table to the block the scan looks for

### LINKED ISSUE #1835: nothing connects the composition table to the block hasCombining scans
The review of #1825 raised `nfcfold.hasCombining` scanning only U+0300–U+036F and asked what happens to marks outside it — U+0483 and the other Cyrillic ones, the extended blocks U+1AB0– and U+1DC0–.

Measured against the table itself:

```
table entries use 29 distinct marks
inside 0300-036F: 29
outside: []
```

So there is nothing to miss: every mark `Compose` can act on is one the scan looks for, and a mark outside the block has no composition entry, so finding it would give `Compose` nothing to do. The narrow scan is correct, not lucky.

What is worth having is the statement of that, because the two are written in different files and nothing connects them. If someone adds an entry keyed on U+0483, `hasCombining` returns false for a string carrying it, `Compose` hands the string straight back, and the new entry silently does nothing — no failure, no signal.

A test that walks the table and asserts each mark is one the scan looks for turns that into a build failure. Control: adding `{0x0413, 0x0483}` to the table makes it report

```
the table composes U+0413 with U+0483, which hasCombining does not look for
```

vshulcz:
Closes #1835.

No behaviour change. The review of #1825 asked what `hasCombining`'s single block misses; the answer, measured against the table, is nothing — all 29 marks it composes are inside U+0300–U+036F, and a mark outside has no entry, so scanning for it would find nothing to do.

The two live in different files with nothing connecting them, though, so an entry added on U+0483 would make `Compose` hand the string straight back and do nothing at all — silently. This walks the table and asserts each mark is one the scan looks for.

Control: adding `{0x0413, 0x0483}` to the table makes it fail with `the table composes U+0413 with U+0483, which hasCombining does not look for`.

### DIFF
--- a/internal/nfcfold/mark_range_test.go
+++ b/internal/nfcfold/mark_range_test.go
@@ -1,0 +1,28 @@
+package nfcfold
+
+import "testing"
+
+// hasCombining scans one block, U+0300–U+036F, and Compose can only act on a
+// mark the table knows. The two have to describe the same set: a table entry
+// keyed on a mark outside that block would never be reached, because the scan
+// returns early and hands the string back untouched.
+//
+// Today every entry is inside it, which is what makes the narrow scan correct
+// rather than lucky. This says so, so that adding U+0483 or a mark from the
+// extended blocks fails here instead of quietly doing nothing (#1835).
+func TestEveryMarkTheTableUsesIsOneTheScanLooksFor(t *testing.T) {
+	seen := 0
+	for pair := range compose {
+		if mark := pair[1]; !hasCombining(string(mark)) {
+			t.Errorf("the table composes %U with %U, which hasCombining does not look for", pair[0], mark)
+		}
+		seen++
+	}
+	if seen == 0 {
+		t.Fatal("the composition table is empty, so this proves nothing")
+	}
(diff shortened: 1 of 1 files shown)


vshulcz:
Measured what this test does *not* say, before it reads as a clean bill of health. Against the Unicode data:

```
canonical two-part decompositions:        1026
  with the mark inside U+0300–U+036F:      827
  covered by deja's table:                 630
  inside the block but not in the table:   197   (Greek dialytika-tonos, ≠ ≮ ≯, …)
  with the mark outside the block:         199   (Arabic hamza/madda U+0653–U+0655, Hebrew points, …)
```

So the scan and the table agree — which is all this test claims — and the table is a partial cover of Unicode, deliberately: the package says "Latin, Greek, Cyrillic". Two different gaps sit behind that number and neither is what this pins:

- 197 sequences whose mark the scan already finds but the table has no entry for
- 199 whose mark the scan does not look for, because the table would not know them anyway

Filing the coverage question separately rather than widening it here; this PR is only the wire between the two files.

Issue #1027, opened 2026-08-06
doctor --json says nothing about the trust policy


vshulcz:
`doctor` prints a block about the trust policy; `doctor --json` has no key for it at all:

```
$ deja doctor
Trust policy:
  search       nothing activates — withholds 5 of 5 indexed sessions
  mcp          nothing activates — withholds 5 of 5 indexed sessions
  auto         nothing activates — withholds 5 of 5 indexed sessions

$ deja doctor --json | python3 -c 'import sys,json;print(list(json.load(sys.stdin)))'
['schema_version', 'stores', 'index', 'mcp', 'sqlite3', 'version']
```

So a script checking a machine's health cannot see that recall is off there — the state every other machine form reports (`policy_withheld` in search, last and stats). Same shape as #999: one command, two answers, and the machine form is the quiet one.

vshulcz:
Verified fixed on current main. doctor --json includes the trust policy (doctor_report.go, collectDoctorPolicy).

PR #1839, opened 2026-08-25
feat(doctor): report the sync state in --json, not only on the screen

### LINKED ISSUE #1838: deja doctor --json has no sync key, so a script cannot see that memory stopped moving between machines
`deja doctor` has a Sync section because "a sync that stops does not announce itself" (`cmd/deja/doctor_peers.go:15`). `deja doctor --json` has no key for it at all, so the one reader that could watch this unattended — a script — cannot see a sync that has stopped.

## Reproduction

A peers file with one machine that has never received anything and whose last attempt failed:

```
$ deja doctor | sed -n '/^Sync:/,/^$/p'
Sync:
  laptop       last exchange 3 days ago, nothing received yet — last attempt failed: ssh laptop: exit status 255
```

```
$ deja doctor --json | python3 -c "import json,sys; print(sorted(json.load(sys.stdin)))"
['index', 'ingest_health', 'mcp', 'policy', 'schema_version', 'sqlite3', 'stores', 'version']
```

No `sync`, no peers, nothing. `doctorReport` (`cmd/deja/doctor_report.go:72`) has ten fields and none of them is the peers list; `runDoctor` encodes the report and returns before `doctorPeers` is ever called (`doctor.go:78`).

Control that the text half is not a special case either: the same command with no peers prints the "no other machines yet" line, and `--json` is identical in both cases.

## Precedent

The same gap was fixed once already, for the policy block, and the comment on the fix says it in as many words (`doctor_report.go:85`):

> The text report has had a block for it since #661 while `--json` had no key at all, so a script could not see that recall is switched off on a machine (#1027).

Sync is the other half of that: whether memory is still moving between machines is exactly the thing a person does not notice by hand.

## What it needs

A `sync` key carrying, per peer, what the text line carries — the host, when it last pushed and pulled, how many sessions arrived from there, and the last error. `docs/json-output.md` is the published contract and `TestDoctorJSONKeysMatchTheDocumentedContract` pins the key set in both directions, so the doc entry lands with the code.

vshulcz:
Closes #1838.

`deja doctor` has a Sync section because a sync that stops does not announce itself. `--json` had no key for it, so the one reader that could watch it unattended saw nothing:

```
$ deja doctor --json | python3 -c "import json,sys; print(sorted(json.load(sys.stdin)))"
['index', 'ingest_health', 'mcp', 'policy', 'schema_version', 'sqlite3', 'stores', 'version']
```

Now:

```json
"sync": {
  "peers": [
    {
      "host": "build-box",
      "last_push": "2026-08-20T10:00:00Z",
      "sessions_from_there": 0,
      "last_error": "ssh build-box: exit status 255"
    },
    {
      "host": "laptop",
      "last_push": "2026-08-22T10:00:00Z",
      "last_pull": "2026-08-22T10:00:00Z",
      "sessions_from_there": 0
    }
  ]
}
```

Same shape as #1027, which added `policy` for the same reason — the comment on that fix says it in as many words.

Decisions worth naming:

- `peers` is never omitted. An empty list means no machines are configured, which a script can tell apart from a deja too old to report at all — the trap #1710 recorded for `stale_stores`.
- `last_push` and `last_pull` are separate keys, each absent when that direction never happened, because they fail apart: a machine that takes what this one sends while sending nothing back is a broken sync that reads as a working one.
- `host` and `last_error` are written elsewhere — a config file, another machine — so they take the same bounds the text report gives them (#1808, #1819).

Tests: `TestTheJSONReportCarriesTheSyncState` and `TestTheJSONReportSaysWhenThereAreNoPeers`, both failing before with `no sync key in the report`. `docs/json-output.md` gains the example and the prose, and the golden and the contract test move with it — `TestDoctorJSONKeysMatchTheDocumentedContract` is what forced the doc entry.

### DIFF
--- a/cmd/deja/doctor_json_contract_test.go
+++ b/cmd/deja/doctor_json_contract_test.go
@@ -51,6 +51,9 @@ func TestDoctorJSONKeysMatchTheDocumentedContract(t *testing.T) {
 		// doctorPolicyReport / doctorPolicyRule
 		"error": true, "activations": true, "ignored": true, "inert": true,
 		"rule": true, "withheld": true,
+		// doctorSyncReport / doctorPeerReport
+		"sync": true, "peers": true, "host": true,
+		"last_push": true, "last_pull": true, "sessions_from_there": true,
 		// index.HarnessIngest
 		"malformed_lines": true, "clipped_messages": true,
 		"failed_files": true, "last_error": true,
--- a/cmd/deja/doctor_report.go
+++ b/cmd/deja/doctor_report.go
@@ -107,8 +107,8 @@ type doctorPeerReport struct {
 	// Sessions is how much of this index came from there, the number the text
 	// line prints as "N sessions from there".
 	Sessions int `json:"sessions_from_there"`
-	// LastError is why the most recent exchange failed, bounded the way the
-	// text report bounds it: the value is written by another machine (#1808).
+	// LastError is why the most recent exchange failed. Bounded, unlike Host:
+	// nothing acts on this string, and a remote can make it arbitrarily long.
 	LastError string `json:"last_error,omitempty"`
 }
 
@@ -119,7 +119,12 @@ func collectDoctorSync(dir string) doctorSyncReport {
 	out := doctorSyncReport{Peers: make([]doctorPeerReport, 0, len(list))}
 	for _, p := range list {
 		row := doctorPeerReport{
-			Host:      hostForEcho(p.Host),
+			// The name as written, not as printed: JSON is read by something
+			// that may act on it — `deja sync ssh <host>` — and a bounded name
+			// names no machine. The encoder escapes a control byte on its own,
+			// which is the reason the text report needs a bound and this does
+			// not.
+			Host:      p.Host,
 			Sessions:  from[p.Host],
 			LastError: safeForStatusline(p.LastError, 200),
 		}
--- a/cmd/deja/doctor_report.go
+++ b/cmd/deja/doctor_report.go
@@ -15,6 +15,7 @@ import (
 	"github.com/vshulcz/deja-vu/internal/index"
 	"github.com/vshulcz/deja-vu/internal/jsonout"
 	"github.com/vshulcz/deja-vu/internal/model"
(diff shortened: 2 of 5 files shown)


vshulcz:
Two of the questions I handed the reviewer, answered on my side while CI ran:

**Schema version.** No bump. The document's own policy says it: "Within a `schema_version`, changes are additive only: new optional fields may appear, but existing field names, types, and meanings stay the same." A new top-level key is exactly that, and `schema_version` stays 2 — which is also what #1027 did when it added `policy`.

**The count.** `sessions_from_there` reads `from[p.Host]` with the raw host; only the rendered `host` is bounded. `ImportedByMachine` keys its map on `meta.From`, the machine name stamped on the session at import, so the lookup and the store agree. Bounding the key instead would have reported 0 for every peer whose name is longer than 80 columns and left the ordinary case working — the kind of thing that only shows up on someone else's machine.

Still open for the review: whether truncating `host` is right for a JSON consumer at all, since a script may want the exact name to act on it. My reasoning for bounding it is that the same value already reaches the terminal bounded, and a consumer that needs the exact string has the peers file — but I am not certain, and it is the one part of this I would change on a second opinion.

vshulcz:
Taking the host finding — you are right, and the repo already said so: `peers_echo.go` states "the name is not altered anywhere it is used as an address", and a script reading this JSON is a caller that may act on it. The encoder escapes a control byte on its own, which is exactly why the text report needs a bound and this does not. Second commit sends the name as written; `last_error` stays bounded, since nothing acts on it and a remote can make it arbitrarily long.

Two new tests, and both controls now bite:

```
host bounded in JSON:   FAIL   (TestTheJSONReportNamesThePeerExactly)
count keyed on display: FAIL   (TestTheJSONReportCountsWhatArrivedFromEachPeer)
```

The second is your untested-value point. It exports a batch under one machine name, imports it under another, and asserts `sessions_from_there` is 1 rather than 0. The machine name in that fixture is deliberately longer than 80 columns — with a short name, keying the count on the displayed host passes by luck, which is what the first version of the test did.

PR #1841, opened 2026-08-25
fix(doctor): say when the peers file cannot be read instead of reporting no machines

### LINKED ISSUE #1840: an unreadable peers file reads as "no other machines yet" on every surface
A peers file that cannot be parsed reads exactly like a machine that has no peers — on both surfaces.

## Reproduction

The file truncated mid-object, which is what a half-finished hand-edit or an interrupted write leaves behind:

```
$ cat peers.json
{"peers":[{"host":"laptop","last_push":"2026-08-24T10:00:00Z"}

$ deja doctor | sed -n '/^Sync:/,/^$/p'
Sync:
  peers        no other machines yet — `deja sync ssh <host>` once, then `deja sync` keeps them all in step

$ deja doctor --json | python3 -c "import sys,json; print(json.load(sys.stdin)['sync'])"
{'peers': []}
```

Same for `{"peers":"not a list"}`. The user is told their sync is fine and empty, while `deja sync` quietly stops walking any machine at all.

Control that the file is otherwise being read: repair the truncation and both surfaces list `laptop` immediately.

## Cause

`internal/peers/peers.go:63` — `Load` returns nil on any read or parse failure, and its comment says where that is meant to surface:

> Any read or parse failure means no peers — a sync must not break because a config file is malformed, and doctor is where that gets reported.

Load's half of that is right and should not change. The other half was never built: doctor has no idea the file failed to parse.

The precedent is the trust policy, which has exactly this shape — `policy.state` is `default`, `active` or `unreadable`, and `unreadable` carries an `error`. Sync deserves the same, in the section that exists because a sync that stops does not announce itself.

## Two smaller things found alongside, recorded not fixed

**The same host twice.** Both rows print, and both appear in the JSON. `peers.Record` updates the first match only, so the second keeps whatever error it had for ever:

```
  laptop       last exchange 21 hours ago, nothing received yet
  laptop       last exchange 235 days ago, nothing received yet — last attempt failed: stale copy
```

**A timestamp in the future.** `last_push: 2099-01-01` prints as "last exchange just now", because the age is negative and everything under a minute is "just now". deja already has `index.StampedAhead` for this shape on the session side (#1753).

vshulcz:
Closes #1840.

A peers file deja cannot parse read exactly like a machine with no peers, on both surfaces, while `deja sync` walked nothing at all:

```
Sync:
  peers        no other machines yet — `deja sync ssh <host>` once, then `deja sync` keeps them all in step

  json: {'peers': []}
```

`peers.Load` treating every failure as "no peers" is right — a malformed config must not stop a sync — and its comment says doctor is where that surfaces. That half was never built. `peers.Problem` is it, and both surfaces now report it, the way `policy.state` has since #1027:

```
Sync:
  peers        /…/peers.json could not be read — unexpected end of JSON input

  json: {'state': 'unreadable', 'error': 'unexpected end of JSON input', 'peers': []}
```

A missing file is not a fault — it is a machine that has never synced — so it still reports `ok` and the "no other machines yet" line.

Tests: `TestAnUnreadablePeersFileIsReportedAsSuch` over a truncated file and a `peers` that is not a list, on both surfaces; `TestAReadablePeersFileReportsOK` as the control, including the absent-file case. `docs/json-output.md`, the golden and the contract test move with it.

Two smaller things found alongside and recorded in the issue rather than fixed here: the same host twice prints twice and `peers.Record` only ever updates the first of them, and a `last_push` dated in the future prints as "last exchange just now" (deja has `index.StampedAhead` for that shape on the session side, #1753).


vshulcz:
Two of the questions I handed the review, answered while it runs.

**Classification**, every shape measured:

```
empty file              unreadable | unexpected end of JSON input
whitespace only         unreadable | unexpected end of JSON input
no peers key ({})       ok         | 0 peers
null peers              ok         | 0 peers
a peer with no host     ok         | 0 peers
top-level array/string/number  unreadable | cannot unmarshal … into peers.file
a directory             unreadable | read …: is a directory
chmod 000               unreadable | open …: permission denied
```

The one worth arguing is the empty file. It reports `unreadable`, and I think that is right: deja never writes an empty peers file — `Record` writes valid JSON or nothing — so an empty one means a write that did not finish, and the reader should look at it rather than be told they have no machines. Pinned by `TestThePeersFileShapesAreClassified`.

**`deja sync` itself**, which is point 4 and the behaviour that started this: it read the same `Load` and told someone with a broken file to "name one once with `deja sync ssh <host>`" — advice they had already followed. Second commit gives it the same distinction, with the absent-file case keeping the invitation. `TestBareSyncSaysThePeersFileIsBrokenRatherThanEmpty`.

vshulcz:
All three taken.

**The double read.** `peers.Snapshot()` answers the list and the reason from one read; `Load` and `Problem` both go through it, and the three callers that report both — the doctor's two surfaces and bare `deja sync` — take the snapshot. The race you describe is gone rather than narrowed: there is no second read to disagree with the first. `TestSnapshotAnswersBothFromOneRead`.

**The asymmetric bound.** Deliberate, and now written down where it happens: `sync.error` is deja's own string — a parse failure or an OS error naming the file — while a peer's `last_error` is written by another machine and can be made arbitrarily long, which is why that one is bounded and this one is not. The encoder escapes either. Same reasoning as the host decision on #1839.

**The missing shapes.** `TestThePeersFileShapesAreClassified` (second commit, before your review landed) covers the empty file, whitespace only, `{}`, `null` peers, a peer with no host, a top-level array and a top-level string. `TestAFileDejaCannotOpenIsReportedToo` adds the two that cannot be written as a file body — permission denied and a directory in its place — and skips as root, where the denied case cannot be built.

On the empty file specifically: it reports `unreadable`, and I would keep it that way. deja never writes an empty peers file, so an empty one means a write that did not finish — telling the reader they have no machines would be the wrong answer.

PR #1843, opened 2026-08-25
docs(termwidth): say that a cut returns composed text, and pin the brief's splice

### LINKED ISSUE #1842: nothing says a cut returns composed text, and one caller splices it into a string it did not compose
Since #1825 `termwidth.Cut` and `CutRight` compose before cutting, so for decomposed input the bytes out are not the bytes in. Nothing documents that, and one caller pastes the result back into a longer string using byte offsets taken from the original.

## Measured

Every caller, checked by call site: `brief.go:384`, `brief.go:553`, `brief.go:587`, `statusline_memory.go:252`, `search.go:1205`, `search.go:1223`. All six append an ellipsis and print. None compares the result with its input, and none uses it as a map key — so nothing is broken today.

The one worth looking at is `fitBriefWhen` (`brief.go:587`), which does

```go
return when[:start+len(in)] + termwidth.Cut(project, keep) + "…" + when[end:]
```

with `start` and `end` computed on `when`. The middle is composed and the two ends are not, so the result mixes normal forms — but the offsets belong to the original string, so the splice stays valid. Measured across decomposed and precomposed project names at four widths: valid UTF-8 every time, no replacement character, and the cut name renders identically either way.

One inconsistency falls out of it, worth naming rather than fixing: when the line is short enough that nothing is cut, `fitBriefWhen` returns the original and a decomposed name reaches the screen as it was stored; when a cut does happen, the same name comes back composed. Both render the same, so this is about bytes rather than about what a reader sees.

## What this asks for

The contract on `Cut`/`CutRight` — they return composed text — written where a caller will read it, and a test holding the splice: valid UTF-8 for a decomposed name at any width, and a cut name coming back composed.

vshulcz:
Closes #1842.

No behaviour change. #1825 made `Cut` and `CutRight` compose before cutting, which is what lets the cut and the measurement agree about where a character ends — and it means the bytes out are not the bytes in for decomposed input. Nothing said so, and `fitBriefWhen` pastes a cut back into a longer string using byte offsets taken from the original.

Checked every caller by call site — `brief.go:384`, `:553`, `:587`, `statusline_memory.go:252`, `search.go:1205`, `:1223`. All six append an ellipsis and print; none compares the result with its input or uses it as a key, so nothing is broken. The splice is safe because the offsets belong to the original string, and this pins it: valid UTF-8 for decomposed and precomposed names at four widths, no replacement character, and a cut name coming back composed.

One thing the test deliberately does not assert: when the line is short enough that nothing is cut, `fitBriefWhen` returns the original and a decomposed name reaches the screen as stored. Both spellings render the same, so this is about bytes rather than about what a reader sees — recorded in #1842 rather than normalised, since normalising everything the brief prints is a different change.

### DIFF
--- a/cmd/deja/brief_nfd_splice_test.go
+++ b/cmd/deja/brief_nfd_splice_test.go
@@ -19,10 +19,21 @@ func TestTheBriefSpliceSurvivesADecomposedProjectName(t *testing.T) {
 		"cafe" + acute + "-deploy-pipeline-project-name-here",
 	} {
 		line := "worked in " + project + " · last worked 3 days ago"
-		for _, room := range []int{20, 30, 40, 60} {
+		// Below 55 the guard keeps the line whole (there would be nothing
+		// readable left of the name); 55 and up are the widths that actually
+		// splice, which is what this test is about.
+		for _, room := range []int{30, 55, 60, 65, 70, 75, 80} {
 			got := fitBriefWhen(line, room)
 			if !utf8.ValidString(got) {
 				t.Errorf("the splice produced invalid UTF-8 at room=%d: %q", room, got)
+			}
+			// The facts after the name are what the line exists for, so the
+			// splice must hand them back whole however the middle was cut.
+			if !strings.HasSuffix(got, " · last worked 3 days ago") {
+				t.Errorf("room=%d: the splice damaged the tail of the line: %q", room, got)
+			}
+			if !strings.HasPrefix(got, "worked in ") {
+				t.Errorf("room=%d: the splice damaged the head of the line: %q", room, got)
 			}
 			if strings.Contains(got, "�") {
 				t.Errorf("the splice produced a replacement character at room=%d: %q", room, got)
--- a/cmd/deja/brief_nfd_splice_test.go
+++ b/cmd/deja/brief_nfd_splice_test.go
@@ -1,0 +1,40 @@
+package main
+
+import (
+	"strings"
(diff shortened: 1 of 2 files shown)


vshulcz:
Instrumented the widths before the review lands, and the first version of this test was nearly vacuous — of 20, 30, 40 and 60, only 60 actually cut:

```
room=10..50  cut=false   (the keep < 6 guard returns the line whole)
room=55..80  cut=true
```

Second commit runs 30, 55, 60, 65, 70, 75, 80 — one width that must return the line untouched, six that splice — and adds two assertions that hold the ends of the line, since the middle is the only part that may change.

Mutations against the widened test:

```
Cut stops composing                       FAIL
splice takes its suffix one byte late     FAIL
splice recomputes the suffix offset from the cut length   ok
```

The third is worth naming: that mutation is not actually wrong — it computes the same slice by a different route — so a test passing against it is correct rather than blind.

vshulcz:
Your one finding is the second commit, which landed before the verdict: the widths are 30, 55, 60, 65, 70, 75, 80 now — one that must return the line whole and six that splice — plus assertions on both ends of the line. Mutation results on that version are in the comment above (no-compose FAILs, a suffix offset one byte late FAILs). Points 3 and 4 match what I found: no call site compares a cut result or keys on it, and the two-spellings case stays recorded rather than fixed, since normalising everything the brief prints is a different change.

PR #1845, opened 2026-08-25
fix(brief): show a project name the same way at every terminal width

### LINKED ISSUE #1844: the brief shows a project name in two spellings depending on the terminal width
`fitBriefWhen` composes the project name when it cuts it and leaves it alone when it does not, so the same project can appear in two spellings in one screen depending on how wide the terminal is.

## Reproduction

A decomposed project name — which is what macOS hands back, since project names come from paths — in a line the brief renders at two widths:

```
room=30   worked in u<U+0308>ber-server-project-with-a-long-name · last worked 3 days ago
room=60   worked in über-server-p… · last worked 3 days ago
```

Both render identically on screen; the bytes differ, and which one a reader gets depends on the width of their terminal. `deja brief` prints several of these lines, and a long project alongside a short one gets the two treatments in the same screen.

## Cause

`cmd/deja/brief.go:569` — `fitBriefWhen` returns `when` unchanged when the line already fits, and only the cut path goes through `termwidth.Cut`, which composes (#1825, #1842).

## What to do

Compose the project name once, before the width check, so the line is the same text at every width. That is one call and it makes the brief's own output consistent with the standard the measurement already uses.

Raised by the review of #1843, which argued it both ways and landed on fixing it: the PR documents that composition is the measurement standard, and a function that applies the standard only on one of its two paths is the odd one out.

vshulcz:
Closes #1844.

`fitBriefWhen` composed the project name when it cut it and left it alone when it did not, so the same project came out in two spellings depending on the width of the terminal:

```
room=30   worked in u<U+0308>ber-server-project-with-a-long-name · last worked 3 days ago
room=60   worked in über-server-p… · last worked 3 days ago
```

Composing once before the width check makes the line the same text at every width. One call, and it puts the function on the same footing as the measurement it uses — `Cut` has composed what it returns since #1825, which is what made the two paths differ in the first place.

Raised by the review of #1843, which argued it both ways: a function whose contract is "make it fit" has no obligation to normalise, but one that applies the standard on only one of its two paths is the odd one out. I recorded it and the reviewer came back for it; they were right.

Test: `TestTheBriefSpliceSurvivesADecomposedProjectName` now asserts one spelling at every width — it fails before at room=30 for both fixtures.

### DIFF
--- a/cmd/deja/brief.go
+++ b/cmd/deja/brief.go
@@ -376,7 +376,10 @@ func trimBriefTitleTo(t string, max int) string {
 		}
 		return r
 	}, t)
-	t = strings.Join(strings.Fields(t), " ")
+	// Composed with the whitespace: the cut composes what it returns, so
+	// without this a title that fits keeps its stored spelling and a cut one
+	// does not — the same title, two ways, depending on the width (#1844).
+	t = nfcfold.Compose(strings.Join(strings.Fields(t), " "))
 	// Columns rather than runes: the budget is what is left of the terminal,
 	// and a Chinese title is one rune and two columns per character, so a
 	// 44-rune cap printed 88 columns and the line the budget exists to fit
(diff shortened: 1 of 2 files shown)


vshulcz:
Second commit widens the fix to the other two helpers that cut — `trimBriefTitleTo` and `fitBriefProject` had the same split, so the brief still showed two spellings elsewhere while this one line was consistent. All three compose now, and `TestEveryBriefHelperShowsOneSpelling` covers the fits and cut paths of each. Controls: reverting any one of the three composes fails a test.

vshulcz:
You read the tree between the two commits — the fitBriefProject and trimBriefTitleTo fixes and TestEveryBriefHelperShowsOneSpelling are commit cd54e596 on the branch, pushed before your verdict landed. `git log --oneline origin/hunt/1844 -2` shows both, and CI is 12/12 on the pushed head.

Your point 1 is the answer I wanted: the offsets survive because `strings.Index` anchors on ASCII (" in " and " · "), which composition cannot move — that is why composing the whole line is safe rather than only the project span. Cost confirmed on your side too: the no-mark path returns early and allocates nothing.

PR #1848, opened 2026-08-25
fix(sync): bound the batch file name an import error reports

### LINKED ISSUE #1847: an import error carries the batch file's name raw, and the name comes from the other machine
Sync errors bound the peer's stdout (#1819, #1833). One layer deeper, `index.Import` builds its errors out of the batch file's *name* — and the name comes from the sending machine, since `scp` copies whatever `*.jsonl` matched on the other side.

## Reproduction

A batch directory holding one file whose name carries an escape byte, a carriage return and 120 padding characters, imported:

```
len=288 escape=true cr=true
msg="nothing was imported from 1 file: deja-sync-\x1b[31mHACK\rrewound-wwwww…w.jsonl line 1 is n…"
```

`deja sync ssh <host> --pull` prints that error through `main`, which prints an error exactly as it is — so the escape recolours the line and the `\r` rewinds it, the same effect the sync path itself no longer has.

Control that this path is otherwise careful about foreign text: `internal/index/sync.go:370` and `:372` already run the *directory* name through `search.SafeLine`, and the metadata inside a batch is sanitised on the way into the index. It is the file name in these three sentences that is not:

- `sync.go:720` / `:722` — `nothing was imported from N files: …`
- `sync.go:797` — `… looks truncated at line N`
- `sync.go:799` — `… line N is not a record deja wrote`

## Impact

A name that long is not a name deja writes — its own batches are `deja-sync-<hash>-<timestamp>.jsonl` — so this only shows up when the other end is odd or hostile. The damage is a rewritten line on the screen someone reads while a sync is failing, which is exactly when they need it to be true.

Found by walking the call graph for #1833 rather than grepping: the sweep for "text deja did not write" had stopped at the sync package's own errors.

vshulcz:
Closes #1847.

The sync path bounds the peer's stdout (#1819, #1833). One layer deeper, `index.Import` built its errors out of the batch file's *name*, and the name comes from the sending machine — scp copies whatever `*.jsonl` matched over there:

```
len=288 escape=true cr=true
msg="nothing was imported from 1 file: deja-sync-\x1b[31mHACK\rrewound-wwwww…w.jsonl line 1 is n…"
```

`main` prints an error as it is, so that reaches the terminal of whoever is watching a sync fail.

Four sentences take the same `search.SafeLine` the directory beside them has taken since it was first printed — the two "nothing was imported from …" and the two per-line ones. An ordinary batch name is untouched, which the second test pins.

Found by walking the call graph rather than grepping: iteration #1833's sweep stopped at the sync package's own errors and never followed `index.Import` out of it.

Tests: `TestAnImportErrorDoesNotHandTheTerminalAPeersFileName` (fails before, with a control that the reader can still tell which file failed) and `TestAnOrdinaryBatchNameIsReportedAsItIs`.

### DIFF
--- a/internal/index/import_name_test.go
+++ b/internal/index/import_name_test.go
@@ -1,0 +1,57 @@
+package index
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// The batch file's name comes from the sending machine — scp copies whatever
+// matched over there — and the import built its errors out of it raw, so a
+// name with an escape byte in it rewrote the screen of whoever was watching a
+// sync fail (#1847).
+func TestAnImportErrorDoesNotHandTheTerminalAPeersFileName(t *testing.T) {
+	dir := t.TempDir()
+	batch := filepath.Join(dir, "batch")
+	if err := os.MkdirAll(batch, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	name := "deja-sync-\x1b[31mHACK\rrewound-" + strings.Repeat("w", 120) + ".jsonl"
+	if err := os.WriteFile(filepath.Join(batch, name), []byte("{not a record}\n"), 0o644); err != nil {
+		t.Skipf("the filesystem refused the name: %v", err)
+	}
+
+	_, err := Import(filepath.Join(dir, "index.db"), batch)
+	if err == nil {
+		t.Fatal("a batch of nonsense imported cleanly, so there is no error to inspect")
+	}
+	msg := err.Error()
+	if strings.ContainsAny(msg, "\x1b\r") {
+		t.Errorf("the error carries an escape or a rewind: %q", msg[:min(len(msg), 120)])
+	}
+	// The control: the reader can still tell which file failed.
+	if !strings.Contains(msg, "deja-sync-") || !strings.Contains(msg, ".jsonl") {
+		t.Errorf("the error no longer names the file at all: %q", msg[:min(len(msg), 120)])
+	}
+}
+
+// An ordinary batch name is untouched, so the bound is not paid by every sync.
+func TestAnOrdinaryBatchNameIsReportedAsItIs(t *testing.T) {
+	dir := t.TempDir()
+	batch := filepath.Join(dir, "batch")
+	if err := os.MkdirAll(batch, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	const name = "deja-sync-9f2a1b-20260825T090000Z.jsonl"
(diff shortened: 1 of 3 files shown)


vshulcz:
Two more commits, from working the call graph further while the review runs.

`skippedError` joined the entries with `"; "` and sanitised the join, so one file whose name carried that separator read as three:

```
nothing was imported from 1 file: one.jsonl; two.jsonl; three.jsonl line 1 is not a record …
```

The count said one, the list showed three. The name is quoted where each sentence is built now — quoting the joined list instead would have wrapped name *and* reason in one pair of quotes, which reads worse than the problem. `TestOneFileCannotReadAsSeveral` covers it.

That changes the wording an existing test pinned (`a-broken.jsonl line 2` → `"a-broken.jsonl" line 2`), so `TestImportKeepsGoingPastAFileItCannotRead` moves with it — the intent it was written for, that the error says which file and which line, is unchanged.

vshulcz:
Finished the call-graph walk on my side while waiting: within `internal/index/sync.go` the only text from the batch that reaches an error is the file name (now bounded and quoted) and the wrapped `encoding/json` error at :809 — Go's syntax errors quote the offending character, and a type error names our own struct field, so neither carries content from the file. The remaining raw values in that file are `inDir`/`outDir` at :136, :375 and :382, which are the local path the user typed rather than anything a peer wrote; :370 and :372 sanitise the same value, so those three are inconsistent with their neighbours but not a leak.

vshulcz:
Three things, one of which you are right about and two I want to push back on with evidence.

**The failing test** is commit `7b20325e`, pushed before the verdict landed — `TestImportKeepsGoingPastAFileItCannotRead` now expects `"a-broken.jsonl" line 2`. `git log --oneline origin/hunt/1847 -3` shows all three commits, and CI is 12/12 on that head.

**`%q` is not redundant.** It is not there for control characters — `batchName` handles those — it is there because the separator can be part of a name. Measured before the change:

```
nothing was imported from 1 file: one.jsonl; two.jsonl; three.jsonl line 1 is not a record …
```

one file, three entries in the list. That is what you rule out in your third bullet ("the separator cannot forge entries since the count is known independently") — the count being independent is exactly what makes the sentence contradict itself: it says one and shows three. `TestOneFileCannotReadAsSeveral` pins it.

**The `-count=2` failure is not this PR.** It reproduces on the base branch:

```
$ git checkout hunt/2026-08-23 && go test ./internal/index/ -run SessionsWithNothingToIndex -count=2
--- FAIL: TestSessionsWithNothingToIndexAreNotCounted
    empty_sessions_test.go:37: empty transcripts reported = 2, want 1
```

A counter that accumulates across runs rather than a real fault; filed separately.

PR #1851, opened 2026-08-25
fix(index): count the empty transcripts of the last build, not of the process

### LINKED ISSUE #1850: ReportEmptySessions counts since the last read, not since the last build
`ReportEmptySessions` says it "returns how many transcripts held nothing to index **since the last build**, and clears the counter" (`internal/index/ingest.go:707`). It returns everything counted since the process started, or since somebody last read it — whichever is later. Nothing resets it when a build begins.

## Reproduction

```
$ go test ./internal/index/ -run SessionsWithNothingToIndex -count=2
--- FAIL: TestSessionsWithNothingToIndexAreNotCounted
    empty_sessions_test.go:37: empty transcripts reported = 2, want 1
```

The second run reads its own empty transcript plus the one left over from the first. It reproduces on the base branch — found while answering the review of #1848, where it looked at first like that PR had broken the package.

Control that the counter itself works: the same test alone (`-count=1`) reports exactly 1.

## Why the CLI never sees it

One `deja index` run is one process, one build, one read, so the leftover has nowhere to come from. It only shows where several builds share a process — the test binary today, and anything long-running that indexes more than once (the MCP server's warmup, a future daemon).

## Cause

`emptied` (`ingest.go:1612`) is a package-level counter incremented by `dropEmptySessions` and drained only by the reader. A build does not zero it, so "since the last build" is true only if the last read happened to be the last build.

Either the counter resets where a build starts writing sessions, which makes the sentence true, or the doc comment stops promising per-build and the test drains it first. The first is what the callers already assume.

vshulcz:
Closes #1850.

`ReportEmptySessions` promised "since the last build" and delivered "since the last read": nothing zeroed the counter when a build began, so a second build in one process reported its own empty transcripts plus whatever an earlier one had left behind.

```
$ go test ./internal/index/ -run SessionsWithNothingToIndex -count=2
--- FAIL: TestSessionsWithNothingToIndexAreNotCounted
    empty_transcripts reported = 2, want 1
```

That reproduced on the base branch — it surfaced while answering the review of #1848, where at first it looked like that PR had broken the package. `./internal/index -count=2` and `-count=3` are clean now.

Both build paths zero it as they start writing — `rebuildWithTombstones` and `writeSessionsWithSync` — so a caller reads that build's number whether or not anyone read the one before. The CLI never saw this because one `deja index` run is one process, one build, one read; the test binary and the server's warmup are where several builds share a process.

Test: `TestTheEmptyCounterBelongsToTheLastBuild` builds a store with two empty transcripts, deliberately does not read the counter, builds another with one, and expects 1 — it reports 3 before the fix. It also checks a build with none reports none rather than the previous build's number.

### DIFF
--- a/internal/index/empty_counter_test.go
+++ b/internal/index/empty_counter_test.go
@@ -96,3 +96,53 @@ func TestTheCollisionCounterBelongsToTheLastBuild(t *testing.T) {
 		t.Errorf("a build with no collisions reported %d", n)
 	}
 }
+
+// An incremental pass is a build too. Two of them with nobody reading in
+// between used to report the first one's colliding ids alongside the second's
+// (#1850) — the reviewer of #1851 found this path after the two full ones were
+// fixed.
+func TestAnIncrementalPassCountsOnlyItsOwn(t *testing.T) {
+	home := t.TempDir()
+	setHome(t, home)
+	t.Setenv("USERPROFILE", home)
+	root := filepath.Join(home, "claude")
+	t.Setenv("DEJA_CLAUDE_ROOT", root)
+	proj := filepath.Join(root, "-tmp-c")
+	if err := os.MkdirAll(proj, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	line := func(id string) string {
+		return `{"type":"user","sessionId":"` + id + `","cwd":"/tmp/c","timestamp":"2026-08-01T10:00:00Z","message":{"role":"user","content":"pool exhausted"}}` + "\n"
+	}
+	pair := func(id string) {
+		t.Helper()
+		for _, half := range []string{"-one.jsonl", "-two.jsonl"} {
+			if err := os.WriteFile(filepath.Join(proj, id+half), []byte(line(id)), 0o644); err != nil {
+				t.Fatal(err)
+			}
+		}
+	}
+	dir := filepath.Join(home, "idx")
+
+	pair("dupa")
+	pair("dupb")
+	if err := Ensure(dir, "", false, nil); err != nil {
+		t.Fatal(err)
+	}
+	if n := ReportCollisions(); n != 2 {
+		t.Fatalf("the full build found %d colliding ids, want the 2 it was given", n)
+	}
+
+	// Two incremental passes, and nothing reads between them.
+	pair("dupc")
+	if err := Ensure(dir, "", false, nil); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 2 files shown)


vshulcz:
Checked the counter family before the verdict, which is point 3 of what I asked for: `ReportCollisions` has exactly the same flaw. Measured on the base branch, two builds in one process, the second with one colliding pair of its own:

```
second build reports 3 collisions (its own is 1)
```

Second commit resets it beside the empty one in both build paths, with `TestTheCollisionCounterBelongsToTheLastBuild` covering the carry-over and the "a build with none reports none" case.

`lastIngestFiles` is the third package-level value of this shape but not of this kind: it is assigned rather than accumulated, so each build overwrites it — no carry-over to fix.

`./internal/index -count=3` is clean.

vshulcz:
Point 1, answered on my side: every path that reaches these two functions is already under the index lock, so two builds cannot overlap in one process and the reset cannot land mid-count.

- `Ensure` (`ingest.go:161`), `EnsureForSearch`, `EnsureForSearchNoWait` and `EnsureForSearchStale` each take `lockDir`/`tryLockDir` before doing anything.
- The two paths into `rebuildWithTombstones` outside ingest are `Forget` (`privacy.go:470`) and `Unforget` (`privacy.go:612`), and both hold the lock taken at `privacy.go:369` and `:583`.

That is also what makes the MCP case safe: the detached warmup and a foreground command contend for the same lock, so one waits rather than interleaving with the other.

Point 3 is in the second commit — `ReportCollisions` had the same carry-over (3 reported where the build's own was 1) and is reset beside it; `lastIngestFiles` is assigned rather than accumulated, so there is nothing to reset there.

vshulcz:
Your incremental finding is real and is fixed in a third commit. Reproduced first, since a single incremental after a drained read reports its own number and hides it — it takes two passes with nobody reading in between:

```
full build:                 2 collisions (read, so drained)
incremental build:          1 collisions (its own is 1)
two incrementals, one read: 2   ← the last build's own is 1
```

Both incremental paths — `appendIncremental` and the replacement branch in `updateIndex` — now reset both counters where they start, so all four build paths agree. `TestAnIncrementalPassCountsOnlyItsOwn` covers exactly that sequence; it reports 2 before the fix.

`emptied` does not accumulate on the append path (it never drops sessions there), but it is reset alongside anyway: four paths with one rule is easier to keep true than three plus an exception.

PR #1854, opened 2026-08-25
fix(sync): treat one machine named twice as one machine

### LINKED ISSUE #1853: the same host twice in the peers file leaves a row nothing can update, and syncs that machine twice
A peers file holding the same host twice keeps a row nothing can ever update, and makes `deja sync` walk that machine twice.

## Reproduction

```json
{"peers":[
  {"host":"laptop","last_push":"2026-08-24T10:00:00Z"},
  {"host":"laptop","last_push":"2026-01-01T10:00:00Z","last_error":"stale copy"}]}
```

```
Load returns 2 rows
   host="laptop" push=2026-08-24 err=""
   host="laptop" push=2026-01-01 err="stale copy"

after a successful exchange:
   host="laptop" push=2026-08-25 err=""
   host="laptop" push=2026-01-01 err="stale copy"     ← untouched

after a failure:
   host="laptop" push=2026-08-25 err="boom"
   host="laptop" push=2026-01-01 err="stale copy"     ← still untouched
```

`peers.Record` (`internal/peers/peers.go:118`) stops at the first match, so the second row is frozen: it will show "last exchange 235 days ago — last attempt failed: stale copy" for ever, next to the same machine reporting that it synced this morning.

Both rows reach both surfaces — the doctor's Sync section and `doctor --json` since #1839 — and `runSyncAll` iterates the same list, so a bare `deja sync` opens two ssh connections to one machine and reports it twice.

## How a file ends up like that

Not by deja writing it: `Record` appends a host only when no row matches. By hand-editing, by merging two machines' configs, or by restoring a backup over a newer file — the same ways the malformed file in #1840 happens.

## What to decide

Fold duplicates when the file is read, or update every match when an exchange is recorded. Folding covers all three symptoms at once — the frozen row, the doubled listing and the doubled sync — and the next `Record` writes the folded list back, so the file heals itself.

vshulcz:
Closes #1853.

A peers file naming one machine twice produced a row nothing could ever update — `Record` stops at the first match — so the doctor showed a months-old failure beside the same machine reporting a sync this morning, and a bare `deja sync` opened two connections to it.

```
Load returns 2 rows
   host="laptop" push=2026-08-24 err=""
   host="laptop" push=2026-01-01 err="stale copy"
after a failure:
   host="laptop" push=2026-08-25 err="boom"
   host="laptop" push=2026-01-01 err="stale copy"     ← untouched
```

`Load` folds duplicates now, so all three symptoms go at once and the next `Record` writes the folded list back — the file heals itself. Each direction keeps its newest timestamp and the error belongs to whichever exchange happened last: an old failure is not news about a machine that has synced since, and a *newer* failure survives, which the second test covers.

```
Sync:
  laptop       last exchange 23 hours ago, nothing received yet

{'state': 'ok', 'peers': [{'host': 'laptop', 'last_push': '2026-08-24T10:00:00Z', …}]}
```

deja never writes such a file: `Record` appends a host only when no row matches. It comes from a hand-edit, a merged config or a restored backup — the same ways the malformed file in #1840 arrives.

Tests: `TestOneMachineNamedTwiceIsOneMachine`, `TestFoldingKeepsTheNewerFailure`, `TestRecordingLeavesOneRowPerMachine`, and `TestFoldingKeepsDistinctMachinesApart` as the control that two machines stay two.

### DIFF
--- a/internal/peers/duplicate_test.go
+++ b/internal/peers/duplicate_test.go
@@ -1,0 +1,95 @@
+package peers
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+)
+
+// writePeers points deja at a peers file with the given body.
+func writePeers(t *testing.T, body string) string {
+	t.Helper()
+	path := filepath.Join(t.TempDir(), "peers.json")
+	t.Setenv("DEJA_PEERS_FILE", path)
+	if err := os.WriteFile(path, []byte(body), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	return path
+}
+
+// One machine named twice used to be two rows: the second was frozen, because
+// Record stops at the first match, so it showed "failed 235 days ago" beside
+// the same machine reporting a sync this morning — and a bare `deja sync`
+// opened two connections to it (#1853).
+func TestOneMachineNamedTwiceIsOneMachine(t *testing.T) {
+	writePeers(t, `{"peers":[
+	{"host":"laptop","last_push":"2026-08-24T10:00:00Z"},
+	{"host":"laptop","last_push":"2026-01-01T10:00:00Z","last_error":"stale copy"}]}`)
+
+	list := Load()
+	if len(list) != 1 {
+		t.Fatalf("the same host came back %d times: %#v", len(list), list)
+	}
+	// The newer exchange wins, and the error belongs to it — the stale row's
+	// failure is not news about a machine that has synced since.
+	if got := list[0].LastPush.Format("2006-01-02"); got != "2026-08-24" {
+		t.Errorf("the folded row kept the older push: %s", got)
+	}
+	if list[0].LastError != "" {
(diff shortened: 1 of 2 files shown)


vshulcz:
Two of the questions I handed the review, answered while CI ran.

**Identity.** `Load`, `Record` and `Forget` all compare the host exactly after `TrimSpace` — the fold uses the same rule, so `Laptop` and `laptop` stay two peers on every path. That is consistent rather than accidental: changing it would mean `deja sync forget laptop` starting to remove a row the user wrote as `Laptop`, which is a product decision, not part of this fix.

**What else keys on a peer.** `sessions_from_there` looks up `index.ImportedByMachine`, which is keyed on the machine name stamped into the session at import (`meta.From`, from the sender's `DEJA_MACHINE`) rather than on the peers file. Two rows for one host therefore looked up the same key and printed the same count twice; folding removes the duplicate reading rather than changing the number.

CI is 12/12.

PR #1856, opened 2026-08-25
fix(doctor): say when a peer's stamp is ahead of this machine's clock

### LINKED ISSUE #1855: a peer stamped in the future reads as "last exchange just now"
A peer whose last exchange is stamped in the future reads as one that just synced.

## Reproduction

```json
{"peers":[{"host":"laptop","last_push":"2099-01-01T10:00:00Z"}]}
```

```
Sync:
  laptop       last exchange just now, nothing received yet

{'state': 'ok', 'peers': [{'host': 'laptop', 'last_push': '2099-01-01T10:00:00Z', …}]}
```

The age is negative, everything under a minute is "just now" (`cmd/deja/doctor_peers.go:104`), so a stamp seventy years out reads as the healthiest possible state on the screen whose entire purpose is to show that a sync has stopped.

Control that the section is otherwise honest: the same file with `2026-01-01` reads "last exchange 235 days ago".

## Where the stamp comes from

`peers.Record` writes `time.Now()` on this machine, so deja does not produce one — a clock set wrong (or set right, after being wrong) does, and so does a peers file copied from a machine whose clock was ahead. That is the same cause deja already names for sessions: `index.StampedAhead` exists for exactly this, and the recall page says "stamped later than this machine's clock — its date cannot place it against the others" (#1753).

## What it needs

The Sync section saying the same thing rather than flattering the number, and `--json` carrying it too so a script watching machines is not told everything is fine.

vshulcz:
Closes #1855.

A peer whose last exchange is stamped in the future read as the healthiest machine on the screen that exists to show a sync has stopped: the age is negative and everything under a minute is "just now".

```
  laptop       last exchange just now, nothing received yet
```

Now:

```
  desk         last exchange 23 hours ago, nothing received yet
  laptop       last exchange stamped 2099-01-01, later than this machine's clock — one of the two is wrong, nothing received yet

{'host': 'laptop', 'last_push': '2099-01-01T10:00:00Z', 'sessions_from_there': 0, 'stamped_ahead': True}
```

Same rule deja already applies to sessions — `index.StampedAhead`, whose sentence on the recall page says a date ahead of the clock cannot place a session against the others (#1753). The stamp itself is still reported as written: `--json` flags it rather than dropping it, since the reader may need the value to work out which of the two clocks is wrong.

`stamped_ahead` is additive, so `schema_version` stays 2; `docs/json-output.md`, the example and the contract test move with it.

Tests: `TestAPeerStampedAheadIsNotReportedAsJustSynced` over both surfaces, with `TestAnOrdinaryPeerIsNotFlaggedAsAhead` as the control that a normal peer keeps its age and gains no flag.

### DIFF
--- a/cmd/deja/doctor_json_contract_test.go
+++ b/cmd/deja/doctor_json_contract_test.go
@@ -54,6 +54,7 @@ func TestDoctorJSONKeysMatchTheDocumentedContract(t *testing.T) {
 		// doctorSyncReport / doctorPeerReport
 		"sync": true, "peers": true, "host": true,
 		"last_push": true, "last_pull": true, "sessions_from_there": true,
+		"stamped_ahead": true,
 		// index.HarnessIngest
 		"malformed_lines": true, "clipped_messages": true,
 		"failed_files": true, "last_error": true,
--- a/cmd/deja/doctor_peers.go
+++ b/cmd/deja/doctor_peers.go
@@ -75,6 +75,12 @@ func peerLine(p peers.Peer, sessions int, now time.Time) string {
 	switch {
 	case p.Last().IsZero():
 		line = "never exchanged"
+	case index.StampedAhead(p.Last(), now):
+		// The age would be negative, and everything under a minute reads as
+		// "just now" — so the peer that most needs looking at read as the one
+		// that just synced (#1855).
+		line = "last exchange stamped " + p.Last().Local().Format("2006-01-02") +
+			", later than this machine's clock — one of the two is wrong"
 	default:
 		line = "last exchange " + doctorAgo(now.Sub(p.Last()))
 	}
--- a/cmd/deja/doctor_peers.go
+++ b/cmd/deja/doctor_peers.go
@@ -68,6 +68,22 @@ const (
 	doctorPeerColumnMax = 32
 )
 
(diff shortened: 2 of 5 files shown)


vshulcz:
Probed the boundary before the review lands, and the first version was too tight — a stamp a millisecond ahead fired the sentence:

```
stamp -1s   flagged=false   last exchange just now
stamp   0s  flagged=false   last exchange just now
stamp  1ns  flagged=false   last exchange just now
stamp  1ms  flagged=true    last exchange stamped 2026-08-25, later than this machine's clock …
```

A peers file copied from a machine a few hundred milliseconds ahead, or an NTP step between the write and the read, would have put "one of the two clocks is wrong" against a healthy peer. Second commit gives it a minute of slack — which is also the granularity the line already speaks in, since everything under a minute reads as "just now".

```
a second behind / exactly now / 1ms / 30s / 1m ahead   not flagged
2 minutes ahead / 2 days ahead                          flagged
```

`TestTheClockSentenceHasSlackForJitter` pins that table. The session-side rule stays as it is: `index.StampedAhead` is bare `After` because those stamps come from other machines' transcripts, where the signal is hours rather than milliseconds.

PR #1858, opened 2026-08-25
fix(sync): sanitise the directory in every sentence that names it

### LINKED ISSUE #1857: three sync sentences print the directory raw while two beside them sanitise it
`internal/index/sync.go` sanitises the import directory in two sentences and prints it raw in three, and the comment on the sanitised pair says why that matters:

> the raw syscall error went back — carrying the unsanitised path with it, which is the escape sequence this wording exists to defuse.

## Reproduction

Importing from a path whose name carries an escape byte and a carriage return:

```
is-a-file:  escape=true cr=true
  "/…/batch\x1b[31mHACK\rrewound is a file; sync import wants the directory a `sync export` wrote"

unreadable: escape=true cr=true
  "cannot read /…/batch\x1b[31mHACK\rrewound-dir — permission denied; check that directory's permissions …"
```

Both reach the terminal through `main`, which prints an error as it is, so the `\r` rewinds the line.

Control that the neighbours are already careful: the same path through the "no such directory" and "cannot read %s: %s" branches at `sync.go:370` and `:372` comes back with the escape stripped.

## Cause

Three sentences take the value raw:

- `sync.go:375` — `%s is a file; sync import wants the directory a `sync export` wrote`
- `sync.go:382` — `cannot read %s — permission denied; …`
- `sync.go:136` — the export-side twin, `%s is a file; sync export wants a directory to write the batch into`

## Impact

Lower than #1847, which carried a name chosen by another machine: this is the path the user typed, so it takes a deliberately odd argument to hit. But the file's own comment already decided the question for the neighbouring branch, and a reader has to work out which of five sentences about one value follows which rule.

vshulcz:
Closes #1857.

Five sentences in `internal/index/sync.go` name the sync directory; two sanitised it and three did not, and the comment on the sanitised pair says exactly why it matters — "the raw syscall error went back — carrying the unsanitised path with it, which is the escape sequence this wording exists to defuse".

```
is-a-file:  escape=true cr=true
  "/…/batch\x1b[31mHACK\rrewound is a file; sync import wants the directory a `sync export` wrote"
unreadable: escape=true cr=true
  "cannot read /…/batch\x1b[31mHACK\rrewound-dir — permission denied; …"
```

All five take `search.SafeLine` now: the two import branches, and the export-side twin at `sync.go:136`.

Lower stakes than #1847, where the name came from another machine — this is the path the user typed, so it takes a deliberately odd argument to hit. What it costs is a reader having to work out which of five sentences about one value follows which rule.

Test: `TestEverySentenceAboutTheDirectoryIsSafeToPrint` covers a file where the directory should be, a directory deja may not read (skipped as root), and the export-side twin; each asserts no escape, no rewind, and that the sentence still says what happened and names the directory. The first two fail before.

### DIFF
--- a/cmd/deja/sync.go
+++ b/cmd/deja/sync.go
@@ -127,21 +127,21 @@ func runSync(dir string, args []string) error {
 			if p := deniedPath(err); p != "" && !strings.HasPrefix(p, out) &&
 				(errors.Is(err, fs.ErrPermission) || writeFailureReason(err) != err.Error()) {
 				if n > 0 {
-					return fmt.Errorf("%d records are written into %s, but deja could not record that they went: %w; the next export sends them again", n, out, ensureError(dir, err))
+					return fmt.Errorf("%d records are written into %s, but deja could not record that they went: %w; the next export sends them again", n, search.SafeLine(out), ensureError(dir, err))
 				}
 				return ensureError(dir, err)
 			}
 			if parent := filepath.Dir(out); !dirExists(out) && !dirExists(parent) {
-				return fmt.Errorf("cannot write the export into %s — %s is not there; the disk it lives on may have been unmounted", out, parent)
+				return fmt.Errorf("cannot write the export into %s — %s is not there; the disk it lives on may have been unmounted", search.SafeLine(out), search.SafeLine(parent))
 			}
 			if errors.Is(err, fs.ErrPermission) {
-				return fmt.Errorf("cannot write the export into %s — check that directory's permissions, or choose one you can write", out)
+				return fmt.Errorf("cannot write the export into %s — check that directory's permissions, or choose one you can write", search.SafeLine(out))
 			}
 			// A full or vanished disk in the same words as everywhere else
 			// (#906); anything deja does not recognise still comes through
 			// whole rather than being guessed at.
 			if reason := writeFailureReason(err); reason != err.Error() {
-				return fmt.Errorf("cannot write the export into %s — %s", out, reason)
+				return fmt.Errorf("cannot write the export into %s — %s", search.SafeLine(out), reason)
 			}
 			return err
 		}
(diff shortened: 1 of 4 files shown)


vshulcz:
Point 2 answered by finding more of them: the CLI wrapper has its own family. `cmd/deja/sync.go` names the export directory in four error sentences (`:130`, `:135`, `:138`, `:144`) and printed it raw in all of them, while the index package's five are now uniform.

Second commit sanitises those four. The fifth mention, at `:153`, deliberately keeps the path as written — it hands over `deja sync export <dir> --full` to paste, and a collapsed path names no directory. Same tension as the recovery sentence in #1820 and the tombstone id in #1794, and now noted in the code.

`TestTheExportSentencesAreSafeToPrint` covers it, with a fixture that blocks `MkdirAll` by putting a file where the parent would go; it fails without the change.

On your point 1, which I want your verdict on rather than mine: SafeLine collapses whitespace, so a path with a double space no longer round-trips through these sentences. My reasoning for accepting that is that these four are diagnoses rather than commands — the one sentence that *is* a command keeps the path exact — but the argument cuts both ways and I would rather hear it from you.

vshulcz:
Keeping SafeLine, and there is a test that settles it rather than my preference: `TestSyncImportErrorKeepsThePathOnOneLine` (internal/index/sync_path_line_test.go:16) feeds a path holding an escape byte, a newline and a bidi override, and asserts the message contains no newline — the fixture's own second line is `deja: store is clean`, a forged sentence in deja's voice.

`SafeText` keeps newlines and tabs by design, so that is precisely the case it would let back through: the path prints a line of its own that reads as deja speaking. Whitespace precision inside a diagnosis is worth less than that.

So the three sentences stay on SafeLine, alongside the two that already used it and the four in the CLI wrapper. Merged.

PR #1860, opened 2026-08-25
test(doctor): pin what a peer row with neither timestamp means

### LINKED ISSUE #1859: nothing pins what a peer row with neither timestamp means
`doctor --json` reports a peer named once and never reached as a row with neither timestamp:

```
--- doctor
Sync:
  desk                last exchange 23 hours ago, nothing received yet
  named-never-reached never exchanged

--- json
{'host': 'desk', 'last_push': '2026-08-24T10:00:00Z', 'sessions_from_there': 0}
{'host': 'named-never-reached', 'sessions_from_there': 0}
```

The text says "never exchanged" in words; the machine-readable half leaves a consumer to infer it from two absent keys. Nothing tests that inference, and `omitempty` on a meaningful state is the shape #1710 recorded as a trap on the store side, where an omitted zero read as "an older deja that does not report this".

## What is actually true here

The inference is safe, and this is the argument for leaving the shape alone rather than adding a key:

- a deja too old to report any of this has no `sync` key at all, which #1839 made the distinguishing signal — so "row present, both timestamps absent" cannot mean "old deja"
- `sync.peers` is always present, so an empty list already means "no machines configured"
- the two timestamps are independent — `last_pull` alone is a machine that sends and never receives — and collapsing them into one "never" flag would lose that

So the fix is not another key; it is a test saying what the row means, since three surfaces now depend on it and none of them pins it.

vshulcz:
Closes #1859.

No behaviour change. A machine named once and never reached is a row with neither timestamp:

```
Sync:
  named-never-reached never exchanged

{'host': 'named-never-reached', 'sessions_from_there': 0}
```

The text says it in words; the JSON leaves a consumer to read it from two absent keys, and nothing tested that reading — `omitempty` on a meaningful state is the shape #1710 recorded as a trap on the store side.

Here the inference is sound and the shape should stay: a deja too old to report any of this has no `sync` key at all (#1839 made that the signal), `sync.peers` is always present so an empty list already means "no machines configured", and the two directions are independent — a machine that only receives keeps its pull — which a single "never" flag would flatten.

So this is the test rather than another key. Controls: emitting both timestamps unconditionally fails it, and dropping the "never exchanged" sentence from the text fails it.

### DIFF
--- a/cmd/deja/doctor_sync_never_test.go
+++ b/cmd/deja/doctor_sync_never_test.go
@@ -1,0 +1,79 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/vshulcz/deja-vu/internal/index"
+)
+
+// A machine named once and never reached is a row with neither timestamp. Three
+// surfaces depend on reading it that way and none of them said so (#1859): the
+// text line, the JSON row, and a consumer deciding whether an absent key means
+// "never" or "a deja too old to report it".
+func TestAPeerNeverReachedIsRecognisableOnBothSurfaces(t *testing.T) {
+	tmp := hermeticEnv(t)
+	path := filepath.Join(tmp, "peers.json")
+	t.Setenv("DEJA_PEERS_FILE", path)
+	body := `{"peers":[
+	{"host":"named-never-reached"},
+	{"host":"desk","last_push":"2026-08-24T10:00:00Z"},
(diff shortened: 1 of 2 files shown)


vshulcz:
Second commit answers point 4 before the verdict: docs/json-output.md said each timestamp is omitted when that direction never happened, but not what a row carrying neither means. It now says — a machine named once and never reached, still a row, which is what separates it from a deja too old to report peers at all (that one has no `sync` key). The test and the document now say the same thing.

PR #1862, opened 2026-08-25
fix(index): clear the evicted-files counter per build, and catch the next one

### LINKED ISSUE #1861: the evicted-files counter carries across builds too, and nothing stops a fourth
#1850 fixed two package-level counters that accumulated across builds and checked a third, `lastIngestFiles`, which is assigned rather than accumulated. It missed `evicted`, which is neither: it accumulates, and it is read by `ReportEvictedFiles` after a build the same way.

## Reproduction

Three transcripts indexed; two files removed and indexed again with nobody reading the counter; one more removed and indexed again:

```
second eviction reports 3 files (its own is 1)
```

`cmd/deja/main.go:394` reads it to decide what to tell someone whose index came back empty — "nothing to index yet" against "the history that was here has gone" (#1762) — so the number carries a sentence, not just a statistic.

## Why the fix is not the same two lines

`evicted.Add` happens in `updateIndex` *before* the incremental resets #1850 added:

```
updateIndex: compute what went away → evicted.Add(len(removed)) → … → the reset added in #1850
```

so resetting it where the other two are reset would zero the count of the build doing the reporting. It has to be cleared where `updateIndex` starts, and in the rebuild path beside the others.

## The wider point

That is three counters found one at a time across two iterations, each by tripping over it. The package has a shape — `atomic.Int64` read and drained by a `Report*` function — and nothing stops a fourth appearing with the same flaw. A test that walks the package for that shape and asserts each counter is cleared where a build begins turns the next one into a build failure instead of a wrong number on someone's screen.

vshulcz:
Closes #1861.

#1850 fixed two counters that accumulated across builds and cleared a third value that turned out to be assigned rather than accumulated. It missed `evicted`, which accumulates and is drained by `ReportEvictedFiles` the same way:

```
second eviction reports 3 files (its own is 1)
```

`cmd/deja/main.go:394` reads that number to choose a sentence — "nothing to index yet" against "the history that was here has gone" (#1762) — so it is a claim about the user's machine, not a statistic.

The fix is not where the other two went: `evicted.Add` happens inside `updateIndex` *before* the incremental resets #1850 added, so clearing it there would zero the count of the build doing the reporting. It is cleared where `updateIndex` starts, and in the rebuild path beside the others, where nothing evicts but a stale number must not survive.

The second test is what this iteration is actually for: three counters were found one at a time by tripping over them, and the package has a shape — an `atomic.Int64` drained by a `Report*` function. `TestEveryReportedCounterIsClearedByABuild` walks `ingest.go` for that shape and asserts each one is cleared somewhere, so a fourth is a build failure rather than a wrong number on someone's screen. It fails on the base branch naming `evicted`.

`./internal/index` is clean at `-count=2`.

### DIFF
--- a/internal/index/counter_family_test.go
+++ b/internal/index/counter_family_test.go
@@ -1,0 +1,88 @@
+package index
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+// buildWithFiles writes the named transcripts into a fresh store and builds it,
+// returning the project directory and the index directory so the caller can
+// take files away and build again.
+func buildWithFiles(t *testing.T, ids ...string) (proj, dir string) {
+	t.Helper()
+	home := t.TempDir()
+	setHome(t, home)
+	t.Setenv("USERPROFILE", home)
+	root := filepath.Join(home, "claude")
+	t.Setenv("DEJA_CLAUDE_ROOT", root)
+	proj = filepath.Join(root, "-tmp-e")
+	if err := os.MkdirAll(proj, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	for _, id := range ids {
+		line := `{"type":"user","sessionId":"` + id + `","cwd":"/tmp/e","timestamp":"2026-08-01T10:00:00Z","message":{"role":"user","content":"pool exhausted"}}` + "\n"
+		if err := os.WriteFile(filepath.Join(proj, id+".jsonl"), []byte(line), 0o644); err != nil {
+			t.Fatal(err)
+		}
+	}
+	dir = filepath.Join(home, "idx")
+	if err := Ensure(dir, "", false, nil); err != nil {
+		t.Fatal(err)
+	}
+	return proj, dir
+}
+
+// The third counter of the family: it says how many indexed files went away
+// with their store, and the command layer turns that into "the history that was
+// here has gone" rather than "nothing to index yet" (#1762). It accumulated
+// across builds like the other two (#1861).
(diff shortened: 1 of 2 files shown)


vshulcz:
Two of the questions I handed the review, answered while CI ran.

**What a caller sees when a build finds nothing to do.** `emptyIndexReason` (`cmd/deja/main.go:3011`) reads the counter only when the index came back empty, and uses it to pick between "N indexed files went away with the store that held them" and "nothing to index yet". With the reset, a build that evicted nothing reports 0 and gets the second sentence. Before, a stale count from an earlier build in the same process could make it claim files had gone away when this build evicted none — the change removes a false sentence rather than adding one.

**Ordering.** The forced path returns `rebuild(...)` from `updateIndex` before the eviction is computed, so on that path the counter is cleared twice and never added to. On the incremental paths the `Add` happens after this reset and before the ones #1850 put in, which is exactly why this one could not go beside them.

CI is 12/12; `./internal/index` is clean at `-count=2`.

vshulcz:
Ran the brittleness cases myself while the review is out — all three behave:

```
a fourth counter added (drained, never cleared):   FAIL
the reset moved into a test file only:            FAIL
the reset put beside the other two, after the Add: FAIL
```

The second one is the case I most wanted to check: the walk reads `ingest.go` alone, so a reset that lives in a `_test.go` file does not satisfy it — a counter has to be cleared in the build path to pass.

The third is your point 5, and it is the behavioural test that catches it, not the walk: put the reset where the other two live and `TestTheEvictedCounterBelongsToTheLastBuild` fails, because that reset runs after this build has already counted what went away. The two tests cover different halves — one that a counter is cleared at all, one that it is cleared in the right place.

PR #1866, opened 2026-08-25
docs(json): say that a peer stamped ahead gets a minute of slack

### LINKED ISSUE #1865: doctor --json: stamped_ahead is documented as the session rule, but a peer gets a minute of slack
`docs/json-output.md` documents `stamped_ahead` as appearing "only when the newer of the two timestamps is later than this machine's clock … the rule recall applies to sessions stamped ahead", and the field's own comment in `doctor_report.go` says "Same rule as sessions (index.StampedAhead, #1753)".

Neither is what the code does. `peerStampedAhead` adds a minute of slack (#1855), so a peer stamped inside that minute is later than the clock and is not flagged:

```
$ deja doctor --json --offline      # peer's last_push 30s ahead
"peers": [{"host":"laptop","last_push":"2026-08-25T10:33:07Z","sessions_from_there":0}]
$ deja doctor --offline
laptop       last exchange just now, nothing received yet

$ deja doctor --json --offline      # 120s ahead
"peers": [{"host":"laptop","last_push":"…","stamped_ahead":true}]
```

The minute is right — a peers file is written by deja itself from `time.Now()`, so a copy from a machine a few hundred milliseconds ahead, or an NTP step between write and read, should not read as a broken clock. What is wrong is that the documented contract promises the session rule, so someone implementing against the doc disagrees with deja over the whole 0–60s band, and the sentence that says "same rule" is the reason the two would be expected to converge.

Fix: say what the peer rule is, and say why it is not the session rule, in the two places that describe the field. Nothing about the behaviour changes.

vshulcz:
Closes #1865.

`stamped_ahead` was documented as the rule recall applies to a session — anything later than this machine's clock — but the peer surface gives a stamp a minute of slack first (#1855). A peer stamped 30s ahead is therefore not flagged and reads as "last exchange just now", which is what the doc says it will not do:

```
$ deja doctor --json --offline     # last_push 30s ahead
"peers":[{"host":"laptop","last_push":"2026-08-25T10:33:07Z","sessions_from_there":0}]
$ deja doctor --json --offline     # 120s ahead
"peers":[{"host":"laptop","last_push":"…","stamped_ahead":true}]
```

The slack is right and stays. What changes is the paragraph, which now names the minute and says why a peer is not judged the way a session is: a peers file is written by deja itself, a transcript is not. The field's comment in `doctor_report.go` said "same rule as sessions" and now points at `peerStampedAhead`.

Tests: `TestTheDocumentedPeerAheadRuleNamesTheSlack` reads the paragraph and fails on the base branch; `TestThePeerRuleAndTheSessionRuleDifferByTheSlack` pins the band where the two rules disagree, including the assertion that the session rule still has no tolerance — without it the doc would be describing a difference that had quietly gone away.

No behaviour change.

### DIFF
--- a/cmd/deja/doctor_ahead_rule_test.go
+++ b/cmd/deja/doctor_ahead_rule_test.go
@@ -1,0 +1,56 @@
+package main
+
+import (
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/vshulcz/deja-vu/internal/index"
+)
+
+// A peer's stamps are written by deja itself, so a copy from a machine a few
+// hundred milliseconds ahead is not a broken clock; a session's stamps come
+// from a transcript on some other machine, where anything ahead is unusable.
+// Two rules, on purpose (#1855, #1753) — this pins the band where they differ,
+// so a change to either is a change to a test rather than a silent one.
+func TestThePeerRuleAndTheSessionRuleDifferByTheSlack(t *testing.T) {
+	now := time.Date(2026, 8, 25, 12, 0, 0, 0, time.UTC)
+	inside := now.Add(peerClockSlack / 2)
+	outside := now.Add(peerClockSlack + time.Second)
+
+	if peerStampedAhead(inside, now) {
+		t.Errorf("a peer stamped inside the slack is flagged, so a clock a moment out reads as broken")
+	}
+	if !peerStampedAhead(outside, now) {
+		t.Errorf("a peer stamped past the slack is not flagged")
+	}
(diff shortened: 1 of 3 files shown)


vshulcz:
Took the wording point — the sentence now reads "The minute is what makes this different from the rule for a session, where anything ahead counts".

Left "a peers file" as it was: the file is deja's own `peers.json` listing every machine, not a file belonging to one peer, and that is how the rest of the page and the tests already name it (`json-output.md:354`, `doctor_peers_unreadable_test.go:15`).

PR #1870, opened 2026-08-25
docs(json): the peer host is reported as written, not bounded

### LINKED ISSUE #1869: docs: doctor --json says the peer host is bounded; it is reported verbatim
`docs/json-output.md` says of the sync rows:

> Both `host` and `last_error` are written elsewhere — a config file, another machine — so they are bounded and stripped of control characters before they are reported.

Only `last_error` is. `host` is reported exactly as written, on purpose: the review of #1839 settled that a name used as an address is never altered — `deja sync ssh <host>` has to work on what the JSON hands back, and a bounded name names no machine. The JSON encoder escapes a control byte on its own, which is why the text report needs a bound and this does not.

```
$ deja doctor --json --offline      # peers.json holds a 409-char host with ESC and BEL in it
host len    409 has BEL/ESC: True True
error len   201
```

So a consumer reading the document allocates for a bounded name, or believes a control byte cannot reach it — the two things the sentence promises.

This is the drift the Sync section invites: it has grown three times in five iterations and nothing checks its prose against the values deja emits, the way the contract test checks its key names. Fix the sentence, and pin the two halves it describes — a host through verbatim, an error bounded — against the emitted JSON.

vshulcz:
Closes #1869.

The Sync section said "Both `host` and `last_error` … are bounded and stripped of control characters before they are reported". Only the error is. The host is reported exactly as the config file spells it — that was the review's call on #1839: a name deja hands back for `deja sync ssh <host>` is not a name once it is truncated, and the JSON encoder escapes a control byte on its own.

```
$ deja doctor --json --offline     # peers.json holds a 409-char host with ESC in it
host len    409
error len   201
```

The paragraph now says which of the two is bounded and why the other is not.

The test is the part worth reading: `TestTheSyncSectionDescribesWhatTheRowsCarry` puts an oversized host and an oversized error through `doctor --json` and asserts the host arrives byte for byte, the error is cut, and no raw escape reaches the output — then checks the paragraph does not still make the old claim. The section has grown three times in five iterations with nothing checking its prose against emitted values, only its key names.

No behaviour change.

### DIFF
--- a/cmd/deja/doctor_sync_prose_test.go
+++ b/cmd/deja/doctor_sync_prose_test.go
@@ -1,0 +1,78 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/index"
+)
+
+// The Sync section of docs/json-output.md described both strings a peer row
+// carries as bounded. Only the error is: a host is a name deja is handed back
+// to act on — `deja sync ssh <host>` — and a bounded name names no machine
+// (#1839). The encoder escapes a control byte on its own, which is why the text
+// report needs a bound and this does not.
+//
+// This pins the two halves the paragraph describes against what deja emits,
+// which is the check the section has not had while it grew (#1869).
+func TestTheSyncSectionDescribesWhatTheRowsCarry(t *testing.T) {
+	tmp := hermeticEnv(t)
+	path := filepath.Join(tmp, "peers.json")
+	t.Setenv("DEJA_PEERS_FILE", path)
+	host := "box" + strings.Repeat("y", 400) + "\x1b[31m"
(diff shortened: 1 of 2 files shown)


vshulcz:
Took the bound point, in a different unit than suggested. `safeForStatusline` bounds by columns (`termwidth.Columns` then `termwidth.Cut`), not by bytes or runes, so a byte figure is the wrong instrument at any size — 800 bytes would pass a bound that had broken to 400 columns of ASCII. The assertion now reads the emitted error in columns, and a second case puts 400 wide-script runes through the same bound: two columns to the rune, so a bound counting bytes or runes cuts in the wrong place and a test fed only ASCII cannot tell the three apart.

On point 3: the assertion that fails on the base branch is the last one — the paragraph still carries "Both `host` and `last_error` are written elsewhere". The value assertions pass on base, on purpose: the code was right and the document was not.

Issue #1098, opened 2026-08-06
the same word in NFC and NFD never match: cross-form recall 0/10


vshulcz:
Store: 20 claude sessions, 20 messages. Ten everyday accented words, each
written twice — once precomposed (NFC), once decomposed (NFD). The two render
identically and nobody can tell them apart on screen.

    word      NFC query -> hits    NFD query -> hits    cross-form
    café      nfc-00               nfd-00               NO
    résumé    nfc-01               nfd-01               NO
    naïve     nfc-02               nfd-02               NO
    piñata    nfc-03               nfd-03               NO
    über      nfc-04               nfd-04               NO
    señor     nfc-05               nfd-05               NO
    façade    nfc-06               nfd-06               NO
    crème     nfc-07               nfd-07               NO
    jalapeño  nfc-08               nfd-08               NO
    Zürich    nfc-09               nfd-09               NO

    same-form recall  10/10 (100%)
    cross-form recall  0/10 (0%)

Same-form recall is perfect, so this is normalization alone: tokens are
compared as bytes and the two forms never meet. It is not hypothetical on
macOS, where APFS and several input paths hand back NFD while a terminal
keyboard produces NFC — the same word typed by the same person can miss.

Two more compatibility cases from the same 20-session unicode fixture, same
root cause at the NFKC level:

    fullwidth latin  "ｆｕｌｌｗｉｄｔｈ" queried as "fullwidth"  -> 0 hits
    eszett           "Straße" queried as "strasse"          -> 0 hits

Not filed as bugs, separate design questions: accent-insensitive matching
(réparé/repare) and Turkish casing (ısıtıcı/ISITICI) also miss, but those are
folding choices rather than the same word in two encodings.

I did not fix this. The repo has no external dependencies — go.mod is three
lines — so the choice is between taking `golang.org/x/text/unicode/norm` as
the first one, or hand-rolling a composition table the way the CJK and
Cyrillic folds already are. That is a maintainer's call, not an audit's.

vshulcz:
Verified fixed on current main. nfcfold.Compose() normalizes NFD→NFC (internal/nfcfold/fold.go:23), used on ingest and retrieval — cross-form recall matches now.

PR #1873, opened 2026-08-25
fix(nfcfold): fold letters that carry two marks

### LINKED ISSUE #1872: search: a Vietnamese word typed NFD does not match the same word stored NFC
`nfcfold` exists so an accented word typed in one normal form matches the same word stored in the other (#1098) — the case macOS produces, where a filesystem path hands back NFD while the keyboard emits NFC. It still misses Vietnamese:

```
$ deja index                      # session holds "bệnh" precomposed
$ deja search bệnh --json         # NFC
tier=exact total=1 hits=1
$ deja search bệnh --json         # NFD, the same word
tier=exact total=0 hits=0
```

The cause is in `tools/gen_nfc_fold.py`. A pair is emitted only when its base is Basic Latin, Latin-1/Ext-A/B (U+00C0–U+024F), Greek (U+0370–U+03FF) or Cyrillic (U+0400–U+04FF). A letter that carries two marks composes in two steps, and the intermediate is often outside those ranges:

```
ệ  =  e + U+0323 + U+0302
      e + U+0323  -> U+1EB9  (in the table)
      U+1EB9 + U+0302 -> U+1EC7  (dropped: base U+1EB9 is Latin Extended Additional)
```

So the fold stops halfway and the token still differs. 147 canonical pairs are dropped this way while their mark is inside the range the scan already looks at: 14 with a Latin Extended Additional base (Vietnamese, and `ṩ` = U+1E63 + U+0307) and 133 with a Greek Extended base. Both scripts are named in the package's own doc comment, so this is a gap inside the stated scope rather than a scope question.

The other pairs the range check drops stay dropped: their base is not a letter (`≠` is `=` + U+0338) or the mark is outside the scanned block (Arabic hamza, Hebrew points) — those are the scope question, and this is not it.

Fix: let a base be a Latin Extended Additional or Greek Extended letter too, and regenerate the table.

vshulcz:
Closes #1872.

`nfcfold` folds NFD to NFC so an accented word matches whichever form it was typed in (#1098). It stopped halfway on a letter carrying two marks, because the generator accepted a base only up to U+024F and the intermediate is a letter of its own:

```
ệ = e + U+0323 + U+0302
    e + U+0323      -> U+1EB9      in the table
    U+1EB9 + U+0302 -> U+1EC7      dropped: base is Latin Extended Additional
```

End to end, against a session holding the precomposed word:

```
before   NFC query  tier=exact total=1      NFD query  tier=exact total=0
after    NFC query  tier=exact total=1      NFD query  tier=exact total=1
```

That is Vietnamese, and NFD is the form a macOS path hands back — the case the package exists for. Greek Extended has the same shape (`ἄ` = α + U+0313 + U+0301). Both scripts are named in the package's doc comment, so this was a gap inside the stated scope.

The fix is two blocks added to `is_composable_base` and a regenerated table: 630 → 771 pairs. What the range check still drops stays dropped — a non-letter base (`≠` is `=` + U+0338) or a mark outside the scanned block (Arabic hamza, Hebrew points). Those are the scope question in the package doc and are not touched here.

`TestComposeFoldsLettersCarryingTwoMarks` fails on the base branch on five of its six cases; the sixth (`ợ`) passes there and is kept as the control, since its intermediate `ơ` is inside the old range.


vshulcz:
Checked the first three questions myself while the review ran, and one of them found something.

Table purity: re-derived every pair from Python's `unicodedata` — 772 entries, none where `NFC(base+mark)` differs from the value, and no pair the old table had is missing.

Fuzz against NFC: 3790 words built from every in-scope precomposed letter (1–4 letters each), normalised to NFD and composed. 13 mismatches, all one letter: `ǯ` = U+0292 + U+030C. Ezh sits in IPA Extensions, outside every base range including the two this PR added, so it was still folding halfway. Added as a base, table 771 → 772, and the fuzz now agrees with NFC on all 3790.

What stays out is now only a non-letter base — `¨` + U+0301 → U+0385, and the two like it. Those are symbols, not the package's business.

PR #1881, opened 2026-08-25
fix(statusline): let the filename give way so the memory half fits the bar

### LINKED ISSUE #1880: statusline: the memory half still runs off the bar when the filename is long
`withFileMemory` says the memory half is the one that has to survive, and fits it before appending any of the usage numbers. It still runs off the edge, because only the title is elastic: the filename is capped at 28 columns and never shrinks, and the budget pass runs once.

Measured on the real renderer (`statuslineMemoryLineTo` + the fitting in `withFileMemory`):

```
case                      bar width   rendered   over
long name, short title    60          64         4
long name, long title     60          72         12
narrow bar, long name     40          64         24
```

Two ways it happens. The title budget is cut by the overflow in columns, but cutting a 5-column title by 12 removes 5, not 12 — and the floor (`statuslineMinTitle`) stops it before that anyway. Nothing then rechecks, so the segment is emitted over-width, which is what #1076 fixed for the line as a whole and what a bar with no horizontal scroll loses the tail of.

`statuslineMemoryLineTo`'s comment says it "bounds the title to maxTitle runes"; `safeForStatusline` bounds in columns and says so in its own comment. Same drift, one call away.

Fix: let the filename shrink too once the title is at its floor, and clamp the finished segment to the width so nothing can leave it over — a bar that cuts a filename is readable, a bar that runs off is not.

vshulcz:
Closes #1880.

`withFileMemory` fits the memory half before appending any usage numbers, because that is the half it says has to survive. Only the title was elastic, though: the filename sat at 28 columns whatever happened, and cutting a five-column title by twelve removes five.

```
case                     width   before   after
long name, short title   60      64       60
long name, long title    60      72       60
narrow bar, long name    40      64       46
```

The filename now gives way after the title reaches its floor, down to a floor of its own.

What has not changed is the older decision underneath: a bar too narrow for both floors still gets an over-width segment, because a title cut to a stub inside quotation marks says less than the count form. `TestStatuslineNarrowBarKeepsAReadableTitle` pins that, and an earlier draft of this fix clamped the segment to the width and broke it — the 40-column row above is 46 for that reason, not by omission.

The fitting moved into `memorySegment`, which is what the tests drive. `statuslineMemoryLineTo`'s comment said it bounds the title "to maxTitle runes"; `safeForStatusline` bounds in columns and always has, so the comment now says columns.

Both new tests fail with the filename pass removed — the fit test on the two 60-column rows, and the narrow-bar test because the segment is then exactly as wide as one rendered with the fixed name.

### DIFF
--- a/cmd/deja/statusline_fit_test.go
+++ b/cmd/deja/statusline_fit_test.go
@@ -66,3 +66,29 @@ func TestAnOrdinaryMemorySegmentIsNotTrimmed(t *testing.T) {
 		t.Errorf("a segment that fits carries an ellipsis: %q", got)
 	}
 }
+
+// The budgets are columns of allowance, not columns of line: a title shorter
+// than its budget gives nothing back when the budget is cut. Subtracting the
+// overflow in one step therefore left bars over-width while both parts were
+// still above their floors — 57 of 2496 measured. This walks the grid and
+// allows an over-width segment only where even both floors would not fit.
+func TestNoBarIsLeftOverWidthWhileThePartsCanStillGiveWay(t *testing.T) {
+	when := time.Date(2026, 8, 20, 10, 0, 0, 0, time.UTC)
+	titles := []string{"retry", strings.Repeat("word ", 12), strings.Repeat("計", 30)}
+	for width := 20; width <= 120; width += 4 {
+		for nameLen := 5; nameLen <= 80; nameLen += 5 {
+			for _, title := range titles {
+				m := fileMemory{Path: "/p/" + strings.Repeat("n", nameLen) + ".go", Sessions: 3, Title: title, Last: when}
+				got := memorySegment(m, width)
+				if barColumns(got) <= width {
+					continue
+				}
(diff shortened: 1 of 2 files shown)


vshulcz:
Ran your point 1 myself while you worked, and it found a second defect in my own fix.

Grid: widths 20–120, filename 5–80 characters, three titles (ASCII short, ASCII long, CJK), with and without a timestamp — 2496 cases. **57 came back over-width while both parts were still above their floors.** A budget is columns of allowance, not columns of line: a title shorter than its budget gives nothing back when the budget is cut, so one subtraction does not converge.

The second commit steps each budget down instead. Same grid now: 0 over-width where the floors would fit, 698 over-width at the floors — the documented case that `TestStatuslineNarrowBarKeepsAReadableTitle` requires. The loop is bounded by the budgets themselves (at most 26 + 18 renders of a short string).

`TestNoBarIsLeftOverWidthWhileThePartsCanStillGiveWay` walks that grid and allows an over-width segment only where even both floors would not fit. Please re-read the pushed branch — points 2 through 6 still stand.

PR #1886, opened 2026-08-25
perf(search): build the query side of the record check once

### LINKED ISSUE #1885: search: the per-record check rebuilds the whole query side for every record
`recordMatchesQueryVariants` runs once per record in the verification loop — for a query with no postings, that is every record in the store. Everything it computes from the *query* is recomputed each time:

- `query.QueryParts(o.Query)`
- `nfcfold.Compose(o.Query)`, twice on the NFD retry path, plus `QueryParts` of the result
- `cjkfold.String(o.Query)`, twice, plus `QueryParts` of that

None of it depends on the record. Measured on 120 sessions × 20 messages:

```
ASCII corpus, ASCII query        1.91–2.01 ms   22984 allocs
NFD corpus, ASCII query          2.05–2.18 ms   22984 allocs
NFD corpus, NFD query            4.11–4.35 ms   49385 allocs
```

A decomposed query doubles the work on the same corpus, because it takes the retry path on every record and rebuilds the query side there each time.

Fix: build the query side once per search and hand it to the per-record check. The record-side work — `Compose(r.Text)`, `cjkfold.String(r.Text)` — stays where it is; it genuinely depends on the record, though the composed text is currently computed twice in the same branch.

vshulcz:
Closes #1885.

`recordMatchesQueryVariants` runs once per record — for a query with no postings, that is every record in the store — and everything it derived from the *query* was rebuilt each time: `QueryParts`, `nfcfold.Compose` (twice on the retry path, plus the parts of the result), `cjkfold.String` (twice, plus its parts). None of it depends on the record.

120 sessions × 20 messages, `Search(dir, {All:true})`:

```
                       before                    after
ASCII query            1.91–2.01 ms  22984       1.30–1.32 ms  6205
NFD corpus             2.05–2.18 ms  22984       1.49–1.54 ms  6205
NFD query              4.11–4.35 ms  49385       1.78–1.83 ms  6208
```

A decomposed query took the retry path on every record, so it paid for the rebuild twice over. The record side — composing the record's own text, folding it for CJK — stays where it is, and the composed text is now computed once in that branch rather than twice.

`recordMatcher` holds the query side; the two scan loops build it once. `recordMatchesQuery`/`recordMatchesQueryVariants` remain as wrappers, so the other callers and the tests are untouched.

Tests: `TestTheQuerySideIsNotRebuiltForEveryRecord` compares allocations of a decomposed query against a plain one on the same corpus — a ratio, so it does not depend on the machine. It reports 2.14x on the base branch and fails; it is 1.0x now. `TestADecomposedQueryStillFindsComposedText` keeps the retry paths honest, with both normal forms written as escapes.

The benchmark used for the numbers is not committed.

### DIFF
--- a/internal/index/query_side_test.go
+++ b/internal/index/query_side_test.go
@@ -48,11 +48,14 @@ func seedQuerySideCorpus(t *testing.T, sessions int) string {
 //
 // Measured as a ratio against the plain query on the same corpus, so the
 // assertion does not depend on the machine or on how much the reader's Go
-// version allocates elsewhere.
+// version allocates elsewhere. The threshold is 1.3 because the two searches
+// allocate the same amount once the query side is shared — the ratio is 1.0
+// here and 2.14 on the branch before this — and because -race and a low GOGC
+// move both sides together (checked).
 func TestTheQuerySideIsNotRebuiltForEveryRecord(t *testing.T) {
 	dir := seedQuerySideCorpus(t, 40)
 	allocs := func(q string) float64 {
-		return testing.AllocsPerRun(2, func() {
+		return testing.AllocsPerRun(5, func() {
 			if _, err := Search(dir, query.Options{Query: q, All: true}); err != nil {
 				t.Fatal(err)
 			}
--- a/internal/index/query_side_test.go
+++ b/internal/index/query_side_test.go
@@ -1,0 +1,85 @@
+package index
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	query "github.com/vshulcz/deja-vu/internal/query"
+)
+
+// seedQuerySideCorpus writes sessions whose text is decomposed (NFD), the form
+// a macOS path hands back, and returns the index dir.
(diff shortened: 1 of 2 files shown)


vshulcz:
Took point 4, measured point 5.

**Point 4.** `AllocsPerRun` now averages five runs, and the comment says why 1.3 is the threshold: the two searches allocate the same amount once the query side is shared (ratio 1.0 here, 2.14 on the base branch), so the headroom is an order of magnitude wider than the noise. Checked both of your failure modes: `-race` passes, `GOGC=20` passes — they move both sides of the ratio together, which is the point of measuring a ratio.

**Point 5.** The eager `cjkfold.String` and `Compose` of the query are not a regression for the ASCII case, because the old code did that work *per record* — `cjkfold.String(o.Query)` twice per record on the reject path, and `QueryParts` of it. Measured on the ASCII query over the same corpus: **1.91–2.01 ms / 22984 allocations before, 1.30–1.32 ms / 6205 after.** The one-time cost is two folds of a short string.

Point 1's branch-by-branch reading matches mine, and it is the part I most wanted a second pair of eyes on.

Issue #1876, opened 2026-08-25
doctor: sessions_from_there reads 0 when the machine spells its name in another case


vshulcz:
#1867 made a host name case-insensitive inside `internal/peers`. The count beside each machine in `deja doctor` is looked up with the raw string:

```go
from := index.ImportedByMachine(dir)   // keyed by the sender's own machine name
… peerLine(p, from[p.Host], now)
```

The two names come from different places. The peers row is the ssh alias someone typed; `From` is what the sending machine calls itself (`index.MachineName`, its hostname's first label). macOS hands back a capitalised one, so `ssh laptop` and a host named `Laptop` are the ordinary case:

```
$ deja sync import ./batch          # three records stamped origin "Laptop"
deja: imported 3 records — 3 sessions from another machine
$ deja doctor
  laptop       last exchange 5 days ago, nothing sent yet
$ deja doctor --json
"peers":[{"host":"laptop","last_pull":"…","sessions_from_there":0}]
```

Three sessions arrived from that machine and both surfaces report none. The text line drops the "N sessions from there" clause entirely, so the row reads like a peer that has sent nothing — which is what someone runs `deja doctor` to find out.

Fix: fold the lookup the same way the peers list folds, so the count finds the row whatever case either side spelled the name in. What deja cannot fix here is a genuinely different name — an alias `work` against a hostname `build-box` — and that stays a miss.

PR #1888, opened 2026-08-25
fix(doctor): learn what a peer calls itself, so its sessions are counted

### LINKED ISSUE #1887: doctor: a machine whose hostname differs from its ssh alias counts zero sessions
`deja doctor` counts a peer's sessions by name: the peers row holds the ssh alias someone typed, while an imported session is stamped with what the sending machine calls itself (`index.MachineName`, its hostname). #1876 made that comparison case-insensitive, which covers `laptop` against `Laptop`. It does nothing when the two names are simply different, which is the normal case for anyone with an ssh config:

```
$ deja sync import ./batch          # three records stamped origin "build-box"
deja: imported 3 records — 3 sessions from another machine
$ deja doctor
  work         last exchange 5 days ago, nothing sent yet
$ deja doctor --json
"peers":[{"host":"work","last_pull":"…","sessions_from_there":0}]
```

Three sessions arrived from that machine and the row says none. The text line drops the "N sessions from there" clause altogether, so the row reads like a machine that has sent nothing — the thing `deja doctor` exists to tell you apart from a machine that has.

deja already has both halves of the pairing at the moment of a pull: the alias it connected to, and the origin stamped on the records that arrived. Nothing writes it down.

Proposal: learn the pairing during an ssh pull and keep it in the peer's row (an optional `machine` field, absent until learned, so an older file reads unchanged), then count by either name. A machine deja has only ever pushed to still counts nothing, because nothing has arrived from it to learn from — which is honest, since the count is about what arrived.

vshulcz:
Closes #1887.

`deja doctor` counts a peer's sessions by name. The row holds the ssh alias someone typed; an imported session is stamped with what the sending machine calls itself. #1876 made that comparison case-insensitive, which covers `laptop` against `Laptop` and does nothing when the names simply differ — the normal case for anyone with an ssh config:

```
$ deja sync import ./batch      # three records stamped origin "build-box"
deja: imported 3 records — 3 sessions from another machine
$ deja doctor
  work         last exchange 5 days ago, nothing sent yet
```

Three sessions arrived and the row says none, dropping the "N sessions from there" clause entirely — so it reads like a machine that has sent nothing, which is the distinction the screen exists to draw.

deja has both halves of the pairing only during an ssh pull: the alias it connected to, and the origin stamped on the records that arrived. `learnPeerMachine` takes the per-machine counts before the import, sees which one grew, and writes it into the row as `machine` — a new optional field, absent until a pull has something to learn from, so an older peers file reads unchanged. Both surfaces then count by either name.

Deliberately not covered: a hand-run `deja sync import` learns nothing, because no alias is involved — deja would be guessing. A machine only ever pushed to still counts zero, which is honest: nothing has arrived from it.

Tests: the pairing is learned and both surfaces count three; a batch from a third machine already in the index does not rename the peer; a pull that brings nothing writes nothing. The first fails with the second lookup removed (checked by neutering it), naming both surfaces.

### DIFF
--- a/cmd/deja/doctor_peers.go
+++ b/cmd/deja/doctor_peers.go
@@ -170,13 +170,19 @@ func peerSessionCount(from map[string]int, p peers.Peer) int {
 // attributed — a batch from a third machine already in the index must not
 // rename the peer.
 func learnPeerMachine(dir, host string, before map[string]int) {
-	best, grew := "", 0
+	best, grown := "", 0
 	for name, n := range importsByPeerName(dir) {
-		if d := n - before[peers.Identity(name)]; d > grew {
-			best, grew = name, d
+		if n > before[peers.Identity(name)] {
+			best = name
+			grown++
 		}
 	}
-	if best == "" || peers.Identity(best) == peers.Identity(host) {
+	// Only when one machine grew. A batch can carry records from more than one
+	// origin — a machine that itself imported from a third — and another
+	// process can import while this runs, so picking the largest of several
+	// would sometimes write down a machine this host never sent. Learning
+	// nothing leaves the count at zero, which is what it was.
+	if grown != 1 || peers.Identity(best) == peers.Identity(host) {
 		return
 	}
 	// Best effort: a sync that worked must not fail because a name could not be
--- a/cmd/deja/doctor_peers.go
+++ b/cmd/deja/doctor_peers.go
@@ -48,7 +48,7 @@ func doctorPeers(w io.Writer, dir string, now time.Time) {
 		}
 	}
 	for i, p := range list {
-		state := peerLine(p, from[peers.Identity(p.Host)], now)
+		state := peerLine(p, peerSessionCount(from, p), now)
 		pad := width - termwidth.Columns(names[i])
 		if pad < 0 {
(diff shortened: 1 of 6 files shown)


vshulcz:
Took 1, 2 and 4; 3 was already the behaviour.

**1 (attribution).** Right that the largest grower is not always this host's — a batch can carry a third machine's work, quite apart from a concurrent import. Learning now happens only when exactly one machine grew; otherwise nothing is written and the count stays what it was. `TestAnAmbiguousPullLearnsNothing` covers it (two origins in one exchange), which is also your point 4 in the form a test can actually pin — a genuinely concurrent import is the same shape and the same answer.

**2 (bound).** Bounded at the write, and worth correcting the size: the origin is not unbounded upstream. `sanitizeSyncField` runs on import — control and Cf runes replaced, 120 runes maximum — so a 10 MB origin never reaches the manifest, let alone this file. The bound is defence in depth rather than a fix for a live overflow, and the comment says so.

**3 (stale pairings).** `Learn` already overwrites when the name differs, so the next pull corrects a repointed alias or a renamed machine; only an unchanged name short-circuits. Added `TestARepointedAliasIsRelearned` so that is pinned rather than assumed.

**5.** Left the doc sentence as it is — it says which of the two names does the matching and where the other comes from, and that is the question a reader of the field has.

Issue #1874, opened 2026-08-25
stats --json: recall.since is documented as omitted and reports year 1 instead


vshulcz:
`docs/json-output.md` says of `deja stats --json`:

> Inside `recall`, `raw_bytes` is the size of the source transcripts the served digests distilled and `since` is the oldest event still in the usage log; both are omitted when zero, so a store with no recall history yet shows neither.

`raw_bytes` is omitted. `since` is not — it is a `time.Time`, and `encoding/json`'s `omitempty` has no effect on a struct, so a store that has never served a recall reports a date in year 1:

```
$ deja stats --json          # fresh store, no recall history
"recall": {
  "recalls_served": 0,
  ...
  "since": "0001-01-01T00:00:00Z"
}
```

A consumer that reads the documented contract branches on the key being absent, and gets a timestamp it will happily subtract from — "recall history since the year 1" on a dashboard, or a two-thousand-year window in whatever computes rates.

Same defect elsewhere in the emitted JSON: `blameSessionJSON.Started` and `.Updated` in `cmd/deja/mcp.go` carry `,omitempty` on a `time.Time`, so a session whose harness records no start stamp reports year 1 to an agent rather than leaving the field out. (`internal/peers` has the same tags, but that file is deja's own and a year-1 stamp round-trips back to a zero `time.Time`, so nothing reads it wrong — noise in the file, not a wrong answer.)

Fix: drop the field when the time is zero, and add a check so the next `time.Time` with `,omitempty` is caught rather than believed.

Issue #763, opened 2026-08-01
"Recalls served" silently drops from 8100 to 101 when the usage log rotates


vshulcz:
The usage log is rewritten once it passes 1 MB, keeping the last 14 days. `deja stats` reports the count with no period attached, so it reads as a lifetime total and then falls:

```
$ deja stats --json   # 8100 events, log at 1.3 MB
Recalls served 8100
$ (one more recall triggers the rotation)
$ deja stats --json
Recalls served 101
```

The number a user watches to see whether memory is paying off went backwards by two orders of magnitude, with nothing saying why. Either it should say what window it covers, or it should survive rotation.

PR #1890, opened 2026-08-25
fix(stats): say which window the impact numbers cover

### LINKED ISSUE #1889: stats --impact: presents a 14-day window as everything that ever happened
`deja stats --impact` presents its numbers as everything that has happened on this machine — the header says "measured on this machine, nothing modeled", the counts carry no period, and `ImpactReport.DejaVuMoments` is commented "all time". They are not. The usage log is rewritten past 1 MB keeping the last 14 days, so the figures are a window whose start moves silently.

One rotation over a log holding 30 days of events:

```
recalls        4000  ->  1872
served bytes   2.0MB ->  936KB
oldest event   2026-07-27 -> 2026-08-12
```

Nothing on the screen changed except the numbers, and they halved.

`deja stats` already says which window it is talking about — "Recalls served 101 since Jun 22", which is #763 — because a count with no period attached reads as a lifetime total and then falls by an order of magnitude when the log rotates. The impact screen never got that half, and `--impact --json` has no `since` field at all.

Fix: carry the oldest surviving event on `ImpactReport` and say it, on both surfaces. The phrasing on the human screen is yours to settle; I will put a plain one in the PR and it can be changed there.

vshulcz:
Closes #1889.

`deja stats --impact` reads as everything that has happened on this machine — the header says "measured on this machine, nothing modeled", the counts carry no period, and `DejaVuMoments` was commented "all time". The usage log is rewritten past 1 MB keeping 14 days, so one rotation over a 30-day log takes the numbers down by half and nothing on the screen says why:

```
recalls        4000  ->  1872
served bytes   2.0MB ->  936KB
oldest event   2026-07-27 -> 2026-08-12
```

`deja stats` has named its window since #763 ("Recalls served 101 since Jun 22"). This is that half for the screen that most looks like a lifetime total:

```
deja impact — measured on this machine since Jul 27, nothing modeled
  recalls served     30 agent-initiated recalls returned matches
```

and `"since": "2026-07-27T13:43:31Z"` in `--impact --json`, absent when the log holds nothing (the shape of #1874 — `omitempty` does nothing to a `time.Time`).

One thing worth knowing: the JSON wrapper embedded `ImpactReport` to add `credited_aloud`, and an embedded type's `MarshalJSON` answers for the whole outer struct, so adding one here silently dropped that key. The two are now marshalled apart and joined; `TestImpactJSONIncludesCreditedAloud` already covered it and caught it.

Wording is yours: "measured on this machine since Jul 2" is what I picked, and the alternative is to leave the header alone and put the window on a line of its own.

### DIFF
--- a/cmd/deja/stats_impact.go
+++ b/cmd/deja/stats_impact.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/vshulcz/deja-vu/internal/index"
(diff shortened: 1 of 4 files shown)


vshulcz:
Point 1 was worth asking, and following it found a second slip of mine.

The cutset trim (`TrimRight(b, "}\n")`) does eat every trailing brace, so a nested value one day would take the object's own with it. Fixed. But trimming exactly one brace then left the newline before it, and the comma landed on a line of its own — valid JSON, ugly output, and the existing tests, which only look for keys, said nothing. Both are fixed in dedb148a, and `TestImpactJSONIsWellFormedAfterTheJoin` now pins `json.Valid`, both keys and the absence of a stray comma line.

Point 3: `Since` is the oldest event in the log, uncounted kinds included, which is what `Totals` does and what the document already says `recall.since` means. The alternative — the oldest *counted* event — would make the window narrower than the file it is describing, and the counts would then look like they cover less than they do.

Point 4: a log holding only uncounted kinds never reaches the header. `printImpact` returns on `Recalls == 0 && Injections == 0` with "no recall activity recorded yet", which is the right sentence for it.

vshulcz:
Point 5, answered honestly before you get to it: the three usage tests cannot run on the base branch at all — they name `ImpactReport.Since`, which does not exist there, so it is a compile failure rather than a red test. What fails on base is the measurement in the issue (4000 recalls to 1872 across one rotation, with nothing on the screen saying so).

And you would be right to call `TestAnEmptyImpactLogHasNoWindow` vacuous against base: with no field there is no key either. It is a guard against the shape of #1874 — `omitempty` doing nothing to a `time.Time` — rather than a proof of this fix; `TestImpactJSONCarriesTheWindow` is the one that shows the field arrives when there is a window to report.

PR #1897, opened 2026-08-25
test(json): catch a struct that embeds a type writing its own JSON

vshulcz:
A guard for the shape #1890 hit, and a sweep saying there is nothing else of it in the tree today.

An embedded type's `MarshalJSON` answers for the whole outer struct, so the outer fields vanish with no error and no warning. That is how `deja stats --impact --json` lost `credited_aloud` the moment `ImpactReport` grew a marshaller of its own. Three types now write their own JSON — `usage.Summary`, `usage.ImpactReport`, `blameSessionJSON` — and all three are reported somewhere.

The sweep first: walking `doctorReport`, `stats.Report`, both usage reports, the MCP blame types, `model.Session` and `index.SessionMeta`, plus everything reachable from them — 35 types, no findings. So this PR changes no behaviour; it adds the check that would have caught #1890 the moment it was written.

Two things worth knowing about the walk:

- it does not spare a struct that appears to marshal itself, because embedding is exactly how that method got there and reflect cannot tell a promoted `MarshalJSON` from a declared one. My first draft did spare it, and the negative control caught that immediately — the walk saw nothing at all.
- the test carries that negative control: a local struct embedding `usage.Summary` must be reported, and marshalling it must genuinely drop the outer field. If Go ever changes that rule, the test says so rather than passing quietly.

### DIFF
--- a/cmd/deja/embedded_marshaller_test.go
+++ b/cmd/deja/embedded_marshaller_test.go
@@ -56,6 +56,8 @@ func TestNoReportedTypeEmbedsAMarshaller(t *testing.T) {
 		doctorReport{}, stats.Report{}, usage.Summary{}, usage.ImpactReport{},
 		blameSessionJSON{}, blameHitJSON{}, model.Session{}, index.SessionMeta{},
 	}
+	// One map across the roots: a type's fields do not depend on which root
+	// reached it, so inspecting it once inspects it for all of them.
 	seen := map[reflect.Type]bool{}
 	var found []string
 	for _, r := range roots {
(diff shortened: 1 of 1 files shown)


vshulcz:
Took point 4 as a comment rather than as separate maps.

A finding is a property of a type, not of the path that reached it: `embeddedMarshallers` reports `T embeds U` while looking at T's own fields, so whichever root gets there first does the whole inspection and the second root has nothing left to find. Separate maps per root would repeat the work and emit the same finding several times. The reason is now written where the map is made.

Point 1: also checked from the other side — across `cmd/deja` and `internal`, excluding tests, there is no embedded field in any struct declaration at all today, so the shape is absent tree-wide rather than only under the eight roots.

Point 5: kept "spare", which is the verb for what the draft did and stopped doing.

PR #1900, opened 2026-08-25
docs(json): describe stats --impact --json and pin its keys

### LINKED ISSUE #1898: docs: stats --impact --json is a machine surface with no documented shape
`deja stats --impact --json` prints a machine-readable object and is documented nowhere: not in `docs/json-output.md`, not in the README, not in `deja help`. The document's own stability policy names the two shapes that return no envelope (`blame`, `log`); this is a third and it is not on the list, so a consumer cannot tell whether the missing `schema_version` is deliberate.

What it emits today:

```json
{
  "recalls": 30,
  "injections": 0,
  "served_bytes": 15000,
  "raw_bytes": 150000,
  "reused_twice": 0,
  "dejavu_moments": 0,
  "since": "2026-07-27T14:33:13Z",
  "credited_aloud": 0
}
```

`since` arrived last night in #1890 and `credited_aloud` was silently dropped for one commit in the same PR before a test caught it — on a surface with no documented shape and, unlike `stats --json`, no contract test pinning its keys.

Fix: document the object beside the others, say that it carries no envelope and why, say which keys may be absent, and pin the key set to the document the way `TestStatsJSONKeysMatchTheDocumentedContract` does.

vshulcz:
Closes #1898.

`deja stats --impact --json` is a machine surface documented nowhere — not in `docs/json-output.md`, not in the README, not in `deja help`. It grew `since` last night in #1890, and `credited_aloud` vanished for one commit in that same change before a test caught it.

Now described beside the others: the object, what each counter means, that `since` is the window the counts belong to rather than a lifetime, and that it is absent when the log holds nothing. The stability policy already named the two shapes that return no envelope (`blame`, `log`); this is the third and says so on the same terms.

`TestImpactJSONKeysMatchTheDocumentedContract` pins the key set to that section in both directions, the way `TestStatsJSONKeysMatchTheDocumentedContract` has since #1710. It fails on the base branch — there is no section to read.

No behaviour change.

### DIFF
--- a/cmd/deja/impact_contract_test.go
+++ b/cmd/deja/impact_contract_test.go
@@ -1,0 +1,60 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/vshulcz/deja-vu/internal/usage"
+)
+
+// `deja stats --impact --json` is a machine surface that was documented
+// nowhere, and it grew `since` in #1890 while `credited_aloud` vanished for one
+// commit in the same change. `stats --json` has had its keys pinned to the
+// document since #1710; this is that pin for the screen beside it (#1898).
+func TestImpactJSONKeysMatchTheDocumentedContract(t *testing.T) {
+	doc, err := os.ReadFile("../../docs/json-output.md")
+	if err != nil {
+		t.Fatal(err)
+	}
+	section := string(doc)
+	i := strings.Index(section, "## `deja stats --impact --json`")
+	if i < 0 {
+		t.Fatal("docs/json-output.md does not describe `deja stats --impact --json`")
+	}
+	section = section[i:]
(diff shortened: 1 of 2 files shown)


vshulcz:
The blocking point is not one: a 27-day-old `since` is ordinary. Rotation is triggered by size, not by age — `rotate` returns immediately unless the file is past 1MB (`internal/usage/usage.go`), and only then drops what is older than 14 days. A machine that serves a few recalls a day keeps months of events, and the probe behind #1889 held 30 days of them in a 1.19MB file before rotating.

But the sentence invited the reading, so it is rewritten: `since` is the oldest event in the log, on a quiet machine that is everything there has ever been, and the window only starts moving once the log passes 1MB. That also answers your ⚠ — the doc no longer says the counts *belong to* that period, only that they cover no more than it, which is exactly true whichever kinds the counters ignore.

Example date kept, for the reason above.

PR #1902, opened 2026-08-25
fix(usage): close a truncated line before appending the next event

### LINKED ISSUE #1901: usage: a log cut off mid-write swallows the next event too
The usage log is appended without a lock by design, and `read` drops any line that does not parse — so a process killed between the record and its newline costs that one event, which is the intended trade. It costs one more than that: the next event deja records is appended onto the partial line, and the whole line then fails to parse.

```
log ends mid-record, no trailing newline
  events readable: 1
$ … one ordinary recall is recorded …
  events readable: 1        <- the recall just written is gone
  totals: recalls=1 bytes=100
```

So a crash mid-write silently eats the next recall as well, and every count that reads this file — `deja stats`, the status bar, `--impact` — is short by it. Later appends land on their own lines, so it is exactly one event per truncation, not an ongoing loss.

Fix: when the file does not end in a newline, write one before the record. It costs a one-byte read per append on a file deja already opens.

vshulcz:
Closes #1901.

The usage log is appended without a lock by design, and `read` drops a line that does not parse — so a process killed between a record and its newline costs that record. That is the trade. It cost one more: the next event was appended onto the partial line, so that line stopped parsing too.

```
before   events readable 1 -> record a recall -> events readable 1   (totals: recalls=1 bytes=100)
after    events readable 1 -> record a recall -> events readable 2   (totals: recalls=2 bytes=300)
```

The recall that went missing had already been served; every count reading this file was short by it — `deja stats`, the status bar, `--impact`.

The append now reads the last byte and closes the line first, which needed the file opened `O_RDWR` rather than `O_WRONLY`. A file it cannot stat or read is treated as ending cleanly: a usage event must never be the reason a recall fails.

Tests: the truncated-tail case fails on the base branch, and a second one keeps the ordinary path honest — two records, two lines, no blank line.

### DIFF
--- a/internal/usage/truncated_tail_test.go
+++ b/internal/usage/truncated_tail_test.go
@@ -1,0 +1,65 @@
+package usage
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// A process killed between the record and its newline costs that record — the
+// log is appended without a lock by design and `read` drops a line that does
+// not parse. It cost one more: the next event was appended onto the partial
+// line, so that line failed to parse too and the recall just served went
+// missing from every count (#1901).
+func TestARecordAfterATruncatedTailSurvives(t *testing.T) {
+	dir := t.TempDir()
+	t.Setenv("DEJA_INDEX_DIR", dir)
+	p := Path(dir)
+	if err := os.MkdirAll(filepath.Dir(p), 0o700); err != nil {
+		t.Fatal(err)
+	}
+	good := `{"t":"2026-08-20T10:00:00Z","kind":"recall","bytes":100,"sessions":1}`
+	partial := `{"t":"2026-08-21T10:00:00Z","kind":"recall","byte`
+	if err := os.WriteFile(p, []byte(good+"\n"+partial), 0o600); err != nil {
+		t.Fatal(err)
+	}
+
+	RecordResult(dir, KindRecall, 200, 1, false)
+
+	tot := Totals(dir)
+	if tot.Recalls != 2 {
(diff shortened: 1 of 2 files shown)


vshulcz:
Measured your point 4 while you read, and checked point 1's premise myself.

**Cost.** 2000 appends × 3, M3 Pro: **30.4–35.1 µs/op without the check, 30.8–33.0 µs/op with it** — the same, within noise, at +208 B and one allocation per append. A 1417537994 339 crw-rw-rw- 1 root wheel 50331650 0 "Aug 25 17:50:38 2026" "Aug 25 17:57:32 2026" "Aug 25 17:57:32 2026" "Jan  1 03:00:00 1970" 65536 0 0 (stdin) and a one-byte  are nothing beside the write and the rotation check the function already does.

**One write.** The newline and the record go out in a single  — the byte is prepended to the marshalled buffer, not written separately — so an interleaved append cannot land between them. That was the thing worth confirming before worrying about two processes.

vshulcz:
Correcting the comment above — my shell ate the backticked words in it.

**Cost (your point 4).** 2000 appends × 3, M3 Pro: **30.4–35.1 µs/op without the check, 30.8–33.0 µs/op with it** — the same, within noise, at +208 B and one allocation per append. A `stat` and a one-byte `ReadAt` are nothing beside the write and the `rotate` check the function already does.

**One write (your point 1's premise).** The newline and the record go out in a single `Write`: the byte is prepended to the marshalled buffer rather than written separately, so an interleaved append cannot land between them. Worth confirming before reasoning about two processes at once.

PR #1909, opened 2026-08-25
docs(usage): write down the rotation window the append accepts

vshulcz:
A measurement and a comment. No behaviour change, no issue — the window is real and the trade is the right one.

`recordFull` rotates the usage log and then opens it. A rotation by another process between that open and the write leaves the descriptor pointing at the file just retired, and the event goes with it:

```
1.1MB log of events older than the keep window
open the file → rotate → append
  size_before=1125KB  size_after=0KB  rotated=true  event_in_log=0
```

Reaching it takes all three at once: the log past 1MB (below that `rotate` returns immediately), two processes recording in the same instant, and the rotation landing inside the open plus a stat and a one-byte read — a few microseconds.

Closing it properly means locking a file this package appends to without a lock on purpose (#1319), which is the trade that keeps a usage event from ever being the reason a recall waits. One event is the cost, and one event is what this file is allowed to lose — the same trade it already makes for a process killed mid-write.

So the change is the comment saying that, where the next reader will meet it as a bug otherwise. The numbers are in the PR rather than the source, since they are about this machine.

### DIFF
--- a/internal/usage/usage.go
+++ b/internal/usage/usage.go
@@ -181,6 +181,16 @@ func recordFull(indexDir, kind string, bytes, sessions int, empty bool, raw int6
 		return
 	}
 	rotate(p)
+	// A rotation by another process between this open and the write below
+	// leaves this descriptor pointing at the file that was just retired, and
+	// the event goes with it. Measured and left alone on purpose: it takes a
+	// log past 1MB, two processes recording in the same instant, and the
+	// rotation landing inside those few microseconds — and closing it properly
+	// means locking a file this package appends to without a lock by design
+	// (#1319), which is the trade that keeps a usage event from ever being the
+	// reason a recall waits. One event is the cost, and one event is what this
+	// file is allowed to lose.
+	//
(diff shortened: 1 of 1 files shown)


vshulcz:
Three of your points taken, one declined with a reason.

**"rotate() in the current process isn't a guard."** That is the sharpest thing in your report and the comment now leads with it — the race does not need our own rotation to have done anything.

**Timing.** You are right that "few microseconds" was doing work it could not support: the span is the open, the stat, the one-byte read and the write, which is tens of microseconds here and more under load. The comment says that now.

**Length.** Twelve lines to seven.

**The test.** Declined, and this is the one I want on the record: a test that asserts the event is lost pins the loss as behaviour, and the next person to close this window properly would have to delete the test that says they must not. A comment says "we accept this today"; a test says "this must keep happening". If it ever becomes worth closing, I would rather the fix arrive without an argument attached.

**Inode comparison being cheaper than I implied** — agreed, and it is the first thing to try if this ever stops being acceptable. It narrows the window to the write itself rather than closing it, which is why it is not worth the code today.

PR #1910, opened 2026-08-25
test(usage): pin that a blank line in the log is read past

vshulcz:
A guard, not a fix. Nothing is broken — no issue filed.

After #1902 two processes can both close a truncated tail, which leaves a blank line in the usage log. Today that is harmless: the events around it are whole and `read` drops any line that does not parse.

```
log with blank lines between records
  read=3   totals(recalls=3 bytes=300)
  append   read=4   blank lines still there, nothing lost
```

What was missing is anything saying so. A blank line in a JSON-lines file is exactly the kind of thing a later reader decides is corruption and makes fatal — and every count deja shows comes through that reader.

Two tests: the counts survive blank lines on both screens and an append onto such a file lands whole; and rotation, which walks the same reader, neither chokes on them nor carries them forward.

Checked by mutation rather than assumed: making an empty line stop the read turns the first test red on all three assertions.

### DIFF
--- a/internal/usage/blank_lines_test.go
+++ b/internal/usage/blank_lines_test.go
@@ -10,10 +10,10 @@ import (
 )
 
 // A blank line is what two processes leave when both close a truncated tail
-// (#1902), and it is not corruption: the events around it are whole. `read`
-// drops any line that does not parse, which covers it — but nothing said so, and
-// a blank line in a JSON-lines file is exactly the kind of thing a later reader
-// makes fatal.
+// (#1902), and it is not corruption: the events around it are whole. Both
+// readers here drop a line that does not parse, which covers it — an empty line
+// is not valid JSON — but nothing said so, and a blank line in a JSON-lines
+// file is exactly the kind of thing a later reader makes fatal.
 func TestBlankLinesAreReadPast(t *testing.T) {
 	dir := t.TempDir()
 	t.Setenv("DEJA_INDEX_DIR", dir)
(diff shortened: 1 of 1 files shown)


vshulcz:
Point 2 was a real gap and is now covered. Point 1 is not a bug — here is why, since it was your blocking one.

**Point 2 (the other reader).** `Events` keeps a line on `e.Kind != ""` where `read` keeps it on a non-zero stamp, so it deserved an assertion rather than an assumption. Both tests now check it: three events through `deja log`'s reader on the blank-lined log, and the same count as the counters on the rotated one.

**Point 1 (midnight).** `rotate` does not work in calendar days: `cutoff := time.Now().UTC().Add(-keepWindow)` is an instant fourteen days back, and the events are built as `now.Add(-i days)` from an instant captured microseconds earlier. Nothing rounds to a day, so a calendar boundary between the two `Now()` calls changes nothing — the gap between them is the same handful of microseconds at 23:59:59 as at noon. Ran it under `TZ=Pacific/Kiritimati` and `TZ=Etc/GMT+12` (UTC+14 and UTC−12, so one of them is always across the date line from the other) and five times in a row: green, ~0.6s.

**Point 3 (json.Unmarshal on empty input).** Written into the comment as the thing the tolerance rests on.

**Point 4.** `read(p) != before` cannot be asserted the way you suggest — `before` is defined as `len(read(p))`, so it would compare a value with itself. The rotated-log assertion you asked for is in.

**Point 5.** The test writes `"\n\n"` itself and `os.WriteFile` does not translate line endings on any platform, so Windows sees the same bytes.

PR #1911, opened 2026-08-25
test(json): pin the keys deja show and deja last emit to the document

vshulcz:
A pin for two sections that had prose and nothing checking it. No issue — the prose turned out to be complete.

The audit first. `deja show --json` and `deja last --json` emit:

```
show   schema_version, session{harness,id,project,path,title,started,updated,words,
       source{origin}, messages{role,text,time}}, window{offset,limit,total,returned}
last   schema_version, sessions{…the same session object…}
```

Against `docs/json-output.md`: nothing is undocumented. The keys their own sections do not name — `path`, `title`, `words`, `started`, `updated` — are the ones the shared "session object" table describes for search, last and show together, which is where they belong.

What was missing is the check. `stats --json` has been pinned key by key since #1710 and `--impact --json` since #1900; these two had nothing, so a field added to the session object reaches three commands' output at once with no document entry and nothing to say so.

`TestSessionJSONKeysAreDocumented` runs both commands against a real index and requires every key, nested ones included, to appear in the document. It checks the whole document rather than the two sections, for the reason above.

Verified by mutation rather than assumed: deleting the `words` row from the session-object table turns it red for both commands. It also refuses to run if the corpus emits fewer than eight keys, so a broken fixture cannot make it pass by finding nothing.

### DIFF
--- a/cmd/deja/session_json_contract_test.go
+++ b/cmd/deja/session_json_contract_test.go
@@ -1,0 +1,98 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/index"
+)
+
+// jsonKeys is every key a value emits, nested ones included, once each.
+func jsonKeys(v any, into map[string]bool) {
+	switch t := v.(type) {
+	case map[string]any:
+		for k, inner := range t {
+			into[k] = true
+			jsonKeys(inner, into)
+		}
+	case []any:
+		for _, inner := range t {
+			jsonKeys(inner, into)
+		}
+	}
+}
+
+// `deja show --json` and `deja last --json` are documented in prose and pinned
+// by nothing, while `stats --json` has been pinned key by key since #1710 and
+// `--impact --json` since #1900. A field added to the session object reaches
+// three commands' output at once, so this is the check that a document entry
+// comes with it.
+//
+// Against the whole document rather than the two sections: the keys those
+// sections do not name are the ones the shared "session object" table describes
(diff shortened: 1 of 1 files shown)


vshulcz:
Point 2 was right and is fixed; point 3 taken with a correction; point 1 was already answered on the branch you read before it moved.

**Point 2 (whole-document search).** Fixed rather than argued: each command is now checked against its own section plus the shared session-object table, which is where the keys those sections do not name live. Your example is the one that matters — `superseded` is documented for `search`, so under the old check `show` could emit it undocumented and pass.

**Point 3 (the guard).** Raised, but not to 15 for both: `last` emits twelve distinct keys, so 15 would fail on a healthy tree. It is now per command — 18 for `show`, 10 for `last` — which is the same idea sized to each.

**Point 1 (optional fields).** Agreed, and the branch already says so: a second commit before your report added the sentence naming exactly which optional keys this corpus does not emit. Growing fixtures for `from`, `kind`, `parent` and the rest is worth doing and is a bigger piece of work than this pin.

Re-checked by mutation after the change: deleting the `words` row from the session-object table still turns both commands red.

vshulcz:
Both of these landed before your follow-up — a9dbee14, merged as part of this PR.

**`score`.** That is the example the change was made for: each command is now checked against its own section plus the shared session-object table, and `score` appears in neither, so a `show` that emitted it would fail rather than borrow the search section's mention.

**The guard.** Now 18 for `show` and 10 for `last`. Your broken-fixture case — a session with no messages — produces about 15 keys on `show`, so it fails the guard instead of running to completion. 15 for both would have been wrong in the other direction: `last` emits twelve distinct keys on a healthy tree.

PR #1914, opened 2026-08-25
fix(nfcfold): fold Arabic hamza and madda

### LINKED ISSUE #1913: search: an Arabic word typed decomposed does not match the same word stored composed
`nfcfold` folds NFD to NFC so a word matches whichever form it was typed in (#1098). An Arabic word does not:

```
$ deja index                  # session holds أحمد precomposed
$ deja search أحمد --json     # NFC
tier=exact total=1
$ deja search أحمد --json     # NFD
tier=exact total=0
```

The scan that decides whether a string can fold at all, `hasCombining`, looks at U+0300–U+036F. Arabic's hamza and madda sit just past it, at U+0653–U+0655, so `Compose` returns the decomposed word untouched and the two forms key apart.

That is the whole of it — eight pairs, none of them a composition exclusion:

```
U+0622 alef with madda        = U+0627 + U+0653
U+0623 alef with hamza above  = U+0627 + U+0654
U+0624 waw with hamza above   = U+0648 + U+0654
U+0625 alef with hamza below  = U+0627 + U+0655
U+0626 yeh with hamza above   = U+064A + U+0654
U+06C0 heh with yeh above     = U+06D5 + U+0654
U+06C2 heh goal with hamza    = U+06C1 + U+0654
U+06D3 yeh barree with hamza  = U+06D2 + U+0654
```

Hebrew, which the queue item lumped in with it, needs nothing: its points never compose into a precomposed letter, so both forms already tokenise the same — checked, and an NFD Hebrew query finds the NFC text today.

Fix: let the scan see U+0653–U+0655 and let the generator emit those eight. The cost is one more range comparison per rune on a path that already walks the string.

vshulcz:
Closes #1913.

An Arabic word typed decomposed found nothing:

```
before   NFC query tier=exact total=1      NFD query tier=exact total=0
after    NFC query tier=exact total=1      NFD query tier=exact total=1
```

`hasCombining` — the scan that decides whether a string can fold at all — looked at U+0300–U+036F, and Arabic writes hamza and madda just past it at U+0653–U+0655, so `Compose` handed the decomposed word straight back.

Eight pairs, none of them a composition exclusion, all Arabic. Hebrew, which the queue item behind this lumped in, needs nothing: its points never compose into a precomposed letter, and an NFD Hebrew query already finds NFC text — measured, and the row is in the before/after above.

The scan is the hot path, so the shape of the test matters. Two independent range checks cost 610 → 908 ns on a 1100-character ASCII string (+49%), because ordinary text then pays two comparisons per rune. Guarding both ranges under one `r >= 0x0300` brings it back to 619 ns — within noise of the old scan, since every mark is above U+0300 anyway.

Table 775 → 783, re-derived from Python's `unicodedata` (nothing NFC disagrees with, no earlier pair lost), and the 3790-word fuzz still matches NFC everywhere.


vshulcz:
Right, and taken: the comment said eight and the table had eight, but the test listed six. The two rare ones — U+06C1 + U+0654 and U+06D2 + U+0654 — are in now, so the count in the comment and the count under test are the same number.

Everything else you checked matches what I measured, including the U+0500 boundary, which was the one I most wanted a second pair of eyes on.

PR #1919, opened 2026-08-25
feat: support authenticated embedding endpoints

chrisgeo:
## Summary

Add bearer-token authentication for remote/OpenAI-compatible embedding endpoints while preserving the existing unauthenticated Ollama/LM Studio flow.

## Changes

- Add `DEJA_EMBED_KEY` as the explicit API key for authenticated embedding endpoints.
- Send `Authorization: Bearer <key>` when configured.
- Fall back to the standard `OPENAI_API_KEY` only when `DEJA_EMBED_URL` points to `api.openai.com`.
- Never forward `OPENAI_API_KEY` implicitly to localhost or arbitrary third-party OpenAI-compatible endpoints.
- Add tests covering authenticated/unauthenticated requests, key precedence, OpenAI fallback, and credential-leak prevention.
- Document OpenAI Platform embedding configuration.

## Example
```sh
export OPENAI_API_KEY='sk-...'
export DEJA_EMBED_URL='https://api.openai.com/v1/embeddings'
export DEJA_EMBED_MODEL='text-embedding-3-small'

deja embed
```

For another authenticated OpenAI-compatible endpoint:
```sh
export DEJA_EMBED_URL='https://example.com/v1/embeddings'
export DEJA_EMBED_MODEL='embedding-model'
export DEJA_EMBED_KEY='...'

deja embed
```

Existing local endpoints continue to work without a key.


vshulcz:
Read the whole diff, ran the tests, and went looking for what they could not
say on their own. No blockers.

**What I checked, and how.** `New()` is the only construction site in the repo —
`cmd/deja/embed.go` (three calls), `doctor_report.go` and `bench.go` all go
through it, so nothing reaches an authenticated endpoint without the header.
`deja doctor --json` against `api.openai.com` with a key set prints neither the
key nor the URL, only state and model.

I did not take the host guard on trust. Replacing the scheme-and-hostname check
with a bare non-nil parse turns five of your cases red — plaintext, localhost,
LAN, third-party and the `api.openai.com.example.com` lookalike. That last one
is the case most reviews of this shape miss, and it is the reason I have no note
here.

**One second-order effect, and it is ours, not yours.** The README now tells
people to run `export DEJA_EMBED_KEY=…`, which puts a credential in a shell deja
indexes. A key with a provider prefix is stripped at ingest, and so is a
high-entropy one — but an opaque token in a variable ending in plain `_KEY` was
not, because every pattern wanted api_key, secret, token or password in the
name:

```
export DEJA_EMBED_KEY=9f2b7c4e1a8d6350bb91ee27ac4d0f83   # stored verbatim
```

That gap predates this PR; the docs here are just the first thing to point at
it. Fixed in #1958 rather than sent back to you.

**The one check that is red** is `windows manifests`, and it is GitHub's
infrastructure rather than your change — the runner could not download
`actions/setup-go` and gave up after three attempts. Rerun triggered.

vshulcz:
Thank you — merging.

The part worth naming: you did not stop at "send the header when a key is set."
The default that would have shipped in most versions of this change is
`OPENAI_API_KEY` going wherever `DEJA_EMBED_URL` points, which quietly hands a
platform key to whatever localhost service someone is testing against. You
scoped it to an HTTPS `api.openai.com` host, put `DEJA_EMBED_KEY` in front for
everything else, and then pinned the boundary with the lookalike-host case —
`api.openai.com.example.com`, the one that passes a `strings.Contains` check and
fails yours.

The credential-leak tests across all three error paths are the other half of it,
and they are why this needed no rework.

PR #1921, opened 2026-08-25
fix(usage): one week for every counter that shows one

### LINKED ISSUE #1920: usage: two different weeks on one status line, an hour apart after a clock change
deja has two ways of saying "a week ago" and shows them side by side. In a zone with daylight saving they differ by an hour for one week in each direction:

```
now=2026-11-03 12:00 EST   now-168h=2026-10-27 13:00 EDT   now-7d=2026-10-27 12:00 EDT   gap −1h
now=2026-03-10 12:00 EDT   now-168h=2026-03-03 11:00 EST   now-7d=2026-03-03 12:00 EST   gap +1h
now=2026-08-25 12:00 EDT   now-168h=2026-08-18 12:00 EDT   now-7d=2026-08-18 12:00 EDT   gap 0
```

Who uses which:

```
internal/usage/usage.go:287    DejaVuWeek     AddDate(0,0,-7)   calendar
internal/usage/usage.go:353    Week           Add(-7*24h)       fixed hours
internal/index/manifest.go:557 Overview       AddDate(0,0,-7)   calendar
internal/stats/stats.go:486    week counts    Add(-7*24h)       fixed hours
```

The status bar prints `Week` and `DejaVuWeek` on the same line, so on the week after a clock change one of them counts an event the other does not — the same kind of drift #1569 fixed for the day counters, which use local midnight and are therefore calendar too.

The brief already says calendar; the day counters already say calendar. Fix: one rule, calendar, read by all four.

vshulcz:
Closes #1920.

deja had two ways of saying "a week ago" and printed them on the same line. In a zone with daylight saving they differ by an hour, one week in each direction:

```
now=2026-11-03 12:00 EST   168h → 2026-10-27 13:00 EDT   seven days → 2026-10-27 12:00 EDT
now=2026-03-10 12:00 EDT   168h → 2026-03-03 11:00 EST   seven days → 2026-03-03 12:00 EST
now=2026-08-25 12:00 EDT   168h → 2026-08-18 12:00 EDT   seven days → 2026-08-18 12:00 EDT
```

`Week` cut at 168 hours, `DejaVuWeek` at seven calendar days, and the status bar shows both — so after a clock change one of them counted an event the other did not. `internal/stats` cut at 168 hours too, while the brief's "this week" and the day counters (local midnight) were already calendar.

`weekCut` says it once, calendar, and the counters read it. `usage.WeekCut` is the same rule for `internal/stats`.

Tests: the cut keeps the wall clock across both changes and an ordinary week, and an event half an hour into the week is inside it. They skip rather than fail where the machine has no tzdata, since that is not something this change can fix.

### DIFF
--- a/internal/stats/stats.go
+++ b/internal/stats/stats.go
@@ -483,7 +483,8 @@ func questionStemFor(text string) string {
 // only signal that separates memory that was used from memory that was merely
 // served, so the impact panel reads it too (#1062).
 func AgentCredits(ss []model.Session, now time.Time) (total, week int) {
-	weekCut := now.Add(-7 * 24 * time.Hour)
+	// The same week the status bar and the brief mean (#1920).
+	weekCut := usage.WeekCut(now)
 	for _, s := range ss {
 		for _, msg := range s.Messages {
 			if msg.Role != "assistant" || !strings.Contains(msg.Text, "deja-vu recalled") {
--- a/internal/usage/usage.go
+++ b/internal/usage/usage.go
@@ -280,26 +280,21 @@ func TodayDemand(indexDir string) (recalls, bytes, injected int) {
 	return recalls, bytes, injected
 }
 
-// weekCut is when "this week" opens: seven calendar days back, the same wall
+// WeekCut is when "this week" opens: seven calendar days back, the same wall
 // time. Not 168 hours — in a zone with daylight saving those differ by an hour
 // for one week in each direction, and the status bar prints the week counters
 // beside the déjà-vu count, so one of them counted an event the other did not
 // (#1920). The day counters here cut at local midnight and the brief cuts at
 // seven calendar days, so this is the rule the rest of deja already speaks.
-func weekCut(now time.Time) time.Time {
-	return now.AddDate(0, 0, -7)
-}
-
-// WeekCut is that rule for the other packages measuring the same week.
 func WeekCut(now time.Time) time.Time {
-	return weekCut(now)
+	return now.AddDate(0, 0, -7)
 }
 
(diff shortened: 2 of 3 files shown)


vshulcz:
Checked my own point 1 before the review lands, and the comment needed sharpening.

Two edge hours, measured in New York:

    now=2026-03-15 02:30 EDT  cut=2026-03-08 01:30 EST  (168h apart)
    now=2026-11-08 01:30 EST  cut=2026-11-01 01:30 EDT  (169h apart)

So the wall clock is kept except inside the hour a spring-forward removes: 02:30 did not happen that day, and Go normalises to the time the clock actually reached. The week is then exactly 168 hours — which is the old behaviour, in the one case where the new rule has nothing better to offer. The autumn case keeps the wall clock and the week is 169 hours long, which is what a person means by 'this week' on that day.

The doc comment said 'the same wall time' flatly; it now says what happens in that hour. No behaviour change — this is the last commit on the branch.

vshulcz:
Your finding was right and the test is rewritten around it.

The old one skipped in the autumn case — where the gap actually exists — and in spring re-asserted what the line above had already checked. There are now two, one per direction, and neither can skip past the thing it is for:

- after the autumn change, the fixed-hours cut lands an hour later than the calendar one, so an event in between is inside this week and outside the old rule's week — asserted both ways;
- after the spring change the disagreement runs the other way: the fixed-hours cut opens an hour early, so an event from before the week starts must not be inside it.

Both fail if `WeekCut` goes back to 168 hours — checked by making it, along with the wall-clock assertions in the first test.

Points 2 and 3 match what I found: every week in the tree now reads the one rule, and `internal/stats` already imported `internal/usage`, so nothing new was added to reach it.

PR #1923, opened 2026-08-25
fix(usage): rotation keeps the newest events rather than emptying the log

### LINKED ISSUE #1922: usage: a fortnight away and the first recall deletes the whole log
Rotation drops everything older than fourteen days. Come back from a fortnight away, and the first recall rotates a log whose events are all older than that — so the file goes to zero bytes and the whole record disappears at once:

```
before: events=4000 recalls=4000 since=2026-07-26 impact_recalls=4000
after:  bytes=0     events=0     recalls=0        since=—  impact_recalls=0
```

`deja stats --impact` then says:

> deja: no recall activity recorded yet — impact numbers appear once agents start recalling

which is not what happened. There was a great deal of activity; deja deleted it. `deja stats` loses its "Recalls served N since …" line the same way, and `since` — which #1889 documents as absent only when the log holds nothing — is now absent because rotation emptied it.

The point of rotation is to bound the file, not to erase the record. Fourteen days is the right window while there is anything inside it; when there is not, the newest events are all deja has and they cost nothing to keep — a few hundred of them is tens of kilobytes against a 1MB threshold.

Fix: rotation must not leave nothing behind. When the window is empty, keep the newest events instead.

vshulcz:
Closes #1922.

Rotation drops what is older than fourteen days. Come back from a fortnight away and every event is older than that, so the first recall on return emptied the file:

```
before  events=4000 recalls=4000 since=2026-07-26
after   bytes=0     events=0     recalls=0        since=—        (was)
after   bytes=59800 events=200   recalls=200      since=2026-08-04 (now)
```

`deja stats --impact` then printed "no recall activity recorded yet — impact numbers appear once agents start recalling" about a log that had held four thousand events, and `deja stats` lost its "Recalls served N since …" line the same way.

Bounding the file is the job; erasing the record is not. When nothing is inside the window, rotation now keeps the newest `keepAtLeast` events — 200, which measured 59.8 KB against the 1 MB threshold that triggers rotation in the first place.

Two tests: everything old keeps the newest (and it really is the newest, not an arbitrary slice — the test checks their age), and the ordinary case still drops what is outside the window when there is anything inside it.

### DIFF
--- a/docs/json-output.md
+++ b/docs/json-output.md
@@ -414,8 +414,10 @@ prior work, and `credited_aloud` the recalls an agent said out loud.
 more than the period from then to now. On a quiet machine that is every event
 there has ever been; once the log grows past 1MB it is rewritten keeping the
 last 14 days, and from then on the figures are a window whose start moves. Read
-`since` rather than assuming either. It is absent when the log holds nothing,
-which is the one case where the counts are all zero anyway.
+`since` rather than assuming either — it can also be older than 14 days, since a
+rewrite that would leave nothing keeps the newest few hundred events instead of
+emptying the file. It is absent only when the log holds nothing at all, which is
+the one case where the counts are all zero anyway.
 
 ## `deja blame <path> --json`
 
--- a/internal/usage/rotate_floor_test.go
+++ b/internal/usage/rotate_floor_test.go
@@ -1,0 +1,113 @@
+package usage
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+)
+
+// agedLog writes a log past the rotation size whose events are all the given
+// number of days old.
+func agedLog(t *testing.T, days int) (dir, path string) {
+	t.Helper()
+	dir = t.TempDir()
+	t.Setenv("DEJA_INDEX_DIR", dir)
+	path = Path(dir)
+	if err := os.MkdirAll(filepath.Dir(path), 0o700); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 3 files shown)


vshulcz:
Answered my own point 2 before the review lands, since the sentence was mine.

`docs/json-output.md` said `since` moves once the log passes 1MB and is "absent when the log holds nothing". After this change it can be older than fourteen days — a rewrite that would leave nothing keeps the newest few hundred instead — and it is absent only when the file really is empty. Both are in the paragraph now.

Point 1, the size, measured rather than reasoned: 200 events with 200-character session ids came to 59.8 KB. The floor only applies when the window is empty, so the file cannot grow by it — the next rotation either finds events inside the window (and keeps those) or keeps 200 again. Worst case is 200 events of whatever size the writer produced, and an event is bounded by what deja writes into it.

vshulcz:
All three taken, two of them before your report landed — the branch has moved three times since you started reading.

**Doc (your HIGH).** Fixed in the second commit: the paragraph now says `since` can be older than fourteen days, because a rewrite that would leave nothing keeps the newest few hundred instead, and that it is absent only when the file is truly empty.

**Stable sort (your MEDIUM).** Third commit, and for the reason you give — two events written in the same second have nothing but their order to tell them apart, and `deja log` prints it.

**Prose (your LOW).** Fair; the comment is four lines now and carries the measured figure rather than the rhetoric.

On your point 4: correct that neither test can fail on base, since the file is new. What fails on base is the behaviour they describe — the probe in the issue, 4000 events to zero bytes. I would rather say that plainly than pretend a new file is a red test.

Issue #733, opened 2026-08-01
log --json prints null on an empty log while every other --json prints a structure


vshulcz:
```
$ deja log --json          # nothing recorded yet
null

$ deja last --json
{"schema_version":2,"sessions":[]}
$ deja search nothing --json
{"schema_version":2,"tier":"exact","total":0,"hits":[]}
```

`null` is not an empty list: `len()` raises, `for e in …` raises, `jq '.[]'` errors. Every other machine-readable output in deja answers with its shape intact when there is nothing to report, and `log --json` is the one a script polls most — it is how you watch what the agent was given.

Smaller thing in the same command: a numeric argument out of range is reported as a flag.

```
$ deja log 0
deja: log: unknown flag "0"
$ deja log abc
deja: log: unknown flag "abc"
```

PR #1933, opened 2026-08-25
docs(json): describe deja log --json and pin its keys

### LINKED ISSUE #1924: docs: deja log --json has no section, and nothing says which lines appear in it
`docs/json-output.md` names `deja log --json` twice — in the stability policy, and in passing beside `blame` — and never says what it emits. It is the surface a script polls to see what deja fed an agent:

```json
[
  { "t": "2026-08-24T12:00:00Z", "kind": "recall", "bytes": 0, "empty": true },
  { "t": "2026-08-24T11:00:00Z", "kind": "hook", "bytes": 400, "sessions": 1, "raw": 4000 },
  { "t": "2026-08-24T10:00:00Z", "kind": "recall", "bytes": 900, "sessions": 2, "raw": 9000,
    "ids": ["s1", "s2"] }
]
```

Nothing documents the keys, that the array is newest first, that it is `[]` and never `null` when there is nothing (#733 fixed that deliberately), or what `--last` returns instead — a single snapshot object with a different shape again.

And since #1918 there is a rule about which lines appear at all: a line needs both a stamp and a kind. Someone reading the file, or writing one for a test, has no way to know that from the document.

Fix: give it a section like the others, and pin the keys to it the way `stats --json` (#1710), `--impact` (#1900) and `show`/`last` (#1911) are pinned.

vshulcz:
Closes #1924.

`deja log --json` was named twice in the document — once in the stability policy, once in passing beside `blame` — and described nowhere. It is the output a script polls to see what deja fed an agent:

```json
[
  { "t": "2026-08-24T12:00:00Z", "kind": "recall", "bytes": 0, "empty": true },
  { "t": "2026-08-24T11:00:00Z", "kind": "hook", "bytes": 400, "sessions": 1, "raw": 4000 },
  { "t": "2026-08-24T10:00:00Z", "kind": "recall", "bytes": 900, "sessions": 2, "raw": 9000,
    "ids": ["s1", "s2"] }
]
```

The new section says what each key is, which kinds appear and that an unrecognised one may too, that the array is newest first and is `[]` rather than `null` when empty (#733 settled that for the script that polls it), and what `--last` returns instead. It also carries the rule from #1918 that nothing else states: a line needs both `t` and `kind` to appear at all, and that is the same rule the counters read by, so the two cannot disagree about whether something happened.

`TestLogJSONKeysMatchTheDocumentedContract` pins the emitted keys to that section — it fails on the base branch because there is no section to read — and a second test holds the two properties a reader would otherwise have to discover: newest first, and an empty array rather than null.

Verified the pin by mutation: removing the `ids` mention from the section turns it red naming that key.

### DIFF
--- a/cmd/deja/log_contract_test.go
+++ b/cmd/deja/log_contract_test.go
@@ -22,11 +22,9 @@ func TestLogJSONKeysMatchTheDocumentedContract(t *testing.T) {
 	if err := os.MkdirAll(filepath.Dir(usage.Path(dir)), 0o700); err != nil {
 		t.Fatal(err)
 	}
-	when := time.Date(2026, 8, 24, 10, 0, 0, 0, time.UTC)
 	usage.RecordServedSessions(dir, usage.KindRecall, 900, 2, false, 9000, []string{"s1", "s2"})
 	usage.RecordResultRaw(dir, usage.KindHook, 400, 1, false, 4000)
 	usage.RecordResult(dir, usage.KindRecall, 0, 0, true)
-	_ = when
 	_ = tmp
 
 	out, err := captureRun(t, "log", "--json")
--- a/cmd/deja/log_contract_test.go
+++ b/cmd/deja/log_contract_test.go
@@ -100,3 +100,41 @@ func TestLogJSONIsAnArrayNewestFirst(t *testing.T) {
 		t.Errorf("the array is oldest first: %s then %s", events[0].Time, events[1].Time)
 	}
 }
+
+// `--last` is a different shape and the document describes it in prose; that
+// prose named Go field names rather than JSON keys until this test existed.
+func TestLogLastJSONKeysMatchTheDocumentedContract(t *testing.T) {
+	hermeticEnv(t)
+	dir := index.DefaultDir()
+	if err := os.MkdirAll(filepath.Dir(usage.Path(dir)), 0o700); err != nil {
+		t.Fatal(err)
+	}
+	usage.SnapshotPolicy(dir, usage.KindHook, "a digest of earlier work", 2, "local-only")
+
+	out, err := captureRun(t, "log", "--last", "--json")
+	if err != nil {
+		t.Fatal(err)
+	}
+	var snap map[string]any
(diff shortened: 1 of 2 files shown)


vshulcz:
Checked my own point 3 before the review reached it, and it was wrong twice.

`usage.Snapshot` marshals `t`, not `time` — I had copied the field name rather than the tag — and the sentence omitted `terms` (the query behind a déjà vu firing) and `into` (the agent session it went to). It also called `bytes` omitted when empty, which it is not: only `sessions`, `policy`, `terms` and `into` carry omitempty there.

Corrected in the second commit. Worth noting the pin does not cover `--last`: the contract test reads the array form only, so that sentence is held by reading, not by a test — which is exactly how this one got in.

vshulcz:
Closed the hole I named in my last comment rather than leaving it: `--last` is pinned too now, third commit.

It writes a real snapshot, reads back `deja log --last --json`, and requires every key to appear in the section — which is what would have caught the mistake I made in the first commit. Verified by mutation: rewording the sentence so `policy` is no longer named turns it red on that key.

vshulcz:
Point 1, checked myself at the writers rather than from the struct: `bytes` is `len(text)` at every call — the digest, the recall answer, the hook block, the tool line — except `remember` (`cmd/deja/mcp.go:520`), where it is the length of the note the agent wrote. So "what was served" was wrong for exactly one kind. The sentence now says the size of the text that changed hands, and names that exception.

Fourth commit. Point 2 is still yours.

vshulcz:
Your first HIGH is right and is fixed: only `raw` carries `omitempty`, so `bytes` is always present — and the example in the section shows `"bytes": 0`, which made the text contradict the block above it. It now says so explicitly, since a recall that served nothing is a fact rather than a missing field.

Your second HIGH (`time` against `t`) and the unused variable were both already handled — the `t` correction in the second commit, and the variable is gone in this one. That is the hazard of a branch that moves while you read; thank you for reporting them rather than assuming.

Points 1 and 2 in your later message match what I found at the writers, including that the doc does not claim `resource` is injected.

PR #1940, opened 2026-08-25
docs(json): say what a week and a day mean in the stats numbers

### LINKED ISSUE #1939: docs: stats --json reports a week and never says what a week is
`deja stats --json` reports `week_recalls`, `week_bytes`, `week_injected` and `week_agent_credits`, and nothing anywhere says what a week is. The keys appear in the example and in the key list; no sentence defines the window, and `docs/` does not mention "week" or "today" at all outside that block.

The rules live only in Go:

- a week is seven calendar days back at the same wall time (`usage.WeekCut`, #1920);
- a day starts at local midnight (`TodayWithInjections`, `TodayDemand`, `TodayRaw`);
- across a clock change the week keeps the wall clock, so it runs 169 hours in autumn and 167 in spring, and in the hour a spring-forward removes it opens at the time the clock actually reached.

A consumer comparing `week_recalls` against its own seven-day figure has no way to know which of those it is being handed, and until #1921 deja itself did not agree — the bar's week and its déjà-vu count used different rules.

Fix: say it where the keys are, and pin the sentence to the code the way the other prose in that file now is.

vshulcz:
Closes #1939.

`deja stats --json` reports `week_recalls`, `week_bytes`, `week_injected` and `week_agent_credits`, and nothing said what a week is. The keys were in the example and the key list; no sentence defined the window, and `docs/` did not mention "week" or "today" anywhere else.

The rules lived only in Go: seven calendar days back at the same wall clock (`usage.WeekCut`), a day from local midnight, and across a clock change the week keeps the wall clock — 169 hours in autumn, 167 in spring, and in the hour a spring-forward removes it opens at the time the clock actually reached. All in the reader's own timezone.

Worth saying in the document rather than only in the code, because deja itself had two rules for a week until #1921 and the two numbers on the status bar disagreed for one week in each direction.

`TestTheStatsSectionSaysWhatAWeekIs` fails on the base branch on all three phrases. It also checks the sentence against `WeekCut` itself — that the cut keeps the wall clock, and that it differs from a fixed 168 hours on the date it uses — so the prose cannot outlive the rule it describes. That second assertion is there because a doc test that only greps the document proves the document says something, not that the something is true.

### DIFF
--- a/cmd/deja/week_docs_test.go
+++ b/cmd/deja/week_docs_test.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/vshulcz/deja-vu/internal/usage"
+)
+
+// `deja stats --json` reports four week figures and the document never said
+// what a week is; the rules lived only in Go, and until #1921 deja itself did
+// not agree with them (#1939). This holds the sentence against the code it
+// describes rather than against itself.
+func TestTheStatsSectionSaysWhatAWeekIs(t *testing.T) {
+	doc, err := os.ReadFile("../../docs/json-output.md")
+	if err != nil {
+		t.Fatal(err)
+	}
+	section := docSection(t, string(doc), "## `deja stats --json`")
+	prose := strings.Join(strings.Fields(section), " ")
+
+	for _, want := range []struct{ what, phrase string }{
+		{"a week is seven calendar days", "seven calendar days"},
+		{"a day starts at local midnight", "local midnight"},
+		{"the wall clock is what is kept across a change", "wall clock"},
+	} {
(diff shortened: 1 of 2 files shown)


vshulcz:
All four taken.

The run-on is split, "opens" is gone — it now says the week *starts* at the time the clock reached, which is the verb that was missing — and the test compares the whole wall time rather than the hour, so a minute or second drifting would fail rather than merely showing up in the message.

The "not pedantry" line went too. You were right that it reads defensively; the fact does the work on its own: deja had two rules until #1921 and the two numbers on the bar disagreed, so the definition is worth stating.

Your point 3 in the earlier list — that all surfaces share one week — is the thing I most wanted checked, since the paragraph would otherwise imply an agreement that #1921 had only just created.

Issue #1941, opened 2026-08-25
search: an Arabic word with harakat is unreachable from the way it is normally typed


vshulcz:
Not a defect report — a measurement and a decision for you.

Arabic is normally typed without harakat, and text quoted from elsewhere often carries them. deja treats the two forms as different words, which is the same rule it applies to Latin. What differs is what the close tier does with them:

```
latin, accent dropped from the query   hits=1  tier=close
latin, accent added                    hits=1  tier=close
arabic, vowels dropped from the query  hits=0  —
arabic, vowels added                   hits=0  —
```

`café` and `cafe` are one edit apart, inside the close tier's limit of one for a short word. `كتب` and `كَتَبَ` are three edits apart, because each harakat is a rune of its own — so a Latin reader gets a second chance that an Arabic reader does not, from one rule applied evenly.

The options, none of which I have taken:

1. Leave it. The rule is consistent and the tier is a typo tier, not a diacritic tier.
2. Let the close tier treat a combining mark as a free edit — so a marked form is reachable from an unmarked query in every script. This widens the tier for Vietnamese and Greek too, and it is the change that most looks like "deja finally finds my Arabic".
3. Strip marks at index and query time for a separate tier of its own, leaving exact and close alone.

Option 2 is the one I would pick if it were mine, but it changes what "close" means for every script, so it is yours. #1940's table is now pinned by a test that asserts the tier as well as the hit, so whichever way this goes, the change will be visible rather than silent.

vshulcz:
Correction to the issue above, from the review of #1942: option 2 as I wrote it would not work.

The close tier does not merely reject the Arabic pair on distance — it never compares them. `candidates` walks only the length buckets within the edit limit of the query's length (`internal/index/manifest_cache.go:102`), so a three-rune query never sees the six-rune stored form, and `damerauDistance` is not reached. Treating a combining mark as a free edit would therefore change nothing on its own; the candidate walk would have to widen for marked forms too.

That makes option 3 — marks stripped into a tier of their own — the cheaper of the two changes, not the more expensive one. Still your call; I have only made the measurement honest.

vshulcz:
[...]

PR #1942, opened 2026-08-25
test(search): record which tier joins a word to its marked form

vshulcz:
A measurement pinned as a test. No behaviour change; the decision it raises is in #1941, for @vshulcz.

The queue item asked whether an Arabic word carrying a mark deja does not fold still tokenises the same on both sides. It does — and that is exactly why it cannot be found:

```
latin, accent dropped from the query   hits=1  tier=close
latin, accent added                    hits=1  tier=close
arabic, vowels dropped from the query  hits=0  —
arabic, vowels added                   hits=0  —
```

deja folds encodings and does not strip diacritics, so `café` and `cafe` are two words on the exact tier, as are `كتب` and `كَتَبَ`. What joins the Latin pair is the close tier — one edit apart, inside its limit for a short word. The Arabic pair is three edits apart, because each harakat is a rune of its own, so the same tier does not reach it. One rule, two outcomes.

`TestWhichTierJoinsAWordToItsMarkedForm` records that table and asserts the *tier* as well as the hit, so whichever way #1941 goes, the change shows up here rather than silently.

### DIFF
--- a/internal/index/diacritics_test.go
+++ b/internal/index/diacritics_test.go
@@ -44,10 +44,14 @@ func seedOneWord(t *testing.T, word string) string {
 // them.
 //
 // What does join them, for Latin, is the close tier: café and cafe are one edit
-// apart, inside its limit of one for a short word. Arabic harakat are separate
-// runes, so the two Arabic forms are three edits apart and the same tier does not
-// reach — one rule, two outcomes, and this records which is which rather than
-// leaving it to be discovered.
+// apart, inside its limit of one for a short word. The Arabic pair is out of
+// reach twice over — the candidate walk only visits tokens within the limit of
+// the query's length, so a three-rune query never sees the six-rune stored form,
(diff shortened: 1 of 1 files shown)


vshulcz:
Your first finding is the better explanation and it is in the comment now. More than a wording fix: it changes the advice in #1941, so I have corrected that too.

The close tier never compares the Arabic pair — `candidates` walks only the length buckets within the edit limit of the query's length, so a three-rune query never sees the six-rune form and `damerauDistance` is not reached. That means the option I said I would pick (treat a combining mark as a free edit) would change nothing by itself; the candidate walk would have to widen as well. The cheaper change is the one I had listed third.

Your second point is written in as the test's own limit: raising the edit limit alone would leave those rows at zero, and the comment says so rather than letting a future reader take the green rows as proof that nothing can help Arabic.

PR #1945, opened 2026-08-25
test(search): record what the close tier's length bucket still reaches

vshulcz:
A measurement, pinned. No behaviour change and no issue — the answer came out benign.

#1941 showed the close tier's length bucket putting an Arabic word's marked form out of reach. The obvious next question is what else that bucket hides in ordinary use. Measured, with a fixture whose only content word is the stored one so a stem elsewhere in the sentence cannot fake a hit:

```
case                            stored                      query             hits tier
suffix, one letter              parser                      parsers           1    close
suffix, two letters             retry                       retries           1    close
suffix, three letters           connect                     connecting        1    close
hyphen dropped                  work-tree                   worktree          1    close
hyphen added                    worktree                    work-tree         1    close
long word, one edit             connection_pool             connection_poll   1    close
long word, three edits          connection_pool             connectionpool    1    close
filename against its stem       parser.go                   parser            1    exact
long name against short stem    connection_pool_manager.go  connection        1    exact
stem against its filename       parser                      parser.go         0    —
```

Nine of ten reach. The one that does not is the direction where the query is the longer token: `parser.go` is one query token of nine runes, `parser` is six, and the walk only visits buckets within the limit — 7 to 11 — so the stored token is never compared.

The reverse works through a different door, worth knowing when reading this file: the index splits a filename into its parts, so a query for the stem meets a sub-token of `parser.go` on the exact tier. The two tokenisers are not symmetric, and that asymmetry is what saves nine of these rows.

The table is now a test, tier included, so a change to either tokeniser or to the bucket shows up as a change here.

### DIFF
--- a/internal/index/close_tier_reach_test.go
+++ b/internal/index/close_tier_reach_test.go
@@ -1,0 +1,80 @@
+package index
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"testing"
+
+	query "github.com/vshulcz/deja-vu/internal/query"
+)
+
+// seedCleanWord indexes a session whose only content word is the one given, so
+// a hit cannot come from a stem sitting elsewhere in the sentence.
+func seedCleanWord(t *testing.T, word string) string {
+	t.Helper()
+	tmp := t.TempDir()
+	claude := filepath.Join(tmp, "claude", "-proj")
+	if err := os.MkdirAll(claude, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	line, err := json.Marshal(map[string]any{
+		"type": "user", "sessionId": "s1", "cwd": "/proj",
+		"timestamp": "2026-08-20T10:00:00Z",
+		"message":   map[string]any{"role": "user", "content": "we looked at " + word + " yesterday"},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(claude, "s1.jsonl"), append(line, '\n'), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	t.Setenv("DEJA_CLAUDE_ROOT", filepath.Join(tmp, "claude"))
+	t.Setenv("XDG_CONFIG_HOME", filepath.Join(tmp, "config"))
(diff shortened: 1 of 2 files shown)


vshulcz:
Checked 1(b) myself at the source while you work, so the claim in the comment is exact rather than plausible.

`internal/index/retrieval.go:2858` — the index tokeniser breaks a run at any rune that is not a letter, a digit, `_` or `-`. A dot is therefore a boundary, so `parser.go` is stored as `parser` and `go`. `query.Tokens` keeps `parser.go` whole (measured). That is the whole of the asymmetry, and it is also why the hyphen rows are genuinely the close tier: `-` is kept by both sides, so `work-tree` and `worktree` are one edit apart rather than two tokens against one.

Nothing else — no touched-files index, no substring path — is doing that work; `touched` is metadata on the session and does not take part in term matching.

Point 2 is the one I still cannot settle from here without reading the tier assignment as carefully as you will.

vshulcz:
This is the finding I wanted and did not have: `close` is one label over three mechanisms, and my comment read as if it were one. Taken in full.

- The suffix rows are the **stem** tier — `oneSuffixStep`, "ies" to "y" — reporting `TierClose` at `retrieval.go:2310`.
- The hyphen and mistyped-word rows are the **fuzzy** tier, the edit-distance walk, reporting the same tier at `retrieval.go:2293`.
- The two exact rows are neither: the index tokeniser breaks a run at any rune that is not a letter, a digit, `_` or `-` (`retrieval.go:2858`), so `parser.go` is stored as `parser` and `go`, with identifier parts on top — while `query.Tokens` keeps `parser.go` whole.

All three are named in the comment now, with line references, and the test is `TestWhatAQueryStillReaches` in `query_reach_test.go` — the old name claimed a single tier that most of its rows do not use. Your point about a developer debugging close-tier logic and landing in stem behaviour is exactly the cost I was setting up.

One correction to your 1(b): `query.Tokens("parser.go")` really does return `["parser.go"]` — measured, not inferred. The splitting on dots you saw is the *index* side (`tokens` in retrieval.go). The two sides differ, which is the whole reason those rows pass, and the comment now says which function does which.

Issue #1917, opened 2026-08-25
usage: the two readers disagree about what an event is, and one of them moves 'since'


vshulcz:
The usage log has two readers and they disagree about what an event is. `read` (the counters) keeps a line with a non-zero stamp; `Events` (`deja log`) keeps one with a non-empty kind.

One line in the log, each way:

```
line              read  events  recalls  since
whole             1     1       1        2026-08-20
kind, no stamp    0     1       0        —
stamp, no kind    1     0       0        2026-08-20
neither           0     0       0        —
unknown kind      1     1       0        2026-08-20
```

Two of those rows are wrong in opposite directions:

- a line with a kind and no stamp is a row in `deja log`, dated January of year 1, and nothing to any counter;
- a line with a stamp and no kind is invisible in `deja log` and can never be counted — no counter has a case for it — yet it sets `since`, which is the window every figure on `deja stats --impact` is measured from. One such line pulls that window back to whatever date it carries.

Neither shape is written by deja: `recordFull` always writes both fields. They come from a hand-edited file, a partial write that happened to parse, or a batch from something that is not deja.

Fix: one rule for what counts as an event — a stamp and a kind — read by both. `deja log` stops showing year-1 rows, and a line that cannot be counted stops deciding what the counts are measured from.

PR #1947, opened 2026-08-25
fix(usage): ask the snapshots file the same question as the log

### LINKED ISSUE #1946: usage: the snapshots file still shows the shapes #1917 removed from the log
#1917 gave the usage log one rule for what a line is — a stamp and a kind — because its two readers had asked different questions and a half-written line was a row on one surface and nothing on the other. The snapshots file beside it (`index.db.injections.jsonl`) has a third reader, with a third rule: keep a line if it parses and carries a digest.

So the shapes #1917 removed are still readable here, through `deja log --last`:

```
newest line has no kind
  human: #  · 2026-08-20 13:00 · 1 session · 100 B
  json  : {"t":"2026-08-20T10:00:00Z","kind":"","sessions":1,"bytes":100,…}

newest line has no stamp
  human: # hook · 0001-01-01 02:30 · 1 session · 100 B
  json  : {"t":"0001-01-01T00:00:00Z","kind":"hook",…}
```

A heading with an empty kind, and a digest dated the year 1 — in local time, which is why it reads 02:30 rather than midnight.

Neither shape is written by deja: every writer passes a kind, and the stamp is `time.Now()`. They come from a half-written line, a hand edit, or a file from something else — the same provenance as the usage log's, and the same answer applies.

Fix: ask this file the same question, plus the digest it already requires. A snapshot with no digest is nothing to show; one with no stamp or kind is a heading deja cannot write.

vshulcz:
Closes #1946.

#1917 gave the usage log one rule for what a line is — a stamp and a kind. The snapshots file beside it had a third reader with a third rule: keep a line if it parses and carries a digest. So the shapes #1917 removed were still readable through `deja log --last`:

```
before, newest line has no kind    #  · 2026-08-20 13:00 · 1 session · 100 B
before, newest line has no stamp   # hook · 0001-01-01 02:30 · 1 session · 100 B
after                              deja: no injected digests recorded yet …
```

A heading with an empty kind, and a digest dated the year 1 — in local time, which is why it read 02:30 rather than midnight.

`Snapshot.usable` now asks the same question as the log, plus the digest this file exists to carry. Neither missing shape is written by deja: every writer passes a kind and the stamp is `time.Now()`.

Tests: the six shapes, and a second one that writes a snapshot through the real writer and reads it back, so the rule cannot tighten into "a kind I know" — an unrecognised kind is still a snapshot, for the reason it is still an event in the log.

The doc-comment guard caught me putting the new function between `snapshotsFrom`'s comment and its declaration — the same trap as #1798, #1811 and #1848. Fixed before pushing.

### DIFF
--- a/internal/usage/snapshot_shape_test.go
+++ b/internal/usage/snapshot_shape_test.go
@@ -1,0 +1,66 @@
+package usage
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// oneSnapshot writes a snapshots file holding exactly the given line.
+func oneSnapshot(t *testing.T, line string) string {
+	t.Helper()
+	dir := t.TempDir()
+	t.Setenv("DEJA_INDEX_DIR", dir)
+	p := SnapshotPath(dir)
+	if err := os.MkdirAll(filepath.Dir(p), 0o700); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(p, []byte(line+"\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	return dir
+}
+
+// The usage log asks one question of a line since #1917 — a stamp and a kind.
+// The snapshots file beside it asked a different one, so `deja log --last`
+// could print a heading with no kind, or a digest dated the year 1 (#1946).
+// Both are things deja cannot write: every writer passes a kind, and the stamp
+// is time.Now().
+func TestASnapshotNeedsAStampAKindAndADigest(t *testing.T) {
+	for _, c := range []struct {
+		name, line string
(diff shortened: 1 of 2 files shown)


vshulcz:
Took your closing note — it was the one thing in the change that would surprise someone later. Rotation rewrites the file from what the reader accepts, so it now also drops a line deja could not have written, and the file can shrink by more than the rotation alone explains. Said so where the rewrite happens rather than leaving it to be found.

On the wording: kept "half-written" over "malformed" deliberately. Malformed suggests the JSON is broken, and these lines parse perfectly — what is missing is a field, which is what a write cut short leaves behind.

PR #1948, opened 2026-08-25
test(stats): hold the card and the page to the week the report carries

vshulcz:
A guard, not a fix. No issue — the thing I went looking for is not happening.

deja had two rules for what a week is until #1921, and the queue item behind this asked whether the card and the page recompute one rather than reading the report. They do not: both render `WeekRecalls` from the report `deja stats --json` serialises, so all four surfaces answer with the same number.

One correction to my own premise while writing it. I had assumed the page prints no week at all — it does, through the same punchline the card's hero uses:

```
card: <text …>37</text> + "recalls handed to your agents this week"
page: "deja handed your agents memory 37 times this week."
```

Both are now pinned: hand the renderers 37 and they show 37; hand them 2 and the 37 is gone. That second half is what makes it a test rather than a screenshot — checked by mutation, replacing the card's source with a constant turns it red on both rows.

### DIFF
--- a/cmd/deja/card_week_test.go
+++ b/cmd/deja/card_week_test.go
@@ -1,0 +1,57 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/stats"
+)
+
+// Four surfaces print this week's recalls — `deja stats`, `--json`, the card and
+// the page — and deja had two rules for what a week is until #1921. They agree
+// now because each renders the field the report carries rather than asking the
+// usage log again. This holds that: hand the renderers a figure and it is the
+// figure they show.
+func TestTheCardAndThePagePrintTheWeekTheyWereHanded(t *testing.T) {
+	r := stats.Report{TotalSessions: 40, TotalMessages: 400, WeekRecalls: 37}
+
+	card := renderStatsCard(r)
+	if !strings.Contains(card, ">37</text>") || !strings.Contains(card, "this week") {
+		t.Errorf("the card does not show the week it was given")
+	}
+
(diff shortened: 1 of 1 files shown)


vshulcz:
Point 2 checked while you read, and it earned a row.

`formatStatNumber` groups thousands (`stats_metrics.go:31`), so a test that only ever passes two-digit figures says nothing about how a four-figure week is rendered — and a renderer that formatted its own number would look identical at 37. There is now a 1,234 row on both surfaces: the card must contain `>1,234</text>` and the page `1,234 times this week`.

That also answers the brittleness half of your question honestly: these are string matches against generated SVG and HTML, and they would break if the card wrapped the hero in a tspan or the punchline changed wording. I would rather they broke loudly there than silently miss a recomputed week — but it is worth knowing which kind of test this is.

vshulcz:
Point 3 taken, and it turned up something better than the gap itself.

The plain-text surface is pinned now, and while writing it: it prints the week with `%d` while the card and the page use `formatStatNumber`. A four-figure week therefore reads `1234 recalls` in the terminal and `1,234 times this week` on the card and the page — the same number, two spellings, on three surfaces a person sees in the same session.

I have pinned it as it is rather than fixing it. What a count looks like on screen is @vshulcz's to settle, and "make them agree" has two answers: group everywhere, or stop grouping. The test says so, and fails if this surface starts grouping on its own, so whichever answer he picks will be a deliberate change rather than a drift.

Your high-severity note on number formatting was already addressed in the second commit — a 1,234 row through both renderers — but you were right to raise it: the two-digit-only version proved nothing about the case that actually differs between surfaces.

Point 1 matches what I found: neither renderer names the `usage` package at all, so there is nothing there to recompute with.

Issue #1949, opened 2026-08-25
usage: 'into' is written on one injection path, read by nothing, and shown to nobody


vshulcz:
`Snapshot.Into` is described as "the agent session this went to, as the harness names it", with the reason spelled out: without it "the log says what was injected and never to whom, so the only way to tell whether a recall was used is the sentence the block asks the agent to say — measured on a real store, that sentence appears after 22 of 1218 injections, which measures reporting rather than use."

Today it does not do that job:

- **written on one path only.** The déjà-vu hook records it (`cmd/deja/hook_prompt.go:346`). The session-start hook records through `RecordDigestPolicy`, which passes `""` (`internal/usage/snapshots.go:97`) — though it is holding `input.SessionID` and uses it two lines away (`hook_context.go:128`, `:313`).
- **read by nothing.** No use of `.Into` anywhere in `cmd/`.
- **shown to nobody.** `deja log --last` does not print it; it reaches a person only through `--last --json`.

Two separable things, and the second is yours:

1. The session-start hook could pass the id it already has, so the field means the same thing on both paths. That is mechanical.
2. Whether `deja log --last` should name the session it went to — and in what words — is a wording call on a line that currently reads `# hook · 2026-08-24 11:00 · 2 sessions · 4.0 KB · policy: local-only`.

If the answer to 2 is "no", the field's comment should stop claiming the job it is not doing, and that is worth saying explicitly rather than leaving the next reader to measure it as I did.

vshulcz:
All three halves are true again, or answered:

- **written on one path.** The session-start hook passes the id it holds: cmd/deja/hook_context.go:313 and :385 both call the `…From` recorders with `input.SessionID`. Two paths still record an empty `into`, both honestly — `handoff` prints a digest for a chat that does not exist yet, and the antigravity hook's payload, as deja decodes it, carries `invocationNum` and `workspacePaths` and no session id. If someone with the host can show an id in that payload, that one is worth wiring; filing a guess would not help.
- **read by nothing / shown to nobody.** #2301 answered the wording question the other way: `deja log` prints ` · into: <id>` on the row (cmd/deja/log.go:103) and `--last` prints it in the header (cmd/deja/log.go:161), both pinned — log_into_row_test.go, log_last_into_test.go.

So the field does the job its comment claims on the paths that have an id to give it, and the surfaces a person types show it.

PR #1950, opened 2026-08-25
fix(usage): record where a session-start digest went

vshulcz:
Part of #1949 — the mechanical half. The wording half is left for you there.

`Snapshot.Into` exists so the log says to whom a digest went, and its comment gives the reason with a measurement: the agent's own "deja-vu recalled" line appears after 22 of 1218 injections, so counting that sentence measures reporting rather than use.

The field was not doing that job. It was written on one path — the déjà-vu hook (`hook_prompt.go:346`) — while the session-start hook, the commonest injection there is, recorded nothing, though it holds `input.SessionID` and uses it two lines away. Nothing in `cmd/` reads the field, and `deja log --last` does not print it.

This fixes the first part: `RecordDigestPolicyInto` takes the destination and the session-start hook passes the id it already has. `RecordDigestPolicy` stays and still writes nothing, so a caller that genuinely has no destination is not made to invent one — there is a test for each.

Still open in #1949, and yours: whether `deja log --last` should name the session it went to, on a line that currently reads

```
# hook · 2026-08-24 11:00 · 2 sessions · 4.0 KB · policy: local-only
```

If the answer is no, the field's comment should stop claiming the job — I would rather that were written down than measured again by the next person.

### DIFF
--- a/cmd/deja/hook_context.go
+++ b/cmd/deja/hook_context.go
@@ -248,7 +248,7 @@ func runHookContext(dir string, plain bool) error {
 		// does not depend on the digest having found anything.
 		if env := environmentBlock(dir, policy.ActivationAuto); env != "" {
 			out := frameRecall(env)
-			usage.RecordDigestPolicy(dir, usage.KindHook, out, 0, 0, policy.Load().Describe(policy.ActivationAuto))
+			usage.RecordDigestPolicyInto(dir, usage.KindHook, out, input.SessionID, 0, 0, policy.Load().Describe(policy.ActivationAuto))
 			if plain {
 				fmt.Fprintln(os.Stdout, out)
 				return nil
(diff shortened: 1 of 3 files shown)


vshulcz:
Reviewer was right on both counts, fixed in ac7fe548:

- the same hook's environment-block branch (line 251) still passed nothing while holding the same id — that is the same bug one branch over, so it now records too;
- I had overwritten `into_test.go` instead of adding to it, dropping the two déjà-vu-path tests. Restored; the new ones sit below them, and the no-destination test now also covers a harness that sends no `session_id` reaching the new form.

On the privacy question: nothing exports `.injections.jsonl`. Sync moves sessions, and session ids are already the index's own keys, so this is not a new kind of data on disk.

Wording half of #1949 still yours.

PR #1952, opened 2026-08-25
test(docs): end a doc section at the heading that ends it

### LINKED ISSUE #1951: docSection reaches past the section it names, so contract tests check most of the document
`docSection` is the helper four contract tests use to say "this key must be documented *here*, not somewhere in the file" (`cmd/deja/session_json_contract_test.go:15`). Its own callers state the intent:

> Not the whole document: a key documented for `search` alone would otherwise pass as documented here.

It does not hold that line. A section is cut only by the next heading of the *same* level, so a `###` section runs past the `##` heading that actually ends it, stopping only at the next `###` — and if there is none, at the end of the file.

Measured on today's `docs/json-output.md` (505 lines):

```
### The session object                           section =  374 lines of 505
## `deja search --json`                          section =   87 lines of 505
### `hits` is not a fixed window across tiers    section =   81 lines of 505
## `deja log --json`                             section =   49 lines of 505
```

`### The session object` is the last `###` in the file, so its "section" is everything from line 132 to the end — through `stats --json`, `doctor --json`, `--impact --json`, `log --json` and `blame --json`. Every contract test that adds that section to its own is really checking the whole back three quarters of the document.

The failure it lets through: a new key on the session object named like something in a later section — `policy`, `bytes`, `kind`, `digest` — passes the contract test with no line written about it anywhere near the session object. That is the exact case the test exists to catch.

Second thing while in there: the cut uses `if end > 0`, so a heading immediately followed by the next heading of its level (an empty section) is not cut at all and swallows the rest of the file too. Same failure, rarer cause.

Fix is the ordinary Markdown rule: a section ends at the next heading of the same **or shallower** level.

vshulcz:
Fixes #1951.

`docSection` is what four contract tests use to say a key must be documented *here* rather than anywhere in the file. It cut a section at the next heading of the same level only, so a `###` ran past the `##` that ends it and stopped at the next `###` — or, for the last one in a file, at the end of it:

```
### The session object                           section =  374 lines of 505
## `deja search --json`                          section =   87 lines of 505
```

Every test that adds the session-object section to its own was really checking the back three quarters of the document. Now a section ends at the next heading of the same or a shallower level, and the empty-section case (`end > 0`) no longer swallows the file either.

Tightening it turned up what it had been hiding: `deja show --json` emits `started` and `updated`, and neither its section nor the session-object table said so. They are always present, so the table — "these are the rest, each omitted when empty" — is the wrong home; they are in the section's example now, where the other always-present fields are.

Three tests: the reach on a fixture, the empty section, and the reach measured on the real `json-output.md` so the next `###` added at the end of the file cannot quietly re-open it.

### DIFF
--- a/cmd/deja/docsection_test.go
+++ b/cmd/deja/docsection_test.go
@@ -60,6 +60,34 @@ func TestAnEmptySectionTakesNothingFromTheNextOne(t *testing.T) {
 	}
 }
 
+// A `#` inside a fenced example is a comment, not a heading. `deja log` prints
+// its rows starting with one, and an example of that output in this document
+// would have ended every section it appeared in.
+func TestAHashInsideAnExampleIsNotAHeading(t *testing.T) {
+	doc := strings.Join([]string{
+		"## `deja log --json`",
+		"",
+		"```",
+		"# hook · 2026-08-24 11:00 · 2 sessions",
+		"```",
+		"",
+		"Log says `bytes`.",
+		"",
+		"## `deja blame --json`",
+		"",
+		"Blame says `occurrences`.",
+		"",
+	}, "\n")
+
+	got := docSection(t, doc, "## `deja log --json`")
+	if !strings.Contains(got, "`bytes`") {
+		t.Errorf("a comment line inside an example ended the section: %q", got)
+	}
+	if strings.Contains(got, "occurrences") {
+		t.Errorf("the section still reached into the next command: %q", got)
+	}
+}
+
 // Prose that names a heading is not the heading. The match was on the raw text
 // anywhere, so a sentence pointing at a section started the section there — mid
 // paragraph, and short by everything up to the real heading.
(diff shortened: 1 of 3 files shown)


vshulcz:
Three reviews, all shipping, three points taken:

- a `#` inside a fenced example is a comment, not a heading — `deja log` prints its rows starting with one, so an example of that output would have ended every section it sat in. Fenced blocks are skipped now, with a test.
- the section-size bound was a share of the document, which loosens as the file grows. It is a cap of 40 lines against a 25-line table.
- line-anchored was not line-terminated: a sentence opening with the heading's own text still started the section. A heading is a whole line now.

Left alone: on a CRLF document a section keeps a trailing `\r`. Nothing asserts on it and the document is LF.

The one open question is the reviewers' third point and it is yours: `started` and `updated` are always present, so they are in the section's example rather than the session-object table, whose preamble says "each omitted when empty". With sections now bounded correctly, that preamble's "the examples above" points at an example outside the table's own section. Worth a line saying which fields are always there, if you want one.

PR #1953, opened 2026-08-25
test(hooks): pin the receiver both injection paths record

vshulcz:
No bug — measured, then pinned.

The question was whether the two writers of `into` record the same kind of id. Driven end to end against one throwaway index, one session id, both hooks:

```
{"t":"…","kind":"hook",  "sessions":1,"bytes":726,"policy":"local+imported","into":"agent-session-777",…}
{"t":"…","kind":"dejavu","sessions":1,"bytes":741,"terms":[…],              "into":"agent-session-777",…}
```

Same string, and by construction: both read `session_id` from their own payload, and `hook_prompt.go:257` already compares that id against the index's own session ids, so it is one namespace.

What was missing is a test. The déjà-vu path has had one since the field was added; the session-start path was pinned only at the usage package's door, so the hook could stop passing the id and nothing would say so — which is how #1949 lived through a release. Two tests here: the session-start hook records the payload's session, and both hooks record the same string for one session. Reverting `hook_context.go:323` to the old call makes both fail, the second with "one session was recorded under two names".

### DIFF
--- a/cmd/deja/hook_context_into_test.go
+++ b/cmd/deja/hook_context_into_test.go
@@ -1,0 +1,131 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/vshulcz/deja-vu/internal/index"
+	"github.com/vshulcz/deja-vu/internal/usage"
+)
+
+// snapshotsInto is every receiver the injection log recorded, by kind.
+func snapshotsInto(t *testing.T, dir string) map[string]string {
+	t.Helper()
+	b, err := os.ReadFile(usage.SnapshotPath(dir))
+	if err != nil {
+		t.Fatalf("nothing was written to the injection log: %v", err)
+	}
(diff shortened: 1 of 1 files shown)


vshulcz:
Reviewer found three, all taken in af892977:

- the payload's `cwd` reaches `adoptHookCWD`, which exports `CLAUDE_PROJECT_DIR` into the process with a bare Setenv. Nothing here pinned it, so every later test in the package inherited a project directory that no longer exists — and the second test's own cwd was ignored, since the first one's was still set. Both tests pin it now.
- my stdout capture drained the pipe only after the hook returned, so a digest over the pipe buffer would have wedged the test. Uses the package's `captureStdout`, which drains while the hook writes.
- keying the snapshot scan by kind kept the last row, so a good row could cover a bad one. Every row of a kind has to agree now, and an unreadable line fails instead of being skipped.

Left as is: the first test is a subset of the second. It names the #1949 regression on its own, which is what I want it to say when it fails.

Kept honest: reverting hook_context.go:323 to the old call still fails both, the second with "one session was recorded under two names".

PR #1955, opened 2026-08-25
fix(log): stop calling an injection that sent bytes an empty result

### LINKED ISSUE #1954: deja log calls an injection that sent 501 bytes an empty result
A session start on a checkout with no sessions of its own still injects the environment block — the "this machine keeps hitting X" list. That injection is logged with `empty: true`, because the flag means "no project sessions went into this", and `deja log` renders that flag as `(empty result)`:

```
2026-08-26 00:26  hook           501 B  (empty result)
```

The same line says 501 bytes were injected and that the result was empty. A person reading their own log to see whether memory is working reads the second half.

The event itself is right, and the two other readers of it disagree about what it means:

```
usage log   {"t":"…","kind":"hook","bytes":501,"empty":true}
Totals      injections=1 injectedBytes=501 injectedSessions=0
Impact      injections=0 servedBytes=0
```

`Impact` skips it deliberately and says why in a comment — counting it made "N session starts began with project memory" claim memory that was not there. That is sound: the block is about the machine, not the project. `deja log` is the surface that is wrong, because it is reporting on the injection rather than on project memory, and 501 bytes did go out.

Mechanical half: the mark must not contradict the bytes on its own line — `(empty result)` belongs on an event that served nothing.

Wording half, and yours: an injection that carried the environment block and no project sessions is a real thing a person might want named, not just unmarked. If it deserves a word of its own on that line, it is your word, not mine.

Reproduced with `seedWalls(t, index.FrictionMinSessions)` plus `runHookContext`; snapshot in the hunt notes.

vshulcz:
Fixes #1954 — the mechanical half. The wording half is left for you there.

A session start on a checkout with no sessions of its own still injects the environment block. The event is logged `empty` because no project session went into it, and `deja log` turned that flag into a mark beside the bytes the block sent:

```
2026-08-26 00:26  hook           501 B  (empty result)
```

The flag is right; the sentence is not. Three readers of that one event, before this:

```
Totals   injections=1 injectedBytes=501 injectedSessions=0
Impact   injections=0 servedBytes=0      (skipped on purpose, comment says why)
deja log 501 B (empty result)
```

`Impact` is right to skip it — the block is about the machine, not the project, and counting it made "N session starts began with project memory" claim memory that was not there. `deja log` reports on the injection itself, and 501 bytes did go out, so the mark now follows the bytes: an event that served nothing keeps it, an event that served something does not.

`Event.Empty` had no comment saying which of the two it meant, and `json-output.md` said "an answer that found nothing". Both say what it is now.

One existing fixture changed: `log_test.go` recorded a recall of 42 bytes with `empty: true`, a shape no caller writes — a recall that found nothing serves nothing. It records a real empty recall and still asserts the mark.

Still yours, from the issue: an injection carrying the environment block and no project sessions may deserve a word of its own on that line rather than no mark at all. That word is yours.

### DIFF
--- a/cmd/deja/log.go
+++ b/cmd/deja/log.go
@@ -78,7 +78,11 @@ func runLogTo(w io.Writer, dir string, args []string) error {
 	}
 	for _, e := range events {
 		mark := ""
-		if e.Empty {
+		// Empty means no session went into the event, not that nothing was
+		// served: a session start on a checkout with no sessions of its own
+		// still injects the environment block, and marking that row put
+		// "(empty result)" beside the bytes it sent (#1954).
+		if e.Empty && e.Bytes == 0 {
 			mark = "  (empty result)"
 		}
 		sess := ""
--- a/cmd/deja/log.go
+++ b/cmd/deja/log.go
@@ -78,11 +78,7 @@ func runLogTo(w io.Writer, dir string, args []string) error {
 	}
 	for _, e := range events {
 		mark := ""
-		// Empty means no session went into the event, not that nothing was
-		// served: a session start on a checkout with no sessions of its own
-		// still injects the environment block, and marking that row put
-		// "(empty result)" beside the bytes it sent (#1954).
-		if e.Empty && e.Bytes == 0 {
+		if e.FoundNothing() {
 			mark = "  (empty result)"
 		}
 		sess := ""
--- a/cmd/deja/log_empty_test.go
+++ b/cmd/deja/log_empty_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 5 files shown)


vshulcz:
The blocker the first review raised is fixed, and the rule changed with it.

Its point was right: keying the mark on `Empty && Bytes == 0` dropped it on every empty recall, which serves the sentence saying nothing matched and so always has bytes. That is the case the flag exists for.

The mark now follows what the event was, not its size — `usage.Event.FoundNothing()` is `Empty && !injectedKind(kind)`. On a lookup the flag does mean the answer found nothing; on an injection it means no project session went in, which the row's own bytes already contradict.

I did not take the other suggestion — recording the environment block with `empty: false` — because `Impact` leans on exactly that flag to keep "N session starts began with project memory" from counting environment-only starts. Flipping it at the record site would have moved the wrong number instead of the wrong word.

Checked every site that writes an injected kind, since the rule turns the mark off for all three:

- `hook_prompt.go:346` (dejavu) records `len(ss)` sessions and the block only goes out when `ss` is non-empty
- `hook_tool.go:100` and `hook_tool_after.go:102` both pass `empty=false`
- `hook_antigravity.go:52` returns before recording when the digest is empty
- `hook_context.go:251` is the environment block — the one event this changes

Both wrong rules are pinned: the old `if e.Empty` fails the injection test, and `Empty && Bytes == 0` fails the recall test, which drives a real empty recall through `callMCPTool` rather than a hand-written event.

Two re-reviews were asked for on the replacement rule and neither reported, so the check above is mine.

PR #1956, opened 2026-08-25
ci: give every package a coverage floor

vshulcz:
The question was whether `internal/prompt` deserves a floor. It does — 84.8% — and it was not alone: eight of the twenty packages deja ships had no floor at all.

```
internal/atomicfile   47.2%   none        internal/mark        98.6%   none
internal/prompt       84.8%   none        internal/cjkfold    100.0%   none
internal/model       100.0%   none        internal/nfcfold    100.0%   none
internal/query       100.0%   none        internal/termwidth  100.0%   none
```

Five of the eight are fully covered, which is where an untested helper is cheapest to catch and where nothing was watching. The 87.5% total gate is not that watcher: prompt has been sitting below it for a while and the sum never said a word.

Floors set just under today's number, as the neighbours are. The fully covered ones get 99 rather than 100 — the point is to catch a package sliding, not to fail a build over one defensive branch. bench and policy keep their 100 because they already had it.

Ran the gate's own loop against a real `go test -coverprofile` run before pushing: every package passes, `fail=0`.

`internal/atomicfile` at 47.2% is pinned where it stands rather than raised here — a package whose whole job is not losing a file on a crash, half covered, is its own piece of work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,24 +130,28 @@ jobs:
       # wrapper with no caller was enough to trip it (#1921). If you are adding
       # a helper to one of those, add its test in the same commit.
       #
-      # The list covers every package deja ships. Eight had no floor at all,
-      # five of them fully covered — the case where an untested helper costs
-      # nothing to notice, and the total floor of 87.5% would not have: prompt
-      # sat at 84.8% under it. The fully covered ones are pinned at 99 rather
-      # than 100 so a single defensive branch does not stop a build; the whole
-      # point is to catch a package sliding, not to freeze it.
+      # Every package with coverable statements is listed. The exception is
+      # internal/jsonout, one const and no statements, which prints no coverage
+      # line at all and so cannot be floored.
+      #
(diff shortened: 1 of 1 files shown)


vshulcz:
Reviewer was right on the number and on the comment, both fixed in 3607ff06.

The 99 was justified by slack that does not exist: these packages are 10 to 74 statements, so one uncovered line costs between 1.4 and 10 points and fails a 99 floor just as it fails 100. They are pinned at 100 now, the same as bench and policy, and the comment says what is actually true — every floor here is today's number rounded down, so in a small package one statement fails the build.

Also corrected: the list does not cover every package deja ships. `internal/jsonout` is one const with no statements, prints no coverage line, and cannot be floored at all; it is named as the exception rather than quietly missing.

Gate re-run against a real coverprofile after the change: `fail=0`.

Windows failed once on the way here for a reason outside the repo — the runner could not resolve GitHub's own host to download `actions/setup-go`. Re-run, green.

PR #1957, opened 2026-08-25
test(atomicfile): cover WriteStream, which had none

vshulcz:
`internal/atomicfile` was the lowest-covered package deja ships, and the reason turned out to be one function:

```
publish      100.0%
sweepStale    87.5%
Write         65.0%
WriteStream    0.0%
```

`WriteStream` is the path the embedding sidecar takes — megabytes of vectors, written through this package precisely because a half-written one is worse than yesterday's: it parses, just shorter. It had no test at all, while `Write` beside it had five.

Four now, mirroring what `Write` is held to: it publishes exactly what the writer wrote and applies the mode (`CreateTemp` makes 0600, the chmod is what fixes it), a writer that gives up halfway leaves the previous file untouched, a rename that cannot land leaves no temp behind, and a missing directory fails before the writer is ever called.

47.2% to 77.4%; the floor moves with it. What is still uncovered in both functions is the same set — `CreateTemp`, `Close` and `Chmod` failing — which needs a filesystem that fails on command rather than a test.

No behaviour change, and nothing surprised: the tests pass as written against the current code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,7 +145,7 @@ jobs:
         shell: bash
         run: |
           declare -A floor=(
-            [cmd/deja]=88 [internal/atomicfile]=47 [internal/bench]=100
+            [cmd/deja]=88 [internal/atomicfile]=77 [internal/bench]=100
             [internal/cjkfold]=100 [internal/digest]=90 [internal/embed]=93
             [internal/index]=90 [internal/mark]=98 [internal/model]=100
             [internal/nfcfold]=100 [internal/peers]=76 [internal/policy]=100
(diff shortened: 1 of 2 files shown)


vshulcz:
Reviewer ran all four mutations against the tests and each one is caught — dropping the chmod, publishing despite the writer's error, skipping the temp removal on a failed rename, and swallowing the writer's error. Both of its points taken in 42d94ee7.

The one that mattered was factual and mine: I wrote that a half-written sidecar "parses, just shorter". It does not. The header carries the vector count and `binary.Read` fails mid-vector, so a truncated sidecar decodes as nothing — which is what `sidecar.go:235` has said all along. The comment says that now, and so does this description.

Also folded: the two inline temp-file loops in the older tests now use the same helper, which sits below the tests rather than between them.

Correction to the body above: a half-written sidecar decodes as nothing, not as a shorter index.

PR #1959, opened 2026-08-25
test(hooks): hold the session start to one project across worktrees

vshulcz:
No bug — measured, then pinned.

The question was `adoptHookCWD`, which exports `CLAUDE_PROJECT_DIR` for the whole process and never puts it back. In production that costs nothing: every hook is its own process and the export dies with it. It only bites tests sharing one process, which #1953 already pinned.

The case where the precedence could actually hurt is a linked worktree — the session runs in one checkout while the project's history was recorded in another. Driven end to end on a real repo with a worktree, both open with the same memory:

```
=== session-start hook run FROM THE MAIN CHECKOUT
✓ recalled from claude session · 3 days ago
  - Session: **…/repo** `m3`

=== session-start hook run FROM THE LINKED WORKTREE (same repo)
✓ recalled from claude session · 3 days ago
  - Session: **…/repo** `m3`
```

`ProjectNameCandidates` is what makes that true and has its own test in `internal/digest`. The hook is where the project is actually decided — it reads the environment, or the payload's cwd, and everything downstream follows that one string — and nothing held it to the same rule. This is that test, both checkouts through `runHookContext` itself.

It is not free-passing: stub `gitWorktreeRoots` to return nothing and the worktree subtest fails with an empty digest, while the main checkout keeps working.

### DIFF
--- a/cmd/deja/hook_worktree_test.go
+++ b/cmd/deja/hook_worktree_test.go
@@ -22,14 +22,14 @@ func TestASessionStartedInAWorktreeGetsTheProjectsMemory(t *testing.T) {
 	}
 	withStatsStores(t)
 	tmp := t.TempDir()
-	main := filepath.Join(tmp, "checkout")
+	repo := filepath.Join(tmp, "checkout")
 	side := filepath.Join(tmp, "checkout-fix")
-	if err := os.MkdirAll(main, 0o755); err != nil {
+	if err := os.MkdirAll(repo, 0o755); err != nil {
 		t.Fatal(err)
 	}
 	git := func(args ...string) {
 		t.Helper()
-		cmd := exec.Command("git", append([]string{"-C", main}, args...)...)
+		cmd := exec.Command("git", append([]string{"-C", repo}, args...)...)
 		cmd.Env = append(os.Environ(),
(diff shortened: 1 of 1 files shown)


vshulcz:
Both taken in 5fca3a80: the two paths go through `filepath.ToSlash`, and the local `repo` no longer shadows `main`.

One correction to the finding, in fairness to the record: the windows leg was green before the fix as well as after, so the invalid-escape failure did not actually reproduce on CI. The reasoning still holds — a backslash path in a JSON string literal is a run of invalid escapes, and `hook_cwd_test.go:52` does exactly this — so it goes in as written rather than waiting for the day it does bite.

The reviewer's stronger mutation is the one worth recording: replacing `git worktree add` with `git init` in a sibling directory of the same name fails the worktree subtest with an empty digest while the main checkout keeps working. That rules out the basename doing the work, which my own stub of `gitWorktreeRoots` only implied.

Noted and not fixed: `gitWorktreeRoots` gives `git worktree list` 400ms, and the windows runner is slow enough that a lost fork would read as a regression here. If this test ever goes red on windows alone, that is the first thing to look at.

PR #1961, opened 2026-08-25
fix(doctor): tell a corrupt sidecar apart from no sidecar

### LINKED ISSUE #1960: doctor reports a corrupt vector sidecar as no sidecar at all
A vector sidecar deja cannot parse looks exactly like one that was never built. `deja doctor` says nothing about the file, and search quietly serves lexical results.

That is the failure #1319 was about: two embeds racing left a sidecar that decodes as nothing. The write side was fixed; the read side still has no way to tell anyone.

Reproduced on a throwaway index, writing a sidecar with a good header and a third of one vector record after it — a process killed mid-write:

```
=== A. no sidecar at all
Embedding:
  endpoint   unavailable

=== B. sidecar truncated mid-file
Embedding:
  endpoint   unavailable          <- same screen, same words

=== C. what a search says about it
[claude] proj · 3d ago · s3 — 8 matches      <- lexical, silently

=== D. doctor --json, embed section
(no embed key)
```

`collectDoctorEmbed` (cmd/deja/doctor_report.go:313) treats every `embed.Read` error the same way: return the bare report, or `nil` when no endpoint is configured. `maybeSemantic` and `maybeRerank` (cmd/deja/embed.go:99, :75) return the lexical hits and say nothing — while the neighbouring failure, an unreachable endpoint, does print "semantic rerank unavailable; using lexical order".

Doctor's own comment says why this matters: it "is the screen someone opens to find out why semantic results stopped".

Every other section of the report already draws this distinction. `sync.state`, `policy.state` and a session store's state are each `unreadable` with an `error` saying why, precisely so a file deja cannot parse is not reported as a file that is not there. The embed section is the one that does not.

Mechanical half: doctor should report an unreadable sidecar as unreadable, with the reason, and not as absent.

Wording half, and yours: whether a search that falls back because the sidecar is corrupt should say so on the spot, the way an unreachable endpoint does. It is one line on a screen a person reads mid-task, so it is your call rather than mine.

vshulcz:
Fixes #1960 — the mechanical half. The wording half is left for you there.

A vector sidecar deja cannot parse looked exactly like one that was never built:

```
=== no sidecar at all                    === sidecar truncated mid-write
Embedding:                               Embedding:
  endpoint   unavailable                   endpoint   unavailable
```

That truncated file is the failure #1319 was about — two embeds racing, one renaming a file the other was still writing. The write side was fixed then; the read side still had no way to tell anyone.

Every other section of the report already draws this line: `sync.state`, `policy.state` and a session store's state are each `unreadable` with an `error` saying why, exactly so a file deja cannot parse is not reported as a file that is not there. Embedding is the one that did not.

Now:

```
Embedding:
  sidecar    unreadable — unexpected EOF

"embed": { "state": "unreadable", "error": "unexpected EOF", "coverage": 0 }
```

It is the sidecar's own line rather than the endpoint's, and it is reported whether or not an endpoint is configured — the file is the evidence, and the endpoint is not what broke it. A missing sidecar still says nothing, which is the other test.

The fixture writes a real header and stops a third of the way into the first vector record, so `Read` fails with `unexpected EOF` rather than on the magic — I had the magic wrong at first, which would have tested a different thing entirely.

Still yours, from the issue: whether a search that falls back to lexical because the sidecar is corrupt should say so on the spot, the way an unreachable endpoint already does.

### DIFF
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -303,16 +303,20 @@ func doctorEmbed(w io.Writer, r doctorEmbedReport) {
 	fmt.Fprintln(w, "Embedding:")
 	// The sidecar's own line, not the endpoint's: the endpoint may be fine and
 	// the file still unparseable, and saying "endpoint unreadable" would send
-	// the reader after the wrong thing (#1960).
-	if r.State == "unreadable" {
+	// the reader after the wrong thing. Both lines print, because whether an
+	// endpoint is configured is what decides if re-running `deja embed` fixes
+	// it (#1960).
+	if r.Sidecar == "unreadable" {
 		fmt.Fprintf(w, "  sidecar    unreadable — %s\n", r.Error)
-		return
 	}
 	if r.Model == "" {
 		fmt.Fprintf(w, "  endpoint   %s\n", r.State)
 		return
 	}
 	fmt.Fprintf(w, "  endpoint   %s/model=%s/dim=%d\n", r.State, r.Model, r.Dim)
+	if r.Sidecar == "unreadable" {
+		return
+	}
 	fmt.Fprintf(w, "  sidecar    coverage=%.1f%%\n", r.Coverage)
 }
 
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -301,6 +301,13 @@ func hookEventWired(hooks map[string]any, event, command string) bool {
 
 func doctorEmbed(w io.Writer, r doctorEmbedReport) {
 	fmt.Fprintln(w, "Embedding:")
+	// The sidecar's own line, not the endpoint's: the endpoint may be fine and
+	// the file still unparseable, and saying "endpoint unreadable" would send
+	// the reader after the wrong thing (#1960).
+	if r.State == "unreadable" {
+		fmt.Fprintf(w, "  sidecar    unreadable — %s\n", r.Error)
+		return
+	}
 	if r.Model == "" {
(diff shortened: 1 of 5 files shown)


vshulcz:
All four taken in 0c020ce0. The first two turned out to be one problem, and the fix is better for it.

Overloading `state` was wrong: a consumer testing `state == "reachable"` to learn the endpoint is up would have started seeing "unreadable", and — as the review found — `model` was still being emitted underneath it from `embed.New`, so the JSON and the text disagreed about the same report. The endpoint's state and the file's are two things, so they are two fields now:

```
Embedding:
  sidecar    unreadable — invalid vector sidecar
  endpoint   unavailable

{ "state": "unavailable", "sidecar": "unreadable", "error": "invalid vector sidecar", "coverage": 0 }
```

Both lines print, since whether an endpoint is configured is what tells the reader if re-running `deja embed` can fix it. The coverage line is suppressed when the sidecar is unreadable — a coverage figure read off a file deja could not parse is a number with nothing behind it.

Mutation (c) surviving was the sharpest catch: `strings.Contains(out, "unreadable")` passes on "endpoint unreadable", which is the exact misdirection the comment beside it warns about. Both assertions are tightened — the sidecar is named, the reason is required, and the endpoint line must still be there. The missing-sidecar test asserts `got == nil` rather than merely not-unreadable.

Doc sentence rewritten; the repeated line is gone from it. And the new key needed adding to the doctor contract test's own list, which is a second place the contract lives — worth knowing.

PR #1963, opened 2026-08-25
fix(usage): count an injection that carried no project session

### LINKED ISSUE #1962: statusline says 0 B injected on a day it injected 502 bytes
On a machine whose project has no history of its own, the session start still injects the environment block. On such a day the statusline says:

```
deja · no recalls yet today · 0 B injected
```

502 bytes were injected. The line three lines above that number in `statusline.go` says why this matters:

> Injections are not recalls, but they are the whole day for someone who lives on auto-recall: saying "0 B injected" while memory has been arriving since morning is the kind of untrue line #1403 is about.

The cause is `Event.Empty`, which means "no session went into this event" and not "nothing was served" — #1954 fixed `deja log` reading it the second way. `TodayDemand` (usage.go:288) and `Week` (usage.go:385) still `continue` on `e.Empty` for every kind, so an injection that carried the environment block and no project session is dropped from both.

Five readers of one event, three answers:

```
event        {"t":"…","kind":"hook","bytes":501,"empty":true}

Totals       injections=1 injectedBytes=501
TodayWith    recalls=1 bytes=501 injected=501
Week         recalls=0 bytes=0 injections=0 injectedBytes=0
TodayDemand  recalls=0 bytes=0 injected=0
Impact       injections=0 servedBytes=0
```

`Impact` is right to abstain — its sentence is "N session starts began with project memory", and this one did not. The other two are not making that claim. `Week`'s own comment says `injected` counts "the hook deliveries deja pushed unprompted", which is exactly what this was, and today's 501 bytes are not inside the week that contains today.

Fix: skip on the lookup meaning of the flag, not on the flag — `FoundNothing()` already exists for this, added in #1954. A recall that matched nothing stays out of the demand-side counts, and an injection stays counted.

Reproduced with `seedWalls(t, index.FrictionMinSessions)` and `runHookContext`; snapshot in the hunt notes.

vshulcz:
Fixes #1962.

On a machine whose project has no history of its own, the session start still injects the environment block, and on such a day that block is the whole of deja's work. The statusline said:

```
deja · no recalls yet today · 0 B injected
```

502 bytes had gone out. Its own comment, three lines above that number, says saying this "while memory has been arriving since morning is the kind of untrue line #1403 is about".

Same cause as #1954: `Event.Empty` means no session went into the event, not that nothing was served. `TodayDemand` and `Week` still skipped on the raw flag, so five readers of one event gave three answers:

```
Totals       injections=1 injectedBytes=501
TodayWith    recalls=1 bytes=501 injected=501
Week         recalls=0 bytes=0 injections=0 injectedBytes=0
TodayDemand  recalls=0 bytes=0 injected=0
Impact       injections=0 servedBytes=0
```

Both now skip on `FoundNothing()`, which #1954 added for exactly this distinction. The demand-side numbers are unchanged — a recall that matched nothing still stays out of `recalls` and `bytes`, and there is a test for that half too. `Impact` keeps its own rule: its sentence is "N session starts began with project memory", and this start did not.

`week_injected` and the day are documented as what they now count.

### DIFF
--- a/cmd/deja/statusline_injected_test.go
+++ b/cmd/deja/statusline_injected_test.go
@@ -29,9 +29,12 @@ func TestTheStatuslineCountsTheBlockItInjected(t *testing.T) {
 	if err := runStatusline(dir, strings.NewReader(""), &line); err != nil {
 		t.Fatal(err)
 	}
-	if strings.Contains(line.String(), "0 B injected") {
-		t.Errorf("the statusline reports nothing injected on a day it injected %d bytes:\n%s",
-			len(out), line.String())
+	// Asserted positively: runStatusline returns early on a warmup, a policy
+	// rule or a quiet week, and every one of those lines would satisfy "does
+	// not say 0 B injected" while saying nothing about the block at all.
+	got := strings.TrimSpace(line.String())
+	if want := "deja · no agent recalls today · 501 B injected"; got != want {
+		t.Errorf("statusline = %q,\n                 want %q", got, want)
 	}
 }
 
--- a/cmd/deja/statusline_injected_test.go
+++ b/cmd/deja/statusline_injected_test.go
@@ -1,0 +1,67 @@
+package main
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/index"
+	"github.com/vshulcz/deja-vu/internal/usage"
+)
+
+// A machine whose project has no history of its own still gets the environment
+// block at every session start, and on such a day that block is the whole of
+// deja's work. The statusline said "0 B injected" through it — the untrue line
+// its own comment cites #1403 for — because `Empty` means no session went in,
+// not that nothing was served (#1962).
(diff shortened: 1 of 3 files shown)


vshulcz:
Point 4 taken in c4f5cc29 — the right catch. The assertion was negative, and `runStatusline` returns early on a warmup, a policy rule and the quiet-week branch, every one of which satisfies "does not say 0 B injected" while saying nothing about the block. It asserts the whole line now:

```
deja · no agent recalls today · 501 B injected
```

Corrected while there: 501 is the event, 502 was the block plus the newline `Fprintln` adds. The description above says 502; the number that matters is 501.

Point 2 is a real finding and not this PR's: `Impact` skips `ServedBytes` and `RawBytes` along with the count, while its comment justifies only the count — those bytes were served. Queued rather than folded in, since it moves a number on a different screen than the one this fixes.

The rest of the review matched what I had: `KindHook` with the flag set is the only shape whose treatment changes, and it is written in one place, inside `if env := environmentBlock(...); env != ""`, so it always carries bytes.

PR #1964, opened 2026-08-25
test(usage): pin why the environment block stays out of the impact ratio

vshulcz:
No bug — and the change that was proposed would have made a different number wrong.

Reviewing #1963 turned up that `Impact` skips `ServedBytes` and `RawBytes` along with `Injections++` for an environment-block event, while the comment justifies only the count. That reads like an oversight. It is not.

The block is a summary of what this machine keeps hitting, not a digest of transcripts, so it is recorded with no raw size behind it — `hook_context.go:251` passes `0, 0`. The impact line it would feed says "N served instead of M of raw transcripts". Adding the block's bytes moves that numerator against an unchanged denominator:

```
three recalls (900 B each, from 9 kB of transcripts) and ten blocks (501 B each):

  as it stands:       served=2700  raw=27000  ratio=10.0
  counting the block: served=7710  raw=27000  ratio=3.5
```

Ten blocks a day is an ordinary day on a project with no history of its own, and it would turn "10× less" into the short-sessions message — deja understating its own saving with bytes that never replaced a transcript.

So both fields stay out, and the comment now says why rather than leaving the next reader to work it out or "fix" it. Two tests: ten blocks move neither figure while the recalls still count, and a session start that did carry project memory counts bytes and all — the skip is about the block, not about session starts.

Mutated to the proposed alternative to check the test earns its place: it fails with exactly those numbers.

### DIFF
--- a/docs/json-output.md
+++ b/docs/json-output.md
@@ -430,7 +430,11 @@ one.
 `recalls` counts agent-initiated recalls that returned matches, `injections`
 session starts that began with project memory. `served_bytes` is what the
 digests actually returned and `raw_bytes` the source transcripts they were
-distilled from, so the ratio is how much reading deja saved. `reused_twice` is
+distilled from, so the ratio is how much reading deja saved. The two go
+together: a session start that carried only the environment block is in neither,
+because that block is a summary of what the machine keeps hitting rather than a
+digest of transcripts. `deja stats` counts its bytes, which is the number for
+everything deja handed over. `reused_twice` is
 sessions agents recalled two or more times, `dejavu_moments` prompts matched to
 prior work, and `credited_aloud` the recalls an agent said out loud.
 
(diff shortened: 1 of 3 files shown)


vshulcz:
All four taken in f2b2ada8.

The one that stings is right: I quoted a line the code does not print. `stats_impact.go:84` formats the ratio with `%.0f`, so 27000/7710 renders "4× less", not "3.5×". The comment says a tenfold saving reads as fourfold and quotes nothing.

The redundant test is gone. `impact_empty_hook_test.go` already holds both halves — I went looking for a gap that was not there, and the ratio test stays only because its value is the numbers in it, not the coverage.

Doc clause added: `served_bytes` and `raw_bytes` go together and the block is in neither, while `deja stats` counts its bytes for anyone who wants everything deja handed over. That asymmetry was the one real argument for counting it, and now it is written down instead of inferred.

Point 2 is the interesting one and I agree with where it lands: the block is derived from transcripts, so a raw size could be attributed — but it would claim replay the agent would never have done, and inflate the ratio in the other direction. Skipping both sides stays the honest choice.

Review meta-commentary trimmed out of the test comment.

PR #1966, opened 2026-08-25
fix(usage): close a truncated snapshot line before appending

### LINKED ISSUE #1965: one interrupted snapshot write loses every digest after it
A process killed between writing a snapshot record and its newline leaves half a line. The next injection appends onto it, so one line holds two JSON objects and parses as neither — and because the reader drops what it cannot parse, both digests are gone:

```
one injection recorded, killed before the newline, then the next injection:

  after the kill:            1 snapshot readable
  after the next injection:  0 readable
  file has 2 lines
```

`deja log --last` then reports no injected digests on a machine that has just served two.

This is #1901 in the sibling file. The usage log learned to close a truncated line before appending — one byte of reading, `endsMidLine` in `internal/usage/usage.go` — with the reasoning spelled out there: "A process killed between a record and its newline costs that record… It cost the next one too: appended onto the partial line, so that line no longer parsed either and a recall deja had just served went missing from every count."

`snapshotWriteInto` (`internal/usage/snapshots.go:118`) opens `O_CREATE|O_WRONLY|O_APPEND` and writes. It never reads the last byte, and cannot: `endsMidLine` needs a readable descriptor, which is why the usage log's own open was changed to `O_RDWR` in #1901.

The snapshot file is the one deja reads the actual injected text back out of, so losing it is more visible than losing a count: the count screens are approximate by nature, `deja log --last` is either the digest or nothing.

Reproduced against `RecordDigestPolicy` in a throwaway index; snapshot in the hunt notes.

vshulcz:
Fixes #1965.

A process killed between a snapshot record and its newline leaves half a line. The next injection appended straight onto it, so one line held two objects and parsed as neither — and the reader drops what it cannot parse, so both digests went:

```
one injection recorded, killed before the newline, then the next injection:

  after the kill:            1 snapshot readable
  after the next injection:  0 readable
```

`deja log --last` then reports no injected digests on a machine that has just served two.

This is #1901 in the sibling file. The usage log closes a truncated line before appending — one byte of reading — and its own comment says why: "It cost the next one too… a recall deja had just served went missing from every count." The snapshot log is the file deja reads the injected *text* back out of, where losing a record is not an approximation but the difference between the digest and nothing, and it never got the same byte.

Same fix, same shape: `O_RDWR` so the descriptor can be read, `endsMidLine` before the write.

Two tests. The second is the rotation half of the queue item that found this: a partial tail plus a rotation in the same run, since rotation rewrites from what the reader accepts. Both fail on base — the second returns a digest from before the rotation as the newest.

### DIFF
--- a/internal/usage/snapshot_tail_test.go
+++ b/internal/usage/snapshot_tail_test.go
@@ -31,23 +31,36 @@ func TestASnapshotAfterAnInterruptedWriteIsStillReadable(t *testing.T) {
 	if got[0].Digest != "the second digest" {
 		t.Errorf("newest digest is %q, want the one just served", got[0].Digest)
 	}
-	// The interrupted record itself is gone, which is the cost of the kill and
-	// not of the write after it.
+	// The interrupted record survives too: what the kill cost it was the
+	// newline, not the JSON, and the write after it no longer takes it down.
 	if len(got) != 2 {
-		t.Logf("readable snapshots: %d (the interrupted line may or may not survive)", len(got))
+		t.Errorf("readable snapshots: %d, want both", len(got))
 	}
 }
 
-// The same file, rotated in the same run: rotation rewrites from what the
-// reader accepts, so a partial tail must not take the rest of the file with it.
+// The same file, rotated by the write that follows a partial tail. Rotation
+// rewrites from what the reader accepts, so it is the second place a
+// half-written line could take the rest of the file with it — and it turns out
+// to be the one place that already healed it: the rewrite terminates every
+// line, so the write after a rotation never glued anything. A guard on a path
+// that was right, not a repeat of the one above.
 func TestRotationSurvivesAPartialTail(t *testing.T) {
 	dir := t.TempDir()
 	big := make([]byte, 4096)
 	for i := range big {
 		big[i] = 'x'
 	}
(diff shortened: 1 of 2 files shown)


vshulcz:
All three test findings taken in 2655c3a1, and the first one changes what this PR can claim.

The rotation test did not rotate. 300 records of 4 kB is 378 kB against a 512 kB threshold, so `rotateSnapshots` was a no-op on the write that followed the truncation — it was the test above it with a bigger file. It now grows the file until `stat` says it is past the threshold and asserts the file shrank, so a rotation demonstrably happened.

With that fixed, the honest result is the opposite of what I implied: **the rotation path was already safe**. Reverting the guard fails the first test and leaves the second passing, because the rewrite terminates every line, so the write after a rotation never had anything to glue onto. The test stays as a guard on a path that was right, and its comment says so rather than claiming a second bug.

The other two: the `usable()` loop could not fail — the reader filters on it already — and my comment said the interrupted record was gone when `b[:len(b)-1]` strips only the newline. Both records survive, and the test asserts that now instead of logging a line that never fired.

Correction to the description above: the second test does not fail on base.

Out of scope and queued rather than folded in: `hook_prompt.go:568` and `:613` append to the `.hookseen` dedup file the same way, with no such guard.

PR #1968, opened 2026-08-26
fix(hooks): close a truncated dedup line before appending

### LINKED ISSUE #1967: an interrupted hook write loses one session from the dedup file
The `.hookseen` dedup file is appended to with `O_WRONLY|O_APPEND` and no check for a truncated tail (`cmd/deja/hook_prompt.go:568` and `:613`). A hook killed between a line and its newline leaves half a line, and the next hook writes straight onto it:

```
ses1 sessA 2026-08-25T23:54:45Zses2 sessB 2026-08-25T23:54:45Z
ses3 fingerprint-c

alreadyInjected(ses1) = map[sessA:true]
alreadyInjected(ses2) = map[]              <- lost
alreadyInjected(ses3) = map[fingerprint-c:true]
```

`alreadyInjected` splits on newlines and takes the first two fields, so the glued line keeps the interrupted record and drops the one appended onto it. `ses2` is then a session deja does not know it has already shown, and the per-prompt hook shows it again — which is what `rememberInjected`'s own comment says the 1MB bug did: "permanently broke dedup — the per-prompt hook then re-injected the same sessions turn after turn". The comment above `blockFingerprint` gives the reason it matters: "the same words are not shown to the same agent session twice".

One record is lost rather than the whole file, because `strings.Fields` tolerates the glue. That is the difference from #1965, not an absence of the same defect — and the fix is the same one byte of reading, `endsMidLine`, which the usage log has had since #1901 and the snapshot log since #1965.

Found while reviewing #1965. Reproduced with `rememberInjected` in a throwaway index; snapshot in the hunt notes.

vshulcz:
Fixes #1967, and puts the check that fixes it in one place.

A hook killed between a dedup line and its newline leaves half a line, and the next hook wrote straight onto it:

```
ses1 sessA 2026-08-25T23:54:45Zses2 sessB 2026-08-25T23:54:45Z
ses3 fingerprint-c

alreadyInjected(ses1) = map[sessA:true]
alreadyInjected(ses2) = map[]              <- lost
alreadyInjected(ses3) = map[fingerprint-c:true]
```

`alreadyInjected` takes the first two fields of each line, so the glued line keeps the interrupted record and drops the one appended after it. `ses2` is then a session deja does not know it has already shown, and the per-prompt hook shows it again — the state `rememberInjected`'s own comment describes: "permanently broke dedup — the per-prompt hook then re-injected the same sessions turn after turn".

One record lost rather than the whole file, because `strings.Fields` tolerates the glue. That is the difference from #1965, not an absence of the same defect.

This is the third file with it — the usage log (#1901), the injection snapshots (#1965), this one — so the check is now `atomicfile.EndsMidLine`, in the package whose comment already opens with "for the writers that are not behind a lock", and the two earlier callers use it instead of their own copy. Its own tests pin the trap the fix keeps hitting: a write-only descriptor cannot read the last byte and reports a clean end, so the open mode is half the fix.

Both writers of `.hookseen` are covered, sessions and block fingerprints, and both tests fail on base.

### DIFF
--- a/cmd/deja/hook_prompt.go
+++ b/cmd/deja/hook_prompt.go
@@ -565,11 +565,18 @@ func rememberInjected(dir, sid string, ss []model.Session) {
 	if fi, err := os.Stat(p); err == nil && fi.Size() > 1<<20 {
 		rotateHookseen(p, sid)
 	}
-	f, err := os.OpenFile(p, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0o600)
+	// O_RDWR so the append can read the last byte first: a hook killed before
+	// its newline leaves half a line, and writing onto it costs the record
+	// written after — a session deja then does not know it has shown, which is
+	// this file's whole job (#1967).
+	f, err := os.OpenFile(p, os.O_CREATE|os.O_RDWR|os.O_APPEND, 0o600)
 	if err != nil {
 		return
 	}
 	defer f.Close()
+	if atomicfile.EndsMidLine(f) {
+		_, _ = f.WriteString("\n")
+	}
 	stamp := time.Now().UTC().Format(time.RFC3339)
 	for _, s := range ss {
 		fmt.Fprintf(f, "%s %s %s\n", sid, s.ID, stamp)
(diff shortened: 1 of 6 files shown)


vshulcz:
Nit taken in d8a348dd. The package doc described replacing a file and not appending to one, and I had quoted the half that fit my case — the reviewer's point. The doc now says why a read-only probe lives there: the writers are the same set, and the reason is the same one.

Its checks match mine on the parts that could have gone wrong: `rememberInjectedIDs` returns on the size cap before the newline write, so the bounded path still writes nothing; `rotateHookseen` ends its join with a newline, so the first append after a rotation sees a clean end; and `ReadAt` with an explicit offset is unaffected by O_APPEND.

Blank lines from two hooks each closing the same tail are inert — `alreadyInjected` and `recentlyInjected` skip anything under two fields, `forgetInjected` and `rotateHookseen` skip empty lines, and nothing outside `hook_prompt.go` reads the file.

Issue #1969, opened 2026-08-26
two agents at once lose the newest injected digests to a rotation


vshulcz:
Two agents starting at once — the case #1319 was about — lose the newest injected digests from `.injections.jsonl`.

Measured, 8 writers × 24 records of 4 kB, against a serial control writing the same 192 records:

```
concurrent:  wrote 192 records, file holds 67
             writers whose final record is missing: 8 of 8

serial:      wrote 192 records, file holds 87, newest is w07-i23
```

Rotation keeps the newest 20 and writers keep appending, so the serial run ends with 87 and its newest record present. The concurrent run ends with 67 and every writer's *final* record gone — the newest records, which is exactly what `deja log --last` exists to show.

The mechanism is the rotation, not the append: `rotateSnapshots` replaces the file through `atomicfile.Write`, and a writer that opened its descriptor before the rename appends into the unlinked inode. Its record is written successfully to a file nothing will ever open again.

`usage.go:191` names this trade for the usage log and bounds it: "another process can rotate between this open and the write below, leaving this descriptor on the file that was just retired, and the event goes with it. Accepted, measured — it needs a log past 1MB and a rotation inside the open, the stat and the write, tens of microseconds here and more under load. One event is the cost."

That reasoning does not carry to the snapshot file. Its records are kilobytes rather than ~100 bytes, so the write window is long and the 512 kB rotation threshold is reached often — and the cost is measured above at every concurrent writer's newest record, not one event.

Not a tear: the same probe at 200 B, 4 kB and 64 kB produced no torn or interleaved lines in three runs, so a single write() to a regular file is not the problem.

What I would like to fix mechanically: after appending, notice that the file was replaced under the descriptor and write the record to the file that is now there. It costs one `fstat` on the common path and touches nothing about the no-lock design the package comment describes.

What is yours: whether this path should take a lock instead. The package says it is "the one writer in deja with no lock at all by design", and a lock on every session start is a different trade than the one that note was making.

vshulcz:
Correcting my own report: the headline number was wrong, and the conclusion changes with it.

"8 of 8 writers lost their newest record" was an artifact of the probe, not of deja. I built the survivor set keyed on nine characters of each digest and looked it up with a seven-character key, so every lookup missed. With the key fixed, the measured loss is **one writer's record in one run out of twelve** — a rare single record, not the newest record of every writer.

[...]

PR #1970, opened 2026-08-26
fix(usage): keep the injection log under its rotation threshold

vshulcz:
No bug, one guard, and a correction to my own issue.

The question was whether two agents starting together can interleave two multi-kilobyte snapshot records into a line that parses as neither. They cannot: sixteen writers at 200 B, 4 kB and 64 kB, repeated and under `-race`, produce no torn line. Worth a test because the reader silently drops anything that would have proved otherwise — a tear would show up as digests quietly missing, not as an error.

While measuring it I reported #1969, claiming every concurrent writer loses its newest record. That was my probe, not deja: I built the survivor set keyed on nine characters of each digest and looked it up with a seven-character key, so every lookup missed. Corrected, the loss is one writer's record in one run out of twelve — the trade `usage.go:191` already documents and bounds at one event.

I also wrote two fixes against the wrong model — carrying the tail across a rotation, and re-appending when the file was replaced under the descriptor — and measured them at one in twelve, unchanged. Neither is here. The issue carries the correction and is closed.

### DIFF
--- a/internal/usage/snapshot_race_test.go
+++ b/internal/usage/snapshot_race_test.go
@@ -16,14 +16,27 @@ import (
 // them into a line that parses as neither — the shape #1319 was about, and the
 // one the reader's own filtering would hide rather than report.
 //
-// They cannot: one write to a regular file is not split. Pinned across three
-// sizes because the answer is about the size, and 64 kB is past any buffer this
-// code has.
+// They cannot here: one write of a whole record reaches the file whole, on the
+// systems this runs on. Not a guarantee — os.File.Write loops on short writes,
+// and a network filesystem promises nothing about an append — which is why this
+// is measured rather than assumed.
+//
+// Sizes stay under the rotation threshold on purpose: past it, a rotation
+// rewrites the file from the lines the reader accepts and would delete the torn
+// line before the assertion ever saw it (#1971).
 func TestConcurrentWritersNeverInterleaveASnapshot(t *testing.T) {
 	for _, size := range []int{200, 4096, 64 << 10} {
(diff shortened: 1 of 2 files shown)


vshulcz:
The review found a real bug behind my corrected non-bug, and this PR now carries the fix.

Its measurement: the loss is not a rate but a cliff at `snapshotsToKeep (20) × digest > snapshotRotateAt (512 kB)`. Below it, 0 lost in 6144 writes. Above it, half of two concurrent injections never reach the log. I reproduced that independently — 100 of 200 missing before the fix, 0 of 200 after.

It is reachable: `mcp_resources.go:109` records a read of `deja://session/…` here, and that record is a whole session through `search.PrintContext` with no budget. One such read parks the log above the threshold, and it stays there.

The fix is the one that does not touch what a person sees: a rotation now drops the oldest records until the rebuild plus the record on its way fits under the threshold, so the next write has nothing to rotate. One record over the threshold is still kept — a log holding the last digest is a log; holding none is not. Filed as #1971 with the numbers.

The three test findings are taken too. The 4 kB and 64 kB cases were vacuous: the run passed 512 kB, and a rotation deletes a torn line before the assertion sees it — measured at 21 of 128 records reaching it. Each size now stays under the threshold and the test asserts the exact line count, which is what makes the emptiness check real.

Prose: "one write to a regular file is not split" overclaimed and now says what it is — an empirical result on the systems this runs on, with the reason it is measured rather than assumed. The opaque sentence is gone.

The PR title and description still describe the tearing test; the rotation fix is the second commit.

PR #1973, opened 2026-08-26
perf(usage): stop re-reading a log that has nothing to drop

### LINKED ISSUE #1972: a busy fortnight makes every recall pay a full read of the usage log
The usage log rotates past 1 MB by keeping the last 14 days. On a machine whose whole log is inside those 14 days, nothing ages out — and `rotate` returns without rewriting anything:

```go
if len(keep) == len(all) {
    return
}
```

So the file never shrinks, and because `rotate` runs before every append and reads the whole file to decide, every event costs a full read of a file that only grows:

```
after 12000 events: 1054650 bytes, threshold 1048576
200 more events:    1072232 bytes now, 6.377716ms per write
same 200 writes on a fresh log:        28.415µs per write
```

224× slower at 1 MB, and climbing. Twelve call sites record events, all on paths a person is waiting on: a recall answering an agent, a session start, a tool call.

The guard is right about what it protects — a rotation must not throw away recent history to make room. It is the fallback that is missing: when nothing is old enough to drop and the file is still over the threshold, the log needs to keep the newest events that fit rather than keep everything and pay for it on every write. That mechanism already exists here for the all-older case, and the injection log got the same shape in #1971.

Reproduced in `internal/usage`; snapshot in the hunt notes.

vshulcz:
Fixes #1972.

A usage log over the 1 MB trigger because it is busy rather than old drops nothing, and #1900 stopped rewriting it for exactly that reason. The read that decides it stayed:

```
after 12000 events: 1054650 bytes, threshold 1048576
200 more events:    1072232 bytes now, 6.377716ms per write
same 200 writes on a fresh log:        28.415µs per write
```

Every event pays for the whole file, on paths a person is waiting on, and it climbs with the file. What one read found now answers for the next: the oldest event it saw, and the size the file had then. Nothing can age out before that stamp does, and a file that shrank was rewritten by someone else, so the memo is dropped.

What this does **not** do, deliberately: bound the file. My first attempt did — a rotation that keeps the newest events that fit — and it failed `TestRecordSkipsRotationWhenNothingWouldBeDropped`, which pins the opposite on purpose. That test is right: a rotation must not throw away a fortnight of recent history to save space. Trading retention for size is a decision, not a patch, and it stays open in #1972.

It also helps only a process that records more than once — the MCP server answering recalls all day. A hook runs once and exits, so it pays the read the first time either way. That half needs the file bounded, which is the decision above.

Three of my own wrong turns are in the issue rather than here: assuming the first line is the oldest event (three existing tests build logs deliberately out of order), and a test asserting the file gets bounded, which encodes the decision that is not mine.

### DIFF
--- a/internal/usage/rotate_growth_test.go
+++ b/internal/usage/rotate_growth_test.go
@@ -33,8 +33,8 @@ func TestAWriteDoesNotPayForTheWholeLog(t *testing.T) {
 	}
 	full := time.Since(start)
 
-	// Ten times is far below the 224× measured before the fix and far above any
-	// ordinary noise between two runs of the same loop.
+	// Ten times is far below the 135× this same test measures on the branch
+	// before the fix, and far above the noise between two runs of one loop.
 	if full > base*10 {
 		t.Errorf("writing to a full log took %v against %v on a fresh one", full, base)
 	}
--- a/internal/usage/rotate_growth_test.go
+++ b/internal/usage/rotate_growth_test.go
@@ -1,0 +1,41 @@
+package usage
+
+import (
+	"testing"
+	"time"
+)
+
+// A log over the size trigger because it is busy rather than old drops
+// nothing, and #1900 stopped rewriting it for that reason. The read that
+// decides it stayed: every event paid for the whole file, 6.4 ms against 28 µs
+// on a fresh log and climbing (#1972). What one read found now answers for the
+// next, so a process that records more than once pays it once.
+//
+// A ratio rather than a duration, so it means the same thing on a slow runner.
+func TestAWriteDoesNotPayForTheWholeLog(t *testing.T) {
+	if testing.Short() {
+		t.Skip("timing")
+	}
+	fresh := t.TempDir()
+	start := time.Now()
+	for i := 0; i < 200; i++ {
+		RecordResultRaw(fresh, KindRecall, 900, 2, false, 9000)
+	}
+	base := time.Since(start)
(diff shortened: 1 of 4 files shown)


vshulcz:
All of it taken in 802c8999, and the first point was a real hole rather than a nit: a backdated event — a clock that stepped back, a log carried from another machine — arrives behind the memo and would have waited out the whole fourteen days. The memo now carries when it was learned and stops answering after five minutes, which bounds that without giving the cost back.

The rest as advised: `cutoff` moved below the early return, the size passed in instead of a second stat, the unreachable zero-time skip gone, and the contradicting sentence three lines above reconciled — it said the read is nearly free, which is the premise this PR disproves.

The figure is corrected to the one this test measures: 135×, not the 224× per-event number from the issue.

Four tests on the memo's own semantics, which was the gap worth naming: a backdated event is still dropped once the memo expires, a log that shrank under this process is read again, a memo past its window is not used, and the counters still answer afterwards.

Separately, CI caught something on Linux that my machine did not: `TestConcurrentWritersNeverInterleaveASnapshot` failed at 200 bytes with "unexpected end of JSON input". Not a torn record — a blank line, from two writers each closing the same truncated tail with their own newline, exactly as the #1968 review predicted. Every reader of that file skips blank lines; my test did not. Fixed here since it is already on the collector branch.

On the PR body: you are right that "the memo is dropped" is wrong — it is declined, not deleted.

Issue #1971, opened 2026-08-26
above 26 kB a digest, half of two concurrent injections never reach deja log


vshulcz:
Once a single record in `.injections.jsonl` is large enough that twenty of them exceed the 512 kB rotation threshold, the file is permanently over it: every write rotates, and every record appended while a rotation is rebuilding is rewritten away. Measured with two agents injecting at once above that line, **199 of 400 injections never appear in `deja log`** — about half, over 200 runs.

Below the line the same measurement loses nothing: 0 of 3072 writes at 200 B and 0 of 3072 at 4 kB. It is a cliff, not a rate, and it sits at `snapshotsToKeep (20) × digest > snapshotRotateAt (512 kB)` — digests over roughly 26 kB.

That is reachable. `mcp_resources.go:109` records a read of `deja://session/…` into this log, and that record is a whole session: `search.PrintContext` with no budget, against the ~4 kB the other tools answer in. One such read on a long session parks the log above the threshold, and from then on the machine is in the losing regime for as long as those twenty records stay newest.

The mechanism is rotation's read-modify-write: `rotateSnapshots` reads the survivors, rebuilds, and renames. Anything appended between the read and the rename is gone — N concurrent agents lose N records, not the one event `usage.go:191` bounds for the usage log.

Two things could close it, and they are different trades:

- make a rotation leave the file *under* the threshold, dropping oldest records until it is. The cliff disappears because the next write no longer rotates. Costs: fewer digests kept when they are huge.
- cap what the snapshot stores per record. Costs: `deja log --last` stops showing the whole text of a big answer, which is a product decision rather than a mechanical one.

I intend the first. The second is yours if you want the log bounded by content rather than by count.

Credit where due: the cliff was found by the reviewer of #1970, after I had reported the same area with a broken measurement in #1969 and corrected it too far in the other direction.

vshulcz:
Correcting one sentence in this issue, which is mine and wrong.

I wrote that a read of `deja://session/…` records "a whole session: `search.PrintContext` with no budget". It has one: `search.go:1355` sets `budget := 8000`, and I measured it — a 600-turn session answers in 8213 bytes, and the same session with turns sixteen times longer answers in 8213 bytes.

Every writer of this log is budgeted:

| path | cap |
|---|---|
| recall / recall_context | 4096, 2048 under a narrowing policy |
| blame | 8192 |
| resource | 8000 |
| déjà vu | 1536 |
| handoff | 6144 |
[...]

PR #1974, opened 2026-08-26
test(usage): pin the record size this log can hold

vshulcz:
No bug, one guard, and a correction to my own issue.

The question was how big a record a read of `deja://session/…` leaves in the injection log — the queue item and #1971 both say "a whole session, `search.PrintContext` with no budget". It has one. `search.go:1355` sets `budget := 8000`, and the measurement agrees: a 600-turn session answers in 8213 bytes, and the same session with turns sixteen times longer answers in 8213 bytes.

Every writer of that log is budgeted:

| path | cap |
|---|---|
| recall / recall_context | 4096, 2048 under a narrowing policy |
| blame | 8192 |
| resource | 8000 |
| déjà vu | 1536 |
| handoff | 6144 |
| tool | 480 |

Twenty of the largest is about 164 kB against the 512 kB threshold, so the cliff fixed in #1971 is not reachable through any of them today. That fix still stands — it costs nothing, and the file's own guarantee should not rest on six numbers in another package staying where they are — but the account of how a store gets there was wrong, and the issue now carries the correction.

What is missing is the line between the two: `snapshotRotateAt / snapshotsToKeep` is the record size this log can hold without rotating on every write, and nothing said so. A test does now, with the budgets listed and the number to raise named — the threshold, not the budget. Its sibling holds the other half: a record over that line does not take the log with it.

### DIFF
--- a/cmd/deja/answer_budget_room_test.go
+++ b/cmd/deja/answer_budget_room_test.go
@@ -1,0 +1,40 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/search"
+	"github.com/vshulcz/deja-vu/internal/usage"
+)
+
+// Every answer deja snapshots lands in one injection log, and that log holds a
+// record of `usage.RecordRoom` before it starts rotating on every write — at
+// which point half of two concurrent injections is rewritten away (#1971). The
+// budgets that decide a record's size live here and in internal/search, which
+// that package cannot import, so this is where the two meet.
+//
+// A budget raised past this line is not wrong on its own; what is wrong is
+// raising it and leaving the log's threshold where it was.
+func TestEveryAnswerFitsTheRoomTheLogHas(t *testing.T) {
+	// The frame, the header and a trailing note ride along with the body: blame
+	// is documented as ending up about its note's length over budget, so the
+	// room has to cover more than the number itself.
+	const overhead = 2048
+
+	for _, c := range []struct {
+		name   string
+		budget int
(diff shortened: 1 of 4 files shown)


vshulcz:
All four taken in 7aa313ca, and three of them were factual errors of mine.

Verified each against the code before accepting: `contextMCPBudget` is 8192 (mcp.go:629), not 4096 — the 4096/2048 pair is `BuildAutoRecall`, which is the session-start hook, not recall_context. And `hook_tool.go:100` / `hook_tool_after.go:102` call `RecordResult` only, so `tool` never reaches this log at all: I listed a writer that is not one and left out `KindHook`, which `snapshots.go:82` calls the commonest injection there is.

The gap you named in finding 2 was the real one: the assertion could not fail when a *budget* grows, which is the thing it claimed to pin, because budgets live in packages `internal/usage` cannot import. So the check moved to where they are visible — `cmd/deja` holds each budget to `usage.RecordRoom`, now exported for exactly that, with 2 kB of headroom for the frame and the trailing note blame is documented as exceeding its budget by. The message names the number to raise: the threshold, not the budget.

`PrintContext`'s 8000 is a named constant now (`search.ContextBudget`) rather than a literal the test would have to repeat.

The duplicate is gone. You were right that it passed on 6acb42fa~1 and discriminated nothing — every record was the same string, so the assertion could not fail on any non-empty log, and `snapshot_race_test.go` already covers both halves properly.

The corrected table is on #1971 as well, since I had put the wrong one there too.

PR #1976, opened 2026-08-26
fix(log): answer --last --json in JSON when there is no digest

### LINKED ISSUE #1975: deja log --last --json answers a script with a sentence
`deja log --last --json` prints a human sentence when there is nothing to show:

```
log --json          json ok: []
log --last --json   NOT JSON: "deja: no injected digests recorded yet — they appear after a hook or MCP recall fires"
```

A script polling that surface gets a parse error rather than an answer. The array form beside it gets this right and the document says why: it is `[]` and never null, "because this is the output a script polls, and `null` raises where an empty list iterates zero times".

`log.go:44` returns the sentence before the `jsonOut` branch is reached, so the flag never gets a say.

The same sentence also comes back for every shape of nothing — no file, one truncated line, a line with no digest, a line with no kind, blank lines only. That is the conflation #1960 fixed for the vector sidecar: a file deja cannot read reported as a file that was never written. Here it is milder, because a snapshot log with unreadable lines is usually a log deja itself is about to rotate, but the reader still has no way to tell "nothing served yet" from "the last thing served is unreadable".

Two things, and only the first is mechanical:

- `--json` should answer in JSON. `null` is what the shape's own absence looks like — the document already says `deja log --last --json` is "one object", and `null` is how a missing object is spelled — or an empty object if you would rather a consumer could index it without a nil check.
- whether the human form should distinguish an unreadable log from an empty one, and in what words, is yours. I would rather ask than invent a second sentence for a screen a person reads mid-task.

vshulcz:
Fixes #1975 — the mechanical half. The wording half is left for you there.

`deja log --last --json` printed a human sentence when there was nothing to show:

```
log --json          json ok: []
log --last --json   NOT JSON: "deja: no injected digests recorded yet — they appear after a hook or MCP recall fires"
```

The sentence went out before the `jsonOut` branch was reached, so the flag never got a say, and a script polling that surface got a parse error rather than an answer. The array form beside it gets this right and the document says why: `[]` and never null, "because this is the output a script polls".

`null` now, which is how a missing object is spelled for a shape the document already calls "one object". The human form is untouched, and there is a test for each — plus one that the ordinary answer still comes back whole, since the branch sits in front of it.

The same sentence still comes back for every shape of nothing: no file, a truncated line, a line with no digest, blank lines only. That conflation — a file deja cannot read reported as a file that was never written — is the one #1960 fixed for the vector sidecar, and whether this screen should draw the same distinction, in what words, is the half left in the issue.

### DIFF
--- a/cmd/deja/log.go
+++ b/cmd/deja/log.go
@@ -42,6 +42,14 @@ func runLogTo(w io.Writer, dir string, args []string) error {
 	if last {
 		snaps := usage.Snapshots(dir, 1)
 		if len(snaps) == 0 {
+			// `--json` answers in JSON, including when the answer is that
+			// there is nothing: this shape is one object, and null is how a
+			// missing object is spelled. The sentence went out under the flag
+			// too, so a script polling this got a parse error (#1975).
+			if jsonOut {
+				fmt.Fprintln(w, "null")
+				return nil
+			}
 			fmt.Fprintln(w, "deja: no injected digests recorded yet — they appear after a hook or MCP recall fires")
 			return nil
 		}
--- a/cmd/deja/log_last_json_test.go
+++ b/cmd/deja/log_last_json_test.go
@@ -1,0 +1,82 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/index"
+	"github.com/vshulcz/deja-vu/internal/usage"
+)
+
+// `--json` has to answer in JSON, including when the answer is that there is
+// nothing. The array form beside this one is `[]` and never null for the reason
+// the document gives — it is the output a script polls — while `--last --json`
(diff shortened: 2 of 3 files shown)


vshulcz:
Doc nit taken in b5461788 — the paragraph leads with the shape again and the null clause sits at the end, where it reads as a detail of that shape rather than as the headline.

The survey in point 1 is the part worth keeping: every other empty `--json` surface holds its shape (`[]` for the two arrays, a zero-valued envelope for the rest), so this is the only bare null in the repo. I am keeping it for the reason you gave — `{}` cannot be told from a real digest whose fields are all empty, and a nonzero exit would break a poller under `set -e` for an ordinary state.

Your pre-existing note is queued rather than folded in: `--last --json` is the one surface with no `schema_version`.

Confirmed independently before merging: nothing else reaches that stdout — `runLog` is the only caller, the lock notice and the policy warning go to stderr, and the `--last` tip only prints on the table path.

PR #1977, opened 2026-08-26
docs: name the one object-shaped answer that carries no schema_version

vshulcz:
No bug in the output; the document's own exception list was short by one.

The stability policy opens with a rule — object-shaped responses carry `schema_version` — and then names the exceptions: the two array forms, and `stats --impact --json`. Measured:

```
log --last --json      object, schema_version=false
log --json             array (documented exception)
```

An object-shaped response with no version and no entry in the list, which is how it was found while reviewing #1975.

Adding the field would be the wrong fix. That object is one snapshot record — the same record `deja log --json` returns in its array, whose elements the policy already covers ("their element shapes are stable; only additive fields inside them are permitted"). A version on the single form and none on the array would be two answers about one record.

So the list names it, and a test holds the two together: every surface checked must either carry the field or be named in the policy section, and an object without one must say which record shape it is. Prose drifts; this is the drift that produced the gap.

### DIFF
--- a/cmd/deja/schema_version_contract_test.go
+++ b/cmd/deja/schema_version_contract_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


vshulcz:
All four taken in 96a19695, and the first was a false claim of mine rather than a nit.

"The same record `deja log --json` returns in its array" is wrong: that array is `[]usage.Event` from the usage log, this is a `usage.Snapshot` from `.injections.jsonl`. They share four keys and disagree on seven, and the document says so 480 lines below — I had it contradicting itself.

The real reasons, both of which I verified rather than took on trust:

- the object is the record deja stores, marshalled from the same struct at `snapshots.go:125`, so a version field on it would be written into every line of that file and into every rotation rewrite;
- the surface answers `null` when nothing has been recorded, and a shape that is sometimes absent cannot be relied on to carry a version.

You are right that additive-ness was not an argument: an inline envelope would be additive and no conforming consumer could tell.

The test was circular and your sabotage proved it. It reads the specific bullet now — sliced from the list by the command name — and requires that bullet to say what it means. I re-ran your exact sabotage against it: it fails on both assertions, naming the offending line.

`deja last --json` and `deja show --json` are in the table now, so the versioned branch is exercised rather than declared. And the bench surfaces are named in the policy with what they are: a report to whoever asked, not a contract anything parses on a schedule.

Issue #1020, opened 2026-08-05
a locked store is reported as no history at all


vshulcz:
A machine whose only transcripts sit in a directory deja cannot open:

```
$ chmod 000 ~/.qwen/projects        # one session inside
$ deja search anything
deja: no matches for "anything" — no agent history was found on this machine; `deja sources` shows where deja looked
$ deja last
deja: no sessions indexed yet — no agent history was found on this machine; `deja sources` shows where deja looked
```

The history is there. `doctor` says `qwen denied … (0 files, cannot be read)`
and `sources` says `(cannot be read — permission denied on ~/.qwen/projects)`,
so deja knows — the two screens a newcomer starts from state the opposite.

`noAgentHistoryFound` counts the files it can see, and a denied directory
yields none.

PR #1978, opened 2026-08-26
test: hold a query and last to naming a store they could not read

vshulcz:
No bug: the family is in good order, and this holds the two surfaces that were not held.

The sweep asked which screens can tell "there and broken" from "not there" — the distinction #1960 had to add for the vector sidecar. Driven against a store deja is not allowed to open, with the sessions in it:

```
readable store, query: [claude] app · Jan 2 · s1 — 1 matches
denied store, query:   no matches for "pgbouncer" — 1 store could not be read (permission denied); `deja doctor` names it
denied store, last:    no sessions indexed yet — 1 store could not be read (permission denied); `deja doctor` names it
doctor:                claude denied … (0 files, cannot be read)
                       warning  claude store cannot be read — permission denied on …; check its permissions
```

Every one of them draws it. The wording lives in `emptyIndexHint`, and #1020 pinned the helper — but a query and `deja last` calling it was pinned nowhere: `empty_store_commands_test.go` covers files, restore and ctx for the empty case, `friction_test.go` covers friction for both. So a wiring change could let the two commonest surfaces call a machine empty while its sessions sit behind a wall.

Not a free pass: stub `deniedStoreCount()` to zero and the test fails with the sentence it would have printed — "no agent history was found on this machine".

The one surface that still does not draw the distinction is `deja log --last`, which is with you in #1975.

Nothing else changed; the probe I used to find this is not part of the diff.

### DIFF
--- a/cmd/deja/denied_store_surfaces_test.go
+++ b/cmd/deja/denied_store_surfaces_test.go
@@ -1,0 +1,68 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/index"
+)
+
+// The wording that separates a locked store from an empty machine lives in
+// `emptyIndexHint`, and #1020 pinned the helper. What was not pinned is the two
+// surfaces a person actually types: a query and `deja last` both answer from
+// that helper, and a wiring change would let either call a machine empty while
+// its sessions sit behind a wall doctor can see.
+func TestAQueryAndLastNameAStoreTheyCouldNotRead(t *testing.T) {
+	if runtime.GOOS == "windows" || os.Geteuid() == 0 {
+		t.Skip("directory permissions do not deny reads here")
+	}
+	tmp := t.TempDir()
+	claude := filepath.Join(tmp, "claude")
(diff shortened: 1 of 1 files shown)


vshulcz:
Merging with my own verification of the two questions I had put to a reviewer, since it has not reported.

Redundancy: nothing already covers this. Six test files lock a store with `chmod 000` — `denied_store_hint_test.go`, four `doctor_denied_*` files and `doctor_peers_unreadable_test.go` — and every one of them drives either `emptyIndexHint` directly or doctor. `empty_store_commands_test.go` covers files, restore and ctx for the empty case only; `friction_test.go` covers friction for both. A query and `deja last` were in none of them.

Missed surfaces: none conflate the two. Driven against a denied store — `stats` names it, `friction` names it, `sources` names it, and the statusline speaks about usage rather than the corpus, which is a different question. My first sweep said `sources` did not, and that was my grep: the hint says "could not be read" and sources says "cannot be read".

That difference is queued rather than changed here — which of the two wordings deja should use on all of them is a product call, not a mechanical one.

The test is not a free pass: stub `deniedStoreCount()` to zero and it fails on both surfaces with the sentence they would have printed instead.

PR #1980, opened 2026-08-26
fix(brief): say a store was locked instead of no agent ran here

### LINKED ISSUE #1979: bare deja tells a new user no agent ran here when the store is just locked
Bare `deja`, on a machine whose session store deja is not allowed to read:

```
deja-vu … · no agent history found yet

deja reads the session stores your agents already write —
Claude Code, Codex, opencode, Cursor, Gemini, Copilot and twelve more.
Nothing was found on this machine, which usually means no agent has
run here yet, or its store lives somewhere else.
```

The sessions are there. `deja doctor` says so in the same breath — `"state": "denied"` with the path — and `deja sources` prints "cannot be read — permission denied on …".

This is the screen a new user sees first, and it tells them the opposite of what is wrong: they go looking for a store that is not missing, when the fix is a permission (on macOS, usually Full Disk Access for the terminal).

Every other surface draws the distinction. `emptyIndexHint` carries the wording — "N stores could not be read (permission denied); `deja doctor` names it" — and a query, `deja last`, `deja stats` and `deja friction` all answer through it. `printNoHistory` (`cmd/deja/brief.go:517`) does not go through it and has no branch of its own, which is #1020's fix having missed one caller.

Found by the reviewer of #1978 while sweeping the family.

Mechanical half: the empty screen should say a store could not be read when one could not, and point at the same place the other surfaces do.

Wording half, and yours: this is the first screen and the only one written as an introduction rather than an answer, so what it should say in that state is not a sentence I want to invent. Something in the shape of "Nothing was found on this machine — but one store could not be read; `deja doctor` names it" is what I would put there if you want it fixed without another round.

vshulcz:
Fixes #1979, and takes the two findings from #1978 that arrived after I merged it.

Bare `deja`, on a machine whose store deja is not allowed to read:

```
Nothing was found on this machine, which usually means no agent has
run here yet, or its store lives somewhere else.
```

The sessions are there. `deja doctor` says `"state": "denied"` with the path in the same breath, and `deja sources` prints "cannot be read — permission denied on …". This is the first screen a new user sees, and it sends them looking for a store that is not missing when the fix is a permission.

`emptyIndexHint` carries the wording every other empty answer uses — a query, `deja last`, `deja stats`, `deja friction`. `printNoHistory` does not go through it and had no branch of its own: #1020's fix having missed a caller. It says it in its own shape now, because that screen is written as an introduction rather than as an answer.

Also here, from the #1978 review:

- that test built its environment from four `Setenv` calls instead of `hermeticEnv`, so `XDG_CONFIG_HOME` leaked from the machine — and a `deja/exclude` file there sends the answer through `hiddenByOwnSettings` before the hint is reached, failing both cases. Reproduced by the reviewer on a real config.
- its comment said the two surfaces were "pinned nowhere". They were pinned at friction, which drives a command through the same helper. What is not covered there is the `hiddenByOwnSettings` branch that a query and `deja last` return through and friction does not, so the test earns its place for a narrower reason than I claimed. The comment says that now.

The wording of the new sentence is the half I would rather you decided, and it is in the issue: this is the one screen written to explain deja to someone who has just installed it.

### DIFF
--- a/cmd/deja/brief.go
+++ b/cmd/deja/brief.go
@@ -529,8 +529,20 @@ func printNoHistory(w io.Writer, stale bool) {
 	fmt.Fprintln(w)
 	fmt.Fprintln(w, "deja reads the session stores your agents already write —")
 	fmt.Fprintln(w, "Claude Code, Codex, opencode, Cursor, Gemini, Copilot and twelve more.")
-	fmt.Fprintln(w, "Nothing was found on this machine, which usually means no agent has")
-	fmt.Fprintln(w, "run here yet, or its store lives somewhere else.")
+	// A store deja is not allowed to open is not a machine no agent has used,
+	// and this is the first screen a new user sees: sending them after a
+	// missing store when the fix is a permission is the worst place to make
+	// that claim. Every other empty answer draws the line through
+	// emptyIndexHint; this screen is written as an introduction rather than an
+	// answer, so it says it in its own shape (#1979).
+	if denied := deniedStoreCount(); denied > 0 {
+		fmt.Fprintf(w, "Nothing was found on this machine — but %d store%s could not be read\n",
+			denied, pluralS(denied))
+		fmt.Fprintln(w, "(permission denied); `deja doctor` names it.")
+	} else {
+		fmt.Fprintln(w, "Nothing was found on this machine, which usually means no agent has")
+		fmt.Fprintln(w, "run here yet, or its store lives somewhere else.")
+	}
 	fmt.Fprintln(w)
 	fmt.Fprintln(w, "  deja sources     what was looked for, and where")
 	fmt.Fprintln(w, "  deja doctor      check the setup")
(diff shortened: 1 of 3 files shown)


vshulcz:
Both taken in 0364a7d2.

The plural was a real bug and mine twice over: I wrote a second copy of a sentence whose original has had `pluralWhich` since it was written, and the copy hardcoded "it". Two locked stores read "2 stores could not be read … names it". A third test drives exactly that case now.

The line also reads better for the rewrite: "Nothing was found on this machine, but N stores could not be read (permission denied) — `deja doctor` names them", so no line starts with the parenthesis.

On the dead setup: I doubted you and re-added `index.Ensure`, because the test failed the moment I removed it. It failed for a different reason — my own rewording wraps "could not be\nread", and my assertion looked for the unwrapped string. With the assertions collapsing whitespace, the setup is dead exactly as you said, and it is gone. The assertions are wrap-insensitive now, which is what copy assertions should be.

The timing measurement is the part I would not have had: 0.5–1.0 ms against a screen that already spends ~13 ms, and only on the branch where the index came back empty. Worth knowing that it opens the newest file of every store to learn a permission fact — that is pre-existing in `emptyIndexHint`, and I have queued it rather than touching it here.

PR #1981, opened 2026-08-26
test(usage): measure what a record weighs instead of guessing

vshulcz:
The 2 kB allowance in #1974 was a guess, and it described the wrong thing. Measured, against the 8192-byte budget it was guarding:

```
plain text                   body= 8190 record=  8358 overhead=  168 (1.02x)
a real digest with newlines  body= 8385 record=  8943 overhead=  558 (1.07x)
every byte a newline         body= 8192 record= 16552 overhead= 8360 (2.02x)
every byte a quote           body= 8192 record= 16552 overhead= 8360 (2.02x)
control bytes                body= 8192 record= 49320 overhead=41128 (6.02x)
cyrillic                     body=15360 record= 15529 overhead=  169 (1.01x)
```

The overhead is not a constant. JSON escaping is a multiplier: a newline and a quote are two bytes each, and a control byte is six. A fixed allowance is the right shape only for the frame, which is 154 bytes and never moves.

So `usage.RecordSize(n)` is `2n + 512` and the guard uses it. At the widest budget that is 16.9 kB against 26.2 kB of room — and a budget twice as wide would not fit, which the test asserts so the bound is not vacuous.

The doubling holds only because control bytes never reach a digest: at six bytes each the model would be wrong by three times. That is `SafeText` and `SafeLine`, and it was pinned for a different reason — an escape byte recolours a terminal — so this adds the case that says why the arithmetic depends on it too, across escape, bell, SOH, NUL, carriage return, vertical tab and delete.

No behaviour change: nothing is over the line today, and nothing here moves a budget or a threshold.

### DIFF
--- a/cmd/deja/answer_budget_room_test.go
+++ b/cmd/deja/answer_budget_room_test.go
@@ -16,11 +16,6 @@ import (
 // A budget raised past this line is not wrong on its own; what is wrong is
 // raising it and leaving the log's threshold where it was.
 func TestEveryAnswerFitsTheRoomTheLogHas(t *testing.T) {
-	// The frame, the header and a trailing note ride along with the body: blame
-	// is documented as ending up about its note's length over budget, so the
-	// room has to cover more than the number itself.
-	const overhead = 2048
-
 	for _, c := range []struct {
 		name   string
 		budget int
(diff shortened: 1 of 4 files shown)


vshulcz:
All five taken, and the review turned a test-model fix into a real one — filed as #1982.

We converged on the same hole independently: `<`, `>` and `&` are six bytes each under `encoding/json`'s default HTML escaping, they survive sanitising because they are ordinary text, and code is made of them. An answer at the widest budget weighs 49 kB against 26 kB of room, which is exactly the #1971 state.

I did not raise the bound to 6n — that would not fit. The escaping is turned off for this file instead: it exists so a document can be embedded in HTML, and this one is read by deja.

Your invalid-UTF-8 finding was the sharper one, and it went further than either of us measured. CI failed where my machine passed: **Go 1.25 escapes the replacement character as six bytes, 1.27 writes it raw as three**. A record whose size depends on the toolchain has no bound at all, so the digest is coerced to valid UTF-8 before marshalling, and there is a test that fails if an escaped replacement character ever appears in a record.

The bound is `3n + 512` now, with the reasons for each class written down rather than assumed.

The 0x01 byte in my comment was real and is gone — an invisible control byte committed to a public repo, inside a comment about stripping control bytes. Checked the file for non-printables afterwards rather than trusting the edit.

The unreachable NUL branch is dropped, and the sanitising test no longer claims more than it proves: it holds `SafeText` and `SafeLine`, and says so — that every digest is built through one of them is two dozen call sites and is not what it shows.

Issue #1982, opened 2026-08-26
an answer full of angle brackets writes a record six times its size


vshulcz:
An answer full of `<`, `>` or `&` — which is to say code — writes a record six times its own size into `.injections.jsonl`, because Go's JSON encoder escapes those three characters as `<`, `>` and `&` unless told not to.

Measured against the widest answer budget (8192):

```
every byte a less-than         body= 8192 record= 49313 (6.02x)
every byte an ampersand        body= 8192 record= 49313 (6.02x)
generics, as code has them     body= 8184 record= 16870 (2.06x)
html in a transcript           body= 8468 record= 15345 (1.81x)
a shell pipeline               body= 8190 record= 11276 (1.38x)
```

`usage.RecordRoom` — the record size this log holds without rotating on every write — is 26214. A single blame or `recall_context` answer over a file of templates, XML or shell redirections can exceed it, and #1971 is what happens then: the file sits above its rotation threshold for good, every write rotates, and half of two concurrent injections is rewritten away.

Nothing sanitises these away, and nothing should: `<` and `&` are ordinary text. The escaping is `encoding/json`'s default, which exists so a JSON document can be embedded in HTML. This log is a file on disk read by deja itself.

There is a second class the same measurement found: a byte that is not valid UTF-8 is written as the replacement character and costs three. A truncated binary paste in a transcript is enough, and nothing strips it.

Fix: `json.Encoder` with `SetEscapeHTML(false)` for the record and for the rotation that rewrites it. That leaves the worst case at three bytes per character — the newlines and quotes a transcript is made of, and the replacement character — which is what `usage.RecordSize` is written for and which fits the room at the widest budget today.

The records already on disk stay readable: the escaped and unescaped forms parse to the same string, and only new lines change.

Found while replacing a guessed overhead constant with a measured one (#1974, and the PR that measures it).

PR #1985, opened 2026-08-26
test: hold every served path to carrying no control byte

vshulcz:
No bug: nothing deja serves carries a control byte. This is the test that says so, on the paths rather than on the helpers.

Two packages strip them — `internal/search` for what search serves, `internal/redact` for what a digest is built from — and each has its own tests. What nothing held is the wiring between them: a transcript full of what an agent captures from a terminal, indexed and then served.

The index keeps those bytes: raw escapes survive into `records.bin` and `sessions.gob`, so the stripping is at serve time and there is something to test.

Six paths driven, each asserted to carry the session and no control character: `recall`, `recall_context`, the session start, the prompt hook, a `deja://session/…` read and the antigravity hook. Plus the snapshots they leave behind, which is where #1982's size bound reads them.

Corrections to my first version, all from review:

- the `blame` case proved nothing. It answers `[]` for this fixture, and my "answered with nothing" guard passed on it. Every case now requires the answer to contain the session before it looks for bytes.
- the resource and antigravity writers were reachable in a few lines each; I had said the resources surface could not be driven, which was about my fixture and stated as if it were about the code.
- the scan was over bytes, so the C1 controls at U+0080–U+009F walked past it — U+009B is a control sequence introducer as surely as an escape is. It reads runes now.
- `MessageText` in `internal/digest` is a third sanitiser, and no path test reaches it: the handoff digest is built through it and is CLI-shaped. Dropping it left every other guard green while a raw escape reached the file a person opens. It has its own test now, which fails on exactly that sabotage.

Still not covered: the handoff writer itself, for the reason above. Queued.

### DIFF
--- a/cmd/deja/digest_control_bytes_test.go
+++ b/cmd/deja/digest_control_bytes_test.go
@@ -17,9 +17,10 @@ import (
 // is six bytes in JSON, and the injection log's size bound assumes three at
 // worst (#1982), so one arriving would make the log's own guarantee wrong.
 //
-// `internal/search` holds the helpers that strip them. This holds the paths:
-// each writer of a snapshot, driven over a transcript full of what an agent
-// captures from a terminal.
+// Two packages strip them — `internal/search` for what search serves and
+// `internal/redact` for what a digest is built from — and each has its own
+// tests. This holds the paths between them: the writers of a snapshot,
+// driven over a transcript full of what an agent captures from a terminal.
 func TestNothingServedCarriesAControlByte(t *testing.T) {
 	hermeticEnv(t)
 	claude := os.Getenv("DEJA_CLAUDE_ROOT")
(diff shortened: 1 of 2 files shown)


vshulcz:
Every finding taken, and the sabotage matrix was the part I could not have produced by asking myself the same question again.

The `blame` case proving nothing is the one that stings: it answered `[]` and my own emptiness guard waved it through. Every case now requires the answer to contain the session before it looks at bytes, which is what that guard should have been.

I also told you the resources surface could not be driven. That was about my fixture and I stated it as if it were about the code — both it and the antigravity hook went in as you showed, a few lines each.

The rune point is right and I would not have found it: U+009B is a control sequence introducer, two bytes in UTF-8, and a byte scan walks past it.

And the sabotage my test missed is closed at its own level rather than pretended away. `MessageText` is a third sanitiser, the handoff digest is built through it, and no path test reaches it — dropping it left everything green while a raw escape reached the file a person opens. It has a test in `internal/digest` now, and I ran your exact mutation against it: it fails, naming the byte and its offset.

Handoff itself stays uncovered for the reason you gave, and is queued rather than claimed.

PR #1986, opened 2026-08-26
test(usage): state the room one event has, and hold the writers to it

vshulcz:
No bug: the usage log's own numbers hold, with a factor to spare. This states the relationship between them and holds the writers to it.

When every event is older than the fourteen days it keeps, the rotation falls back to the newest 200. Those have to fit under the 1 MB that triggered the rotation — or it leaves the file over its own threshold and the next write rotates again, which is what #1971 measured in the injection log.

Measured, one event by what it carries:

```
an ordinary recall             106 bytes
a recall over many sessions   7898 bytes   (200 ids of 36)
ids as long as a path        15247 bytes   (50 ids of 300)
```

Only `ids` grows, and it grows with the sessions an answer holds, which the answer's own byte budget bounds. Real session ids are 35 to 42 characters for Claude and up to 70 for a codex or gemini transcript name, so a widest answer carries on the order of a kilobyte of them.

Driven rather than argued: a recall with `limit: 100` over sixty sessions whose ids are 62 characters writes an event of **420 bytes**, against the 4096 this now states as the room. 200 of those is 800 kB against a megabyte.

Two tests: the arithmetic between the three constants, and a real recall held to the bound. Neither is a free pass — raise `keepAtLeast` past 256 and the first fails; make an answer carry its ids unbounded and the second does.

### DIFF
--- a/cmd/deja/usage_event_room_test.go
+++ b/cmd/deja/usage_event_room_test.go
@@ -23,12 +23,13 @@ func TestARealEventFitsItsRoom(t *testing.T) {
 	t.Setenv("HOME", tmp)
 	t.Setenv("USERPROFILE", tmp)
 
-	// Sixty sessions, each with a name as long as the longest a real store
-	// produces — a codex or gemini transcript filename.
-	long := strings.Repeat("the pgbouncer pool kept timing out in transaction mode ", 4)
+	// Shaped for the worst case rather than around it: an event grows with the
+	// number of ids, an answer holds more sessions when each says less, and an
+	// id is as long as a filename may be. Long turns would cut the answer to a
+	// handful of sessions and measure nothing.
 	for i := 0; i < 60; i++ {
-		id := strings.Repeat("a", 60) + string(rune('a'+i%26)) + string(rune('a'+(i/26)%26))
-		seedClaude(t, claude, "app", id, "turn "+long, "reply "+long)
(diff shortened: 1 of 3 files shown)


vshulcz:
All four taken. The second was the one worth the review: my test was shaped **around** the worst case rather than for it — long turns cut the answer to six sessions, so 420 bytes measured almost nothing. Short turns and filename-length ids, as you had it: **2857 bytes of 4096, carrying 11 ids**. That is a bound with a factor of 1.4 rather than 10, which is the number worth writing down.

The failure message named the sixty sessions I seeded rather than the ids the event carried — the count is the thing that grows, so it is the thing to print. It parses the event now, and refuses to pass at all if the widest one carried fewer than two ids.

"Measured" was doing work it had not earned: 1.4 kB came from the synthetic unit test below it, not from any answer. The comment cites the 2.9 kB a real recall writes.

Your fourth point is the one I want on the record rather than fixed here. The invariant covers the fallback branch only; the ordinary rotation keeps whatever is inside the window and has no size rule, so it can leave the file over the threshold — transient, since the next write finds nothing to drop and stops re-reading. Closing that properly means dropping recent history to save space, which is exactly the retention decision #1972 leaves with Vlad. The comment now says so, and names the other road: the injection log loops the rebuild until it fits (#1971).

Also fixed: two comments disagreed about how long a real session id gets, and both were wrong.

PR #1987, opened 2026-08-26
test(usage): measure the shape that costs three bytes a character

vshulcz:
The question was whether the two constants behind the injection log's size bound can drift apart without a test noticing. They cannot — but the model between them could, and that is what this fixes.

Mutated each in turn:

```
snapshotsToKeep 20 -> 60         CAUGHT
snapshotsToKeep 20 -> 5          CAUGHT
snapshotRotateAt 512k -> 128k    CAUGHT
keepAtLeast 200 -> 400           CAUGHT
EventRoom 4096 -> 16384          CAUGHT

RecordSize 3n -> 2n              passed — nothing noticed
RecordSize envelope 512 -> 0     passed — nothing noticed
```

The constants are held; the model was not. The case that was supposed to justify the third byte is 8192 invalid bytes in a row — and `strings.ToValidUTF8` replaces a *run* with one replacement character, so that case measures 91 bytes for an 8 kB body. 0.01×, proving the opposite of what it claimed.

The number itself is right, and here is what makes it right:

```
all invalid, one run              body=8192 record=    91 (0.01x)
invalid alternating with text     body=8192 record= 16472 (2.01x)
invalid alternating with newline  body=8192 record= 20568 (2.51x)  over 2n
invalid, quote, newline           body=8190 record= 19197 (2.34x)  over 2n
```

Alternating is what costs three bytes a character, because keeping invalid bytes apart is what stops them collapsing into one. 2.51× is the worst a body reaches, so `3n + 512` is the bound and not a guess.

Both shapes are cases now, and weakening the model to `2n + 512` fails on the alternating one. The comment says why a solid run is the best case rather than the worst, which is the thing that fooled me.

### DIFF
--- a/internal/search/control_bytes_test.go
+++ b/internal/search/control_bytes_test.go
@@ -6,8 +6,8 @@ import (
 
 // Control bytes never reach the text deja hands an agent, and two things rest
 // on that. The one this file was written for: a control byte is six bytes in
-// JSON, so a digest carrying them would weigh three times what
-// `usage.RecordSize` says, and the injection log's own bound would be wrong.
+// JSON, so a digest carrying them would weigh twice what
+// `usage.RecordSize` says — six bytes against the three it allows — and the log's own bound would be wrong.
 // This holds SafeText and SafeLine to it; that every digest is built through
 // one of them is about two dozen call sites and is not what this proves.
 // The older one is why the stripping is there at all — an escape byte
--- a/internal/usage/envelope_test.go
+++ b/internal/usage/envelope_test.go
@@ -1,0 +1,46 @@
+package usage
+
+import (
+	"strings"
+	"testing"
+	"time"
+)
+
+// `RecordSize` allows 512 bytes for the envelope — the stamp, the kind, the
+// policy name, the terms, the receiving session — and every case that measures
+// the escaping uses a body big enough that 3n swamps that half. This measures
+// the envelope on its own.
+//
+// The terms are the part that grows: a prompt containing one long word puts
+// that word in the record, and `prompt.Terms` treats a path or a hash as a
+// single token (#1988).
+func TestTheEnvelopeFitsWhatItIsAllowed(t *testing.T) {
+	body := "a digest"
+	for _, c := range []struct {
+		name string
+		snap Snapshot
+	}{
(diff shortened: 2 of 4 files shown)


vshulcz:
My own answer to the question I put to review, since it is the one that matters: I could not find a shape that beats 2.51×.

Thirteen tried under the real writer — invalid bytes alternating with a newline, a quote, a backslash, a tab and 0x7f; lone surrogate bytes; an overlong encoding; a truncated four-byte lead; U+2028 and U+2029; solid backslashes; and three-way mixtures:

```
invalid + newline / quote / backslash / tab   2.51x
invalid, quote, newline                       2.34x
backslashes, U+2028, U+2029, invalid + 0x7f   2.01x
lone surrogates, overlong, truncated lead     0.01x
```

None over `3n + 512`. The last row is the same collapse that fooled me: any run of bytes that is invalid end to end becomes one replacement character, so the exotic encodings measure nothing at all.

The 2.51 family is flat at 2.51 whichever escaped character separates the invalid bytes, which is what you would expect — three bytes for the replacement character and two for the escape, per two bytes in.

vshulcz:
Your brute force settled the escaping half — the ceiling is exactly 2.5, three bytes out for the invalid byte and two for the cheapest neighbour that stops it collapsing — and the comment says that instead of my sampled 2.51, which was a property of my probe's envelope rather than of any shape.

Your third finding turned out to be a live bug, so this PR now carries a fix and #1988.

You wrote that the envelope half is untested because every case uses a body where 3n swamps 512. It is worse than untested: the envelope is not bounded at all. `Terms` are capped at six by count and at nothing by length, and `prompt.Terms` treats a pasted path or hash as one token. Driven through the real prompt hook with one 4000-character word:

```
before   record 4827 bytes   terms 4003   RecordSize says 2612
after    record  956 bytes   terms  132
```

1.8× its own bound, from a prompt anyone might type. The record now cuts each term so the set fits 256 bytes, and cuts rather than drops: a missing term reads as a term that did not match, while a cut one still explains a wrong "you have been here". `RecordSize` stays a function of the digest alone, which is what lets the writers be held to it.

The short-body case you asked for is the new test's first case, and it is what caught this.

Also fixed: `control_bytes_test.go` said a control byte would weigh three times the bound. Twice — six against three.

PR #1989, opened 2026-08-26
test: drive the handoff too

vshulcz:
Two writers of a snapshot that the control-byte test did not drive: the handoff digest, and blame — which I had dropped from #1985 rather than fixed.

The handoff, over the same transcript full of captured terminal output:

```
handoff printed 924 bytes, keeps a control byte: false
snapshot kind=handoff  digest=924  keeps a control byte: false
```

It is the path `internal/digest.MessageText` serves, and the reason it had no path test is that `deja handoff` looked CLI-shaped. It is three lines. Worth having for a specific reason: dropping that sanitiser is the sabotage that left every other case in that test green while a raw escape reached the text a person pastes into another agent. It fails here now, and only here.

Blame answered `[]` for that fixture, which is why I dropped it — the fixture never named a file. It does now, and blame answers about it. An empty answer no longer passes: point it at a file nothing mentions and the case fails with `"[]"`.

Eight writers driven: recall, recall_context, blame, the session start, the prompt hook, a `deja://session/…` read, the antigravity hook and the handoff. The ninth line — the session start's environment-block branch — goes through the same writer as its digest branch.

Also corrected: `message_text_test.go` said its subject was the one no path test reaches, which this PR makes untrue.

### DIFF
--- a/cmd/deja/digest_control_bytes_test.go
+++ b/cmd/deja/digest_control_bytes_test.go
@@ -133,6 +133,22 @@ func TestNothingServedCarriesAControlByte(t *testing.T) {
 		}
 	})
 
+	t.Run("a handoff", func(t *testing.T) {
+		// The path `internal/digest.MessageText` serves, and the one #1985 left
+		// out for being CLI-shaped: dropping that sanitiser left every other
+		// case here green while a raw escape reached the text a person pastes.
+		out, err := captureRun(t, "handoff", "s1")
+		if err != nil {
+			t.Fatal(err)
+		}
+		if !strings.Contains(out, "pgbouncer") {
+			t.Fatalf("the handoff carried none of the session:\n%q", out)
+		}
+		if at := controlByteAt(out); at >= 0 {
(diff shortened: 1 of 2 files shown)


vshulcz:
Both taken, and the third one I had already found the same way you did — which is why it is fixed rather than reworded.

`blame` was a writer this test did not drive because I dropped it in #1985 instead of fixing what was wrong with it: the fixture never named a file, so it answered `[]`. The transcript names one now, blame answers about it, and pointing the case at a file nothing mentions fails with `"[]"` rather than passing. So "the last writer" is now true rather than edited to be defensible.

Your first answer is the one I could not have got by asking myself again: that `printSanitized` redacts secrets and strips bidi and zero-width but not C0/C1, so the assertion is not laundered by the very function under it. That is the failure mode this test would have had and I had not checked.

The stale clause in `message_text_test.go` is gone — it said its subject was the one no path test reaches, which this PR makes untrue.

Eight writers driven now. The ninth line, the session start's environment-block branch, goes through the same writer as its digest branch.

PR #1990, opened 2026-08-26
perf(main): stop parsing every store to count its files

### LINKED ISSUE #1991: the empty screen parses every store to read a number it already has
`noAgentHistoryFound` (cmd/deja/main.go:3065) calls `inspectDoctorStore` for every store and then reads exactly one field of the result: `store.Files`. That field is set at `doctor_report.go:471` as `len(check.files)` and never mutated anywhere in the tree.

Everything the inspection does in between — a stat per path, a listing per directory, opening the newest file and running the store's own parser over it, which for opencode and cursor means SQLite and for others zstd — is discarded.

Measured by the reviewer of #1990 on a real home directory, 21 stores: **514 ms per call against 6.6 ms** when the count is read directly, an identical answer either way. The 6.6 ms is `doctorStoreChecks()` enumerating the stores, which happens first regardless.

This is on the screen a person sees when deja has nothing to show — bare `deja`, and every empty answer that goes through `emptyIndexHint`. `main.go:393` calls it once and `emptyIndexHint` again, so up to two of those per run before `deniedStoreCount` adds its own walk, which does need the inspection.

The same misuse is at `install.go:259` and `:415`, which also read only `store.Files`.

Found while measuring #1990, where my own numbers turned out to be an artefact of the test sandbox: `hermeticEnv` points every store root at an empty temp directory, so nothing has files to parse and the cost disappears.

vshulcz:
Fixes #1991. What started as a comment recording a negative result turned into a real fix, because the negative result was measured wrong.

`noAgentHistoryFound` called `inspectDoctorStore` for every store and read exactly one field of the answer: `store.Files`. That field is `len(check.files)`, set at `doctor_report.go:471` and never mutated anywhere. Everything the inspection did to produce it — a stat per path, a listing per directory, the newest file opened and run through that store's own parser, SQLite for opencode and cursor — was discarded.

On a real home directory with 21 stores, measured by the reviewer of this PR: **514 ms against 6.6 ms** for the same answer. It runs on bare `deja` and on every empty answer that goes through `emptyIndexHint`.

The test that pins it compares the question against the enumeration that precedes it rather than against the clock, so it means the same thing on a loaded runner. On this branch: 10.7 ms before, 150 µs after, against 500 µs to enumerate.

My own first measurement said microseconds and was an artefact of `hermeticEnv`, which points every store root at an empty temp directory — nothing to parse, so the cost disappears. The fixture now writes files big enough to put it back.

`install.go:415` reads `State` and `Paths` as well, so it needs the inspection. `install.go:259` reads only the count, but a full index build follows it immediately, so there is nothing to win there and I have left it alone.

The comment this PR started as is gone: it reasoned about which of two walks to share, and there is one walk now.

### DIFF
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -3063,8 +3063,14 @@ func deniedStoreCount() int {
 // opposed to an index that merely has not been built yet.
 func noAgentHistoryFound() bool {
 	for _, check := range doctorStoreChecks() {
-		store, _ := inspectDoctorStore(check)
-		if store.Files == 0 {
+		// The count, not the inspection. `store.Files` is `len(check.files)`
+		// and nothing else ever sets it (doctor_report.go:471), so asking
+		// `inspectDoctorStore` for it paid a stat per path, a listing per
+		// directory, and the newest file of the store opened and run through
+		// that store's parser — SQLite for opencode and cursor — to learn a
+		// number already in hand. Measured on a real home, 514 ms against
+		// 6.6 ms for the same answer (#1991).
+		if len(check.files) == 0 {
 			continue
 		}
 		// By content, not by existence: an empty notes file — one `deja
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -3024,6 +3024,12 @@ func emptyIndexReason(b index.BuildSummary, evicted int) string {
 // so telling someone to run `deja index` sends them to do again what just
 // happened, and they are left where they started. When no agent history was
 // found at all, the useful next step is finding out where deja looked.
+// Both questions below walk every store — stat each path, list each directory,
+// and open the newest file (doctor_report.go:530) — and on a machine with
+// history the first returns at the first store that has some, which is why
+// answering them from one shared walk is slower rather than faster: measured on
+// 21 stores, 0.4 ms as it stands against 1.0 ms for a walk that always
+// completes. The two walks only both happen when there is no history at all.
 func emptyIndexHint(what string) string {
 	if noAgentHistoryFound() {
 		// "no agent history was found" is a claim about the machine, and it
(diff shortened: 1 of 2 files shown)


vshulcz:
Correcting my own comment before the review lands, because I re-measured and it overclaimed.

I had written "0.4 ms as it stands against 1.0 ms for a walk that always completes". Timing the pieces separately on 21 stores, three runs each:

```
doctorStoreChecks()      214µs / 379µs / 591µs
inspect every store      306µs / 364µs / 638µs
inspect the first store  176µs / 185µs / 194µs
noAgentHistoryFound      349µs / 364µs / 418µs
deniedStoreCount         469µs / 500µs / 705µs
```

The spread between runs of the same call is wider than the difference I quoted, so the two figures had no business being in the source. The shape of the finding survives — the first question stops at the first store with history, and answering both from one completed walk trades that away, which measured slower — and the comment now says that plus the honest scale: a fraction of a millisecond, not worth restructuring in either direction.

There is a second thing the measurement showed that my first version missed: `doctorStoreChecks()` enumerates every store's files eagerly and is *not* short-circuited, so it is paid by both calls. Sharing that specific piece could win where sharing the inspection loses — but at this scale it is noise, which is why the comment now says the whole path is not worth restructuring rather than pointing at one half of it.

vshulcz:
You were right on both counts and the second one is a real fix, so this PR is no longer a comment.

My microseconds were the sandbox: `hermeticEnv` points every store root at an empty temp directory, so there is nothing to parse and the cost I was reasoning about does not exist there. Quoting that as a machine's cost was the mistake, and the empty-machine comparison being backwards followed from it. Both figures are out of the source.

The finding underneath is the one worth having: `noAgentHistoryFound` reads `store.Files`, which is `len(check.files)` and never anything else, through an inspection that stats every path, lists every directory and parses the newest file of each store. I verified the field is set once at `doctor_report.go:471` and mutated nowhere before changing anything.

On this branch, with a fixture big enough to make parsing visible: **10.7 ms before, 150 µs after**, against 500 µs to enumerate. The test asserts against the enumeration rather than the clock so it survives a loaded runner, and it fails on base.

Your ordering point is in the fix rather than worked around: the short-circuit landing on `claude` is what made it expensive, and with the count read directly it stops mattering which store is first.

`install.go:415` reads `State` and `Paths` too, so it keeps the inspection. `install.go:259` reads only the count, but a full index build follows it immediately — nothing to win, so I left it.

PR #1992, opened 2026-08-26
test(index): pin that the index keeps what the transcript brought

vshulcz:
The serving-path cases in `cmd/deja` index a transcript full of captured terminal output and then drive each writer over it. That only proves something if the index keeps those bytes — and nothing said so.

It does: a session read back through `FindByIdentity` carries the escapes, because messages come from `records.bin` and the stripping happens where the text is served. This pins it, and it fails when `loadSessionMeta` is made to sanitise: every case in `TestNothingServedCarriesAControlByte` would keep passing while proving nothing, which is the shape of a test that answers `[]` and calls it clean.

The fixture uses the same dirty string the serving-path cases use, carriage return included, and asserts per message and per byte: concatenating the messages would let a strip that touches one role through, and "any of these bytes" would let a strip that touches one byte through. Both were true of the first version.

### DIFF
--- a/internal/index/keeps_bytes_test.go
+++ b/internal/index/keeps_bytes_test.go
@@ -1,0 +1,80 @@
+package index
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"testing"
+)
+
+// The index keeps a transcript as it found it, control bytes included, and the
+// stripping happens where the text is served. That is not a preference — it is
+// what `TestNothingServedCarriesAControlByte` in cmd/deja rests on: those cases
+// index a dirty transcript and then drive each writer, so an indexer that
+// sanitised on the way in would leave every one of them passing while proving
+// nothing.
+//
+// Which is the failure this file exists to catch, and the same shape as a blame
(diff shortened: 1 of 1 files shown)


vshulcz:
Both sabotages that passed were holes in my assertion, and both are closed — I re-ran your exact mutations against the new version:

- stripping only the carriage return at ingest: now fails, naming the byte and the role. My fixture had dropped `\r` from the dirty string, so that byte was unguarded in the serving-path cases too.
- stripping only assistant-role text: now fails, three times, once per byte. The old version concatenated the messages and asked for any of the bytes, so either kind of partial strip walked through.

Your first answer is the one that makes this worth keeping: no source parser touches message text, and the only ingest-side stripping is on note tags and on sync metadata fields, never on `rec.Text`. So the premise holds for every source, not only for a claude fixture.

Your third is bigger than this PR and is queued rather than folded in: a transcript line carrying a raw control byte is invalid JSON, gets dropped through `diagMalformedLine`, counted in `IngestHealth.MalformedLines`, and the index run still reports success. Real turns can vanish with a signal that only `deja doctor` shows.

The process narrative is out of the description, as you asked.

Issue #1928, opened 2026-08-25
remember --tag is how notes get found later, and no shell completes it


vshulcz:
`deja remember` takes a repeatable `--tag` (`cmd/deja/remember.go:44`), and
tags are how a stored note is found later. Completion offers only `--project`:

```
$ deja completion bash | grep -A6 'remember)' | grep -- '--'
--project
```

Same in zsh (`cmd/deja/completion.go:245`), fish (`:322`) and powershell
(`:403`).

Add `--tag` to all four. Verify with `deja completion <shell> | grep -- '--tag'`
for each shell.

Issue #1927, opened 2026-08-25
forget tells you to pass --all-matches, a flag completion never offers


vshulcz:
`deja forget` refuses an ambiguous prefix and tells you to pass
`--all-matches` (`cmd/deja/main.go:2345`), a flag it accepts at
`cmd/deja/main.go:2359`. Completion never offers it:

```
$ deja completion bash | grep -A8 'forget)' | grep -- '--'
--list --dry-run --session --project --before --unforget
```

Same six-flag list in zsh (`cmd/deja/completion.go:227`), fish (`:306`) and
powershell (`:390`).

Being told to use a flag that Tab does not know is the worst version of this —
add `--all-matches` to all four blocks. Verify with
`deja completion <shell> | grep all-matches` for bash, zsh, fish and
powershell.

Issue #1925, opened 2026-08-25
deja last takes --json and --from, and no shell completes either


vshulcz:
`deja last` accepts `--json`, but no shell will ever suggest it:

```
$ deja last --json
{"schema_version":2,"sessions":[...]}

$ deja completion bash | grep -A8 'last)' | grep -- '--'
--harness --project --since --role
```

The parser at `cmd/deja/main.go:1668` takes `--json` and `--from`; the four
completion blocks list four flags and stop. Bash is at
`cmd/deja/completion.go:110`, zsh at `:242`, fish at `:318`, powershell at
`:398`.

Add the two missing flags to all four. Verify with
`deja completion <shell> | grep -A8 'last)'` for each of bash, zsh, fish and
powershell — every one should now name `--json` and `--from`.

PR #2021, opened 2026-08-26
fix(completion): give deja show its own flag completions

### LINKED ISSUE #1926: deja show has four flags and Tab offers none of them
`parseShow` (`cmd/deja/main.go:557`) takes `--json`, `--harness`, `--offset`
and `--limit`. All four shells put `show` in the bucket for commands that have
no flags at all, so pressing Tab after it offers nothing:

- bash `cmd/deja/completion.go:143` — `check|ctx|embed|...|show|...) COMPREPLY=()`
- zsh `cmd/deja/completion.go:265` — same grouping
- powershell `cmd/deja/completion.go:416` — same grouping, returns `@()`
- fish has no rule for `show` anywhere in the file

```
$ deja show <id> --offset 10 --limit 5 --json   # works
$ deja completion bash | grep -c -- '--offset'  # 0
```

Give `show` its own case in each of the four blocks, the way `last` and
`forget` have one. Verify by running `deja completion <shell>` for all four and
checking each names the four flags.

vaibhav8a:
Closes #1926.

## The change

`show` had its own flags but sat in every shell's "commands with no flags" bucket, so Tab after it offered nothing. It now has its own case in all four generators, matching how `last` is handled:

| Shell | Change |
|---|---|
| bash | own `show)` case; removed from the `check\|ctx\|…` bucket |
| zsh | own `show)` case with `_arguments`; removed from the bucket |
| powershell | own `'show'` branch; removed from the `$_ -in @(…)` list |
| fish | four `__fish_seen_subcommand_from show` rules (fish had none at all) |

bash, zsh and powershell also complete harness names after `--harness`, the way they already do for `last` — the flag takes a value from a known set, so offering the bare flag and then nothing would be a half-finished completion.

## Verified

Built the binary and ran `deja completion <shell>` for all four, checking each names all four flags — and that `show` no longer appears in any no-flag bucket:

```
  bash        --json=ok --harness=ok --offset=ok --limit=ok    (bucket: removed)
  zsh         --json=ok --harness=ok --offset=ok --limit=ok    (bucket: removed)
  powershell  --json=ok --harness=ok --offset=ok --limit=ok    (bucket: removed)
  fish        --json=ok --harness=ok --offset=ok --limit=ok
```

One note on that check: fish declares options as `-l json`, not `--json`, so a naive grep for `--json` reports fish as missing even when the rules are correct. My first pass did exactly that and I had to correct it — flagging it in case the same trap shows up in a CI assertion for the sibling issues.

`go build ./...`, `go test ./cmd/deja/` and `go vet ./cmd/deja/` all pass.

## Scope

Only `show`. The sibling issues (#1925 `last`, #1927 `forget`, #1928 `remember --tag`) are the same shape and I'm happy to take them, but kept separate so each stays reviewable.

### DIFF
--- a/cmd/deja/completion.go
+++ b/cmd/deja/completion.go
@@ -140,7 +140,14 @@ _deja_completion() {
                 COMPREPLY=( $(compgen -d -- "$cur") )
             fi
             ;;
-        check|ctx|embed|hook-precompact|hook-prompt|mcp|share|show|sources|statusline|update|version|warmup)
+        show)
+            if [[ "$prev" == "--harness" ]]; then
+                COMPREPLY=( $(compgen -W "$harnesses" -- "$cur") )
+            else
+                COMPREPLY=( $(compgen -W "--json --harness --offset --limit" -- "$cur") )
+            fi
+            ;;
+        check|ctx|embed|hook-precompact|hook-prompt|mcp|share|sources|statusline|update|version|warmup)
             COMPREPLY=()
             ;;
         *)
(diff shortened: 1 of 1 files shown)


vshulcz:
Read the diff and generated the completions from a build of this branch. All
four shells offer the four flags `parseShow` accepts, and `--harness` completes
harness names in the three shells that can do value completion:

```
bash        --json ok  --harness ok  --offset ok  --limit ok
zsh         --json ok  --harness ok  --offset ok  --limit ok
fish        --json ok  --harness ok  --offset ok  --limit ok
powershell  --json ok  --harness ok  --offset ok  --limit ok
```

Two details worth naming because they are the ones that get missed. You removed
`show` from the no-flags group in all three places it appears rather than
leaving a stale entry behind — in powershell that matters, since `switch` runs
every matching clause and a leftover `show` in the array would have appended an
empty list after yours. And fish had no rule for `show` at all, so it needed
adding rather than editing; the `-r` on `--harness`, `--offset` and `--limit` is
right, they all take a value.

No findings.

vshulcz:
Merged main into your branch — the `windows manifests` check was red for a reason that was not yours: v0.19.0 went out while this was open, and the manifests in the repository still pointed at 0.18.0 until #2019 repinned them. Nothing of yours changed; saying so here so it is not a silent edit.

vshulcz:
Thank you — merging.

Of the four this is the one that took the most reading: `show` was not one line
short of complete, it was sitting in the no-flags bucket in three shells and
absent from fish entirely, so it needed a case written from scratch in each and
removing from the group in the same edit. Getting that consistent across four
syntaxes without leaving a stale entry behind is the whole job, and it is right
in all four.

PR #2026, opened 2026-08-26
fix(completion): complete --json and --from for deja last

### LINKED ISSUE #1925: deja last takes --json and --from, and no shell completes either

vaibhav8a:
Closes #1925.

## The change

`deja last` accepts `--json` and `--from` (parser at `cmd/deja/main.go:1668`), but all four completion blocks listed only the other four flags. Added both to each:

| Shell | Location |
|---|---|
| bash | `completion.go:110` |
| zsh | `:242` |
| fish | `:318` |
| powershell | `:398` |

`--from` takes a value, so it is declared as value-taking (`-r` in fish, `=[…]` in zsh) rather than as a bare switch — otherwise Tab would stop dead after the flag.

## Verified

Built the binary and checked each shell names both new flags:

```
  bash        --json=ok --from=ok
  zsh         --json=ok --from=ok
  powershell  --json=ok --from=ok
  fish        --json=ok --from=ok
```

Note for anyone scripting this check: fish declares options as `-l json`, not `--json`, so grepping for `--json` reports a false miss on fish. The verification above accounts for that.

`go build ./...` and `go vet ./cmd/deja/` pass.

### DIFF
--- a/cmd/deja/completion.go
+++ b/cmd/deja/completion.go
@@ -113,7 +113,7 @@ _deja_completion() {
             elif [[ "$prev" == "--role" ]]; then
                 COMPREPLY=( $(compgen -W "%ROLES%" -- "$cur") )
             else
-                COMPREPLY=( $(compgen -W "--harness --project --since --role" -- "$cur") )
+                COMPREPLY=( $(compgen -W "--json --from --harness --project --since --role" -- "$cur") )
             fi
             ;;
         remember)
@@ -240,7 +240,7 @@ _deja() {
       _arguments '--no-guidance[skip guidance files]' "1:target:($install_targets)"
       ;;
     last)
-      _arguments '--harness=[filter by harness]:harness:($harnesses)' '--project=[filter by project]:project:' '--since=[filter by age]:duration:' '--role=[filter by role]:role:(%ROLES%)' '1:count:'
+      _arguments '--json[print JSON]' '--from=[start from a session]:session:' '--harness=[filter by harness]:harness:($harnesses)' '--project=[filter by project]:project:' '--since=[filter by age]:duration:' '--role=[filter by role]:role:(%ROLES%)' '1:count:'
       ;;
     remember)
       _arguments '--project=[note project]:project:' '1:text:'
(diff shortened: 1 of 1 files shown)


vshulcz:
Read the diff and checked all four shells from a build of this branch — bash,
zsh, fish and powershell each offer `--json` and `--from` now, and nothing else
moved. That is the issue closed.

One thing to fix before it goes in, and it is in the description rather than the
wiring.

`--from` does not take a session. It takes the machine a session came from, with
`local` as the magic value for this one — `filterRecentSources` at
`cmd/deja/main.go:1726` returns nothing unless `From` is empty or equals
`local`, and `From` is documented at `internal/index/index.go:175` as "the
machine this session was worked on". Reproduced on this branch:

```
$ deja last --from local
[grok · pin-manifests · 2026-08-24 · 01a03437-…] hi
…

$ deja last --from 2915986c        # a real session id prefix
deja: no sessions indexed yet
```

So the zsh line

```
'--from=[start from a session]:session:'
```

teaches the wrong concept, and the `:session:` tag tells the shell to complete
the wrong kind of value. Something like `'--from=[only sessions from this
machine]:machine:'` is the shape; `deja sync forget` already completes host
names if you want a value list, though a plain description is enough for me.

Blocker on that one line, nothing else. The rest — including remembering that
fish needs `-r` on `--from` because it takes a value — is right.

vshulcz:
Merging. All four shells carry both flags and nothing else moved — the fish block declaring `--from` with `-r` is the detail that makes Tab keep going after the flag instead of stopping there, and it is easy to leave out.

The one thing I raised, the zsh description saying `--from` starts from a session when it takes `machine|local`, I am fixing myself rather than sending you back for a word. Thanks for the fix.

vshulcz:
Merged as #2148 — your commit, your authorship, carried onto a branch I could rebase since this one is behind main and conflicts in `completion.go`. I fixed the `--from` description in the same commit rather than sending you back for a word.

Thanks. If you want another, #1925's neighbours are the same shape: the completion blocks drift behind the parsers whenever a flag lands.

PR #2027, opened 2026-08-26
fix(completion): offer --all-matches for deja forget

### LINKED ISSUE #1927: forget tells you to pass --all-matches, a flag completion never offers

vaibhav8a:
Closes #1927.

## The change

`deja forget` refuses an ambiguous prefix with a message telling you to pass `--all-matches` (`cmd/deja/main.go:2345`), and accepts it (`:2359`) — but no shell offered it. Being told to use a flag that Tab does not know is the sharpest version of this gap, since the user has already been sent looking for it.

Added to all four blocks: bash (`completion.go:91`), zsh (`:227`), fish (`:306`), powershell (`:390`).

It is a boolean switch, so it is declared as a bare flag — no value completion.

## Verified

```
  bash        1 occurrence(s)
  zsh         1 occurrence(s)
  powershell  1 occurrence(s)
  fish        1 occurrence(s)
```

(fish declares it as `-l all-matches`; the check accounts for that spelling.)

`go build ./...` and `go vet ./cmd/deja/` pass.

### DIFF
--- a/cmd/deja/completion.go
+++ b/cmd/deja/completion.go
@@ -89,7 +89,7 @@ _deja_completion() {
             COMPREPLY=( $(compgen -W "--json --offline --deep" -- "$cur") )
             ;;
         forget)
-            COMPREPLY=( $(compgen -W "--list --dry-run --session --project --before --unforget" -- "$cur") )
+            COMPREPLY=( $(compgen -W "--list --dry-run --session --project --before --unforget --all-matches" -- "$cur") )
             ;;
         handoff)
             if [[ "$prev" == "--to" ]]; then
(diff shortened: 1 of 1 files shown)


vshulcz:
Read the diff and generated the completions from a build of this branch; all
four shells offer `--all-matches` now.

Checked the thing the issue was really about, too: the error at
`cmd/deja/main.go:2345` tells a user to pass this flag when a prefix is
ambiguous, and until now the shell it told them that in did not know the flag
existed. That is the version of a missing completion that costs the most, and
it is closed.

No findings.

vshulcz:
Thank you — merging.

This is the one of the four I would have picked first: deja's own error told
people to type a flag that the shell had never heard of, which is worse than a
missing completion because the tool sounds confident while the terminal
disagrees. Closed now in all four shells.

PR #2028, opened 2026-08-26
fix(completion): offer --tag for deja remember

### LINKED ISSUE #1928: remember --tag is how notes get found later, and no shell completes it

vaibhav8a:
Closes #1928.

## The change

`deja remember` takes `--tag`, and tags are how a note is found again later — so it is arguably the more important of its two flags. Completion offered only `--project`. Added `--tag` to all four blocks: bash (`completion.go:119`), zsh (`:245`), fish (`:322`), powershell (`:403`).

## One detail worth flagging

`--tag` is **repeatable** — `remember.go:44` does `tags = append(tags, args[i+1])`, so `--tag a --tag b` is valid. In zsh, a plain `'--tag=…'` in `_arguments` means "at most once", and zsh stops offering it after the first use. It is declared `'*--tag=…'` here so it keeps being offered:

```
'*--tag=[tag the note, repeatable]:tag:'
```

The other three shells have no equivalent notion, so they are plain value-taking flags.

## Verified

```
  bash        ok
  zsh         ok
  powershell  ok
  fish        ok
```

`go build ./...`, `go test ./cmd/deja/` and `go vet ./cmd/deja/` pass.

### DIFF
--- a/cmd/deja/completion.go
+++ b/cmd/deja/completion.go
@@ -117,7 +117,7 @@ _deja_completion() {
             fi
             ;;
         remember)
-            COMPREPLY=( $(compgen -W "--project" -- "$cur") )
+            COMPREPLY=( $(compgen -W "--project --tag" -- "$cur") )
             ;;
         resume)
             COMPREPLY=( $(compgen -W "--exec" -- "$cur") )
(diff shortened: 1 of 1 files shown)


vshulcz:
Read the diff and generated the completions from a build of this branch; all
four shells offer `--tag` now.

The detail worth naming is the `*` in the zsh line:

```
'*--tag=[tag the note, repeatable]:tag:'
```

`runRemember` at `cmd/deja/remember.go:44` collects `--tag` in a loop, so it is
repeatable, and without that prefix zsh stops offering it after the first one.
Most versions of this change would have left it off and nobody would have
noticed until they tried to put two tags on a note.

No findings.

vshulcz:
Thank you — merging.

The repeatable marker is the part I want to name: `*--tag=` is the difference
between a completion that works once and one that works the way the flag does,
and it is not the kind of thing that shows up in a screenshot. You read the
parser rather than the issue text.

PR #2124, opened 2026-08-27
test(bench492): keep the rig that measured the real store

AliceLJY:
Upstreaming the two files the real-store run in #492 used, as asked there.

## `--bin LABEL=PATH`, repeatable

`tools/bench492/run_bench.py` fixed the matrix at two binaries. The real-store measurement needed five in one window — legacy, legacy with bigrams stubbed, new, new with bigrams stubbed, and one built `-trimpath` as a layout probe — and separate runs are not comparable on a loaded machine: the same v0.15.5 on the same corpus measured 38 s and 52.5 s six hours apart. Anything being compared has to share the window, so the window has to hold more than two.

The first `--bin` is the baseline the summary subtracts from, which generalises the old head-minus-stub line without changing what it computes. `--head-bin` / `--stub-bin` stay as shorthand for `head=` / `stub=`, so the invocation in the README still runs and prints the same numbers with the same sign. A label names index directories and log files, so it is rejected unless it is a bare word, and repeats are rejected rather than silently overwriting a cell.

## A real-store replay

`internal/index/cjkindex_realcorpus492_test.go`, behind the `realcorpus492` build tag.

The per-message figures in #1220 come from `benchJAText`, which is `strings.Repeat(sentence, 4)`: every bigram in it occurs exactly four times. That is the ceiling of the repetition range rather than a sample of it — and repetition is exactly what the two dedupe halves are paid for.

| corpus | bigram repetition rate |
|---|---|
| `benchJAText` | 4.000 |
| `gen_corpus.py` ja, 5,000 sessions | 1.640 |
| a real 365,642-message Chinese store | 1.376 |

At 1.376 the same code keeps 2.44× rather than 4.86× on allocations and −30.4% rather than −45.4% on time. Whole builds still recover about a third of the bigram cost, which is the part of the estimate that survives. No generator can be reasoned into a repetition rate, so this measures one on whatever store the machine already has.

It replays what a real build walks before `indexKeys` sees a byte — `ParseClaudeFile` → `preRedactSessions` (NFC, redact, 64 KiB cap) → skip-if-empty → `tokenizedPart` — because skipping any of those measures a different string than the index does. The cross-file duplicate filter is the one step left out; it removes messages rather than rewriting them.

## Coverage is unchanged

The file is a `_test.go` behind a build tag, so it contributes no statements to the profile: `grep -c cjkindex_realcorpus492 coverage.out` is 0.

`internal/index` measures 90.3–90.4% either way. Six runs interleaved, three with the file and three without, produce both numbers in both states:

| round | with | without |
|---|---|---|
| 1 | 90.3% | 90.4% |
| 2 | 90.3% | 90.3% |
| 3 | 90.3% | 90.3% |

So the 0.1 pp is the package's own jitter, not this file. (Worth knowing separately: the floor is 90 and the package sits 0.3–0.4 pp above it while moving 0.1 pp on its own.)

## Checks

- `gofmt -s` clean; `go vet ./internal/index/` and `go vet -tags realcorpus492 ./internal/index/` both clean.
- `go test -race -count=1 ./internal/index/` passes.
- With the tag on and no corpus pointed at, the three tests that need one `SKIP` and the fixture test passes. A normal `go test ./...` never sees the file at all.
- `--bin` exercised end to end: three labels × two arms × two rounds, separate index dirs and logs per label, plus the `--head-bin`/`--stub-bin` path, plus four rejection cases (no `--bin`, malformed spec, label containing a separator, duplicate label).

Everything in the new file is read-only except one sampled-message cache, written only to a path given on the command line.

Refs #492.

### DIFF
--- a/internal/index/cjkindex_realcorpus492_test.go
+++ b/internal/index/cjkindex_realcorpus492_test.go
@@ -1,0 +1,438 @@
+//go:build realcorpus492
+
+// Real-corpus instrumentation for upstream issue #492.
+//
+// Opt-in via the realcorpus492 build tag so a normal `go test ./...` never
+// touches a live store. Everything here is read-only: it parses transcripts
+// and runs the two key emitters against them. It writes exactly one file, the
+// sampled message cache, and only to the path given on the command line.
+//
+// The pipeline reproduced here is the one a real build walks before
+// indexKeys() sees a byte (ingest.go:404,448-468):
+//
+//	ParseClaudeFile -> preRedactSessions (NFC + redact + 64 KiB cap)
+//	  -> skip messages that trim to empty -> tokenizedPart(role, text)
+//	  -> indexKeys(text)
+//
+// Skipping any of those measures a different string than the index does. The
+// cross-file duplicate filter (seenMsgs.dup) is the one step left out; it
+// removes messages, it does not rewrite them.
+package index
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/cjkfold"
+	"github.com/vshulcz/deja-vu/internal/sources"
+)
+
+type sampleMsg struct {
+	Role string `json:"role"`
+	Text string `json:"text"`
+}
+
+func envOr(key, def string) string {
+	if v := os.Getenv(key); v != "" {
+		return v
+	}
+	return def
+}
+
+func envInt(t testing.TB, key string, def int) int {
+	v := os.Getenv(key)
+	if v == "" {
+		return def
+	}
+	n, err := strconv.Atoi(v)
+	if err != nil {
+		t.Fatalf("%s=%q: %v", key, v, err)
+	}
+	return n
+}
+
+// jsonlFiles lists every transcript under root, sorted, so two runs walk the
+// corpus in the same order.
+func jsonlFiles(t testing.TB, root string) []string {
+	var out []string
+	err := filepath.Walk(root, func(p string, info os.FileInfo, err error) error {
+		if err != nil {
+			return nil
+		}
+		if !info.IsDir() && strings.HasSuffix(p, ".jsonl") {
+			out = append(out, p)
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("walk %s: %v", root, err)
+	}
+	sort.Strings(out)
+	return out
+}
+
+// eachIndexedText replays the ingestion pipeline and hands fn the exact string
+// indexKeys() would receive for every message the index would store.
+func eachIndexedText(t testing.TB, root string, fn func(role, text string)) (files, sessions, messages int) {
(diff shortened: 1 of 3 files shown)


Sora-bluesky (approved):
Reviewed at 06580c8. Nothing blocking; two notes below the verification.

What I checked locally (Windows, go1.26.4):

- `go vet -tags realcorpus492 ./internal/index/` is clean. With the tag on and no corpus pointed at, `TestRealCorpusStats492`, `TestRealCorpusEquivalence492` and `TestSampleRepresentativeness492` skip as described and `TestFixtureStats492` passes; a plain `go test ./internal/index/` never compiles the file.
- The replay order is faithful to `ingest.go`: the real path is dup filter → trim-to-empty skip → `tokenizedPart` (ingest.go:448-468), and the file reproduces it with only the dup filter left out, which the header discloses. `preRedactSessions(nil, ss)` is also safe and text-identical: `countClipped` returns on a nil manifest and the stats writes are guarded, so nil changes counters, never bytes.
- `parse_bins` rejects what the description says it rejects, and the old `--head-bin`/`--stub-bin` invocation still computes the same subtraction with the same sign — head is inserted first, so it stays the baseline.
- One environment note for whoever reviews on Windows: a clone with `core.autocrlf=true` makes `gofmt -l` flag the whole new file. The blob itself is LF and clean; the flag is the checkout's, not the PR's.

Two notes, neither blocking:

1. `allTokens` counts `distinct(indexKeys(text))` as bucket() calls, but `eachIndexKey` also hands `bucket()` the three `dateTokens` of every timestamped message — on a 365k-message store that is up to roughly 1.1M calls the counter never sees, about 2.6% at most (less where a year or ISO date in the text already produced the same key). The issue thread used this counter for the decomposition cross-check quoted at 2% agreement, so one comment line saying the counter sits below the true call count by construction would keep a future reader from over-reading that margin.
2. The header's pipeline list names NFC + redact + 64 KiB cap; `preRedactSessions` also runs `stripSelfRecall`. The replay is right, since it calls the real function — the enumeration is just one item short.

And thank you for leaving `cjkindex_diff_test.go` untouched; I will send the mark-only-run fixtures against it separately, as offered on the issue.

vshulcz:
Merged this without a word, which is on me — you should get a review, not a silent squash.

Reading it properly now, and running `parse_bins` rather than reading it: `--head-bin`/`--stub-bin` still resolve to `[head, stub]` in that order, mixing shorthand with `--bin` puts head first, and all three rejections fire — a label with a slash, a duplicate label, a spec with no `=`. So the README invocation keeps printing the same numbers with the same sign, which is the part that had to hold.

The thing worth naming is the repetition-rate measurement, not the runner. `strings.Repeat(sentence, 4)` being the ceiling rather than a sample is the kind of thing that quietly turns a microbenchmark into a wrong headline, and 4.86× → 2.44× is a large correction to publish about your own earlier number. Replaying the real pipeline — NFC, redact, the 64 KiB cap, skip-if-empty — instead of tokenising directly is what makes it a measurement of the index rather than of a string.

Thanks. Both are in.

PR #2131, opened 2026-08-27
test(index): pin what keeps the mark-guard asymmetry out of the index

Sora-bluesky:
The mark-only fixtures promised on #492, plus one correction to what I claimed there.

`cjkfold.Bigrams` grew an all-marks flush guard in #1392 and `cjkIndexKeys` never did, so the two emitters disagree on mark-only runs. AliceLJY's real-store run found five such messages and none of the difference reached the index; nothing pinned why. Writing the test forced the why to be stated precisely, and it came out narrower than my comment on the issue said:

- On one- and two-mark runs, `tokens()` keeps the run as an ordinary token and emits the identical key, so the guard asymmetry is invisible there. All five of AliceLJY's real messages were this shape.
- On runs of three or more marks it is not: `cjkIndexKeys("ーーーー")` emits the interior bigram `tーー`, which the token path does not produce, so the index carries `{tーーーー, tーー}`. **Mirroring the guard into `cjkIndexKeys` would drop real keys — "adding the guard changes no observable output" was wrong**, and I have said so on the issue. It held for every message on the measured store, not for the construction.

Both directions are pinned as a golden table in `TestCJKMarkOnlyRuns492`: covered fixtures assert the token path produces every key `cjkIndexKeys` emits; non-covered fixtures name their extra keys and assert the token path does not produce them. If `tokens()` ever stops keeping kana marks, the covered rows fail; if someone mirrors the guard, the non-covered rows fail. Comparisons are on sets — `indexKeys` itself does not dedupe (on `ー` its raw slice carries `tー` twice; the first collapse is `eachIndexKey`).

`TestCJKMarkGuardSweep492` derives the mark set from behavior — every valid rune where `IsCJK` holds and `Bigrams` emits nothing — so a future addition to the unexported `isKanaMark` list is tested without reaching into `cjkfold`, and asserts the floor of five so a refactor that empties the predicate fails loudly. The existing differential table gains `テストーー`, the mixed-run shape where both emitters agree; pure mark runs stay out of that table because `legacyCJKKeys` delegates to the guarded `Bigrams` and is empty there by design.

Verified: full `internal/index` suite passes; mutating one golden value (removing `tーー` from the `ーーーー` row) fails the test, restored after.

Refs #492.

### DIFF
--- a/internal/index/cjkindex_diff_test.go
+++ b/internal/index/cjkindex_diff_test.go
@@ -6,6 +6,7 @@ import (
 	"math/rand"
 	"strings"
 	"testing"
+	"unicode/utf8"
 
 	"github.com/vshulcz/deja-vu/internal/cjkfold"
 )
@@ -125,6 +126,7 @@ func TestCJKIndexKeysDifferential492(t *testing.T) {
 		{name: "two-rune run", text: "装订"},
 		{name: "supplementary-plane Han", text: "𠀾界𠀀"},
 		{name: "kana only", text: "ひらがなカタカナ"},
+		{name: "mark run adjacent to CJK", text: "テストーー"},
 		{name: "Hangul only", text: "한글만사용"},
 		{name: "mixed CJK and Latin boundaries", text: "abc装订def 관계 xyz"},
 		{name: "long CJK run across tokenizer byte boundary", text: longRun},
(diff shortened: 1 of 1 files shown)


vshulcz:
I merged this silently, which is not the review you should get. Reviewing it now, after the fact.

Checked the central claim by doing the thing the test says would be wrong: mirrored the `Bigrams` all-marks guard into `cjkIndexKeys` and ran the table. `ーーーー` and `ｰﾞﾟ` fail, exactly the two fixtures marked as not covered by the token path, and nothing else moves. So the asymmetry is load-bearing and the test proves it in the direction that matters — mirroring drops real keys.

What I would not have thought to do is derive the mark set from behaviour with the full-rune sweep instead of reading `isKanaMark`. A test that reaches into the other package's table passes forever after that table changes; this one notices.

The one- and two-mark cases resting on `tokens()` — a pass neither emitter knows about — is the sort of property that holds by accident until it doesn't, and it had nothing pinning it. Good catch, and thanks.

Sora-bluesky:
The sweep was the piece I was least sure would earn its place, so it is good to see it named. Deriving the set from behaviour was the only way I could find to have a future `isKanaMark` change exercised in this package at all — a new mark gets the coverage checks run against it, and emptying the predicate fails the floor. Thanks for the merge, and for checking the table by mutation — the two uncovered rows failing with nothing else moving is exactly the failure it was built to produce.

PR #2143, opened 2026-08-27
feat(hooks): give a spawned agent the memory its parent has

vshulcz:
Most of the agent work on this machine happens inside spawned subagents, and none of it had memory.

Counted on the real store, over sessions where deja was installed and available:

```
                        sessions   got a recall   called the tool
own session                   39            10%               33%
Task/Agent subagent          289             1%                0%
```

The 1% is measurement noise — those files contain the words because the agent was reading deja's own source. The real number is zero, and the reason is structural rather than a bar set too high: the per-prompt hook fires on what a person types, and a subagent's prompt is not typed, it is a tool input the parent wrote. Session start does not fire for it either. So the surface that carries memory never sees the agent doing the reviewing, the hunting and the migrating.

PreToolUse is the one event that can still reach it, because its reply can rewrite the tool's input. `hook-tool` now answers a spawn by running the parent's instructions through the same recall the per-prompt hook uses — same bar, same dedup, same trust policy — and appending what comes back to the prompt the subagent is about to receive.

Fed the 622 real spawns in this store, it carries memory into 14 of 25 sampled, at 800–1600 bytes each.

Three details worth reviewing.

The reply returns the whole tool input, not just the prompt: a PreToolUse reply replaces the input rather than merging into it, so a dropped field would silently lose the subagent_type or the model the parent chose. The test asserts every field survives.

Dedup is keyed on the parent plus the instructions, not on the parent's session id. Sharing the parent's key would withhold from an agent that has never seen anything because the parent saw it a minute ago, and would make ten agents spawned together count as one reader.

`permissionDecision` is `allow`. This changes what the agent is told, not what it may do, and a permission prompt on every spawn would make the memory the most annoying thing in the session.

The matcher gains `Task|Agent` — the same event under two names, depending on the host.

### DIFF
--- a/claude-plugin/.claude-plugin/plugin.json
+++ b/claude-plugin/.claude-plugin/plugin.json
@@ -50,7 +50,7 @@
     ],
     "PreToolUse": [
       {
-        "matcher": "Bash|Edit|Write|MultiEdit|NotebookEdit",
+        "matcher": "Bash|Edit|Write|MultiEdit|NotebookEdit|Task|Agent",
         "hooks": [
           {
             "type": "command",
--- a/cmd/deja/hook_spawn.go
+++ b/cmd/deja/hook_spawn.go
@@ -1,0 +1,133 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"strings"
+)
+
+// A spawned agent starts with no memory and no way to be given one.
+//
+// Measured on this machine's store, 289 of 328 sessions that had deja
+// available were Task subagents, and they received a recall in 1% of them and
+// called the tool in none. The reason is structural rather than a bar being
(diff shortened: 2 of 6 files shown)


vshulcz:
Correcting the numbers in the description — they were measured wrong, and the corrected ones are cleaner.

The hook's block is stored in the transcript as its own `attachment` record, not inside the user message. I counted only `user` records, which missed every injection and made both columns far too low.

Counted properly, over sessions where deja was installed:

```
                     sessions   carried an injected block
own session                39                        87%
Task/Agent subagent       292                         0%
```

So it is not 10% against 1% but 87% against zero — 0 of 292, no noise to explain away. Per person turn rather than per session, own sessions carry a recall on 14.2% of turns (1038 of 7314), which is the surface working as intended.

Nothing in the change depends on the old figures; the premise is the same and now has no soft edge.

PR #2149, opened 2026-08-27
test(search): pin when a question stands on one rare word

vshulcz:
No defect behind this one, and the measurement is worth writing down because it fooled me first. A question standing on one rare word looked unreachable:

```
"retry_budget"                                  target   exact
"why did retry_budget break production"         —        nothing
"what happened with the retry_budget last week" target   relevance
```

Two phrasings of the same question, one answered and one not. The rule explains all of it: a lone rare anchor is served when the query has three or more informative words **and at least two of them are words the corpus knows at all** — "one known anchor among typos is noise, not a question". My fixture was eight sentences that had never written "break" or "production", so those read as typos; the phrasings that worked carried "last week", which the time parser turns into a term the corpus does know. Put a release and a break into the store and the silent query answers.

So this adds the test rather than a change. Nothing outside the package could see either half of the rule: served among words the corpus knows, withheld among words it never saw, and the tier is asserted too, so a spelling guess from a rung above cannot pass for it. It fails if the bar drops to one known word.

Review caught that the two-word case in the first draft was refused by the arity clause rather than the one under test; it is labelled as its own rule now.

### DIFF
--- a/cmd/deja/search_lone_anchor_test.go
+++ b/cmd/deja/search_lone_anchor_test.go
@@ -1,0 +1,79 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+)
+
+// A question carrying one rare word the store holds is answered — and only
+// when the words it stands among are words the corpus knows. That second half
+// is what makes the first safe: among words nothing has ever written, a single
+// surviving anchor is as likely to be a typo as a question, and the tier owes
+// silence rather than a guess. Both halves are one condition in
+// relevanceSearch (`len(terms) >= 3 && termsKnown >= 2`) and neither was
+// visible from outside.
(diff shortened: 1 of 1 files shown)


vshulcz:
Reviewed by mutation rather than by reading, and one half of the claim did not hold.

`termsKnown >= 2` → `>= 1`: caught, four assertions fail. Good.

`len(terms) >= 3` → `>= 2`: **passed**. The case meant to pin it, `retry_budget kkjh`, is withheld by the known-words half anyway — `kkjh` is in no session, so `termsKnown` is 1 and the floor never decides anything. The comment calling it "the other half of the same condition" was describing an assertion that could not fail.

Pushed the case that does pin it: `retry_budget production`, both words in the corpus, two terms. Only the floor withholds that one, and lowering the floor to two now fails four assertions instead of none.

Everything else stands — the tier assertion in particular, since a hit from the rung above would mean the answer came from a spelling guess and the test would be measuring something else.

PR #2202, opened 2026-08-27
docs(index): the token counter is a lower bound, not the count

AliceLJY:
Two comments in `internal/index/cjkindex_realcorpus492_test.go` that @Sora-bluesky's review of #2124 caught after it merged. Neither changes what the file measures.

## The token counter is a lower bound, not the count

`allTokens` sums `distinct(indexKeys(text))` per message and is logged as `bucket() calls`. `eachIndexKey` also hands `bucket()` the three `dateTokens` of every timestamped message (`ingest.go:944,1076`), and those never pass through the counter. On the 365k-message store in #492 that is up to ~1.1M calls it does not see — about 2.6%, and fewer wherever a year or an ISO date in the text already produced the same key and deduped against them.

That matters because the decomposition cross-check in #492 reads this counter as agreeing to 2% on the non-bigram leg. The margin carries this much slack by construction, and a reader should not have to rederive the gap to know it is there. The comment now says so, with the file and line that produce the uncounted calls.

## The pipeline list was one step short

The header names NFC, redact and the 64 KiB cap. `preRedactSessions` also runs `stripSelfRecall`, and runs it first (`ingest.go:2868`). The replay was always right — it calls the real function rather than reimplementing it — so this is the enumeration catching up with the code, now in applied order.

## Checks

- `go vet -tags realcorpus492 ./internal/index/` — clean, exit 0, no output
- `TestFixtureStats492` passes; a plain `go test ./internal/index/` still never compiles the file
- coverage unchanged by construction: `grep -c cjkindex_realcorpus492 coverage.out` is 0

### DIFF
--- a/internal/index/cjkindex_realcorpus492_test.go
+++ b/internal/index/cjkindex_realcorpus492_test.go
@@ -10,9 +10,9 @@
 // The pipeline reproduced here is the one a real build walks before
 // indexKeys() sees a byte (ingest.go:404,448-468):
 //
-//	ParseClaudeFile -> preRedactSessions (NFC + redact + 64 KiB cap)
-//	  -> skip messages that trim to empty -> tokenizedPart(role, text)
-//	  -> indexKeys(text)
+//	ParseClaudeFile -> preRedactSessions (stripSelfRecall -> NFC -> redact
+//	  -> 64 KiB cap) -> skip messages that trim to empty
+//	  -> tokenizedPart(role, text) -> indexKeys(text)
 //
 // Skipping any of those measures a different string than the index does. The
 // cross-file duplicate filter (seenMsgs.dup) is the one step left out; it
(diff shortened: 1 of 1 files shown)


Sora-bluesky (approved):
Both notes landed better than I wrote them. Checked against main at 9ed8498: `dateTokens` sits at ingest.go:944 and the append that hands its three keys to the dedupe-and-bucket path at :1076, so the cited lines are the producing ones; `stripSelfRecall` is the innermost call at :2868, so the header's applied order matches the code. The lower-bound comment keeps both the cap and the direction — at most ~1.1M calls, ~2.6%, less wherever the text already produced the key — which is what a future reader of the 2%-agreement cross-check needs to see without rederiving it.

`go vet -tags realcorpus492 ./internal/index/` is clean here and a plain `go test ./internal/index/` still never compiles the file. (Same Windows note as last time: `gofmt -l` flags the file on a `core.autocrlf=true` checkout; the blob is LF and clean.)

vshulcz:
Both corrections check out against the code, which is the only thing worth saying about a comment-only change.

`eachIndexKey` really does `append(indexKeys(text), dateTokens(when)...)` before deduping (ingest.go:1228, and the same shape at 2550), so those three keys per timestamped message reach `bucket()` without ever passing the counter. And `preRedactSessions` really is `stripSelfRecall` first, then NFC, then redact, then the cap — the pipeline line was one step short.

The reason this matters is the one you gave: a number quoted as "bucket() calls" invites a decomposition checked against it, and 2.6% of slack that nobody wrote down is exactly the kind of thing that gets discovered halfway through the next measurement. Naming it as a lower bound costs nothing and saves that.

Thanks — and thanks to @Sora-bluesky for catching it after the merge rather than letting it stand.

Issue #2238, opened 2026-08-27
deja stats groups redactions under the file kind, not the store


vshulcz:
`deja stats` groups redaction counts under the file kind, so a cline user reads a heading no other screen uses.

A secret in a claude transcript and the same secret in a cline session, indexed together:

    redaction rule keys: map[claude:github-token:1  cline-sdk:github-token:1]

and the screen:

    Redactions
      Total       2
      claude
        github-token         1
      cline-sdk
        github-token         1

Five stores have kinds by another name — cline (`cline-sdk`, `cline-vscode`), goose (`goose-jsonl`, `goose-db`), cursor (`cursor-db`), codex (`codex-history`), hermes (`hermes-pg`) — so for those the heading is a word the index run, doctor and `deja sources` never print.

Same shape as #2234, which fixed it for `ingest_health` and did not touch this counter: the key is built at `internal/index/ingest.go:2055` with `harnessForPath`, which answers with the kind, and split back at `internal/index/manifest.go:193`.

vshulcz:
Fixed by #2239 (fix(stats): group redactions by the store, not the file kind).

That PR merged into `hunt/2026-08-23` rather than into the default branch, so nothing here acted on its `Fixes` line. The branch went into main as #2351, so the fix is shipped and this can close.

PR #2241, opened 2026-08-27
fix(index): match the carry-forward to the key it filters

### LINKED ISSUE #2240: an incremental pass counts redactions twice after the key fold
A regression from #2239, on the hunt branch: an incremental pass over a re-read file counts its redactions twice for the five stores whose file kinds have another name.

    after the first build:              map[cline:github-token:1]
    after an incremental pass over it:  map[cline:github-token:3]

Two secrets in the file, three on the counter.

`carryRedactionRules` decides which of the old per-rule counts to drop by asking `harnessForPath` for each re-read file — that answers with the file kind — and compares it against the first segment of the rule key, which #2239 changed to the store:

    internal/index/ingest.go:2105   h := harnessForPath(path)   // "cline-sdk"
    internal/index/ingest.go:2113   !skipHarness[parts[0]]      // "cline"

Before #2239 both sides were kinds and matched. Whichever name is chosen, the two sides have to be the same one.

Affects cline, goose, cursor, codex and hermes; claude and the rest are unaffected because their kind carries the store's name.

vshulcz:
Closes #2240 — a regression from #2239, found by carrying that change's own question one step further.

    after the first build:              map[cline:github-token:1]
    after an incremental pass over it:  map[cline:github-token:3]

Two secrets in the file, three on the counter, and one more on every pass after that.

`carryRedactionRules` drops the old counts of a file the pass is re-reading, so the fresh read is the whole story. It decided that by file kind — `harnessForPath` answers `cline-sdk` — while #2239 made the key the store, so for the five stores whose kinds have another name nothing matched and the old counts were carried on top of the new ones. Both sides are the store now.

The test writes one secret, then two, then indexes again with nothing new to read: 1, 2, 2. The third pass is there because a carry that only shows on the second repeat is exactly the shape this was.

### DIFF
--- a/internal/index/ingest.go
+++ b/internal/index/ingest.go
@@ -2102,7 +2102,11 @@ func carryRedactions(m *Manifest, old Manifest, skip map[string]bool) {
 		if !skipped {
 			continue
 		}
-		h := harnessForPath(path)
+		// The store, matching the key the rules are filed under since #2238:
+		// asking for the file kind here left "cline-sdk" against a "cline" key,
+		// so nothing matched and every incremental pass carried the old counts
+		// on top of the fresh ones (#2240).
+		h := sources.HarnessForKind(harnessForPath(path))
 		if h == "" && path == sources.OpencodeDB() {
 			h = "opencode"
 		}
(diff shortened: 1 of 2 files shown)


vshulcz:
Follow-up: the same mismatch bit the other direction on an upgraded index. Rules written before #2238 are keyed by kind, and the drop-list now holds store names, so the stale `cline-sdk:` count survived its file being re-read and `Redactions()` folded it into the fresh one — two secrets on disk, three on the screen. The carry now folds the old key before the lookup; second test covers it.

PR #2248, opened 2026-08-28
fix(sources): subtract excluded projects from the opencode row

### LINKED ISSUE #2247: sources: the opencode row counts sessions the exclude list removed
`deja sources` counts opencode straight out of sqlite, so its row keeps sessions that exclusion patterns removed everywhere else.

Two opencode sessions with one in an excluded project, plus a claude session in the same project:

    claude    …/claude/-work-secret  sessions=0 messages=0 …  excluded-patterns=1  excluded-sessions=1
    opencode  …/opencode.db          sessions=2 messages=2 …  excluded-patterns=1

Every store in the table filters and reports what it dropped:

    cmd/deja/main.go:2424   raw := it.load()
    cmd/deja/main.go:2425   ss := sources.FilterSessions(raw)
    cmd/deja/main.go:2426   excluded := len(raw) - len(ss)

opencode is printed separately at the end of `printSources` from `sources.OpencodeCounts()`, which is `select count(*) from session` — nothing filtered, nothing reported. The screen that answers "what does deja hold" therefore counts opencode sessions that are not indexed, not searchable and not exported.

aider is also printed by hand there, but it loads and filters, so its row is right.

vshulcz:
Closes #2247.

Two opencode sessions with one in an excluded project, and a claude session in the same project:

    claude    …/claude/-work-secret  sessions=0 messages=0 …  excluded-patterns=1  excluded-sessions=1
    opencode  …/opencode.db          sessions=2 messages=2 …  excluded-patterns=1

opencode is printed outside the table that loads and filters, from `select count(*) from session`, so its row counted sessions that are not indexed, not searchable and not exported.

The store is loaded and filtered only when a pattern is actually set — counting by SQL is what keeps this row cheap on a large database — and the row now says what it dropped, like the others.

### DIFF
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -2488,12 +2488,30 @@ func printSources(dir string) {
 		size = fi.Size()
 	}
 	s, m, _ := sources.OpencodeCounts()
+	// The counts come out of sqlite, which knows nothing about the exclude
+	// list, so with a pattern in force this row kept reporting sessions that
+	// are not indexed, not searchable and not exported while every other row
+	// subtracted them (#2247). Only then is the store loaded: counting by SQL
+	// is why this row is cheap on a large database.
+	opencodeExcluded := 0
+	if len(sources.ExclusionPatterns()) > 0 {
+		raw := sources.LoadOpencode()
+		kept := sources.FilterSessions(raw)
+		opencodeExcluded = len(raw) - len(kept)
+		s, m = sources.CountSessions(kept), 0
+		for _, x := range kept {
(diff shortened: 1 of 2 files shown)


vshulcz:
Follow-up before merge: my first version recounted the store from the loader when a pattern was set, and that moved numbers for a reason unrelated to the exclude list — the loader drops a session holding no text, which `select count(*) from session` includes, so a pattern matching nothing took the row from `sessions=3` to `sessions=2`. It now subtracts the excluded sessions and their messages from the SQL numbers, and a second test pins the no-op-pattern case.

PR #2252, opened 2026-08-28
fix(show,resume): name the mistyped harness instead of the id

### LINKED ISSUE #2251: show and resume blame the id when the harness is mistyped
Two faults in the commands that take an id-prefix, both of which report a missing session when the session is there.

Three indexed sessions, ids `sess1`..`sess3`, one mistyped harness:

    $ deja show sess1 --harness cluade
    deja: no session matches "sess1"

    $ deja resume sess1 --harness cluade
    deja: no session matches "cluade"

For comparison, the commands that check:

    $ deja last 2 --harness cluade
    deja: "cluade" is not a harness deja knows — one of: claude, codex, ...

**show never validates `--harness`.** The value goes straight to `FindByIdentity` / `findByPrefixHarness`, which find nothing under a harness that does not exist, and the refusal blames the id — which was correct:

    cmd/deja/main.go:442   if o.harness != "" { ... }
    cmd/deja/main.go:462   return fmt.Errorf("no session matches %q%s", o.id, ...)

`checkHarness` is right there and is called by search, last, blame, stats and the MCP tools. The comment on `checkRole` beside it states the rule: a typo must not be indistinguishable from a real value with no matches.

**resume takes the last argument that is not `--exec` as the id.** Any stray token silently replaces the id:

    cmd/deja/resume.go:27   for _, a := range args {
    cmd/deja/resume.go:32       prefix = a

So `deja resume sess1 --harness claude` looks up "claude", and so does `deja resume sess1 extra`.

vshulcz:
Closes #2251.

    $ deja show sess1 --harness cluade
    deja: no session matches "sess1"

    $ deja resume sess1 --harness cluade
    deja: no session matches "cluade"

The session exists both times. show never ran `checkHarness`, so an unknown harness matched nothing and the refusal blamed the id; resume took the last argument that was not `--exec` as the id-prefix, so any flag or stray word replaced it.

show now names the harness it does not know, like search, last and blame; resume rejects a flag it does not take and a second id-prefix instead of quietly preferring one.

### DIFF
--- a/cmd/deja/handoff.go
+++ b/cmd/deja/handoff.go
@@ -52,6 +52,11 @@ func runHandoff(dir string, args []string, stdout io.Writer) error {
 			if strings.HasPrefix(args[i], "-") {
 				return fmt.Errorf("handoff: unknown flag %q", args[i])
 			}
+			// The last one used to win, so a stray word replaced the id and
+			// the refusal named it as the session that was missing (#2251).
+			if prefix != "" {
+				return fmt.Errorf("handoff takes one id-prefix — got %q and %q", prefix, args[i])
+			}
 			prefix = args[i]
 		}
 	}
--- a/cmd/deja/id_command_flags_test.go
+++ b/cmd/deja/id_command_flags_test.go
@@ -1,0 +1,76 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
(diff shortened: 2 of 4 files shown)


vshulcz:
Ran the same input at the sibling commands: `deja handoff sess1 extra` said `no session matches "extra"` — the same last-argument-wins shape, now refused the same way. `share` ignores a stray word (first wins) and `ctx` folds it into the selector, which is a different question and not this fix.

PR #2258, opened 2026-08-28
fix(hook): recall the only session a new store holds

### LINKED ISSUE #2257: hook-prompt: a store with one session never recalls it
A store holding exactly one session never recalls it: `deja hook-prompt` stays silent on a prompt repeating that session's rarest words, and starts working as soon as a second, unrelated session exists.

    session: "gateway_timeout on the reconnect_loop keeps dropping heartbeats"
    prompt:  "seeing gateway_timeout in the reconnect_loop again"

    corpus of 1 session:  nothing on stdout
    corpus of 2 sessions: the recall is injected

The ranking is not what fails — instrumented, both rank the session and count the same terms:

    corpus 1: ranked=1 matched=[2] strong=[2] idf=map[gateway_timeout:0 reconnect_loop:0]
    corpus 2: ranked=1 matched=[2] strong=[2] idf=map[gateway_timeout:0.405 reconnect_loop:0.405]

`byIdentifying` sorts the prompt's terms by idf to choose the one word the session must have actually said:

    cmd/deja/hook_prompt.go:398   sort.SliceStable(out, func(i, j int) bool { return idf[out[i]] > idf[out[j]] })
    cmd/deja/hook_prompt.go:205   leadTerms = ordered[:1]
    cmd/deja/hook_prompt.go:264   if !search.SpeechCarriesAnyTerm(s, leadTerms) { continue }

In a one-session store every ratio collapses to zero, every term ties, the stable sort leaves the prompt's own word order, and the lead term becomes the first word typed — "seeing". The session never said "seeing", so the only session deja holds is dropped.

Elsewhere the tiny-corpus case is handled on purpose (`informative := idf >= floor || minSess <= 2`, and `idf = 0.1` where the ratio collapses). This gate reads the raw map and inherits the zeros.

It is the worst place to be silent: the second session someone ever has is the first time deja could say "you have been here".

vshulcz:
Closes #2257.

    session: "gateway_timeout on the reconnect_loop keeps dropping heartbeats"
    prompt:  "seeing gateway_timeout in the reconnect_loop again"

    one session in the store:  nothing
    two sessions in the store: the recall is injected

The ranking was never the problem — both corpora rank the session with the same two matched terms. `byIdentifying` sorts the prompt's terms by idf to pick the word the session must actually have spoken, and in a one-session store every ratio is zero, so the stable sort left the reader's word order and the lead term became "seeing".

Ties now break toward the word that reads like a term of art, and then toward the longer one, which is what separates `gateway_timeout` from `seeing`. `deja bench recall` still reports recall@5 = 1.0 on the 500-session corpus.

Worth naming: this is the first recall someone can ever get — the second session they have — and it was the one case that never fired.

### DIFF
--- a/cmd/deja/hook_one_session_test.go
+++ b/cmd/deja/hook_one_session_test.go
@@ -1,0 +1,77 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/vshulcz/deja-vu/internal/index"
+)
+
+// The second session someone ever has is the first time deja can say "you have
+// been here", and it said nothing: in a one-session store every idf collapses
+// to zero, so the term the session had to have SPOKEN was picked by the prompt's
+// word order and came out "seeing" (#2257).
+func TestAStoreOfOneSessionStillRecallsIt(t *testing.T) {
+	inject := func(t *testing.T, fillers int) string {
+		t.Helper()
+		tmp := t.TempDir()
+		home := filepath.Join(tmp, "home")
+		if err := os.MkdirAll(home, 0o755); err != nil {
+			t.Fatal(err)
+		}
+		t.Setenv("HOME", home)
+		t.Setenv("USERPROFILE", home)
+		claude := filepath.Join(tmp, "claude")
+		if err := os.MkdirAll(filepath.Join(claude, "beta"), 0o755); err != nil {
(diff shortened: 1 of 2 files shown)


vshulcz:
Refined before merge: length alone would have picked `dashboards` over `s3` when the idf ties, so a symbol shape (underscore, dot, slash, digit) now ranks first, then the term-of-art test, then length. `deja bench recall` is unchanged at recall@5 = 1.0.

PR #2261, opened 2026-08-28
fix(share,ctx): refuse an empty selector

### LINKED ISSUE #2259: share and ctx accept an empty selector and hand back a session nobody chose
`deja share ''` and `deja ctx ''` hand back a session for a selector that names nothing — the shape a shell produces when a variable is unset.

Every sibling refuses an empty argument:

    deja search ''    deja: query required
    deja show ''      deja: show needs id-prefix (see `deja last`)
    deja resume ''    deja: resume needs id-prefix (see `deja last`)
    deja promote ''   deja: promote needs a session id prefix (see `deja last`)
    deja blame ''  files ''  how ''  fix ''  restore ''  handoff ''   — all refuse

These two do not:

    $ deja share ''
    # deja share: sess1
    - Project: work/app
    - Harness: claude
    …
    deja: 0 secrets masked in this share. pattern redaction is a floor — review before sending.
    (exit 0)

    $ deja ctx ''
    # deja context: claude · work/app · sess1 · updated 2026-08-28
    ## user
    gateway_timeout on the reconnect_loop number 1
    (exit 0)

An empty prefix matches the first session there is:

    cmd/deja/share.go:27    s, ok, err := findByPrefix(dir, args[0])
    cmd/deja/main.go:1783   index.FindByPrefix(dir, p)     // "" matches anything
    cmd/deja/main.go:674    q := strings.Join(rest, " ")   // "" then takes the first hit

`deja share "$ID"` with `ID` unset is the case that matters: it prints a shareable digest of a session nobody chose, with the redaction line under it, and exits 0.

vshulcz:
Closes #2259.

    $ deja share ''
    # deja share: sess1
    - Project: work/app
    …
    deja: 0 secrets masked in this share. pattern redaction is a floor — review before sending.
    (exit 0)

    $ deja ctx ''
    # deja context: claude · work/app · sess1 · updated 2026-08-28
    (exit 0)

An empty prefix matches the first session there is, and an empty query returns the first hit. `search`, `show`, `resume`, `promote`, `blame`, `files`, `how`, `fix`, `restore` and `handoff` all refuse an empty argument; these two handed back a session's content instead.

The shape that matters is `deja share "$ID"` with the variable unset — a shareable digest of a session nobody chose, with the redaction line under it. Both now give the same refusal as the no-argument case, blank strings included.

### DIFF
--- a/cmd/deja/empty_selector_test.go
+++ b/cmd/deja/empty_selector_test.go
@@ -74,3 +74,42 @@ func TestAnEmptySelectorIsNotASession(t *testing.T) {
 		t.Errorf("share of a blank selector printed: %q", out.String())
 	}
 }
+
+// And the lookup itself: every id has the empty string as a prefix, so the
+// guard belongs where the class lives rather than only at the callers (#2259).
+func TestFindByPrefixMatchesNothingForABlankPrefix(t *testing.T) {
+	tmp := t.TempDir()
+	home := filepath.Join(tmp, "home")
+	if err := os.MkdirAll(home, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	t.Setenv("HOME", home)
+	t.Setenv("USERPROFILE", home)
+	claude := filepath.Join(tmp, "claude", "-work-app")
+	if err := os.MkdirAll(claude, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	t.Setenv("DEJA_CLAUDE_ROOT", filepath.Join(tmp, "claude"))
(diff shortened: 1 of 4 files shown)


vshulcz:
Followed the class down a level before merging: the MCP resource reader already guards this exact thing (#1728, "every id has the empty string as a prefix"), which means the callers keep re-learning it. `index.FindByPrefix` now returns no match for a blank prefix, so share and ctx are belt and braces rather than the only line of defence. Checked every caller — none pass an empty prefix meaning "anything".

PR #2262, opened 2026-08-28
fix(recall): stop Japanese grammar bigrams from triggering auto-recall

### LINKED ISSUE #2260: auto-recall fires on Japanese grammar bigrams: すべ / べて / ての as via-terms
A Japanese-primary session on v0.17.0 (Windows 11) got this UserPromptSubmit injection:

> deja-vu: you have been here — 「見送り（deja-vu + auto-memory で十分）」(Jul 21 · via: すべ, べて, ての)

The via-terms are 「すべての」 ("all the …") sliced into overlapping bigrams. They are pure grammar, shared by more or less every Japanese sentence, and the recalled session had no topical connection to the prompt. Over a long Japanese session most injections looked like this: recall announced on fragments of function words, paying the injection's token cost on prompts it had nothing to say about.

Reproduced on `main` at 9ed8498 — the extraction moved from `cmd/deja/hook_prompt.go` into `internal/prompt` in #2136, but behaves the same:

```
prompt.Terms("すべてのテストを直してください")
  → [すべ べて ての のテ テス スト]
prompt.Terms("この変更をレビューして、問題があれば教えて")
  → [この の変 変更 更を をレ レビ]
```

Where it comes from, read on `main`:

- A CJK prompt falls back to the relevance tokenisation (`internal/prompt/prompt.go:66`): expand each CJK run to bigrams, drop "pure grammar" pairs via `CJKFunctionBigram`.
- `cjkFunctionRunes` (`internal/query/cjk.go`) is the closed class for three Chinese scripts — Mandarin simplified, traditional, and Cantonese. It has no Japanese entries at all, so every hiragana grammar bigram (すべ, ての, して, …) passes as content.
- Bigrams also cross word boundaries inside a run — ての spans 「すべて|の」, のテ spans 「の|テスト」 — so even a Japanese function-word list would not catch everything the slide produces.
- `Terms` caps at six in text order, so a prompt that opens with grammar spends the whole cap before the content arrives: in the second example above, 問題 and 教え never reach the list at all.

That combination is why Japanese specifically is noisy: Chinese grammar is caught by the rune list, while Japanese (and, from reading the same list, likely Korean — untested) has no closed class here.

## Proposal

PR follows, with the failing test first: in the prompt-terms fallback, skip bigrams that contain a hiragana rune. At the rune level, hiragana is close to being the Japanese closed class — content is carried by kanji and katakana runs, hiragana carries particles and inflection. On the probes above the surviving terms become テス / スト for the first prompt and 変更 / レビ / ビュ / ュー / 問題 for the second — the ones that identify something — and the cap stops being eaten by sentence grammar. (Okurigana pairs like 直し and 教え go too; a verb rarely answers for a topic, but it is part of the same trade.)

The cost, stated up front: a content word written entirely in hiragana (ひらがな, ちらつき) loses its trigger capacity. For a hook that fires on every prompt and already demands co-occurring terms before announcing anything, precision looks like the right side of that trade — and ranking and the index are untouched, since this is the trigger-extraction path only.

The alternative shape is extending `cjkFunctionRunes` with the Japanese closed class. That kills the field case (すべ/べて/ての) but keeps the boundary-crossers alive (のテ, 更を), which is why the PR takes the script route. Happy to reshape it if you would rather have the list.

Environment of the original sighting: deja v0.17.0, Windows 11, Japanese-primary prompts, MCP + hooks setup. `main` at 9ed8498 reproduces as above.

Sora-bluesky:
The fix promised on #2260, failing test first.

`prompt.Terms` falls back for CJK prompts to the relevance tier's bigrams, filtered only by `cjkFunctionRunes` — a closed class for three Chinese scripts with no Japanese entries. Japanese grammar lives in hiragana, and the bigram slide also crosses word boundaries (ての spans 「すべて|の」), so a Japanese prompt spent the six-term cap on sentence grammar and auto-recall said "you have been here" about sessions that shared nothing else.

One added condition in the fallback: skip a bigram that contains a hiragana rune. Measured on the probes from the issue:

| prompt | before | after |
|---|---|---|
| すべてのテストを直してください | すべ べて ての のテ テス スト | テス スト |
| この変更をレビューして、問題があれば教えて | この の変 変更 更を をレ レビ | 変更 レビ ビュ ュー 問題 |
| すべての結果を保存する方法は? | すべ べて ての の結 結果 果を | 結果 保存 方法 |

The junk stops earning recalls, and content later in the prompt reaches the cap instead of being crowded out by the opening grammar — the third row's 保存 and 方法 never made the list before.

The trade, stated in the comment as well as the issue: an all-hiragana content word (ひらがな, ちらつき) and okurigana pairs (直し, 教え) no longer trigger on their own. For a hook that fires on every prompt and already demands co-occurring terms before speaking, precision is the right side. Scope is the trigger-extraction path only — index, ranking and the query side are untouched, and Chinese and Korean prompts carry no hiragana, so nothing changes for them; the Chinese fixture pins that. ー U+30FC is a katakana mark rather than hiragana, so loanword bigrams like ュー survive, consistent with the mark handling from #1392.

Tests, written before the fix and measured failing against it reverted: `TestTermsJapaneseGrammarBigrams2260` (the probe goldens, plus the field case — no すべ/べて/ての from a prompt starting 「すべての」), `TestTermsJapaneseKeepsContent2260` (サーバーのログを確認 → サー ーバ バー ログ 確認), `TestTermsChineseUnaffected2260` (装订计数是什么 unchanged; it also passes on the unfixed tree, which is the point). Targeted suites across `cmd/deja`, `internal/query` and `internal/prompt` pass.

Fixes #2260.

### DIFF
--- a/internal/prompt/japanese_terms_test.go
+++ b/internal/prompt/japanese_terms_test.go
@@ -1,0 +1,66 @@
+package prompt
+
+import (
+	"slices"
+	"testing"
+)
+
+func TestTermsJapaneseGrammarBigrams2260(t *testing.T) {
+	tests := []struct {
+		name   string
+		prompt string
+		want   []string
+	}{
+		{
+			name:   "test request",
+			prompt: "すべてのテストを直してください",
+			want:   []string{"テス", "スト"},
+		},
+		{
+			name:   "review request",
+			prompt: "この変更をレビューして、問題があれば教えて",
+			want:   []string{"変更", "レビ", "ビュ", "ュー", "問題"},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got := Terms(tc.prompt)
+			if !slices.Equal(got, tc.want) {
+				t.Fatalf("Terms(%q) = %v, want %v", tc.prompt, got, tc.want)
+			}
+		})
+	}
+
+	got := Terms("すべての変更を確認")
+	for _, grammar := range []string{"すべ", "べて", "ての"} {
+		if slices.Contains(got, grammar) {
+			t.Errorf("Terms beginning with すべての = %v, contains grammar bigram %q", got, grammar)
+		}
+	}
+	for _, content := range []string{"変更", "確認"} {
+		if !slices.Contains(got, content) {
+			t.Errorf("Terms beginning with すべての = %v, missing content bigram %q", got, content)
+		}
+	}
+}
+
+func TestTermsJapaneseKeepsContent2260(t *testing.T) {
+	prompt := "サーバーのログを確認"
+	want := []string{"サー", "ーバ", "バー", "ログ", "確認"}
+
+	got := Terms(prompt)
+	if !slices.Equal(got, want) {
+		t.Fatalf("Terms(%q) = %v, want %v", prompt, got, want)
+	}
+}
+
+func TestTermsChineseUnaffected2260(t *testing.T) {
+	prompt := "装订计数是什么"
+	want := []string{"装订", "订计", "计数", "数是"}
+
+	got := Terms(prompt)
+	if !slices.Equal(got, want) {
+		t.Fatalf("Terms(%q) = %v, want %v", prompt, got, want)
+	}
+}
--- a/internal/prompt/prompt.go
+++ b/internal/prompt/prompt.go
@@ -60,12 +60,18 @@ func Terms(prompt string) []string {
 	// field, and techTerm rejects every rune above 127 — a Chinese, Japanese
 	// or Korean prompt therefore yields no terms at all and auto-recall can
 	// never fire for it. Fall back to the terms the relevance tier already
-	// ranks on: per-run bigrams with pure-grammar pairs dropped. A bigram is
-	// as specific as the identifiers techTerm looks for, and the caller still
+	// ranks on: per-run bigrams with pure-grammar pairs dropped. That filter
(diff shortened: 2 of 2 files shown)


vshulcz:
Checked by running it rather than by reading, and the claim holds in both directions.

Korean, Chinese and ASCII come through untouched — 한글로 → `[한글 글로 질문 문입 입니 니다]`, 装订计数是什么 → `[装订 订计 计数 数是]` — and the mixed Japanese case keeps its content: サーバーのログを確認 → `[サー ーバ バー ログ 確認]`.

The cost you named is real and it is the right side to be on: ひらがなだけのことば now yields nothing at all. A hook that fires on every prompt should say nothing rather than claim a déjà vu on sentence grammar, and an all-hiragana content word is rarer than a hiragana particle by a wide margin.

What makes this a good fix rather than a wider filter is the reason for it: the bigram slide crosses word boundaries, so ての in 「すべて|の」 is not even a word — no closed-class list could have caught it, only the script it is written in. That is the observation the change rests on, and the table in the description is what proves it.

Merging.

Sora-bluesky:
Checking by running it is the review every fix wants, thank you. Agreed on the failure mode: for a hook that fires on every prompt, silence beats a false déjà vu. If the all-hiragana gap ever turns out to matter in real recall data, the way out would be proper word segmentation rather than a longer skip list. Nothing so far says it is worth the dependency.

PR #2264, opened 2026-08-28
fix(forget): refuse an unforget that restored nothing

### LINKED ISSUE #2263: forget --unforget exits 0 when it restored nothing
`deja forget --unforget <id>` exits 0 when it restored nothing, so a script cannot tell a restore from a miss.

    $ deja forget --session sess1
    sessions dropped: 1  messages dropped: 1  tombstones added: 1        exit 0
    $ deja forget --unforget sess1
    restored 1 session and rebuilt the index — claude:sess1              exit 0
    $ deja forget --unforget sess1
    no tombstone matches "sess1" — `deja forget --list` shows what is forgotten     exit 0
    $ deja forget --unforget nosuchsession
    no tombstone matches "nosuchsession" — …                             exit 0

So `deja forget --unforget "$id" && echo back` prints "back" for a typo, for an id that was never forgotten, and for a session that is still gone.

    cmd/deja/main.go:2718   fmt.Fprintf(os.Stdout, "no tombstone matches %q …")
    cmd/deja/main.go:2719   return nil

Every neighbour refuses when asked for something that is not there — `deja promote nosuch`, `deja show nosuch`, `deja sync forget nosuchhost` all exit 1. The sentence here is right; the exit code is the part a script reads.

(The `--dry-run` arm at main.go:2703 prints the same line and should presumably answer the same way, since it is reporting the same miss.)

vshulcz:
Closes #2263.

    $ deja forget --unforget sess1
    restored 1 session and rebuilt the index — claude:sess1     exit 0
    $ deja forget --unforget sess1
    no tombstone matches "sess1" — …                            exit 0
    $ deja forget --unforget nosuchsession
    no tombstone matches "nosuchsession" — …                    exit 0

So `deja forget --unforget "$id" && echo back` printed "back" for a typo, for an id that was never forgotten, and for a session that is still gone. `promote`, `show` and `sync forget` all refuse what they cannot find.

Both arms — the restore and its `--dry-run` — now return the same sentence as an error. Two existing tests pinned the old exit code incidentally while asserting the message; they now assert the refusal and keep the message check.

### DIFF
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -2700,8 +2700,7 @@ func runForget(dir string, args []string) error {
 			// this path went straight to Unforget, which lifts the tombstone
 			// and rebuilds. Name what it would restore and stop (#1066).
 			if len(lifting) == 0 {
-				fmt.Fprintf(os.Stdout, "no tombstone matches %q — `deja forget --list` shows what is forgotten\n", unforget)
-				return nil
+				return fmt.Errorf("no tombstone matches %q — `deja forget --list` shows what is forgotten", unforget)
 			}
 			fmt.Fprintf(os.Stdout, "dry run — nothing was changed\nwould restore %d tombstone%s and rebuild the index — %s\n",
 				len(lifting), pluralS(len(lifting)), joinCapped(lifting, 5))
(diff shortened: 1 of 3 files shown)


vshulcz:
Tried extending this to the other zero-restore case — tombstone lifted but no session came back — and backed it out. For an imported session the lift IS the action: `deja forget --unforget <id>` followed by `deja sync import` is the documented way to get it back, and two tests exercise exactly that sequence. Refusing there would break a working flow to make an error code tidier. The miss this PR fixes is different: nothing was lifted and nothing happened.

PR #2277, opened 2026-08-28
fix(stats): say what the shared page holds instead of denying it holds anything

### LINKED ISSUE #2275: stats --html says it holds no message text while embedding the opening line of every session
The footer of `deja stats --html` is the sentence someone reads before sharing the file:

    3 metadata-only sessions embedded. No message text is included in this file.

The file embeds the first user message of every session as its title:

    user: why does the acme-prod payroll export keep dropping rows for contractor 4821

    "title":"why does the acme-prod payroll export keep dropping rows for…"

It is redacted for secret patterns and cut to sixty runes, but it is message text, and it is the part that says what the session was about:

    cmd/deja/statshtml.go:96      Title: search.SafeLine(stats.Title(s))
    internal/stats/stats.go:311   TrimRunes(strings.Join(strings.Fields(redact.SafeForDisplay(m.Text)), " "), 60)

The comment on the data block one line below the footer is accurate — "redacted first-user titles. No message text." — so the code knows what it embeds; the sentence shown to the reader does not.

    cmd/deja/statshtml.go:140   <footer>… No message text is included in this file.
    cmd/deja/statshtml.go:141   // Only metadata is embedded below: dates, harnesses, projects, counts, and redacted first-user titles.

`deja stats --card` does not carry titles, so only the HTML page is affected.

Two ways out: say what the file actually holds, or stop embedding the titles. The first keeps the page useful — the TITLE column is what makes the session table readable — and the second is the stronger promise.

vshulcz:
Closes #2275.

The footer is what someone reads before handing the file to anyone else:

    3 metadata-only sessions embedded. No message text is included in this file.

Every row carries the first user message of its session:

    "title":"why does the acme-prod payroll export keep dropping rows for…"

Redacted for secret patterns and cut to sixty runes — but message text, and the part that says what the session was about. The comment on the data block one line below already described it correctly; only the sentence shown to the reader was wrong.

Now:

    3 sessions embedded: dates, harnesses, projects, message counts, and the opening line of each session, redacted and shortened. No other message text is in this file.

Behaviour is unchanged — this is the claim catching up with the file. Dropping the titles instead would be the stronger promise and a different change: the TITLE column is what makes the session table readable, so that is your call rather than mine.

`TestStatsHTMLMetadataEscapingAndPrivacy` pinned the old wording through the substring "metadata-only"; it now pins the new sentence.


vshulcz:
Checked the enumeration myself rather than waiting to be asked: the embedded rows carry exactly five fields — date, harness, project, title, message count (statshtml.go:93-97) — no ids and no paths, so the sentence is complete. But `stats.Title` prefers a harness-supplied session title and only falls back to the first user message (internal/stats/stats.go:305-313), so "the opening line of each session" was imprecise; it now says "each session's title or its opening line".

PR #2279, opened 2026-08-28
fix(promote): refuse a session the exclude list covers

### LINKED ISSUE #2278: promote reports a recall that cannot happen for an excluded project
With an exclude pattern set and the index not yet rebuilt, `share` refuses and `promote` proceeds:

    $ export DEJA_EXCLUDE_PROJECTS=secretproj
    $ deja share sec1
    deja: sec1 is in a project your exclude list covers — `deja index --rebuild` drops it …   exit 1

    $ deja promote sec1 --state accepted
    promoted claude:sec1 as accepted: the payroll export for acme-prod drops rows
    the note now outranks the raw transcript in recall                                        exit 0

The note is written, and then nothing can read it:

    $ deja index && deja search payroll --harness deja --all
    (nothing)

The note inherits the excluded project, so every reading surface filters it out again. The last line promote prints describes an outcome that cannot occur for this session.

The exclusion itself holds — I checked that the note reaches neither search nor `deja sync export` — so this is the report being wrong, not a leak.

`share` grew this guard for the same reason (#1307):

    cmd/deja/share.go:46   // The exclude list is a privacy control … It only ran at ingest
    cmd/deja/share.go:49   if sources.ExcludedProject(s.Project) { … }

promote copies the session's opening line into deja's own store and says nothing about the pattern. Either it should refuse the way share does, or it should say the note is excluded too — but not promise a recall that will not happen.

vshulcz:
Closes #2278.

    $ export DEJA_EXCLUDE_PROJECTS=secretproj
    $ deja share sec1
    deja: sec1 is in a project your exclude list covers …                 exit 1
    $ deja promote sec1 --state accepted
    promoted claude:sec1 as accepted: the payroll export for acme-prod drops rows
    the note now outranks the raw transcript in recall                    exit 0

The note is written and then nothing can read it: it inherits the excluded project, so every reading surface filters it out again. `deja index && deja search payroll --harness deja --all` returns nothing.

share grew this guard for the same reason (#1307) — the pattern only runs at ingest, so an index built before it still holds the session. promote now refuses in the same shape, and says why the note would not have helped:

    deja: sec1 is in a project your exclude list covers — a note promoted from it is filtered out of recall too; `deja index --rebuild` drops the session, or remove the pattern to promote it

The exclusion itself was holding: I checked that the note reached neither search nor `deja sync export`. What was wrong was the report — and that promote copies the session's opening line into deja's own store on the way.

### DIFF
--- a/cmd/deja/promote.go
+++ b/cmd/deja/promote.go
@@ -84,6 +84,15 @@ func runPromote(dir string, args []string, stdout io.Writer) error {
 	if err := denyPolicyHidden(prefix, s, os.Stderr); err != nil {
 		return err
 	}
+	// The exclude list is a privacy control that only runs at ingest, so an
+	// index built before the pattern still holds the session — share refuses
+	// it for that reason (#1307). promote copied the session's opening line
+	// into deja's own store and reported that the note now outranks the
+	// transcript in recall, while every reading surface filters that note out
+	// again for the same project (#2278).
+	if sources.ExcludedProject(s.Project) {
+		return fmt.Errorf("%s is in a project your exclude list covers — a note promoted from it is filtered out of recall too; `deja index --rebuild` drops the session, or remove the pattern to promote it", prefix)
+	}
 	src := s.Harness + ":" + s.ID
 	if s.Harness == "deja" {
 		// A correction on a promoted note is what `promote` tells you to write,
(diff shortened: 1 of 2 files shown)


vshulcz:
Followed my own guard one step further before merging: it also refused a correction to a note that already existed (`deja promote deja-note-claude-sec1 --state rejected`), since the note carries the excluded project. That is the wrong side of the line — the guard is about copying new text out of an excluded project, and changing a state field copies nothing, while retracting a mark on work you have since excluded is exactly the cleanup someone would want. Corrections are exempt now; `--to <path>` and promoting a source session still refuse. Also worth its own issue: `deja handoff` prints the digest of a session in an excluded project without a word.

PR #2283, opened 2026-08-28
fix(fix): say deja is waiting for a second sighting, not that nothing ran

### LINKED ISSUE #2282: fix says nothing was ever run after an error it is holding a candidate for
One session that hit an error and ran something after it:

    go build ./...
    undefined: snorblefunc in vendor/blarg/api.go
    go mod vendor && go build ./...
    ok

    $ deja fix 'undefined: snorblefunc in vendor/blarg/api.go'
    deja: no session on this machine ran a command after that error

A session did, and deja recorded it — as a candidate, because one sighting is not evidence the remedy worked:

    internal/index/fixpair.go:450   // One session doing something after an error is not evidence that it
    internal/index/fixpair.go:452   if p.Candidate { continue }

The same error and remedy in a second session produces the answer:

    undefined: snorblefunc in vendor/blarg/api.go · 2026-08-28
      ran next: $ go mod vendor && go build ./...

The withholding is deliberate and right — a single sighting is a guess. The sentence is what is wrong: it says nothing was ever run after that error, while deja is holding exactly that and waiting for a second sighting. Someone who knows they fixed it an hour ago reads that line as "deja lost it".

    cmd/deja/fix.go:79   "deja: no session on this machine ran a command after that error"

vshulcz:
Closes #2282.

One session that hit an error and ran something after it:

    $ deja fix 'undefined: snorblefunc in vendor/blarg/api.go'
    deja: no session on this machine ran a command after that error

A session did. deja recorded the pair and is holding it as a candidate, because one sighting is not evidence the remedy worked (fixpair.go:450). A second sighting produces the answer.

The withholding is right; the sentence was not. Someone who fixed that error an hour ago reads it as deja having lost it.

    deja: one session ran something after that error, and nothing has confirmed it worked — deja waits for a second sighting before naming a remedy

An error nobody has hit still gets the plain line, and a confirmed pair still names the remedy — both covered by the test, which was also run red against the unpatched tree.

`FixCandidateSeen` takes the same project gate as `FixesFor`, so a candidate from a project the trust policy withholds does not change the answer either.

### DIFF
--- a/cmd/deja/fix.go
+++ b/cmd/deja/fix.go
@@ -76,6 +76,15 @@ func runFix(dir string, args []string, stdout io.Writer) error {
 		// "never seen it", but the test it used rejects `Error: …`, `npm ERR!`
 		// and any line with a quote — the exact output people paste — so it
 		// accused users of not pasting an error when they had.
+		// Held-but-unconfirmed is not the same as never seen, and saying the
+		// second about the first tells someone who fixed this an hour ago that
+		// deja lost it (#2282).
+		if index.FixCandidateSeen(dir, text, func(project string) bool {
+			return pol.Allows(policy.ActivationSearch, project)
+		}) {
+			fmt.Fprintln(stdout, "deja: one session ran something after that error, and nothing has confirmed it worked — deja waits for a second sighting before naming a remedy")
+			return nil
+		}
 		fmt.Fprintln(stdout, "deja: no session on this machine ran a command after that error")
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


vshulcz:
Checked the neighbours before merging: the MCP `fix` tool printed the same false sentence to the agent — the one caller that would otherwise re-derive the remedy — so it now gives the same answer, gated by ActivationMCP rather than ActivationSearch. Left alone: `hook-tool-after`, which fires after every failing Bash command; a line on every first sighting would be noise, and it only reads confirmed pairs. Cost of the extra lookup is one gob read of at most 4,000 entries, on the empty path of an interactive command.

Issue #834, opened 2026-08-01
on an empty machine, files, ctx and restore answer as if the query was the problem


vshulcz:
Clean machine, nothing indexed:

```
$ deja files anything      → no sessions mention "anything"
$ deja ctx anything        → deja: no session matches "anything"
$ deja restore some/file.go → no replaced spans recorded for "some/file.go"
```

The commands that handle it, same machine:

```
$ deja last     → no sessions indexed yet — no agent history was found on this machine; `deja sources` shows where deja looked
$ deja anything → no matches for "anything" — no agent history was found … (#832)
$ deja friction → nothing recurring — no sessions are indexed yet
```

`emptyIndexHint` exists for exactly this and is used by `last`, `blame` and now search. Three commands still phrase an empty store as a miss on the argument, which reads as "deja looked and your thing is not there" when deja has nothing to look in.

PR #2289, opened 2026-08-28
fix(stats): the card says the policy is withholding, not that nothing is indexed

### LINKED ISSUE #2288: the stats card says nothing is indexed when the policy is withholding it
One session indexed, and a trust policy that withholds every local session:

    {"activations":{"search":{"local":false},"mcp":{"local":false},"auto":{"local":false}}}

The three surfaces of `deja stats` disagree:

    terminal:  deja: nothing to report — the trust policy withholds every indexed session from this path
    --json:    {"total_sessions":0,"policy_withheld":1,…}
    --card:    nothing indexed yet — run deja index

The session is indexed. `deja index` again changes nothing, so the card sends the reader to fix a state the tool is not in — the shape rejected for other commands in #637 and #834.

    cmd/deja/statscard_term.go:221   default: return "", "nothing indexed yet — run deja index"

The report already carries what tells the two apart:

    internal/stats/stats.go:27   PolicyWithheld int `json:"policy_withheld,omitempty"`

The terminal line and the JSON both use it; the card is the one that does not.

vshulcz:
Closes #2288.

One session indexed, a policy that withholds every local session, and the three surfaces of `deja stats` disagreeing:

    terminal:  deja: nothing to report — the trust policy withholds every indexed session from this path
    --json:    {"total_sessions":0,"policy_withheld":1,…}
    --card:    nothing indexed yet — run deja index

Indexing again changes nothing in that state, which is the shape #637 and #834 rejected elsewhere. The card now says what the other two already say:

    3
    sessions your trust policy keeps to itself

An empty store keeps the old sentence, and a store with sessions is untouched — both pinned by the test.

### DIFF
--- a/cmd/deja/statscard_term.go
+++ b/cmd/deja/statscard_term.go
@@ -218,6 +218,11 @@ func heroStat(r stats.Report) (string, string) {
 		return formatStatNumber(handed), "recalls handed to your agents"
 	case r.TotalSessions > 0:
 		return formatStatNumber(r.TotalSessions), "sessions of agent history, all searchable"
+	case r.PolicyWithheld > 0:
+		// Indexed and withheld is not empty: `deja index` changes nothing in
+		// that state, and the terminal line and the JSON both already tell the
+		// two apart (#2288).
+		return formatStatNumber(r.PolicyWithheld), "sessions your trust policy keeps to itself"
 	default:
 		return "", "nothing indexed yet — run deja index"
 	}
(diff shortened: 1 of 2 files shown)


vshulcz:
Checked the two things that decide whether this is complete. Mixed state — some sessions visible, some withheld — keeps the ordinary "N sessions of agent history, all searchable" line, because that case sits above the new one and is the true one there; the new sentence only appears when everything is withheld. And the SVG takes a different path (`cardPunchline`, statscard.go:153) whose empty case is "Your coding-agent memory, indexed and searchable." — vague rather than false, so it needed no fix, and I would not want an SVG people post publicly to announce how many sessions their policy hides.

PR #2291, opened 2026-08-28
fix(sync): count records in singular on the ssh push too

### LINKED ISSUE #2290: sync ssh says "exported 1 records"
`deja sync ssh` says "exported 1 records".

    $ deja sync ssh nosuchhost.invalid
    deja: exported 1 records
    deja: ssh nosuchhost.invalid: exit status 255: ssh: Could not resolve hostname …

The local path counts properly for the same act:

    $ deja sync export /tmp/batch
    deja: exported 1 record

    cmd/deja/sync_ssh.go:193   fmt.Fprintf(os.Stdout, "deja: exported %d records\n", n)
    cmd/deja/sync.go:154       fmt.Fprintf(os.Stdout, "deja: exported %d record%s\n", n, pluralS(n))

Small, and it is the line printed on the first sync somebody runs against a new machine.

The rest of that surface holds, for the record: `doctor` reports an unreachable peer as "never exchanged — last attempt failed: …", and a batch whose records carry no session id is reported as left out rather than silently dropped.

vshulcz:
Closes #2290.

    $ deja sync ssh <host>
    deja: exported 1 records

The local path counts properly for the same act — `deja sync export` prints "exported 1 record" — and the two lines are one file apart. It is the line printed on the first sync anyone runs against a new machine.

The count now goes through a small helper so both spellings cannot drift again.

The rest of that surface was checked and is fine: `doctor` reports an unreachable peer as "never exchanged — last attempt failed: …", `deja sync` with only that peer says so and does not pretend, and a batch whose records carry no session id is reported as left out rather than dropped in silence.

### DIFF
--- a/cmd/deja/sync_ssh.go
+++ b/cmd/deja/sync_ssh.go
@@ -190,7 +190,7 @@ func syncSSHPush(dir, host string, full bool) error {
 	if err != nil {
 		return err
 	}
-	fmt.Fprintf(os.Stdout, "deja: exported %d records\n", n)
+	fmt.Fprint(os.Stdout, sshExportedLine(n))
 	if n == 0 {
 		fmt.Fprintln(os.Stdout, "deja: nothing new to push")
 		return errNothingToSend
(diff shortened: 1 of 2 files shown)


vshulcz:
Swept the rest of the file before merging rather than fixing only what I tripped over: the pull side had the same bare format string (`deja: imported %d records`, sync_ssh.go:293) and now goes through the same helper. Those are the only two in the sync/ssh path — every other count there already uses pluralS. The tests I found asserting "exported 4 records" and "exported 0 records" all exercise the local path, which was never wrong and is untouched.

Issue #1004, opened 2026-08-04
one unwritable index, three different stories


vshulcz:
An index whose parent directory is read-only, with a new transcript since the last build:

```
$ deja search "third session"
deja: answering from the index as it was — cannot write the index at …/idx — check the directory's permissions, or point DEJA_INDEX_DIR somewhere writable

$ deja stats
deja: mkdir …/idx.tmp: permission denied            exit=1

$ deja doctor
  freshness 1 store changed since last build — run `deja index`
```

search words it. stats hands back the syscall, which #798 replaced everywhere
else. doctor calls it ordinary staleness and advises the one command that
cannot succeed here: `deja index` fails with the same permission error.

Issue #735, opened 2026-08-01
doctor reports a damaged index as "built, up to date"


vshulcz:
```
$ rm -rf ~/.deja/buckets     # postings gone; the index cannot answer anything
$ deja doctor
Index:
  status   built (size=2.5 KB, updated=2026-08-01 08:25)
  freshness up to date
$ echo $?
0
```

Same for a corrupt `sessions.gob`. The store is unusable until something rebuilds it, and the command whose job is "is my setup right" says it is fine.

The check already exists — `recordsIntact` in internal/index/manifest.go verifies the record log's size and the presence of `buckets/`, and it is what makes the next search silently rebuild. `doctor` does not consult it.

It self-heals on the next query, so nothing is permanently lost; what is wrong is the answer. Someone running `doctor` after a partial copy, an interrupted sync, or a `find -delete` that caught too much is told the index is current.

PR #2293, opened 2026-08-28
fix(doctor): report a damaged index in --json too

### LINKED ISSUE #2292: doctor --json calls a damaged index "ok" while the human line says it is damaged
`deja doctor` names a damaged index; `deja doctor --json` calls the same store "ok".

With `records.bin` replaced by seven bytes while the manifest still says 58:

    Index:
      status   built (size=2.1 KB, updated=2026-08-28 08:00)
      integrity damaged — records or postings are missing; the next search rebuilds the index

    "index": {"state":"ok","path":"…/index.db","stale_stores":0,"sessions_stamped_ahead":0}

No integrity field appears anywhere in the JSON report, so a script watching it for index health is told the store is fine while it cannot answer a single query. The human line for this exists because of #735 — "built, up to date" about a store that could not return a result.

    cmd/deja/doctor.go:1270        if index.Damaged(dir) { … }      // the human line
    cmd/deja/doctor_report.go:619  result.State = "ok"              // the JSON, no such check

The JSON's index states are already "missing", "ok", "stale" and the unwritable case from #1004; damage is the one condition the two surfaces disagree about.

Adding a state (or a boolean) is a schema change, so which shape you want is your call — `"state":"damaged"` reads consistently with the others, and a separate `"damaged":true` would keep existing consumers of `state` working.

vshulcz:
Closes #2292.

An index whose `records.bin` was replaced with seven bytes while the manifest still says 58:

    doctor:        integrity damaged — records or postings are missing; the next search rebuilds the index
    doctor --json: "index": {"state":"ok", …}

The human line exists because of #735 — "built, up to date" about a store that could not return a result — and the JSON never got it, so a script watching index health was told the store was fine.

`index.state` is now `damaged` in that case, ahead of `stale`: a store that cannot answer is not merely behind. The states are documented now too; `stale-readonly` from #1004 was undocumented as well and is named in the same sentence.

Two notes on the test change: `TestDoctorIndexStates` built its fixture by writing two files called `manifest.gob` and `sessions.gob` containing the word "fixture". That is a damaged index by deja's own definition, so it now reports `damaged` and the staleness arithmetic the test is about could not be reached — it builds a real index instead.

### DIFF
--- a/cmd/deja/doctor_json_damaged_test.go
+++ b/cmd/deja/doctor_json_damaged_test.go
@@ -1,0 +1,80 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/vshulcz/deja-vu/internal/index"
+)
+
+// The human doctor names a damaged index (#735) and the JSON called the same
+// store "ok", so a script watching it for index health was told the store was
+// fine while it could not answer a query (#2292).
+func TestDoctorJSONReportsADamagedIndex(t *testing.T) {
+	tmp := t.TempDir()
+	home := filepath.Join(tmp, "home")
+	if err := os.MkdirAll(home, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	t.Setenv("HOME", home)
+	t.Setenv("USERPROFILE", home)
+	claude := filepath.Join(tmp, "claude", "-work-app")
+	if err := os.MkdirAll(claude, 0o755); err != nil {
+		t.Fatal(err)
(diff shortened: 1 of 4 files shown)


vshulcz:
Two things checked against the repo's own rules before merging. Schema: the stability policy says a breaking change is a field removal, rename or type change (docs/json-output.md:9-13) — a new value in an existing enum is none of those, and `stale-readonly` was added the same way by #1004 without a bump, so this stays at schema_version 2. Placement: the damaged check sits after `stale_stores` and `sessions_stamped_ahead` are populated, so the early return keeps both fields for a consumer that reads them alongside the state.

PR #2295, opened 2026-08-28
fix(usage): stamp an injection once for both logs

### LINKED ISSUE #2294: one injection is logged twice with two timestamps, so the event and its digest cannot be joined
One injection is written twice with two different timestamps, so `deja log --json` and `deja log --last --json` cannot be joined.

    index.db.usage.jsonl       {"t":"2026-08-28T05:31:53.604308Z","kind":"dejavu","bytes":578,"sessions":1,…}
    index.db.injections.jsonl  {"t":"2026-08-28T05:31:53.604364Z","kind":"dejavu","sessions":1,"bytes":578,…}

Same kind, same size, same session count, 56 microseconds apart — and neither record carries an id. So:

    $ deja log --json          newest entry t = …604308Z
    $ deja log --last --json   the digest   t = …604364Z

A consumer reading the event list and then asking for the digest has no way to tell which event that digest belongs to; matching on (kind, bytes, sessions) is the best it can do, and two injections of the same size in the same second defeat that too.

Both stamps are taken separately inside one function:

    internal/usage/snapshots.go:126   RecordResultRaw(...)     → time.Now() at usage.go:235
    internal/usage/snapshots.go:127   snapshotWriteInto(...)   → time.Now() at snapshots.go:197

and the same pairing is in `RecordDigestInto` (snapshots.go:113-114).

Stamping once and passing that instant to both writers would make the two logs joinable, costs nothing, and changes no field names or types.

vshulcz:
Closes #2294.

One injection wrote two records with two `time.Now()` calls:

    usage.jsonl        {"t":"…604308Z","kind":"dejavu","bytes":578,"sessions":1}
    injections.jsonl   {"t":"…604364Z","kind":"dejavu","sessions":1,"bytes":578}

so `deja log --json` and `deja log --last --json` described the same injection at two different instants, and neither record carries an id to join on.

Both pair writers — `RecordDigestInto` and `RecordDigestPolicyInto` — now take the instant once and hand it to both logs. No field names, types or meanings change; the two logs simply agree.

    after: event t = 2026-08-28T05:36:00.230518Z
           digest t = 2026-08-28T05:36:00.230518Z

The public API is untouched: `recordFull` and `snapshotWriteInto` keep their signatures and delegate to `…At` variants.

### DIFF
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -311,8 +311,7 @@ func callMCPTool(dir, name string, raw json.RawMessage) (string, error) {
 		if err == nil {
 			text = frameRecall(text) + env
 			deliverEnv()
-			usage.RecordServedSessions(dir, usage.KindRecall, len(text), sessions, sessions == 0, raw, ids)
-			usage.SnapshotPolicy(dir, usage.KindRecall, text, sessions, policy.Load().Describe(policy.ActivationMCP))
+			usage.RecordServedSnapshot(dir, usage.KindRecall, text, sessions, raw, ids, policy.Load().Describe(policy.ActivationMCP))
 		}
 		return text, err
 	case "recall_context":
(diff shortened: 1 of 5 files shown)


vshulcz:
Swept for the rest of the pattern before merging and found the fix was half of it: the MCP surfaces write both logs from the call site — `recall` and `recall_context` (mcp.go:314), `blame` (mcp.go:380) and the resource reader (mcp_resources.go:117) each called an event writer and then a snapshot writer, which is two instants for one act. That is in fact the pair I measured first, since the probe that started this used an MCP recall. They now go through one `RecordServedSnapshot`, and the test covers all three shapes.

Issue #2324, opened 2026-08-28
a stored digest records no projects, so the page cannot check it against the policy


vshulcz:
A stored digest carries no projects, so the page cannot check it against the policy on its own.

#2316 keeps withheld projects off the view page by recognising their names inside the digest text, and it learns those names from the sessions still in the index. Forget the sessions and the name is unknown:

    $ deja forget --project imported:secretclient/api
    messages dropped: 2 · tombstones added: 2

    (policy tightened to local-only)
    $ deja view --no-open
    page mentions secretclient:      1
    page mentions the peer sentence: 1

A `Snapshot` records `t`, `kind`, `sessions`, `bytes`, `policy`, `terms`, `into` and `digest`. The projects the digest was built from are not among them, and the trust policy can answer for a project name with no index at all — so recording them closes this without depending on what the store still holds. Additive, the way `into` went onto the event in #2307.

Separate, and worth its own issue rather than folding in here: `deja forget` drops the messages and adds tombstones while the injection log keeps the text, so the page republishes forgotten content even with no policy in play.

PR #2350, opened 2026-08-28
fix(hook): record the projects a session-start digest came from

### LINKED ISSUE #2349: the environment block survives forgetting the project it came from
Four sessions in `clients/acme/api` hit the same wall. A session start in a fresh checkout gets the environment block:

    This machine, from deja's index of past sessions across every agent used here:
    - 4 separate sessions hit `panic: acme ledger overflow in ledger/quaxbolt.go`

The stored record:

    kind: hook | projects: (none recorded) | carries the client's error: true

Forget that project:

    $ deja forget --project clients/acme/api
    (no "removed N stored digest" line — the digest stays)

The block holds content derived from the project, names the project nowhere, and #2329's sweep can only match a record by its `projects` field or by the project name inside its prose. This prose names the error, not the project, so the text outlives the forget.

The session-start and antigravity hooks build their digest through `cachedHookDigestFor` and record no projects. #2324 added the field and wired recall, recall_context, resource reads, handoff and the prompt hook, leaving these two on the argument that their digests are about the checkout the reader is in — an argument #2333 already disproved for the project list, and which does not hold at all for a block aggregated across every project on the machine.

vshulcz:
#2324 gave a stored digest its `projects` and wired recall, recall_context, resource reads, handoff and the prompt hook, leaving the session-start and antigravity hooks on the argument that their digests are about the checkout you are in. The environment block is the counter-example: it is aggregated from every project's walls, names none of them in its text, and so survived `deja forget --project X` — the sweep in #2329 can only match by the field or by a project name in the prose.

`environmentBlockFrom` now reports the projects behind its walls, the digest path carries them (through the hook cache, whose gate gained a shape marker so an entry written by an older binary is not served as a digest from nowhere), and both recorders pass them on.

Fixes #2349.

### DIFF
--- a/cmd/deja/environment.go
+++ b/cmd/deja/environment.go
@@ -40,14 +40,25 @@ const (
 // received error text drawn from the sessions that denial had just hidden, plus
 // the harnesses and dates behind it (#659). The block reads the manifest
 // directly, so nothing upstream could have filtered it.
+// environmentBlock is environmentBlockFrom without the projects, for callers
+// that only print it.
 func environmentBlock(dir, activation string) string {
+	text, _ := environmentBlockFrom(dir, activation)
+	return text
+}
+
+// environmentBlockFrom also reports the projects whose sessions the walls came
+// from. The block names none of them in its text — it is about the machine —
+// so a record without them could not be reached when one of those projects was
+// forgotten (#2349).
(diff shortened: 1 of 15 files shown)


vshulcz:
Rebuilt on the updated hunt branch, where main's session-id work landed in the same functions — reopened as a fresh branch.

PR #2371, opened 2026-08-28
test: guard MCP manifest server names

### LINKED ISSUE #1830: seven manifests all name the MCP server "deja" and nothing keeps them that way
Seven files in this repository declare an MCP server, and every one of them names it `deja`:

```
claude-plugin/.claude-plugin/plugin.json   deja   ${CLAUDE_PLUGIN_ROOT}/mcp/deja-mcp.sh
codex-plugin/.mcp.json                     deja   deja mcp
extensions/grok/.mcp.json                  deja   node ${GROK_PLUGIN_ROOT}/bin/deja-mcp.mjs
extensions/kimi/kimi.plugin.json           deja   node ./bin/deja-mcp.mjs
gemini-extension.json                      deja   deja mcp
kimi.plugin.json                           deja   node ./extensions/kimi/bin/deja-mcp.mjs
mcp.json                                   deja   deja mcp
```

That shared name is not cosmetic — it is the thing that makes a double install harmless. Hosts key MCP servers by name, so when `deja install` has written one and a plugin carries another, the two collapse into a single server instead of listing every tool twice. Verified on Gemini CLI and Qwen Code (`… mcp list` reports one `deja`), and Grok Build refuses a second extension under a name it already has.

Nothing tests it. Ship one of these as `deja-vu` or `dejavu` and a user with both wirings quietly gets six tools twice — the fault Zed had before one shared context-server id made it impossible (#1571).

Individual manifests are tested (`plugin_manifest_test.go` checks the Claude one, `agent_plugin_test.go` checks `mcp.json`), but each in isolation, and none of them asserts the name is the same across all seven.

## What a fix looks like

One test in `cmd/deja/` that finds every JSON file in the repository carrying an `mcpServers` object and asserts each declares exactly one server named `deja`. Finding them by walking rather than by a hardcoded list is the point: a file added later is the case that would otherwise slip through, and the walk catches it for free.

Watch for two things:

- `mcpServers` is nested at different depths — top level in `mcp.json` and `.mcp.json`, under the manifest root in `plugin.json` and `kimi.plugin.json`. A recursive search is simpler than encoding each shape.
- Do not walk `node_modules/` or `.claude/worktrees/` if they exist locally.

Check it by mutation before you open the PR: rename the server in one manifest and make sure the test fails naming that file.

kkkhs:
## Summary
Adds a repository-walk test that finds JSON manifests with `mcpServers` objects and fails if any manifest declares anything other than the single `deja` server.

Fixes #1830.

## Test plan
Added `TestMCPServerNamesStayConsistentAcrossManifests`; mutation-checking one manifest server name to `dejavu` fails on that file.

## Notes
None.


vshulcz:
reviewed and merging — thank you for taking this one.

ran it rather than read it, and checked both directions myself:

- renaming the server in `mcp.json` to `deja-vu` fails with the file and both names: `mcp.json declares [deja-vu], want [deja]`. that is the drift the issue was about.
- adding an eighth manifest trips the count and lists all eight, so whoever added it sees which one is theirs.

the count of seven is the part i looked at hardest, because a test that fails on *addition* can read as stale rather than as a finding. it earns it here: a new manifest is exactly the moment someone can name the server something else, and the message lists the files instead of only the number.

one thing i verified rather than assumed, in case it comes up later: `codex-plugin/.codex-plugin/plugin.json` carries `"mcpServers": "./.mcp.json"` — a string pointing at another file, not an object — so the walk correctly does not count it, and the file it points at is one of the seven. if that pointer ever becomes an inline object, the guard picks it up as an eighth. nothing to change; i just wanted it written down.

green here: `go test ./...`, `go vet`, and the guard itself.

closes #1830, which had sat as a `good first issue` for a month. exactly what that label is for.

PR #2648, opened 2026-08-30
fix(install): update the entry that already runs deja instead of adding a second

### LINKED ISSUE #2269: install adds a second registration when the server is already wired under another name, and doctor cannot see the pair
Found while preparing #2260, on the same setup: both `~/.claude.json` and Codex's `config.toml` carried **two** MCP registrations for the same deja binary — one named `deja`, one named `deja-vu`, sha256-identical command, identical args (v0.17.0 at the time). Every session started the server twice, carried the six-tool schema twice (`mcp__deja__*` and `mcp__deja-vu__*` side by side), and got the server instructions injected twice. `deja doctor` reported everything wired and healthy.

How the `deja-vu`-named entries got there is not tracked on this machine — the only first-party writer names the server `deja`, and the shipped plugin manifests do too, so they were presumably wired by hand at some point. But a hand add under that name is exactly the case doctor already anticipates: its comment calls `deja-vu` "the obvious other choice" and accepts any key whose command runs deja (`doctor.go:1004-1012`, pinned by `doctor_mcp_name_test.go`). The project is, after all, called deja-vu.

What turns that reasonable hand add into a silent duplicate is that install, doctor and uninstall each handle it one notch worse than doctor's read path does:

- `installClaude` sets `m["deja"]` without looking at what else is in `mcpServers` (`cmd/deja/install.go:936-938`), and `installCodex` manages only the `[mcp_servers.deja]` block (`install.go:1148-1152`). So a config with a working `deja-vu` entry gains a second registration the moment `deja install` runs for any other reason — a hook update, a new harness.
- `doctorJSONWired` answers "wired?" as a boolean and returns on the first hit, so two entries running deja look exactly like one (`doctor.go:989-1014`; `doctorTOMLWired` at `:1057` has the same shape). The duplicated setup never finds out.
- `deja uninstall` deletes only the key `deja`, so the duplicate survives uninstall too.

Two halves to a fix, and I am happy to send the PR if the shape sounds right:

- **install**: before adding `deja`, scan the existing entries for one whose command already runs this binary — `mcpEntryRunsDeja` (`doctor.go:1020`) is that test for the JSON configs, `commandIsDeja` (`doctor.go:1044`) for the TOML side — and if found, update that entry in place under its existing key instead of adding a second.
- **doctor**: when more than one entry in the same config runs deja, say so. That is a line the current boolean cannot produce, and it is the only way an already-duplicated setup ever learns it is paying for two.

Environment of the sighting: Windows 11, deja v0.17.0 at the time, Claude Code and Codex configs affected the same way; cleaned up by hand since. Read against `main` at 9ed8498.

Sora-bluesky:
Fixes #2269, in the shape confirmed there.

**install** finds the entry whose command already runs this binary and updates it in place under its existing key, instead of writing a second one named `deja` beside it. The JSON side is the change you predicted: a `dejaEntryKey` lookup in front of the five `mergeDejaEntry` call sites, so a hand-wired `deja-vu` keeps its key, its `env`, and anything else `mergeDejaEntry` already preserves. A literal `deja` key still wins when both exist — install has no way to know which of the two the person means to keep, so it updates its own and leaves the naming question to doctor.

The codex/grok TOML needed more scaffolding than the JSON half, since `installTOML` managed exactly one block by name. It now recognizes a foreign `[mcp_servers.X]` block that runs deja and rewrites only the lines deja owns (`type`, `command`, `args`) inside it, keeping a hand-set timeout, an env, or a comment where they were. Trailing comments on the header line survive too — the first parser draft dropped `[mcp_servers.deja-vu] # hand-wired` whole, which would have re-created exactly the duplicate this change removes; a test pins that now.

**doctor** keeps its wired row and gains the line the boolean could not produce. When two or more entries in one config run deja it says so under the row, with your wording:

```
two entries in this config run deja (`deja`, `deja-vu`) — every session starts the server twice
```

The TOML probe attributes commands to their `[mcp_servers.X]` block rather than scanning the whole file, which also stops a hook line elsewhere in codex's config from reading as MCP wiring. Args count as well as commands, since Windows wiring runs deja behind the `cmd /c` shim.

**uninstall** deletes only the literal `deja` key, as before, and now says what it left: `left "deja-vu", which also runs deja — deja did not write it, so uninstall does not remove it`.

Out of scope, deliberately: the zed/goose/hermes writers still touch only a literal `deja` (their configs go through format-specific editors this change does not extend), and `doctorOpenClawWired` still answers by key. The sighted configs from the issue — `~/.claude.json` and codex's `config.toml` — plus every harness sharing the common JSON shape are covered.

Tests: seven new ones in `mcp_duplicate_registration_test.go`, each failing on main before the change except the both-keys case, which pins that install leaves the foreign entry untouched. `go test ./cmd/deja` passes with the same set of environment-specific failures this Windows machine shows on an unmodified checkout.

### DIFF
--- a/cmd/deja/coverage_extra_test.go
+++ b/cmd/deja/coverage_extra_test.go
@@ -156,7 +156,7 @@ func TestInstallWriteAndJSONEdges(t *testing.T) {
 	if made, err := backupOnce(filepath.Join(dir, "missing")); err != nil || made {
 		t.Fatal(err)
 	}
-	if _, _, err := updateOpencodeJSON([]byte(`{"mcp":`), "/bin/deja", false); err == nil {
+	if _, _, err := updateOpencodeJSON([]byte(`{"mcp":`), "", "/bin/deja", false); err == nil {
 		t.Fatal("expected malformed opencode json error")
 	}
 	if b, _, err := updateOpencodeJSONC([]byte("{}\n"), "/bin/deja", true); err != nil || string(b) != "{}\n" {
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -1300,6 +1300,7 @@ func mcpEntryRunsDeja(v any) bool {
 			return true
 		}
 	case []any:
+		// Opencode's entry keeps the command as one list, written by deja's own installer.
 		for _, word := range cmd {
 			if s, ok := word.(string); ok && commandIsDeja(s) {
 				return true
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -982,6 +982,16 @@ func doctorMCP(w io.Writer) {
 			status = "plugin"
 		}
 		fmt.Fprintf(w, "  %-12s %-14s guidance %-11s %s\n", c.name, status, guidanceStatus(guidanceHarness(c.name)), reportPath(c.path))
+		// One "wired" can be two registrations: a hand add under another name
+		// — the project is called deja-vu, after all — plus the `deja` a later
+		// install wrote beside it. Each session then starts the server twice
+		// and carries the tool schema twice, and the boolean above cannot say
+		// so (#2269).
+		if status == "wired" && c.dupes != nil {
+			if keys := c.dupes(c.path); len(keys) >= 2 {
+				fmt.Fprintf(w, "  %-12s %s\n", "", doctorMCPDuplicateNote(keys))
+			}
+		}
 		// "Wired" says the server is declared, not that it can start. A config
 		// naming a binary that is gone — a restored backup, a hand edit, a
 		// machine where deja moved and one file was fixed by hand — read as
@@ -996,6 +1006,22 @@ func doctorMCP(w io.Writer) {
 			fmt.Fprintf(w, "  %-12s %s\n", "", note)
 		}
 	}
+}
+
+// doctorMCPDuplicateNote is the line a duplicated setup never got to read: it
+// names every key so the reader can decide which one is theirs. Which one to
+// remove is not deja's call — one of them may be a hand add carrying fields
+// deja does not know about (#2269).
+func doctorMCPDuplicateNote(keys []string) string {
+	quoted := make([]string, len(keys))
+	for i, key := range keys {
+		quoted[i] = "`" + strings.ReplaceAll(key, "`", "'") + "`"
+	}
+	names := strings.Join(quoted, ", ")
+	if len(keys) == 2 {
+		return fmt.Sprintf("two entries in this config run deja (%s) — every session starts the server twice", names)
+	}
+	return fmt.Sprintf("%d entries in this config run deja (%s) — every session starts the server %d times", len(keys), names, len(keys))
 }
 
 // dejaCommandMissing returns the deja binary a config names when that file is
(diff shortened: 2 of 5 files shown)


vshulcz:
read the whole diff and ran it against a real config rather than only the tests. the install half does exactly what it says: a hand-wired `deja-vu` keeps its key and its `env`, and only `type`/`command`/`args` are rewritten. the TOML parser earning its own header-comment test is the right instinct — that draft would have re-created the duplicate it removes.

**one blocker, and it is the case your tests do not cover.**

`TestUninstallClaudeLeavesAndNamesTheForeignEntry` sets up `deja` *and* a genuinely foreign `deja-vu` (`/hand/bin/deja serve`), and leaving that one is right. the case this PR creates is the other one: a config with **only** the foreign entry, which install now adopts.

    # before: {"deja-vu": {"command": "/usr/local/bin/deja", "args": ["mcp"], "env": {"MY_FLAG": "1"}}}
    deja install claude
    # after:  deja-vu now has deja's own type/command/args, env kept
    deja uninstall claude
    # after:  unchanged — the entry is still there, still runs deja

so `deja uninstall` leaves deja wired and starting on every session. worse, the person is not told: i captured stdout on that run and the note never appears — the `.claude.json` row is not printed at all when the file comes back unchanged, so `leftDejaEntriesNote` reaches nobody in the flow that needs it.

my "do not delete what deja did not write" from #2269 was right for an entry deja only reads. it stops being right once install has written into it: after adoption, the command line is ours.

the repo already has the memory for this — `noteBlockAdded`/`blockWasAdded`/`forgetBlockAdded`, used a few lines above to decide whether an empty `mcpServers` block was deja's to remove. an adopted key wants the same treatment: record it on adoption, and on uninstall remove it (and forget the record) when it is the one deja adopted, leaving a key deja never touched alone. that keeps both halves honest and needs no new mechanism.

**two notes, not blockers.**

- `mcpEntryRunsDeja` now accepts `command` as a list. worth a line saying which client writes it that way — if it is speculative, i would rather it were not there than there untested.
- the duplicate line is per config. a machine with the pair in both `.claude.json` and codex's `config.toml` — which is what #2269 actually found — gets it twice with no hint they are the same mistake. fine as is; mentioning in case you want a summary line.

everything else i checked passes: `go test ./...` clean on your branch, e2e green, and your seven new tests all run.

fix the uninstall half and i will merge it. thank you for taking the TOML side seriously — that was the part i expected to come back half-done, and it did not.

Sora-bluesky:
Fixed in 25815f2, the way you prescribed — the ownership record does all of it, no new mechanism.

Install records an adoption as `<container>.<key>` through the same `noteBlockAdded` the empty-block cleanup uses (`mcpServers.deja-vu`, `mcp.servers.X` for openclaw, `mcp.X` for opencode, `mcp_servers.X` on the TOML side), at every site `dejaEntryKey` or the TOML in-place update can adopt. Uninstall then walks the remaining keys: one with a record is removed and forgotten; one without stays and is named. The left-note is computed after the removals, so it can no longer name a key that just went. Your empty-block cleanup fires unchanged when the adopted entry was the last thing in a block deja created.

Your repro is now the first test in `mcp_adopted_uninstall_test.go`, and it fails on the code you reviewed — I ran the four new tests against the pre-fix tree to check: the adopt-then-uninstall cases (claude, codex TOML, and a mixed config where an adopted `deja-vu` goes while an untouched `memory` stays and is the only one named) all fail there, and the boundary case — foreign key, no record, uninstall leaves it — passes there, pinning that the record is what changed the behavior. One honest limit: the tests exercise the in-run buffer, since install and uninstall share the process; the cross-process persist rides the same `recordWiring` fold your container records already use.

On the notes:

- `command` as a list is not speculative — it is deja's own opencode shape (`updateOpencodeJSON` writes `"command": []string{exe, "mcp"}`). The list case in `mcpEntryRunsDeja` now says so in a comment, so the next reader does not have to wonder either.
- Per-config duplicate lines: leaving as is, per your call. A cross-config summary would need doctorMCP to hold state across rows for one sentence; happy to add it if the twice-told pair bothers anyone in practice.

One behavior change to flag from the delta, in the safe direction: block removal now reads headers through the same comment-stripping the parser uses. For adopted keys that is the `removeTOMLMCPBlock` loop, so uninstall can remove an adopted `[mcp_servers.deja-vu] # hand-wired` block the same way install adopted it; for the literal block it means `removeCodexDejaBlock` now also matches `[mcp_servers.deja] # …`, which only removes more of deja's own table. A fully commented-out header still matches neither.

vshulcz:
merging.

ran my own repro against your fix rather than reading it: a config holding only a hand-wired `deja-vu`, install adopts it and keeps the `env`, uninstall takes it and leaves `mcpServers` empty as the reader's own block. then the boundary — `memory` untouched, a hand-wired `deja-vu` that install never adopted still there and named in the note. both right.

using the ownership record for it was the point and you took it further than i described: computing the left-note after the removals, so it can no longer name a key that just went, is a bug i would have shipped.

running the four new tests against the pre-fix tree to prove they fail there — including the boundary case that passes there — is the part that makes this reviewable. thank you for doing it rather than asserting it.

one thing i fixed rather than bouncing it back: `entryRunsDeja` was left unused after the doctor probe took over, and golangci-lint refused the build. removed in its own commit on your branch (and one clumsy first attempt of mine that took the comment and left the body — both on the branch, sorry for the noise).

**one thing i found that is not yours, filing separately.** removing a TOML block walks to the next line starting with `[`, so a server with a sub-table leaves the sub-table behind:

    [mcp_servers.deja-vu]      → removed
    [mcp_servers.deja-vu.env]  → left, pointing at a server that is gone

that is `removeCodexDejaBlock` on main word for word, so it predates this PR and applies to deja's own `[mcp_servers.deja]` too. your change generalises the function, it does not introduce the hole. issue coming with the repro.

on `command` as a list: good, and thank you for saying which shape writes it — that is what i wanted, not the removal.

Sora-bluesky:
Thank you for the merge, and for cleaning up `entryRunsDeja` yourself — the rework took its last caller and my gate (go vet plus the tests) does not count an unused function the way golangci-lint does. It runs locally before my next PR here; that lesson is now written down twice on this side.

On the sub-table hole: when the issue lands, I would like to take it. The walk-to-the-next-`[` boundary is the shared shape of the block helpers this PR generalised — `tomlMCPBlocks` even reads `[mcp_servers.deja-vu.env]` as an MCP server named `deja-vu.env` — so the fix wants to land once, in the boundary, rather than per function. Your repro would be the fail-before test.

PR #2649, opened 2026-08-30
fix(mcp): exit when the parent is gone and nobody has spoken for five minutes

### LINKED ISSUE #2397: `deja mcp` outlives its client on Windows when another process inherited its stdin pipe handle
On Windows 11, four `deja mcp` processes were still running with parent PIDs that no longer existed. They sat blocked and had to be stopped by hand. Earlier the same day a Codex desktop restart had reaped ~23 servers cleanly, so most exits work; these four stayed. The machine's MCP clients (Claude Code, Codex, grok CLI) are all registered with the `cmd /c deja mcp` entry that `deja install` writes on Windows (`mcpCommandArgs`, cmd/deja/install.go). They also run hooks, which on Windows means a steady stream of short-lived cmd/pwsh children; that is conjecture about the vector, but it is a lot of chances for a handle to travel.

I did not catch the holder for those four in the act, so the field case is consistent with the mechanism below rather than proof of it. The mechanism itself reproduces deterministically.

## Why it can happen

`serveMCP` has exactly one way out: the stdin read returns EOF or an error (cmd/deja/mcp.go). That path works. In a clean test, an abrupt client exit with no cleanup makes the server exit within a couple of seconds.

The gap is Windows handle inheritance. If the write end of the server's stdin pipe is inheritable in the client, any later child the client spawns with inheritance on gets a copy. `ReadFile` on the server's end returns EOF only when every write handle is closed, so as long as one such child lives, the client's death closes nothing that matters: the server stays blocked with no living client. Whether a real client leaks this way depends on how it manages handles; the repro below shows the shape, not any particular client's guilt.

## Minimal repro (Windows 11, main @ a52af00)

Python driver standing in for a leaky client (`deja` here is the built binary on PATH):

```python
import msvcrt, os, subprocess, sys

proc = subprocess.Popen(["cmd", "/c", "deja", "mcp"],
                        stdin=subprocess.PIPE, stdout=subprocess.PIPE)
proc.stdin.write(b'{"jsonrpc":"2.0","id":1,"method":"initialize"}\n')
proc.stdin.flush()
assert b'"serverInfo"' in proc.stdout.readline()

# the leak: mark the pipe's write handle inheritable, then spawn a
# sibling with inheritance on
os.set_handle_inheritable(msvcrt.get_osfhandle(proc.stdin.fileno()), True)
subprocess.Popen([sys.executable, "-c", "import time; time.sleep(600)"],
                 close_fds=False)

os._exit(0)  # die like a killed client: no cleanup
```

Observed:

- Without the two leak lines, driver exit ends the server within ~2 seconds. The EOF path is healthy.
- With them, the driver dies and `cmd`, `deja` and the sibling all stay. Killing `cmd` too reproduces the field tree exactly: `deja` alive, recorded PPID nonexistent, blocked in the stdin read.
- Killing the sibling releases the pipe and `deja` exits within ~2 seconds. In the lab, the inherited handle is the whole story.

## Possible direction

LSP handles this case in the protocol: `initialize` carries the client's `processId`, and the spec tells servers to exit when that process is not alive. MCP's `initialize` has no such field, so a deja-side equivalent would have to watch a process it picks itself (the grandparent through the `cmd /c` shim on Windows, the parent elsewhere), and that has edge cases, such as launchers that legitimately exit after spawning. So I am flagging the mechanism first rather than sending a patch. Happy to write one in whatever shape you prefer, or leave it at documented-and-known if you judge the blast radius too small.

Sora-bluesky:
Fixes #2397, in the both-and-only-both shape decided there: the server exits when the recorded parent pid is gone **and** nothing has been read for a grace period. A live client that keeps talking is never touched however its process tree looks; a launcher that exits while its client keeps talking is never touched either.

`serveMCP` itself is untouched — the command table now calls a `serveMCPProcess` wrapper that records `os.Getppid()`, stamps an atomic last-read time on every successful read through a thin `io.Reader`, and runs the watch on a 30-second tick beside the blocking read. Every existing serveMCP test reads the same server it always did. When both conditions hold, the server says why on stderr and exits:

```
deja: mcp parent pid 38656 is gone and nothing has been read for 5m0s; exiting the orphaned server (#2397)
```

**The grace period, answered rather than assumed.** Five minutes, and the comment on the constant says what it protects: it is the window a quiet-but-live client whose launcher already exited has before deja could misread it as an orphan — longer than any request gap a real session produces, short enough that leaked servers do not accumulate across a working day. A client that dies cleanly never reaches this path; EOF still ends the server in seconds. On this machine's three clients, registered through the `cmd /c` entry the issue describes, cmd waits on deja, so in the trees observed here the dead-parent half stays false while the client lives.

**"Does this pid still exist" needed one platform split after all.** The obvious Windows answer — `os.FindProcess` erroring for a missing pid — is wrong in exactly this bug's neighborhood: an exited pid stays openable while any handle to it is held, and the first draft's test caught its own implementation reporting a waited-on child as alive. The Windows side now asks `GetExitCodeProcess` for `STILL_ACTIVE`; unix probes with signal 0, EPERM counting as alive. PID reuse fails safe in both directions: a recycled pid reads as alive and the server merely keeps waiting.

**Verified on the repro, on Windows 11.** Built this branch, ran the issue's driver against it — inheritable write handle, sibling spawned with inheritance on, driver dead via `os._exit`, then the cmd shim killed to match the field tree. The sibling held the pipe the whole time, and the server exited on its own when the grace ran out, 4m46s after the parent died, with the stderr line above. Without the leak lines the EOF path still ends it — measured at 0.2 s on this build after an abrupt client kill.

Tests are testable-off-Windows as intended: a helper process stands in for the parent, killed or waited to make it dead, and the three-way table (live parent + silence, dead parent + recent read, dead parent + silence) pins that only the last one exits. `TestProcessAlive` fails on the FindProcess-only implementation.

On the `cmd /c` question from the thread: the shim traces to #170, wired for a field report on #9, so whether any of today's clients accepts a bare exe path needs checking per client — happy to probe that separately, since dropping it only shortens the tree and neither causes nor fixes the inheritance leak.

### DIFF
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -150,7 +150,7 @@ var commands = map[string]command{
 	"remember":      runRemember,
 	"promote":       func(dir string, rest []string) error { return runPromote(dir, rest, os.Stdout) },
 	"forget":        runForget,
-	"mcp":           func(dir string, _ []string) error { return serveMCP(dir, os.Stdin, os.Stdout) },
+	"mcp":           func(dir string, _ []string) error { return serveMCPProcess(dir, os.Stdin, os.Stdout) },
 	"hook-prompt": func(dir string, rest []string) error {
 		if sayIfTypedByHand("hook-prompt") {
 			return nil
--- a/cmd/deja/mcp_orphan.go
+++ b/cmd/deja/mcp_orphan.go
@@ -68,7 +68,13 @@ func serveMCPProcess(dir string, r io.Reader, w io.Writer) error {
 	var lastRead atomic.Int64
 	lastRead.Store(time.Now().UnixNano())
 
+	// Windows reports -1 when the parent cannot be read; processAlive calls
+	// that dead, and the AND would quietly collapse into the idle timeout the
+	// design rejected. No recorded parent, no watch — EOF still works.
 	ppid := os.Getppid()
+	if ppid <= 0 {
+		return serveMCP(dir, r, w)
+	}
 	stop := startOrphanWatch(
 		ppid,
 		func() time.Time {
--- a/cmd/deja/mcp_orphan.go
+++ b/cmd/deja/mcp_orphan.go
@@ -1,0 +1,87 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// mcpOrphanGrace is the window a quiet but live client gets after its launcher
+// exits before deja can mistake it for an orphan. Five minutes is longer than
+// any request/response gap a real session should produce, but short enough that
+// leaked servers do not accumulate across a working day. A client that closes
+// cleanly reaches EOF within seconds; this only bounds how long an inherited
+// write handle can keep the server alive after its parent is gone (#2397).
+const mcpOrphanGrace = 5 * time.Minute
+
+const mcpOrphanTick = 30 * time.Second
+
+// startOrphanWatch stops the leaked-handle orphan from #2397. A dead parent
+// alone means nothing because launchers legitimately exit, and silence alone
+// means nothing because quiet clients stay connected. Only both together mark
+// the orphan; a client that dies without leaking a handle already ends the
+// server through EOF within seconds.
+func startOrphanWatch(ppid int, lastRead func() time.Time, grace, tick time.Duration, exit func()) (stop func()) {
+	ticker := time.NewTicker(tick)
+	done := make(chan struct{})
+	var once sync.Once
+
+	go func() {
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				if !processAlive(ppid) && time.Since(lastRead()) > grace {
+					exit()
+					return
+				}
+			case <-done:
+				return
+			}
+		}
+	}()
+
+	return func() {
+		once.Do(func() {
+			close(done)
+		})
+	}
+}
+
+type mcpStampReader struct {
+	r        io.Reader
+	lastRead *atomic.Int64
+}
+
+func (r *mcpStampReader) Read(p []byte) (int, error) {
+	n, err := r.r.Read(p)
+	if n > 0 || err == nil {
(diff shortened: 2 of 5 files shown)


vshulcz:
merging this.

read the whole diff and ran it here: the handshake still answers, `go test ./...` and e2e are clean, and your five tests all run off Windows the way you designed them to. `TestProcessAlive` failing against the FindProcess-only implementation is the part i would have missed — an exited pid staying openable while a handle is held is exactly this bug's neighbourhood, and catching your own first draft with a test is the best thing in the PR.

the grace-period comment is what i asked for and better than i asked: it says what the number protects rather than what it is.

one thing i fixed rather than sending you back for it — `p.Release` was unchecked and golangci-lint refused the build (`mcp_orphan_unix.go:24`, errcheck). pushed to your branch as its own commit; everything else is yours untouched.

two notes for the record, neither blocking:

- `os.Exit(0)` from the watch skips any pending write, which is right here — it only fires when nothing has read for five minutes — but worth remembering if the server ever buffers its output.
- on unix a process whose parent dies is reparented and `os.Getppid()` would return 1 afterwards; recording it once at start, as you do, is what makes the check mean anything. worth a line in the comment so nobody 'simplifies' it into a live call later.

thank you for going to the Windows box and running the actual repro. i cannot do that half, and a fix for this shipped on reasoning alone would have been a guess.

Sora-bluesky:
Thank you for the merge, and for the `Release` fix — my gate here was `go vet` and the test suite, not golangci-lint, which is how an unchecked return walked past me. Noted for the next one.

Both notes taken. The `os.Exit` one is recorded where it matters least until it matters a lot — agreed that the day this server buffers output is the day that call needs revisiting. The reparenting one deserved to live in the code rather than in a merged thread, so it does now: #2711 adds the line saying only the pid recorded at start can ever go dead — a live `Getppid` answers the adopter, init or a subreaper, which is alive by construction. Comment only.

The Windows half is the part I can keep doing — if a field orphan ever shows up again after this ships, the repro harness and the box are still here.

PR #2690, opened 2026-08-30
docs(bench492): what the pass costs, on both corpora

AliceLJY:
@vshulcz asked for this in #492 after posting the generated-corpus whole build. The README carried the rig but not its results: the repetition table was short the incidental store, and the question in the issue title — what the bigram pass costs on a whole build — had been answered twice in the thread and written down nowhere.

## What went in

**The whole-build cost, from both corpora, in one section.** 4.289 s / 4.87% of a bigram-free build on the real store (`main` at `9ed8498`, 372,048 indexed messages, 55.4% carrying CJK), and +0.561 s / 68% on the generated all-Japanese arm, with the `en` control at no cost. Neither is the answer alone: a reader who takes one without the other is off by more than an order of magnitude in whichever direction they picked. That framing is yours from the issue; I have kept it as the point of the section rather than a footnote to it.

**The rejected window, kept.** The real-store run took two windows, +5.814 s and +4.289 s, and only the `-trimpath` cell separates them — 1.337 s of spread against 0.250 s, and window 1's per-round deltas changed sign twice, which the stub cannot do since its index is 114.3 MB smaller. Publishing window 1 would have been 36% high. Showing only the surviving number would teach the wrong lesson about why the third cell is in the summary at all, which was the reason for putting it there.

**A third lesson: the stub arm has to be a build, not a runtime switch.** You lost a build to this and it belongs in the file.

**The repetition table now has all four rates** — 4.000 fixture, 1.640 generator, 1.685 incidental, 1.376 native — with the sentence about which direction matters, since the dedupe halves are worth least exactly where the text is most CJK.

## Two places I departed from the thread — say the word and either goes back

**The bracket is labelled by CJK rune share, not repetition rate.** Your sentence reads "4.87% on your real store at 1.376 repetition and 68% on an all-Japanese generated one at 1.640". Both pairs are true of these two corpora, but the 4.87→68 gap is a gap in how much of the build is CJK work at all (16.6% of runes against all of them), and labelling it by repetition rate invites the reader to attribute it to dedupe efficiency instead. The repetition rates keep their own paragraph at the end, where they are what is being measured.

**The third lesson gives a different reason than you did.** You gave the runner's per-cell environment. What I could point at in the file is narrower: `run_bench.py` never compiles anything — its docstring says it only measures, it runs `<bin> index --rebuild` per cell, and `cell_env` only overrides `HOME`, the two store paths and `GOMAXPROCS`, so it does not eat an outside variable. The conclusion is the same one you reached; I wrote the mechanism a reader can verify in the file rather than the one I could not.

## Checks

- Documentation only — `git diff --stat` is `tools/bench492/README.md | 71 +++--`, nothing else in the tree.
- Every figure traced to the comment it was measured in: real store to [`5442468812`](https://github.com/vshulcz/deja-vu/issues/492#issuecomment-5442468812), generated to [`5465139373`](https://github.com/vshulcz/deja-vu/issues/492#issuecomment-5465139373), incidental store to [`5437923324`](https://github.com/vshulcz/deja-vu/issues/492#issuecomment-5437923324). Nothing rounded, rescaled or restated.
- Code facts checked against this branch rather than the thread: `cjkIndexKeys` is called from `indexKeys` (`internal/index/retrieval.go:3611`, `indexKeys` opens at 3597), and `internal/index/cjkindex_realcorpus492_test.go` is still the read-only counterpart the last paragraph points at.
- Nothing in the original text changed except the three lines that had to: `Two lessons` → `Three lessons`, and the repetition sentence split to take the fourth rate.

### DIFF
--- a/tools/bench492/README.md
+++ b/tools/bench492/README.md
@@ -22,7 +22,7 @@ runs are not comparable, for the reason in the first lesson below. The first
 reports min/median per cell. Interleaving keeps a machine-wide slowdown from
 landing on one cell; read the min column first, contention only ever adds.
 
-Two lessons from using it, so the next measurement does not relearn them:
+Three lessons from using it, so the next measurement does not relearn them:
 
 - Wall-clock deltas between two *different binaries* carry binary-layout
   noise — around ±15% of a small corpus's build time on the machine this was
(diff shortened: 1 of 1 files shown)


vshulcz:
both departures are improvements on what i wrote, and i want to say why rather than just accept them.

**the bracket labelled by CJK rune share.** you are right and my sentence was wrong in a way that would have misled people. i wrote "4.87% at 1.376 repetition against 68% at 1.640", which reads as though dedupe efficiency explains the gap. it does not: 16.6% of runes against all of them explains it, and repetition moves the per-message cost inside each corpus. putting the rates in their own paragraph is the correct separation.

**the third lesson's reason.** mine was the runner's per-cell environment, which is true and which i could not point at in the file. yours is that `run_bench.py` never compiles anything — docstring, `<bin> index --rebuild` per cell, `cell_env` overriding four things — so a reader can check it without taking either of us on trust. that is the better version of the same lesson.

checked before merging rather than after: diff is the one file, `cjkIndexKeys` is still called from `indexKeys` at retrieval.go:3611 as you say, and the generated-corpus figures are mine to the digit — 1.383 / 0.822 / +0.561 / 68%, control 0.294 against 0.316, 14.0 MB against 10.6.

keeping the rejected window was the whole point and you kept it with its reason: 1.337 s of `-trimpath` spread against 0.250 s, and per-round deltas changing sign twice where the stub cannot. a reader now learns why the third cell exists instead of inheriting a number.

thank you for waiting for the second measurement before writing any of it. the file says something neither of us could have said alone, which is what you predicted in the issue.

PR #2711, opened 2026-08-30
docs(mcp): say why the parent pid is read once

Sora-bluesky:
The line the merge review of #2649 asked for, on its second note: why the parent pid is read once. On unix an orphaned process is reparented — to init or a subreaper — and a live `Getppid` answers that adopter from then on, so only the pid recorded at start can ever go dead; inlining the call would make the watch follow a process that is alive by construction, and never fire. Comment only, no behavior change.

### DIFF
--- a/cmd/deja/mcp_orphan.go
+++ b/cmd/deja/mcp_orphan.go
@@ -68,6 +68,10 @@ func serveMCPProcess(dir string, r io.Reader, w io.Writer) error {
 	var lastRead atomic.Int64
 	lastRead.Store(time.Now().UnixNano())
 
+	// Read once, deliberately: on unix an orphaned process is reparented and a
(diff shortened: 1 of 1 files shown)


vshulcz:
merging — and the second version is right where the first was only mostly right.

"answers 1 forever after" holds where init is the reaper. with a subreaper — a systemd user session, a container init, anything that called `PR_SET_CHILD_SUBREAPER` — the orphan is adopted by that instead and `Getppid` answers a pid that is neither 1 nor the client. "the adopter from then on" covers both, and it is the version that stops someone reading it, checking `!= 1`, and building the same trap one level up.

comment only, confirmed on the diff: every changed line is a comment. green.

this is the third round on #2397 and each one was smaller than the last — mechanism, then fix with the platform trap in it, now the sentence that keeps the fix from being simplified away. thank you.

Issue #2243, opened 2026-08-27
three quarters of injected blocks carry a mention rather than a conclusion


vshulcz:
The question this answers is not "does recall find the right session" — it does, 85.3% hit@1 on LongMemEval — but "does the block it hands over carry what that session settled".

Measured over 834 blocks deja actually injected on this machine, read out of the transcripts where they landed and judged by deja's own `digest.CarriesDecision`:

```
blocks injected                 823
carrying a decision             215   (26%)
```

Three quarters carry a mention rather than a conclusion. Rebuilt as a controlled measurement — every session in the store that settles something, asked with the terms the hook would extract from the question that session actually opened with — the same picture:

```
sessions that settled something                      646
the block built from them carried it                 232   (36%)
the block came back empty                             12
a decision line naming a question word, not taken    150
```

**Where those 150 go**, attributed rather than guessed:

```
tool/other   84    the line is in tool output — a printout, not a decision
user         48    the person said it
assistant    18    the agent settled it and the block missed it
```

Two things this corrects about my own numbers, since the first pass of each was wrong in deja's favour and then against it. Asking with a single frequent word instead of the hook's terms read 25% rather than 36%. And counting a term as matched by substring — `retries` for `retry` — inflated the misses from 150 to 211. The numbers above use `prompt.Terms` and `termHits`, which is what the product uses.

**What I tried and backed out.** Within a message the densest line already yields to a concluding one (#1488). Between messages it does not, so the user slot goes to whichever message names the subject most often. Making a concluding user line win that slot changed nothing measurable: 232 → 232 carried, 150 → 151 missed. So the 48 are not lost there, and I have not found where.

That is the fifth ranking change this session that measurement rejected, which is the argument for writing this down instead of trying a sixth. The instrument is the useful part: replay every session with the terms its own question yields, judge the block with `CarriesDecision`, attribute the misses by role. Anyone attacking this should start by moving the `assistant 18` — the only group where the agent demonstrably settled the asked question and the block still missed it.

vshulcz:
Retracting the headline. I wrote this morning that three quarters of injected blocks carry a mention rather than a conclusion. Re-run on today's main and taken apart case by case, the defect is 4 sessions in 703.

The count itself barely moved:

```
                                     this morning      now
sessions that settled something              646       703
the block carried it                    232 (36%)  250 (36%)
a matching decision line, not taken          150       167
```

What changed is that I looked at the 167 instead of counting them:

```
[...]

PR #2732, opened 2026-08-30
feat: let a work record into the digest when it carries the query

vshulcz:
Draft — this reverses a rule the repo states on purpose, so it is a call to make, not a fix to merge.

Today the context digest drops every work record: tool output, commands, file lists, replaced spans. `TestContextDigestLeavesTheMachineryOut` guards that, and the reasoning holds — before those roles were labelled honestly (#560) they arrived as `user` and filled digests with blocks nobody means by context.

But that is where the evidence is. Across twelve real questions on an isolated index, of 184 messages the search matched, **148 sit in work records** the digest never looks at. Asking whether the returned digest contains the query's words at all:

    10/12  before today
    11/12  after #2729
    12/12  with matched work records let in

The change is narrow: a work record is kept **only when it carries the query** — never as filler, which is what #560 was actually about — and capped at 600 characters so a matched build log cannot eat the budget the conversation needs. Unmatched tool output and commands stay out, and the test now guards that instead.

Sizes stay inside the budget: the twelve answers run 5.9–7.8 KB, same band as before.

The argument against: `deja ctx` is also a briefing a person reads, and a command line in the middle of a conversation reads worse than prose. If that matters more than the recall, the alternative is to let work records in only for the MCP tool an agent calls and keep the CLI digest as it is — say which and I will cut it that way.

### DIFF
--- a/internal/search/context_work_records_test.go
+++ b/internal/search/context_work_records_test.go
@@ -13,9 +13,11 @@ import (
 // honestly (#560) they arrived as `user` and filled this with blocks nobody
 // meant by context.
 //
-// Measured by removing the filter: with isWorkRecord returning false, the
-// digest grew "## files" and "## edit" blocks and no test in this package
-// failed. This is that test.
+// One exception, measured: a work record that carries the query is evidence,
+// not machinery. Four in five of everything a query matches lives in these
+// records, and holding all of them back left the digest with nothing about the
+// subject on 1 of 12 real questions. They come in only when matched, never as
+// filler, and capped so a log cannot take the budget the conversation needs.
 func TestContextDigestLeavesTheMachineryOut(t *testing.T) {
 	s := model.Session{
 		Harness: "claude", Project: "work", ID: "w1",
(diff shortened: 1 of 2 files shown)


vshulcz:
Evidence against this from my own earlier measurement, which I should weigh before it gets merged.

#2243 attributed 150 misses — cases where a session settled something and the block did not carry it — by where the missed line lives:

    tool/other   84    the line is in tool output — a printout, not a decision
    user         48    the person said it
    assistant    18    the agent settled it and the block missed it

That reading judged the 84 as correctly skipped: a printout is not a conclusion, even when it holds the words.

This PR measures something different — whether the returned digest contains the query's words — and on that it goes 11/12 to 12/12. The two can diverge exactly here: a build log carries the terms and settles nothing. So the improvement I reported may be words arriving, not answers.

That does not sink it. A matched command line is often the answer to "how did we run this", which is a question about a printout. But it does mean the honest version of this change is narrower than "work records when matched": the case for it rests on commands and error text, not on tool output at large, and the metric to move is #2243's — does the block carry what the session settled — not the one I used.

Leaving it draft. If you want it, the version I would write now keeps `command` and error-bearing output and still holds back plain `tool-output` and `files`.

vshulcz:
Measured the narrow version I said I would write, on the metric I said was the right one, and it does not pay. Closing.

The stand: 120 real user turns out of a store, each searched, the top hit rendered through `PrintContext`, and the block judged twice — does it say half the query's terms (what this PR measured), and does `digest.CarriesDecision` fire on it (what #2243 measures).

```
                                  says the terms   carries a decision   avg bytes
main                                   100                103              7960
+ matched commands and error output    102                100              7983
+ matched commands only                101                104              7996
```

So the divergence I wrote about above is what actually happens. Letting matched work records in moves the word metric up by two and the decision metric **down by three** — the words arrive and displace a conclusion out of the budget. Narrowed to commands alone it is flat: one either way, which is nothing on 120 questions.

The error-output half is the part that costs. That is consistent with #2243's attribution of the 84: a printout holds the terms and settles nothing, and the digest was right to skip it.

Nothing here justifies reversing a rule the repo states on purpose, so `TestContextDigestLeavesTheMachineryOut` stays as it is.

Two caveats on my own numbers, since they are what closes this. `CarriesDecision` applied to a whole block measures the budget as much as the content — the caveat #2243 carries and this inherits. And the stand judges the top hit only; a change that helps the second-best session would not show. Neither would turn a −3 into a reason to merge.

Issue #1639, opened 2026-08-24
deja friction says the policy hides 10 sessions when it hid one


vshulcz:
`deja friction` counts withheld **records** and prints the number as **sessions**, so one hidden session with ten error records is reported as ten.

Fixture: one session, ten tool-output records, a policy that withholds everything from search.

```
$ cat policy.json
{"activations":{"search":{"*":false}}}

$ DEJA_POLICY_FILE=policy.json deja friction
deja: the trust policy hides 10 matching sessions on this path (search: local+imported) — see …/policy.json

nothing recurring — none of the 1 indexed session recorded tool output, which is what friction reads errors from
```

The store holds exactly one session, and the line under it says so.

Cause: `cmd/deja/friction.go:60` does `withheld++` inside `index.EachRecordOfRole`, which fires once per record, and hands that to `policyHiddenNote` (`cmd/deja/policyfilter.go:38`), whose sentence is `hides %d matching session%s`. Every other caller of that helper passes a session count — `policyFilterHitsCounted` and `policyFilterSessionsCounted` both count hits or sessions — so the helper's wording is right and friction's number is wrong.

Expected: the number of sessions withheld, counted the way the same function's other callers count them.

Impact: it is a sentence about the trust policy, which is the one place a wrong number is expensive — it tells the reader how much of their own history a rule is keeping from them, and it overstates it by however many error lines each session recorded.

Found during the bug hunt, item `[friction-threshold]`. The threshold itself is clean and worth recording: exactly three sessions is reported and two is not; ten sightings inside one session do not reach it, and the empty line says why ("no error hit 3 separate sessions"); the counts on screen are session counts, not sighting counts.

vshulcz:
Fixed by #1640, on main since the collector merge in #1912.

PR #2760, opened 2026-08-30
fix(how): count the sessions the policy hid, not their commands

### LINKED ISSUE #1641: deja how tells you the policy hid N sessions when it hid one
`howEntries` counts withheld **records** and the sentence it feeds says **sessions**, so one hidden session with many commands is reported as many.

Same defect as #1639, in a different command, found by the reviewer on that fix and kept out of it:

> `cmd/deja/how.go:134` has the same records-as-sessions bug — `howEntries` increments `hidden` per record fired. This affects `how.go:87` and `mcp.go:341`, which pass `hidden` to `policyHiddenNote`.

The note is `the trust policy hides %d matching session%s on this path`, and every other caller passes a session or hit count. Two callers here pass a record count: the CLI and the MCP `how` tool, so an agent gets the inflated number too.

The fix is the one #1639 used: collect `meta.Harness + ":" + meta.ID` into a set and pass its length. `how` already has that key at hand for the counts it prints.

Reproduction is the same shape as #1639's — a store where one session runs several indexed commands, and a policy that withholds it:

```
{"activations":{"search":{"*":false}}}
```

Follow-on from #1639.

vshulcz:
Closes #1641. Follow-on from #1639, which fixed the same shape in friction.

`howEntries` incremented a counter per withheld record, and the sentence it feeds says "matching sessions" — so one hidden session that ran a command five times was reported as five. Both callers pass that number on: the CLI and the MCP `how` tool.

### DIFF
--- a/cmd/deja/entry_switched_off_test.go
+++ b/cmd/deja/entry_switched_off_test.go
@@ -52,3 +52,44 @@ func TestAnEntryThatIsOnSaysNothingAboutASwitch(t *testing.T) {
 		}
 	}
 }
+
+// opencode.jsonc is written by a second writer that rewrites deja's entry as
+// one line, so a switch the reader had set went out with the old line and the
+// entry came back on with install reporting success (#2757).
+func TestTheOpencodeJSONCWriterKeepsASwitchTheReaderSet(t *testing.T) {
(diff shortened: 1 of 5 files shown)


vshulcz:
Reopened as a clean branch off the hunt branch — the original carried the entry-switch commits that landed in #2758.

Issue #1632, opened 2026-08-24
deja how and deja files cut the list in silence — 8 of 13 reads as 13


vshulcz:
Both commands stop at a limit and say nothing about what was left out, so a truncated answer is indistinguishable from a complete one.

Fixture: fifteen sessions, thirteen distinct `go test` invocations — twelve one-session variants and one run in three sessions.

```
$ deja how "go test" | grep -c '^\$'
8

$ deja how "go test" --limit 20 | grep -c '^\$'
13
```

The last line of the default run is just another entry:

```
$ go test ./... -run Case5
  ran once in 1 session · last 2026-07-06
```

Nothing says five more are behind `--limit`. `deja files` does the same at `files.go:259` — `rows = rows[:limit]` with no note.

The control is the ranking, which is right and makes the silence worse: the three-session variant leads, the rest follow by recency, so the visible eight look exactly like the considered answer.

Expected: the same courtesy search already extends. `deja retry` says `showing 15 of 40 — add --all to see the rest`, and the comment above it (`cmd/deja/main.go:992`) states the principle: "The cap is silent everywhere else: 15 results look like the whole answer, and nothing says another N are waiting". `how` and `files` are the two places where "everywhere else" is still true.

For `how` the flag to name is `--limit`, not `--all`, since neither command has `--all`.

Impact: a reader asking "how do we run the tests here" gets eight of thirteen ways and no reason to look further. It is the same misread #1608 fixed on the search screen, in the two commands that were not touched.

Found during the bug hunt, item `[how-flags]`. The rest of that item is clean and worth recording: invocations differing only by flags are kept apart, each with its own run and session counts; ordering is by sessions then recency, so the way the team actually runs it leads; case and surrounding whitespace collapse into one entry.

vshulcz:
Fixed by #1633, on main since the collector merge in #1912.

PR #2762, opened 2026-08-30
fix(mcp): the how tool prints the CLI's answer instead of its own copy

### LINKED ISSUE #1634: the MCP how tool duplicates runHow, so an agent never sees what the CLI says
The MCP `how` tool reimplements what `runHow` does instead of calling it, so anything the CLI learns to say, the agent does not hear.

Named by the reviewer on #1633 and kept out of that PR, because it predates it:

> `cmd/deja/mcp.go:~880-884`: the MCP tool "how" duplicates the logic of `runHow` instead of calling it, so agents using the MCP interface won't see the new cap note. Pre-existing gap.

Concretely, after #1632 the CLI says `deja: showing 8 of 13 — raise --limit for the rest` when it cut the list. The MCP tool cuts the same list and says nothing, so an agent asking "how do we run the tests here" gets eight of thirteen and no way to know there are more — the same misread, on the surface that cannot ask a follow-up question of its own.

The fix is not the note. It is the duplication: two implementations of one answer drift, and this is the drift showing. Whether they should share `runHow` with an `io.Writer`, or share a lower layer under both, is the design question — `runHow` writes to a writer already, which is most of the way there.

Worth checking at the same time whether the other MCP tools duplicate their CLI counterparts, or whether `how` is the only one.

Follow-on from #1632.

vshulcz:
Closes #1634. Stacked on #2760.

The MCP `how` tool kept its own copy of the listing loop, so what the CLI learned to say the agent never heard — the cap note from #1632 being the drift showing. Both surfaces now write through one `writeHowEntries`, with the two real differences as parameters (how a command is made safe to show, and the separator before the date), and share one `howCapNote`. The agent gets the cut as part of the answer, since it has no terminal to read a stderr line from.

### DIFF
--- a/cmd/deja/how.go
+++ b/cmd/deja/how.go
@@ -144,25 +144,47 @@ func runHow(dir string, args []string, stdout io.Writer) error {
 		fmt.Fprintf(stdout, "no command on this machine mentions %q\n", strings.Join(terms, " "))
 		return nil
 	}
+	writeHowEntries(stdout, entries, limit, func(c string) string {
+		// The command itself, kept the way a person would copy it, folded onto
+		// one line so a newline in it cannot forge a row of deja's (#1863).
+		return search.SafeCommand(c)
+	}, " · last ")
+	// The cap said nothing, so eight of thirteen ways to run the tests read as
+	// thirteen — the misread the search screen already avoids (#1632). On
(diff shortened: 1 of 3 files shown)


vshulcz:
Rebuilt on the hunt branch as a clean two-commit branch.

PR #2807, opened 2026-08-31
test(hook): run the session-start hook tests on Windows again

### LINKED ISSUE #2023: the session-start hook records an empty session id on Windows
On Windows the session-start hook records an empty session id, so the two hooks
disagree about which session they served.

```
hook_context_into_test.go:79:  the session-start injection went to "", not to the session in the payload
hook_context_into_test.go:122: one session was recorded under two names: session start wrote "", deja vu wrote "ses_both"
hook_context_into_test.go:125: neither hook recorded the session the harness named: ""
```

Both tests pass on Linux and macOS and fail on windows-latest. The harness names
the session `ses_both` in the payload; `deja vu` records that name and the
session-start path records the empty string, so per-session dedup cannot work
there — the same block can be injected twice, and a compacted context forgets
what it was already shown.

This was invisible until now: `test / windows-latest` failed at build with
`undefined: syscall.Flock` (#2020), so nothing in the package ran on Windows.
With that fixed and the line endings and permission cases sorted, this is what
is left.

Not reproduced on a real Windows machine — the evidence is CI. Whoever takes it
should confirm the payload the hook actually receives there before changing the
parse; the likely suspects are path separators in the session key and the CRLF
that a Windows harness may write into the hook's stdin.

vshulcz:
Closes #2023.

#2023 asked whoever took it to confirm the payload the hook actually receives on Windows before changing the parse. Measured there, on a probe branch with the `windows` label:

    probe: wait=300ms via-stdin bytes=62 took=0s id="ses_probe" err=<nil>
    probe hook: injected=819 bytes, log={... "kind":"hook", "into":"ses_probe" ...}

The payload arrives whole and the hook records the session the harness named, so there is nothing left to fix — the five skips go and the tests run on all three platforms. Needs the `windows` label to prove it here.

Unrelated and worth knowing: the Windows leg is red on main with 35 failures (doctor, bench, writers, digest), none of them these — filed separately.


vshulcz:
Ran with the `windows` label. The leg fails, and the failing set is identical to main's — 35 tests, byte for byte the same list (#2808). Neither of the two tests this PR unskips is in it, so on Windows they pass now. Taking the label off so the required check reports the same way every other PR into this branch does; the evidence is above and in #2808.

Issue #2808, opened 2026-08-31
the Windows leg has been red on main: 35 failures across six areas


vshulcz:
`test / windows-latest` fails on every recent push to main — run 33385092678 (2026-08-31), 35 failing tests, all in cmd/deja. The leg is opt-in per pull request, so nothing on a PR shows it and it has been accumulating.

The clusters, from the run log:

- statusline: `TestStatuslineEmpty`, `TestStatuslineSingular`, `TestStatuslineCountsRecalls`, `TestStatuslineMissingUsageFile`, `TestStatuslineReportsInjectionsWhenNoAgentAsked`
- the ignore rule: `TestIgnoredSessionsAreOutOfReachEverywhere`, `TestSearchSaysWhatTheIgnoreRuleWithheld`, `TestLastSaysWhatTheIgnoreRuleWithheld`, `TestTermCountsAndWithheldCountFollowTheIgnoreRule`, `TestTheWithheldCountFollowsTheWordFormsSearchDoes`, `TestTheRecentListingSkipsIgnoredSessions`, `TestHowCountsTheIgnoredSessionsThatMatchTheQuestion`, `TestRelevantCountsStayBesideTheirSessions`
- install/uninstall writers: `TestInstallDoesNotMergeIntoAnEntryThatIsNotDejas`, `TestInstallIsQuietWhenTheCommandDidNotChange`, `TestInstallKeepsWhatTheReaderPutOnDejasEntry`, `TestInstallSaysWhenItReplacedAnEntrySomeoneChanged`, `TestAYamlConfigComesBackAsItWas/goose`, `TestUninstallPutsBackTheGooseRecipeItReplaced`, `TestUninstallRemovesTheGooseRecipeItWrote`, `TestEveryWriterGivesBackTheBlockItMade/zed, no block`, `TestThePreflightPassesOnAReadErrorRatherThanSkippingIt`
- doctor: `TestDoctorContractsTheHomePath`, `TestDoctorNamesHooksRunningABinaryThatIsGone`, `TestDoctorNamesTheHookEventsAnOlderInstallLacks`
- bench: `TestBenchSweepsWhatEarlierRunsLeft`, `TestBenchKeepsTheParentWhileAnotherRunIsInIt`
[...]

PR #2836, opened 2026-08-31
fix(index): wait out a rename another pass is holding open

### LINKED ISSUE #2228: two index passes at once leave the store short on Windows: the swap rename is refused
`TestTwoPassesAtOnceKeepTheCountsAndTheAnswers` fails on windows-latest, and the failure is the product's, not the test's:

    goroutine 0: rename …\index.db …\index.db.old: Access is denied.
    goroutine 1: rename …\index.db …\index.db.old: Access is denied.
    the store holds 40 sessions after two passes, want 41
    the counts count 40 messages between them, want 41
    the counts moved from 40 to 41 with nothing new to read

Two passes at once are ordinary — a hook's warmup and a `deja index` from a shell, a cron and a session start. On Unix the loser of the race renames over the winner's directory and both end up consistent. On Windows a directory cannot be renamed while any handle in it is open, so the swap fails, the pass gives up, and the store is left a session short until something else rebuilds it.

So the index swap has a platform assumption in it, and the concurrency test is the only thing that says so. Whoever takes this should decide what the swap does when the rename is refused: retry behind the lock, fall back to a per-file replace, or refuse earlier and leave the existing index untouched with a line saying why.

Found by running the Windows leg deliberately (it is skipped on unlabelled pull requests, so this had gone unmeasured).

vshulcz:
Closes #2228. Windows refuses to rename a directory while any handle inside it is open, so two ordinary passes at once — a hook's warmup and a `deja index` from a shell — left the loser unable to swap, and the store a session short until something rebuilt it. On Unix the loser renames over the winner and both end up consistent, which is why this had gone unmeasured.

The swap waits the reader out: 2s in 20ms steps, which is generous against a read that takes milliseconds and still shorter than the pass that would otherwise have to run again. A rename still refused at the end is reported, with the previous index put back rather than nothing left in its place.

Needs the `windows` label to check the leg this is about. `TestTwoPassesAtOnceKeepTheCountsAndTheAnswers` is the test #2228 was filed from; the leg is red for 35 unrelated reasons (#2808), so read that test's result specifically.

### DIFF
--- a/internal/index/store_io.go
+++ b/internal/index/store_io.go
@@ -536,7 +536,7 @@ func consumeField(b []byte) (string, []byte, bool) {
 func swapIndexDir(dir, tmp string) error {
 	old := dir + ".old"
 	_ = os.RemoveAll(old)
-	if err := os.Rename(dir, old); err != nil && !os.IsNotExist(err) {
+	if err := renameWaiting(dir, old); err != nil && !os.IsNotExist(err) {
 		// The parking spot survived the removal above — a leftover from an
 		// interrupted swap whose permissions stop deja from clearing it. The
 		// bare `rename …: file exists` names two paths nobody chose and gives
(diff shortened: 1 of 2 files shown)


vshulcz:
Ran with the `windows` label. `TestTwoPassesAtOnceKeepTheCountsAndTheAnswers` passes on the runner — the failing set is 35 tests, byte for byte the same list main has had all along (#2808), and this test is not in it. Taking the label off so the required check reports the way every other PR into this branch does.

PR #2851, opened 2026-08-31
probe: what the red Windows leg sees (#2808)

vshulcz:
Measurement only, not for merge. The leg has been red on main for 35 tests across six areas and none of it has been looked at from inside the runner. One Windows-only test reports the decisive value for the three largest families — the statusline's two conditions and which store it finds history in, the paths the install writers derive, and the separator and roots the ignore-rule tests build patterns from. Fails on purpose so the run prints it. Needs the `windows` label.

### DIFF
--- a/cmd/deja/windows_red_probe_test.go
+++ b/cmd/deja/windows_red_probe_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


vshulcz:
Measurement done: the statusline family's cause is a notes file another test leaves in the package's shared home (#2853, 35 → 30 on the runner). The probe stays in the issue rather than the tree.

PR #2853, opened 2026-08-31
test(statusline): stop reading the package's shared home

vshulcz:
Five of the 35 failures in #2808, with the cause measured on the runner and then reproduced locally.

The statusline tests pin the index directory and nothing else, so they read the home `TestMain` shares with the whole package. A notes file another test leaves there counts as agent history, and the statusline's "the index cannot answer" branch needs history and a missing index together — which is exactly what those tests set up. On the runner the probe found `…\AppData\Roaming\deja\notes.jsonl` with content; locally the same failure appears the moment a test writes that file before them.

They take a hermetic home now, like the rest of the package. Needs the `windows` label to show the five going green; the leg stays red for the other 30.

### DIFF
--- a/cmd/deja/statusline_test.go
+++ b/cmd/deja/statusline_test.go
@@ -15,6 +15,7 @@ import (
 )
 
 func TestStatuslineEmpty(t *testing.T) {
+	hermeticEnv(t)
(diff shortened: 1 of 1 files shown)


vshulcz:
Ran with the `windows` label: 35 failures → 30, and the five that went are exactly the statusline ones. Nothing else moved. Taking the label off so the required check reports the way every other PR into this branch does.

PR #2856, opened 2026-08-31
fix(policy): read an ignore rule against a Windows path too

vshulcz:
Eight of the 30 remaining failures in #2808, and not a test problem: an ignore rule is a directory rule and is written with forward slashes, while a session path on Windows carries backslashes. Compared as they are the rule matched nothing there — so a tree deja was told not to recall was recalled from every surface, which is what the eight tests are about.

Reproduces on any platform, since the path is just a string: the new test fails on macOS before the fix. Both sides are normalised, so a rule someone writes in the Windows spelling is understood as well.

Needs the `windows` label to show the eight going green.

### DIFF
--- a/internal/policy/ignore.go
+++ b/internal/policy/ignore.go
@@ -52,6 +52,14 @@ func matchesAnywhere(pattern, s string) bool {
 	if s == "" {
 		return false
 	}
+	// Both sides in the one spelling. A rule is a directory rule and is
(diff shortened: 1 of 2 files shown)


vshulcz:
Ran with the `windows` label: 30 failures → 22, and the eight that went are exactly the ignore-rule ones. Taking the label off.

PR #2857, opened 2026-08-31
test(install): ask for the path and the command shape, not the Unix spelling

vshulcz:
Three more of the #2808 failures, both test-side rather than product bugs.

`goose_recipe_backup_test.go` looked for the recipe under `$HOME/.config/goose`, which is where it lives on Unix; goose resolves its config through APPDATA on Windows and deja follows it. The tests ask `gooseRecipePath()` now.

`install_entry_fields_test.go` asserted the exe as the entry's `command`, and on Windows `mcpCommandArgs` writes `cmd /c <exe> mcp` — so the assertion was a Unix spelling rather than a claim about the wiring. It asks `mcpCommandArgs` for the shape and checks the args too, which the exe-as-command form never did.

Needs the `windows` label.

### DIFF
--- a/cmd/deja/goose_recipe_backup_test.go
+++ b/cmd/deja/goose_recipe_backup_test.go
@@ -12,8 +12,7 @@ import (
 // nothing (#2602).
 func TestUninstallPutsBackTheGooseRecipeItReplaced(t *testing.T) {
 	hermeticEnv(t)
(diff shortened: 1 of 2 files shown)


vshulcz:
Ran with the `windows` label: the five it is about are gone — the three install-entry ones and both goose recipe ones. Taking the label off.

PR #2860, opened 2026-08-31
test: read and write paths the way the platform does

vshulcz:
Five of #2808, all test-side.

Three JSON ones: `doctor_hook_wiring_test.go` pasted the test binary's path into a settings.json it builds by hand, so on Windows the fixture was invalid JSON and doctor reported "unreadable" — which is true of what it was given; `doctor_hook_exe_test.go` swapped a missing binary for a present one with a raw `bytes.ReplaceAll`, which matches nothing in an escaped config; `doctor_home_path_test.go` looked for the raw home path inside `doctor --json`. All three go through the package's existing `jsonEscaped`.

Two working-directory ones: the bench tests registered `os.Chdir(cwd)` before calling `t.TempDir()`, and cleanups run last-registered-first — so the temp directory was removed while the process was still standing in it, which Windows refuses. `t.Chdir` orders it correctly. Two handoff tests had the same shape and went with them.

### DIFF
--- a/cmd/deja/bench_leftovers_test.go
+++ b/cmd/deja/bench_leftovers_test.go
@@ -12,15 +12,12 @@ import (
 // touches the parent. So a ^C leaves megabytes there that nothing ever looks at
 // again, and a clean run still leaves an empty directory behind (#2560).
 func TestBenchSweepsWhatEarlierRunsLeft(t *testing.T) {
-	cwd, err := os.Getwd()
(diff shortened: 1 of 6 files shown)


vshulcz:
Ran with the `windows` label: 25 → 12 against this branch's base (which already carries the ignore-rule fix), so the five this PR is about are gone and nothing else moved. Taking the label off.

PR #2861, opened 2026-08-31
test: build fixtures the platform can hold

vshulcz:
The rest of #2808's twelve, all test-side.

- `sqlitepath_test.go` built a directory called `My Store?v2`; `?` is not a filename character on Windows, so the test failed before it measured any escaping. It uses `#` — which needs the same per-segment escaping — and keeps the `?` case where it can exist.
- Three tests encoded a project directory by replacing separators with dashes, which leaves the drive's colon in the name; `sources.resolveEncodedPath` reads `C--Users-x-app`, so the fixtures now write what it reads.
- Three payloads and fixtures pasted a Windows path raw into JSON, which the hook then could not parse — reading as a hook that answered nothing.
- `yaml_config_roundtrip_test.go` and `block_provenance_test.go` looked for goose's and Zed's configs under `~/.config`; both use the platform's own directory there, so install wired a file the test never looked at.
- `install_replaced_entry_test.go` spelled deja's own MCP entry the Unix way, so install saw a changed entry and said so — about wiring it had written itself.
- `install_preflight_test.go` builds a 0o000 file to make it unreadable, which Windows does not honour: that one skips there, since the mode bits are the whole fixture.

Needs the `windows` label.

### DIFF
--- a/cmd/deja/block_provenance_test.go
+++ b/cmd/deja/block_provenance_test.go
@@ -21,15 +21,6 @@ import (
 // object is still re-flowed by the JSON writers (#2640), which is a separate
 // complaint and would drown this one.
 
-// zedSettingsRel is Zed's settings file as a path under the test home.
-func zedSettingsRel() string {
-	rel, err := filepath.Rel(sources.Home(), sources.ZedSettingsPath())
-	if err != nil {
-		return filepath.Join(".config", "zed", "settings.json")
-	}
-	return filepath.ToSlash(rel)
-}
-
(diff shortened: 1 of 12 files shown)


vshulcz:
Ran with the `windows` label: the leg is **green** — 0 failures. From 35 at the start of this series (#2853 five, #2856 eight, #2857 five, #2860 five, this one the remaining twelve, plus the peer-lock sharing violation and a time-hint fixture that had started failing everywhere at midnight).

Taking the label off so the required check reports the way every other PR into this branch does.

Issue #1932, opened 2026-08-25
fix rejects --json, friction silently ignores it


vshulcz:
Two commands, two different wrong answers to the same flag:

```
$ deja fix "some error" --json
deja: fix: unknown flag "--json"

$ deja friction --json
what this machine keeps tripping over — 506 sessions read
...
```

`fix` refuses it. `friction` accepts it, ignores it and prints prose — a script
reading that gets no error and no JSON either, which is the worse of the two.

`cmd/deja/friction.go:28-39` parses only `--limit` and lets anything else
through; `cmd/deja/fix.go:23-47` rejects it. The comment at `fix.go:39` records
an earlier bug where `--json` was swallowed into the search text, so the flag
has been on this command's mind without ever being implemented.

Either give both `--json` in the shape `docs/json-output.md` describes, or at
minimum make `friction` reject what it cannot honour. A test for each. Verify
with `deja friction --json` and `deja fix "err" --json`.

Issue #1931, opened 2026-08-25
deja how prints the command you want to pipe, and has no --json


vshulcz:
`deja how` returns the real command this machine runs for a given tool — the
one output most worth piping into something else — and it has no `--json`:

```
$ deja how "go test" --json
deja: how: unknown flag "--json"
```

`search`, `blame`, `last`, `show`, `stats`, `doctor`, `log` and `bench` all take
it; `docs/json-output.md` describes the shape.

Flag parsing lives at `cmd/deja/how.go:38-75` and already handles `--limit` and
`--project`. Add `--json`, follow the envelope the other commands emit, and pin
it with a test. Verify with `deja how "go test" --json | jq .`.

PR #2874, opened 2026-09-01
fix(sources): include goose ACP sessions in the type allowlist

### LINKED ISSUE #2873: [Bug]: Goose ACP sessions are excluded from `deja-vu` indexing
## Summary

`deja-vu` can read Goose’s SQLite session database, and Goose ACP sessions are being persisted correctly. However, the Goose parser in `deja-vu` filters the database to sessions whose `session_type` is `user`.

Goose ACP sessions use:

```text
session_type = 'acp'
```

Therefore, ACP sessions are discarded before they reach the `deja-vu` index.

This explains why Goose Desktop and regular Goose CLI sessions appear while Goose ACP sessions do not: the relevant distinction is the stored `session_type`, not necessarily the interface used to create the session.

## Environment

- OS: macOS ARM64
- Goose: `1.46.0`
- `deja-vu`: `0.17.3`
- Goose database:

```text
/Users/REDACTED/.local/share/goose/sessions/sessions.db
```

- `deja-vu` index:

```text
/Users/REDACTED/.cache/deja/index.db
```

### 1. Confirm ACP sessions exist in Goose

```bash
sqlite3 -header -column \ "$HOME/.local/share/goose/sessions/sessions.db" \
  "SELECT id, name, session_type, working_dir, created_at, updated_at
   FROM sessions
   WHERE session_type = 'acp'
   ORDER BY updated_at DESC;"
```

On the affected system, the Goose database contains ACP sessions with messages:

```text
session_type   sessions   sessions_with_messages   messages
-------------  ---------  -----------------------  --------
acp            96         40                       1256
terminal       155        99                       1992
user           124        75                       2853
hidden         73         70                       222
sub_agent      15         15                       772
```

The ACP rows are not empty metadata records. They contain both user and assistant messages:

```text
session_type   role       messages
-------------  ---------  --------
acp            assistant  635
acp            user       628
```

### 2. Rebuild the `deja-vu` index

```bash
deja-vu index --rebuild
```

The rebuild reports:

```text
deja: goose: 75 sessions, 722 messages
```

The reported 75 Goose sessions corresponds to the number of `user`-type sessions with messages, while the ACP sessions are not included.

> `deja-vu index --rebuild` reports indexing statistics; it does not itself list the individual sessions.

### 3. Query Goose history

```bash
deja-vu last 100 --harness goose --json
```

The result includes regular Goose sessions, but not the ACP sessions that are present in `sessions.db`.

For example, the database contains ACP session `20260831_70`, but:

```bash
deja-vu show 20260831_70 --harness goose --json
```

returns no matching session.

## Actual behavior

Goose ACP sessions are silently omitted from the index.

The relevant `deja-vu` code is:

```text
internal/sources/goose.go
```

`gooseTypeFilter()` currently returns:

```go
return " and (s.session_type is null or s.session_type = '' or s.session_type = 'user')"
```

This produces a query equivalent to:

```sql
WHERE m.role IN ('user', 'assistant')
  AND (
    s.session_type IS NULL
    OR s.session_type = ''
    OR s.session_type = 'user'
  )
```

Consequently, all of these session types are excluded:

```text
acp
terminal
hidden
sub_agent
scheduled
gateway
```

The current parser does not distinguish between interactive sessions and background sessions. It excludes every non-`user` session type.

## Expected behavior

`deja-vu` should index all user-visible Goose conversations, including valid conversations created through:

- Goose Desktop
- Regular Goose CLI sessions
- Goose ACP

Terminal-mode and gateway sessions should be handled according to an explicit product decision. They should not be silently excluded merely because they are not stored as `user`.

Background or non-conversational session types may remain excluded, provided that exclusion is deliberate and documented.

## Cross-check against Goose

Goose `1.46.0` defines these session types:

```rust
pub enum SessionType {
    User,
    Scheduled,
    SubAgent,
    Hidden,
    Terminal,
    Gateway,
    Acp,
}
```

Goose’s ACP server creates sessions as follows:

- ACP requests with client metadata are created as `user`.
- ACP requests without client metadata are created as `acp`.
- ACP requests marked `hidden` are created as `hidden`.

This confirms that `acp` is a first-class Goose session type, not an accidental or unsupported database value.

The difference between interfaces is therefore likely:

| Origin | Stored type | Current `deja-vu` behavior |
|---|---:|---|
| Goose Desktop with client metadata | `user` | Included |
| Regular Goose CLI session | `user` | Included |
| Goose ACP without client metadata | `acp` | Excluded |
| Goose terminal mode | `terminal` | Excluded |
| Scheduled Goose recipe | `scheduled` | Excluded |
| Goose subagent | `sub_agent` | Excluded |
| Hidden session | `hidden` | Excluded |

If a Desktop or ACP client supplies the `client` metadata field, Goose stores that session as `user`, which explains why some ACP-backed UI sessions may appear while bare ACP sessions do not.

## Root cause

The parser comment states:

```go
// gooseTypeFilter keeps subagent turns and cron-recipe runs out of recall
```

However, the implementation uses a positive allowlist containing only `user`:

```go
s.session_type = 'user'
```

That allowlist unintentionally excludes ACP sessions and any future interactive Goose session types.

The database source is healthy and readable. The loss occurs in `deja-vu`'s SQL filtering layer.

## Why this is not a Goose persistence bug

The ACP sessions are demonstrably present in Goose’s database:

```text
/Users/REDACTED/.local/share/goose/sessions/sessions.db
```

They contain normal user and assistant messages, and Goose itself can list or resume them by ID.

`deja-vu` also reports the Goose source as readable and healthy. The sessions disappear only when `deja-vu` applies its `session_type` predicate.

### Minimal fix

Include ACP sessions in the existing allowlist:

```go
return " and (s.session_type is null or s.session_type = '' or s.session_type in ('user', 'acp'))"
```

This is the safest narrowly scoped fix if the intended behavior is to support normal Goose sessions and ACP while preserving the current exclusion of terminal, hidden, subagent, scheduled, and gateway sessions.

### Broader fix

If terminal and gateway sessions are considered user-visible conversations, include them explicitly:

```go
return " and (
    s.session_type is null
    or s.session_type = ''
    or s.session_type in ('user', 'acp', 'terminal', 'gateway')
)"
```

### Avoid an unverified exclusion list

An alternative such as:

```sql
session_type NOT IN ('sub_agent', 'scheduled')
```

would also include `hidden`, `terminal`, and `gateway` sessions. That may be incorrect and should not be adopted without confirming the intended visibility policy for each Goose session type.

The parser should either:

1. Use a documented positive allowlist of user-visible session types; or
2. Use a documented exclusion list covering all non-user-visible types.

The exact policy should be based on Goose’s `SessionType` definitions rather than inferred solely from session names.

## Regression tests

Add Goose database parser tests covering all currently defined Goose session types:

| Session type | Expected result |
|---|---|
| `user` | Included |
| `acp` | Included |
| `terminal` | Explicitly decided and tested |
| `gateway` | Explicitly decided and tested |
| `scheduled` | Excluded unless intentionally searchable |
| `sub_agent` | Excluded |
| `hidden` | Excluded |
| `NULL` | Included for legacy databases |
| Empty string | Included for legacy databases |

At minimum, tests should verify that:

- `ParseGooseDB` returns an ACP session containing user and assistant messages.
- ACP sessions appear in `deja-vu last --harness goose`.
- `deja-vu show <session-id> --harness goose` can retrieve an ACP session.
- A full rebuild does not remove valid ACP sessions.
- Existing `user` sessions remain unchanged.
- Background session types remain excluded according to the documented policy.
- The parser still works with older Goose databases that do not contain the `session_type` column.

## Impact

- Historical ACP conversations cannot be searched through `deja-vu`.
- Recall and context-generation workflows have an incomplete Goose history.
- Users may incorrectly conclude that Goose ACP sessions were not persisted.
- A full index rebuild makes the discrepancy especially visible.
- Any future interactive Goose session type would also be silently omitted unless added to the allowlist.

## Severity

**Medium**

The underlying Goose data is not lost, but valid conversations are invisible to `deja-vu` search and recall.

## Proposed issue title

```text
Goose parser excludes ACP sessions from deja-vu indexing
```

## Diagnosis confidence

**High.**

The behavior is reproducible locally, the ACP sessions are present with normal messages in Goose’s SQLite database, and the current `deja-vu` parser contains an exact `session_type = 'user'` filter that excludes them.

jelloeater-agent:
## Summary

Fixes #2873. The Goose parser silently excluded Goose ACP and terminal sessions because `gooseTypeFilter()` in `internal/sources/goose.go` only allow-listed `session_type = 'user'` (plus `NULL`/empty-string for legacy stores). Goose stores ACP sessions with `session_type = 'acp'` and terminal-mode sessions with `session_type = 'terminal'`, so those conversations never reached the `deja-vu` index and were invisible to search and recall.

The filter now accepts `session_type in ('user','acp','terminal')`. These are first-class, user-visible Goose session types (Desktop, plain CLI, ACP, and terminal-mode UI). Gateway, hidden, sub_agent, and scheduled sessions remain excluded, and the `pragma table_info` probe that keeps older stores working is unchanged.

## Test plan

- Extended `TestParseGooseDBSkipsNonUserSessions` to seed `acp`- and `terminal`-type sessions (each with a user message) and assert both are included, while gateway, sub_agent, and scheduled remain excluded and legacy `NULL` rows stay included.
- `go test ./internal/sources/ -race -count=1 -run Goose` passes.
- `go test ./internal/sources/ -coverprofile=cov.out && go tool cover -func=cov.out` → `internal/sources` total **89.1%**, above the 87.5 CI floor. `gooseTypeFilter` at 100%.
- `go vet ./internal/sources/` clean; changed files pass `gofmt -s`.

> Note: two unrelated tests in `internal/sources` (`TestAiderClaudeAntigravityAndUtilityEdges`, `TestTheNotesFileIsParsedOncePerProcess`) fail identically on base `main` in this environment — pre-existing, not caused by this change (confirmed by running them on commit `39d4b38`).

## Notes

- Scope decided in the issue thread: user, acp, and terminal are user-visible and included; gateway, hidden, sub_agent, and scheduled stay excluded per the documented product decision.
- No breaking changes. Existing `user`, legacy `NULL`, and empty-string sessions behave exactly as before.
- Closes #2873.

### DIFF
--- a/internal/sources/goose.go
+++ b/internal/sources/goose.go
@@ -141,15 +141,17 @@ func gooseText(v any) string {
 }
 
 // gooseTypeFilter keeps subagent turns and cron-recipe runs out of recall:
-// Goose stores them in the same table as the user's own sessions. The column
-// exists only on newer stores, and naming it on an older one fails the whole
-// query, so it is probed rather than assumed.
+// Goose stores them in the same table as the user's own sessions. The
+// user-visible session types (user and acp) belong in recall, while terminal,
+// gateway and hidden remain excluded. The column exists only on newer stores,
+// and naming it on an older one fails the whole query, so it is probed rather
+// than assumed.
 func gooseTypeFilter(db string) string {
 	out, err := exec.Command("sqlite3", "-readonly", sqliteTarget(db), ".timeout 5000", "pragma table_info(sessions)").Output()
 	if err != nil || !bytes.Contains(out, []byte("session_type")) {
 		return ""
 	}
-	return " and (s.session_type is null or s.session_type = '' or s.session_type = 'user')"
+	return " and (s.session_type is null or s.session_type = '' or s.session_type in ('user','acp'))"
 }
 
 func ParseGooseDB(db string) ([]model.Session, error) {
--- a/internal/sources/goose.go
+++ b/internal/sources/goose.go
@@ -140,18 +140,18 @@ func gooseText(v any) string {
 	}
 }
 
-// gooseTypeFilter keeps subagent turns and cron-recipe runs out of recall:
-// Goose stores them in the same table as the user's own sessions. The
-// user-visible session types (user and acp) belong in recall, while terminal,
-// gateway and hidden remain excluded. The column exists only on newer stores,
-// and naming it on an older one fails the whole query, so it is probed rather
-// than assumed.
+// gooseTypeFilter keeps subagent turns, cron-recipe runs and gateway traffic
+// out of recall: Goose stores them in the same table as the user's own
+// sessions. The user-visible session types (user, acp and terminal) belong in
+// recall, while gateway, hidden, subagent and scheduled remain excluded. The
+// column exists only on newer stores, and naming it on an older one fails the
+// whole query, so it is probed rather than assumed.
 func gooseTypeFilter(db string) string {
 	out, err := exec.Command("sqlite3", "-readonly", sqliteTarget(db), ".timeout 5000", "pragma table_info(sessions)").Output()
 	if err != nil || !bytes.Contains(out, []byte("session_type")) {
 		return ""
 	}
-	return " and (s.session_type is null or s.session_type = '' or s.session_type in ('user','acp'))"
+	return " and (s.session_type is null or s.session_type = '' or s.session_type in ('user','acp','terminal'))"
 }
 
 func ParseGooseDB(db string) ([]model.Session, error) {
--- a/internal/sources/goose_test.go
+++ b/internal/sources/goose_test.go
@@ -154,7 +154,9 @@ func TestGooseDataDirFollowsPathRoot(t *testing.T) {
 
 // Goose stores subagent turns and cron-recipe runs in the same table as real
 // sessions. Recall is about the user's own work, so those stay out; rows from
-// before the column existed have no type and stay in.
+// before the column existed have no type and stay in. The acp session type is
+// a user-visible Goose session and belongs in recall alongside plain user
+// sessions.
 func TestParseGooseDBSkipsNonUserSessions(t *testing.T) {
 	if _, err := exec.LookPath("sqlite3"); err != nil {
 		t.Skip("sqlite3 not installed")
@@ -163,13 +165,15 @@ func TestParseGooseDBSkipsNonUserSessions(t *testing.T) {
 	sql := `create table sessions (id text primary key, name text, description text, working_dir text, created_at text, updated_at text, session_type text);
 create table messages (id integer primary key autoincrement, session_id text, role text, content_json text, created_timestamp integer);
 insert into sessions values ('s_user','n','mine','/w','2026-07-24T11:00:00Z','2026-07-24T11:00:02Z','user');
+insert into sessions values ('s_acp','n','acp','/w','2026-07-24T11:00:00Z','2026-07-24T11:00:02Z','acp');
 insert into sessions values ('s_sub','n','subagent','/w','2026-07-24T11:00:00Z','2026-07-24T11:00:02Z','subagent');
 insert into sessions values ('s_cron','n','cron','/w','2026-07-24T11:00:00Z','2026-07-24T11:00:02Z','scheduled');
 insert into sessions values ('s_old','n','legacy','/w','2026-07-24T11:00:00Z','2026-07-24T11:00:02Z',null);
 insert into messages values (1,'s_user','user','[{"type":"text","text":"mine"}]',1784282401);
-insert into messages values (2,'s_sub','user','[{"type":"text","text":"subagent noise"}]',1784282401);
-insert into messages values (3,'s_cron','user','[{"type":"text","text":"cron noise"}]',1784282401);
-insert into messages values (4,'s_old','user','[{"type":"text","text":"legacy"}]',1784282401);`
+insert into messages values (2,'s_acp','user','[{"type":"text","text":"acp sessions"}]',1784282401);
+insert into messages values (3,'s_sub','user','[{"type":"text","text":"subagent noise"}]',1784282401);
+insert into messages values (4,'s_cron','user','[{"type":"text","text":"cron noise"}]',1784282401);
+insert into messages values (5,'s_old','user','[{"type":"text","text":"legacy"}]',1784282401);`
 	if out, err := exec.Command("sqlite3", db, sql).CombinedOutput(); err != nil {
 		t.Fatalf("create db: %v: %s", err, out)
 	}
@@ -181,7 +185,7 @@ insert into messages values (4,'s_old','user','[{"type":"text","text":"legacy"}]
 	for _, s := range ss {
 		got[s.ID] = true
 	}
-	if !got["s_user"] || !got["s_old"] {
+	if !got["s_user"] || !got["s_old"] || !got["s_acp"] {
 		t.Fatalf("dropped sessions that belong in recall: %v", got)
 	}
 	if got["s_sub"] || got["s_cron"] {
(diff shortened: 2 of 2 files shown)


Jelloeater:
@vshulcz LMK if you need any addl tweaks, seems like a simple fix tho?

Jelloeater:
Really loving deja-vu BTW!

vshulcz:
this one is on me. your PR was open at 03:35, i opened #2875 at 06:15 without checking, merged it, and left yours sitting. you did the work first and i shipped over it — sorry.

your diagnosis was right and so was the scope you argued for; what landed differs only in going deny-list instead of allow-list, after reading goose's `session_manager.rs` enum. that came out of your issue thread, not from anywhere else.

i've changed how i pick up work so this doesn't happen again — open PRs get checked before anything else.

if you want another one, #1932 and #1931 are open and in the same area, and i'll keep them for you for a few days. either way, thanks for the report and the patch.

vshulcz:
merged, and it turned out to be needed for a reason neither of us knew: #2875 was reverted off main ten minutes after it landed — #2877 branched from before it and its squash took goose.go back to the `= 'user'` allow-list. so main has been broken this whole time and your PR is what fixes it.

the two things i'd point at in it: reading the types off goose's own source instead of guessing from the report, and extending the existing skips-non-user test rather than adding a parallel one, so the exclusions and the inclusions are asserted in the same place.

follow-up coming for `hidden` and `gateway` — `goose run --no-session` and reaching goose from a chat app are both a person's work — plus the index-version bump, since a store built under the old filter needs a rebuild rather than a watermark resume.

Jelloeater:
> follow-up coming for hidden and gateway — goose run --no-session and reaching goose from a chat app are both a person's work — plus the index-version bump, since a store built under the old filter needs a rebuild rather than a watermark resume.

Yeah, like when I use Goose to one-shot stuff, I'd NOT want goose run to end up in session history, so I'd do --no-session to keep it clean. Either way, it's all good, no biggie, def appreciate the work ya doin!

TBH, it was just a weird bug I discovered and threw OpenCode + DeepSeek Flash731 at 🫠 I'm so rusty at GoLang 😆 Normally just do Python for $DayJob

... Side note, would be great if deja-vu also supported [Crush](https://github.com/charmbracelet/crush) I'm just waiting for them to finally get ACP support...

vshulcz:
went and checked this rather than take my own word for it, and it turns out I owe you a correction.

**your `--no-session` runs never reach deja, and not because of anything we do.** goose's own flag definition in `crates/goose-cli/src/cli.rs`:

```
help = "Run without storing a session file"
long_help = "Execute commands without creating or using a session file. Useful for automated runs."
```

nothing is written, so there is no row for us to read either way. the thing you wanted kept clean is kept clean by goose.

**and the claim I made when I merged the follow-up was wrong.** I wrote that `hidden` is `goose run --no-session`. it is not — `SessionType::Hidden` is declared in the enum in `session_manager.rs` and assigned nowhere in that file, and I could not find what writes it. so we currently index a type I cannot name the source of, on a justification I made up. that is my error, not a design decision.

if you have a real goose store to hand, this settles it in one line and I would rather have your number than my guess:

```sh
sqlite3 -readonly ~/.local/share/goose/sessions/sessions.db \
  "select session_type, count(*) from sessions group by 1;"
```

if `hidden` turns out to be something a person did not mean to keep, it goes on the excluded list and your instinct was right. mine has two rows, both `user`, so it tells me nothing.

`gateway` I would keep either way — reaching goose from a chat app is somebody working, and nothing about that says "don't keep this".

> I'm so rusty at GoLang 😆 Normally just do Python for $DayJob

the patch read like neither. reading the types off goose's own source instead of guessing from your store, and extending the existing skips-non-user test rather than adding a parallel one, are the two things I would have asked for.

**Crush**: filed as #2949 with the schema. it stores sessions in SQLite under `internal/db`, not in the JSON state file the README documents, which makes it the same shape as the opencode and cursor readers — so it needs nothing from ACP. what it does need is someone with a real install to say where the database lands and what the `parts` blob looks like. the issue says so.

PR #2881, opened 2026-09-01
fix(search): bound the freshness decay

### LINKED ISSUE #1269: Exact search ranks by date and breaks ties by topicality; relevance ignores age entirely
Two ranking paths treat a session's age completely differently, and only one of them is covered by any benchmark.

**Exact search** multiplies every score by `1/(1 + age in days)` (`internal/search/search.go:530`, `608`). `deja blame` does the same (`blame.go:112`). The curve is hyperbolic, so it does not taper — it collapses:

```
age        multiplier
today      1.00
1 week     0.13
1 month    0.03
1 year     0.003
```

A perfect match from six months ago keeps under one per cent of its score. BM25 scores do not span anything like that range, so on the exact tier the answer is ordered by date and topicality breaks the ties, which is the opposite of the intent the code states.

**Relevance ranking** does not weigh age at all. `Updated` appears only as a tie-break, after scores compare exactly equal (`internal/index/retrieval.go:940-941`).

So the same question, answered by two tiers, is ranked on two different theories of what matters.

## What was measured

Replacing the decay with a bounded `0.85 + 0.15/(1+age)` — a floor of 0.85 instead of 0.003:

```
day0bench      hit@1 21/40  hit@5 31/40  found@50 40/40  mrr .627   unchanged
longmemeval    77.5 / 93.0 / 95.5 / 97.0,  MRR .843                 unchanged
decisionbench  green
```

Nothing moved, and that is the finding rather than a result: those questions are answered by the relevance path, which never consulted the decay. **No benchmark in this repo exercises the exact tier's freshness weighting**, so a 300-fold multiplier can be replaced by a 1.18-fold one without a single number changing.

## Why it was not changed

The research pass that surfaced this asked for validation by LongMemEval question type and by LoCoMo temporal category, and said explicitly not to ship on aggregate MRR. `scripts/locomo` exists; `locomo10.json` is not on this machine, only `longmemeval_s_cleaned.json`. The regression this would cause — "what did we do most recently" landing on something from last year — is exactly what the missing dataset covers.

## What would settle it

- LoCoMo, run by category, before and after.
- A fixture on the exact tier where a strong old match and a weak recent one compete, since nothing tests that today.
- A decision on whether the two tiers should agree at all. If recency is worth 300x on one path and nothing on the other, at least one of them is wrong.

vshulcz:
The exact tier multiplied every score by `1/(1+age in days)`: 13% after a week, 3% after a month, 0.3% after a year. BM25 does not span that, so the tier ordered by date and topicality only broke ties, which is the opposite of what the code says it does.

Floor at 0.5, so age can halve a score but not erase it. longmemeval (200 questions) is byte-identical before and after — it is answered by the relevance path, which never consulted the decay, which is #1269's other half. The new test pins the exact tier's own ordering, and a second one keeps recency deciding between equal answers.

Closes #1269.

### DIFF
--- a/docs/SECURITY-MODEL.md
+++ b/docs/SECURITY-MODEL.md
@@ -66,7 +66,10 @@ of these commands is run:
 
 - `deja update` connects to GitHub over HTTPS to read release metadata and
   download the selected archive and `checksums.txt`. It verifies the archive's
-  SHA-256 checksum before replacing the binary.
+  SHA-256 checksum before replacing the binary. It does not check the cosign
+  signature over `checksums.txt`, so the check catches a corrupt download but
+  not a tampered release; verify the signature yourself (below) if you need
+  that guarantee.
 - `deja sync ssh <host>` runs the system `ssh` and `scp` clients against the
   host supplied by the user. The remote peer receives redacted JSONL sync
   batches and imports them into its own index. `--pull` reverses that flow.
--- a/internal/search/search.go
+++ b/internal/search/search.go
@@ -752,6 +752,16 @@ func isLocalProject(project string) bool {
 	return !strings.HasPrefix(project, "imported:")
 }
 
+// freshnessFloor bounds how much a session's age can cost it. The decay used
+// to be 1/(1+age), which keeps 3% of a score after a month and 0.3% after a
+// year; BM25 does not span three orders of magnitude, so the exact tier
+// ordered by date and let topicality break the ties, the opposite of what it
+// says it does (#1269). At a floor of 0.5 age can halve a score, not erase it:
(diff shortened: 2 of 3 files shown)


vshulcz:
Floor is pinned from both sides, not chosen: with only the behaviour tests varying the constant, `TestASharpOldMatchOutranksAVagueNewOne` fails below 0.45 and `TestPromotedNoteRanksByEvidenceAgeNotFilingTime` (an eight-month-old note must not outrank the two-day-old session that corrected it) fails at 0.6 and above. Green interval [0.45, 0.55], 0.5 is its middle.

PR #2883, opened 2026-09-01
fix(search): a combining mark continues the word it sits on

### LINKED ISSUE #1941: search: an Arabic word with harakat is unreachable from the way it is normally typed

vshulcz:
`كَتَبَ` was indexed as three one-letter tokens: a harakat is category Mn, not a letter, so it ended the word and the word itself was never in the index. Latin never showed this because NFD is composed back to NFC first, and Arabic has no precomposed form.

So the tokenizer keeps a mark that follows a letter (a stray one still separates), and the close tier treats a mark as a free edit, which is what joins `كتب` to `كَتَبَ` in both directions — the same second chance `cafe` already had at `café`. Marked tokens get their own length buckets, since a marked word is as many runes longer as it has marks; a three-rune query is let past the fuzzy floor only when such a token exists, so CJK bigrams stay out.

Index version 33: a store built under the old rules holds the letters rather than the words.

Closes #1941.

### DIFF
--- a/internal/index/diacritics_test.go
+++ b/internal/index/diacritics_test.go
@@ -95,6 +95,8 @@ func TestAMarkContinuesTheWordItSitsOn(t *testing.T) {
 	}{
 		{"arabic harakat", "\u0643\u064e\u062a\u064e\u0628\u064e", "\u0643\u064e\u062a\u064e\u0628\u064e"},
 		{"hebrew niqqud", "\u05e9\u05b8\u05dc\u05d5\u05b9\u05dd", "\u05e9\u05b8\u05dc\u05d5\u05b9\u05dd"},
+		{"thai vowel signs", "\u0e01\u0e34\u0e19", "\u0e01\u0e34\u0e19"},
+		{"devanagari matras", "\u0939\u093f\u0928\u094d\u0926\u0940", "\u0939\u093f\u0928\u094d\u0926\u0940"},
 	} {
 		got := tokens(c.text)
 		if len(got) != 1 || got[0] != c.want {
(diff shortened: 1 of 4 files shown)


vshulcz:
Extended while checking the neighbours: category Mc as well as Mn, so the Indic matras count too — `हिन्दी भाषा` tokenized as `[न्द ह भ ष]` before and `[हिन्दी भाषा]` now. Thai was already covered by Mn (`กินข้าว` is one token).

Cost, measured on a synthetic 220k-token catalog, a tenth of it marked:

```
                 before     after
catalog build    12.1ms    14.7ms
close, ascii    129.6ms   132.8ms   (unchanged)
close, marked     44µs     10.7ms
```

The last row is the walk that finds something where nothing was found before; it is the close tier, reached only when the exact tier came back empty.

Issue #2253, opened 2026-08-28
friction and restore ignore flags they do not take, and a flag with no value


vshulcz:
`friction` and `restore` ignore flags they do not take, and ignore a flag whose value is missing, so a mistake exits 0 with prose on stdout.

Every neighbour refuses:

    $ deja how deploy --json        deja: how: unknown flag "--json"     exit 1
    $ deja fix "..." --json         deja: fix: unknown flag "--json"     exit 1
    $ deja files retry --json       deja: files: unknown flag "--json"   exit 1
    $ deja view --json              deja: view: unknown flag "--json"    exit 1

These two do not:

    $ deja friction --json           nothing recurring — none of the 1 …   exit 0
    $ deja friction --limt 3         nothing recurring …                   exit 0
    $ deja friction --limit          nothing recurring …                   exit 0   (value missing, limit stays 10)
    $ deja friction --limit 3 extra  nothing recurring …                   exit 0
    $ deja restore api/upload.go --json    no replaced spans recorded …    exit 0

friction looks only for `--limit` with a value and drops everything else:

    cmd/deja/friction.go:30   for i := 0; i < len(args); i++ {
    cmd/deja/friction.go:31       if args[i] == "--limit" && i+1 < len(args) {

restore ignores any dashed token it does not recognise, and skips `--span` and `-o/--out` when the value is missing:

    cmd/deja/restore.go:51    if i+1 < len(args) { ... }     // --span
    cmd/deja/restore.go:60    if i+1 < len(args) { ... }     // -o/--out
    cmd/deja/restore.go:64    default: if !strings.HasPrefix(args[i], "-") && path == "" {

The `-o` case is the sharp one: `deja restore file.go -o` prints the recovered file to stdout while the reader believes it was written to a file.

vshulcz:
Fixed by #2254 (fix(friction,restore): refuse a flag they do not take).

That PR merged into `hunt/2026-08-23` rather than into the default branch, so nothing here acted on its `Fixes` line. The branch went into main as #2351, so the fix is shipped and this can close.

Issue #1930, opened 2026-08-25
deja files has no --json while every sibling command does


vshulcz:
`deja files` answers "which files did this topic touch" and prints only prose:

```
$ deja files sometopic --json
deja: files: unknown flag "--json"
```

Its siblings all have one — `search` (`cmd/deja/main.go:1801`), `blame`
(`:1896`), `last` (`:1668`), `show` (`:561`), `stats`, `doctor`, `log`, `bench`.
`docs/json-output.md` describes the shape the others follow.

The command is `cmd/deja/files.go:44-83`, where flag parsing already handles
`--limit` and `--project`. Add `--json`, emit the same envelope style the other
commands use, and add a test beside `cmd/deja/blame_test.go`'s JSON case.
Verify with `deja files sometopic --json | jq .`.

Issue #1929, opened 2026-08-25
search --limit exists and the word never appears in any completion script


vshulcz:
`deja search` takes `--limit` (validated 1–100 at `cmd/deja/main.go:1832`), and
the word never appears in any generated completion script:

```
$ deja search deja --limit 1     # works
$ deja completion bash | grep -c limit
0
```

Search falls into the default case, which lists ten flags without this one:
bash `cmd/deja/completion.go:150`, zsh `:268`, fish `:281`, powershell `:355`.

Add `--limit` to all four. Verify with `deja completion <shell> | grep -- --limit`
for bash, zsh, fish and powershell.

PR #2895, opened 2026-09-01
feat(cli): --json for fix and friction

### LINKED ISSUE #1932: fix rejects --json, friction silently ignores it

vaibhav8a:
## What

Both commands answer `--json`, in the shape `docs/json-output.md` describes.

Fixes #1932, taking the option you list first.

## State when I picked this up

The issue describes `friction` accepting `--json`, ignoring it and printing prose. That half is already closed — its unknown-flag guard (#2253) turned the silent ignore into a refusal. So what was left was two commands refusing the same flag, and the issue's own preference: implement it rather than document the refusal.

## The envelopes

Object-shaped, `schema_version` 2, per the stability policy.

```json
{ "schema_version": 2, "sessions_read": 506, "min_sessions": 3,
  "total": 8, "truncated": true, "rows": [ … ] }
```

**The empty result keeps the shape.** That is most of the value here: the prose paths answer "nothing recurring" **five** different ways and "nothing to fix" three, depending on which rule emptied the store — trust policy, ignore rule, no tool output, below threshold. Every one of those is a good sentence for a human and unparseable for a script.

`friction` carries `total` and `truncated` for the reason version 2 gave the search envelope its counts: reading `rows.length` cannot tell you what `--limit` hid. And `min_sessions` because without it a consumer cannot separate *"nothing recurs on this machine"* from *"nothing recurred **twice**"* — different facts about the same store.

`fix` carries `candidate`, the half-evidence flag the prose renders as *"ran next, unconfirmed"*. A caller acting on a fix automatically needs to know which half it is holding, so it is a field rather than a wording difference.

Both use the same `search.SafeLine` / `SafeCommand` sanitisers as the prose path. A recorded error line is untrusted text, and a JSON consumer pasting it into a shell is no safer than a human reading it.

`friction`'s row struct is hoisted to a package-level type so the two renderers cannot disagree about what a row is.

## Two existing tests changed

`TestFixRejectsFlagMistakes` and `TestFrictionAndRestoreRefuseAFlagTheyDoNotTake` pinned the refusals. Updated, not deleted — their property is unchanged and still covered: **a flag a command cannot honour must be refused, never ignored**, now asserted with `--limt` and `--jsonn`. `restore` still refuses `--json` because it does not implement one.

`fix` keeps a case pinning that **`--json` alone is not swallowed into the query** — that is the bug the comment in `fix.go` records, and the reason the flag was refused rather than parsed. It would have been easy to reintroduce while adding the flag.

## Gates

- [x] `go build ./...`
- [x] `go vet ./...` — clean
- [x] `gofmt -l` — clean
- [x] `go test ./cmd/deja/` — ok, 88s
- [x] Driven by hand: `deja friction --json` and `deja fix "some error" --json` on an empty store both return the envelope

Documented in `docs/json-output.md` beside the other surfaces, including why each field is there.

### DIFF
--- a/cmd/deja/fix.go
+++ b/cmd/deja/fix.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -9,6 +10,7 @@ import (
 	"time"
 
 	"github.com/vshulcz/deja-vu/internal/index"
+	"github.com/vshulcz/deja-vu/internal/jsonout"
 	"github.com/vshulcz/deja-vu/internal/policy"
 	"github.com/vshulcz/deja-vu/internal/search"
 )
@@ -23,9 +25,15 @@ import (
 
 func runFix(dir string, args []string, stdout io.Writer) error {
 	limit := 3
+	asJSON := false
 	var parts []string
 	for i := 0; i < len(args); i++ {
 		switch args[i] {
+		case "--json":
+			// Rejected outright until now. The comment below records an earlier bug where
+			// `--json` was swallowed into the search text, so the flag has been on this
+			// command's mind since without ever being answered (#1932).
+			asJSON = true
 		case "--limit":
 			if i+1 >= len(args) {
 				return fmt.Errorf("fix: --limit wants a number")
@@ -72,6 +80,9 @@ func runFix(dir string, args []string, stdout io.Writer) error {
 	pairs := index.FixesFor(dir, text, limit, func(project string) bool {
 		return pol.Allows(policy.ActivationSearch, project)
 	})
+	if asJSON {
+		return writeFixJSON(stdout, pairs)
+	}
 	if len(pairs) == 0 {
 		// One honest line. The old code tried to tell "not an error" apart from
 		// "never seen it", but the test it used rejects `Error: …`, `npm ERR!`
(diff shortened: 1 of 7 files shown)


vshulcz:
read the whole diff including the tests. the envelope design is right and the reasoning for `min_sessions` and `total`/`truncated` is the part i'd have had to ask for otherwise — a consumer really cannot tell "nothing recurs" from "nothing recurred twice" without it.

one blocker.

**`friction --json` hands back a display-truncated error line.** `trimFriction` is `truncatePlanBytes(l, 79)` — a terminal-width cut, added for the prose renderer so a wall fits an 80-column line. the JSON path calls it too, so a recorded error arrives clipped with an ellipsis. reproduced on a fixture rather than read:

```
recorded len=140
json     len=79 value="psql: error: connection to server at \"db-primary-replica-eu-central.internal…"
```

that breaks the only thing a JSON consumer is for — the string can no longer be matched against the error it came from, or passed to `deja fix`. the row should carry the full `r.line`; 79 is a fact about terminals.

one note, not blocking.

the comment above that line says *"Same sanitiser the prose path uses. A recorded error line is untrusted text"*, and the PR body says both commands use `search.SafeLine`/`SafeCommand`. that is true of `fix` and not of `friction` — `trimFriction` truncates and sanitises nothing, and neither does anything upstream of it: `index.FrictionLine` only normalises the shell prefix. so the comment claims a guarantee that is not there. worth fixing the comment; the missing sanitiser on the friction path is pre-existing and ours, not yours — i'll take that one separately.

the two changed tests are right to change and the property they existed for is still pinned. `restore` still refusing `--json` is the correct line to hold.

vshulcz:
the sanitiser half is done on our side — #2910, `trimFriction` now runs the line through `SafeLine` before truncating. rebase on main and the comment in your JSON path becomes true as written.

the truncation blocker still stands: `trimFriction` is the display cut, and the JSON row wants the full `r.line`.

vshulcz:
Took the blocker myself rather than send you back for it — pushed `3ea64fb` onto this branch. The JSON row now carries `search.SafeLine(r.line)`: sanitised, which is what your comment there says, and not cut to a terminal row. `trimFriction` keeps its 79-byte bound on the prose path, where the bound means something.

Three mutations on it: truncating the row again, dropping the sanitiser, and dropping the prose bound. The fixture carries an ANSI escape now so the sanitiser is pinned by an assertion rather than by the import going unused.

The sanitiser half you flagged in the prose path landed separately as #2910, so once this rebases the comment reads true on both sides. Everything else in the PR stands as you wrote it — merging once CI is green.

vshulcz:
merged. thank you — this one was worth the wait.

the thing i'd name: `min_sessions` in the envelope. a consumer that cannot tell "nothing recurs here" from "nothing recurred twice" would have come back as a bug report months later, and you put the denominator in before anyone asked. same instinct behind `total`/`truncated`. that reasoning is now the argument i used on #2919 as well.

if you want another: #1930 (`deja files` has no --json) and #1929 (`search --limit` missing from the completions) are both open and in the same area.

Issue #1336, opened 2026-08-18
a conclusion cut to fit the budget can arrive meaning the opposite


vshulcz:
`digest.Conclusions` picks the decision-carrying lines of a session, newest
first, and is what recall and `deja share` hand to an agent instead of the
transcript. When the next line did not fit the byte budget it was cut to fit:

```go
if spent+len(line) > budget {
    line = UTF8SafeCut(line, budget-spent)
```

Mid-sentence, with nothing to say so. Everywhere else in this code a truncation
is marked with an ellipsis; here the line simply stops.

## Reproduction

Two long conclusions, 400-byte budget:

```
we capped retries at three having weighed the options having weighed the opt
```

The message it came from ends "and left it there." The clause that says what the
session settled on is what falls off, because it is at the end — which is where
a session usually says it changed its mind. A conclusion reading

> we decided to cap retries at three after weighing the options

when the session actually said

> we decided to cap retries at three after weighing the options and then reverted that

is not a shortened answer. It is the opposite answer, handed to an agent as
memory.

Whole sentences or nothing is enough to fix it: keep the first sentence if it
fits, otherwise leave the conclusion out. The line dropped is the older one —
the loop spends the budget newest-first.

PR #2906, opened 2026-09-01
fix(digest): quote the sentence that concludes, not only the opening

### LINKED ISSUE #2243: three quarters of injected blocks carry a mention rather than a conclusion

vshulcz:
The residual of #2243, and the thing its own follow-up comment names: `CarriesDecision` answers differently for a line and for an excerpt of that line, and 8 of the 12 remaining assistant misses were blocks that *did* carry the conclusion with the marking words cut off.

`Conclusions` quoted `firstSentences(text, 2)`. A reply is diagnosis-first, which is why the opening is the default and why #1336's whole-sentences rule is built on it — but when the outcome lands further in, the block handed over the symptom and dropped what the session settled.

Measured through the real path, one scratch index over this machine's stores, `Conclusions(s, 600, 3)` per session, judged by `CarriesDecision`:

```
sessions that settled something   490     490
  the block carried it            275     421
  the block missed it             211      62
  the block came back empty         4       7
```

The head stays the default: a message that already concludes in its opening keeps both sentences, and a message that settles nothing still quotes its opening. Only when the head does not conclude does the lead move to the first sentence that does — one sentence, not the head plus it, since the budget is why it fell off to begin with.

`sentencesOf` splits on exactly what `firstSentences` counts, including the space-after rule that keeps `v5.4.3` whole and the CJK stops that have no space after them; the two disagreeing about where a sentence starts is its own bug.

Three benches unchanged — prompt 13/13 with 6 cross-paired false fires, recall@5 1.00, context median 294 tokens.

**Two of the five mutations were blind at first.** The head-is-kept fixture concluded in its *first* sentence, so removing the guard produced the same string; it now concludes in the second, and the test asserts both sentences survive. And the no-outcome test skipped itself when nothing came back instead of failing. All five red now.

Does not close #2243 — the tool-output and brief-quoting groups there are unchanged.

### DIFF
--- a/internal/digest/decision_lead_test.go
+++ b/internal/digest/decision_lead_test.go
@@ -1,0 +1,89 @@
+package digest
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/model"
+)
+
+func conclusionSession(text string) model.Session {
+	return model.Session{
+		Harness:  "claude",
+		Project:  "p",
+		ID:       "s1",
+		Messages: []model.Message{{Role: "assistant", Text: text}},
+	}
+}
+
+// A reply is diagnosis-first, and the block quoted its opening. When the
+// outcome is further in, that handed an agent the symptom and dropped what the
+// session settled — 211 of 490 sessions on a real store (#2243).
+func TestTheBlockQuotesTheConclusionAndNotOnlyTheDiagnosis(t *testing.T) {
+	got := Conclusions(conclusionSession(
+		"The pool kept dropping connections under load. "+
+			"Both replicas showed the same pattern in the logs. "+
+			"The fix was raising pgbouncer default_pool_size to 40."), 600, 3)
+	if len(got) == 0 {
+		t.Fatal("nothing was quoted")
+	}
+	joined := strings.Join(got, " ")
+	if !strings.Contains(joined, "raising pgbouncer default_pool_size to 40") {
+		t.Errorf("the outcome is not in the block:\n%s", joined)
+	}
+	if !CarriesDecision(joined) {
+		t.Errorf("the block does not read as a conclusion:\n%s", joined)
+	}
+}
+
+// The opening stays the default. A head that already concludes must not be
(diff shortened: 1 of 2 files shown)


vshulcz:

PR #2908, opened 2026-09-01
fix(goose): keep a session unless goose wrote it for itself

vshulcz:
On top of #2874, which fixed the report. Two things it left, plus the index bump.

`hidden` and `gateway` are still dropped. Reading goose's enum (`session_manager.rs`: user, scheduled, sub_agent, hidden, terminal, gateway, acp), `hidden` is `goose run --no-session` — a person working without keeping a session — and `gateway` is a person reaching goose from a chat app. Both are somebody's work.

And the filter stays an allow-list, which is the shape that produced #2873: a type goose ships next is dropped silently and nothing says why. Turned round, it excludes what goose writes for its own work — a subagent it spawned, both spellings, and a scheduled run — and keeps everything else. The test now seeds a `somethingnew` type and asserts it survives.

Index version 32: the db path re-reads only what has been touched since its watermark, so a store built under the old filter keeps missing those sessions until each is used again.

**Why this is a re-land.** #2875 did this and was merged at 09:34. #2877 was branched before it, touched the same file, and its squash took `gooseTypeFilter` back to `= 'user'` at 09:44 — so main lost the fix ten minutes after getting it, with every check green on both. #2874 is what put the ACP half back. Worth a repo setting rather than more care: "require branches to be up to date before merging" would have made #2877 rebase first.

Five mutations, all red: back to the allow-list, excluding gateway, excluding hidden, dropping either subagent spelling, and dropping the filter entirely.

### DIFF
--- a/internal/index/index.go
+++ b/internal/index/index.go
@@ -77,7 +77,13 @@ import (
 // re-derives them — `deja friction` reads the records and sees the difference,
 // while `hook-tool-after` and the environment block read the manifest and stay
 // silent, which is the state this bump exists to end (#2444).
-const version = 31
+// 32: goose sessions are kept unless goose says it wrote them for itself. The
+// filter was an allow-list, so a way of reaching goose that it did not name
+// was dropped. A store built under the old rules is missing those sessions,
+// and the db path only re-reads what has been touched since its watermark, so
+// without the bump they stay missing until each is used again — the same shape
+// as 21, 22 and 23 (#2873).
+const version = 32
 const maxIndexedText = 64 * 1024
 
 // maxRecordSize bounds a single serialized record. A record is one message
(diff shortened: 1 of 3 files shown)


vshulcz:
Correcting this PR's reasoning where it names `hidden`.

I wrote that `hidden` is `goose run --no-session`, "the reader working without keeping a session". I cannot support that. `SessionType::Hidden` is declared in goose's enum and assigned nowhere in `session_manager.rs`, and `--no-session` writes no session at all — its own flag help is "Run without storing a session file", so there is no row of any type to classify.

So the list keeps a type whose source I cannot name, on a justification I invented. The deny-list shape stands on its own — a type goose adds next is a person's work until goose is known to write it for itself — but `hidden` should be decided on evidence rather than on that sentence. Asked @Jelloeater for a count off a real store in #2874; mine has only `user` rows and settles nothing.

`gateway`, `acp` and `terminal` are unaffected: each of those is a person reaching goose a particular way, which is checkable and was checked.

PR #2913, opened 2026-09-01
fix(digest): a code fence must not take the slot from the conclusion

vshulcz:
Follows #2906, which left 62 of 490 sessions handing over a block that did not read as a conclusion. Attributed those rather than guessing, and they are two different things:

```
misses                                        62
  the marker lives only in text the block strips   34
  the block picked a message that settles nothing  28
```

The 34 are the block being right — the marker is inside a code block or an artifact `MessageText` removes, which is the caveat #2243's own follow-up put on this measurement. Left alone.

The 28 are a selection bug. `selectConclusions` keeps a message for three unrelated reasons — it carries a decision marker, it contains a code fence, or it is the last one — and the walk is newest-first, so a message kept for its fence that happens to be newer took the slot and the budget. `concludingFirst` moves the ones that settle something to the end of the list the walk reads backwards, leaving the order inside each group alone, so "the last thing concluded outranks the first thing tried" still holds.

```
sessions that settled something   490
  the block carries it            421  ->  445
  the block misses it              62  ->   35
```

Two of the four mutations were blind at first, and one of those was my code rather than my test: the early-return guard for the all-one-group cases could not change an answer, since appending an empty half leaves the order as it was, so it is gone rather than papered over with a test. The newest-still-wins fixture had no plain message in it, which meant the grouping never ran and the assertion was about nothing. Three mutations now, all red.

### DIFF
--- a/internal/digest/concluding_first_test.go
+++ b/internal/digest/concluding_first_test.go
@@ -1,0 +1,72 @@
+package digest
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/model"
+)
+
+func assistantSession(texts ...string) model.Session {
+	s := model.Session{Harness: "claude", Project: "p", ID: "s1"}
+	for _, t := range texts {
+		s.Messages = append(s.Messages, model.Message{Role: "assistant", Text: t})
+	}
+	return s
+}
+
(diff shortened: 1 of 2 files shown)


vshulcz:

Issue #1738, opened 2026-08-24
doctor never names three stores it indexes from — zed, deepseek and omp are missing from "Harness stores"


vshulcz:
`deja sources` lists 21 stores; `doctor`'s "Harness stores" block prints 18. Missing: **zed**, **deepseek** (dsh) and **omp**. `doctor --json` names 19 — omp is there, zed and deepseek are not. doctor is the surface that answers "where did deja look", so a Zed or dsh user whose sessions are not in recall has no row to check.

Repro on a machine where all three hold history:

```
$ deja sources | awk '{print $1}' | tr '\n' ' '
claude codex gemini cursor antigravity grok qwen kimi goose hermes copilot cline roo pi omp openclaw deepseek zed deja aider opencode

$ deja doctor | sed -n '/Harness stores:/,/^$/p' | awk 'NR>1{print $1}' | tr '\n' ' '
claude codex opencode aider gemini cursor antigravity grok qwen kimi goose hermes cline roo pi openclaw copilot deja

$ deja doctor --json | jq -r '[.stores[].name]|join(" ")'
claude codex opencode aider gemini cursor antigravity grok hermes qwen kimi goose pi omp openclaw copilot cline roo deja
```

Control that they are real stores and not a naming difference: on this machine `deja sources` reports omp 12 sessions, zed 30 sessions / 3711 messages, deepseek 16 sessions — all indexed and recallable. And all three appear in doctor's *other* blocks (`zed` and `omp` under MCP wiring, `omp` and `deepseek` under hooks), so it is only the store rows they are absent from.

Expected: one row per store deja reads, in both the text and the JSON form.

Cause: the rows are a hand-written list — `printFiles`/`printRow` calls in `cmd/deja/doctor.go:576-629` — and a second hand-written list in `doctorStoreChecks` (`cmd/deja/doctor_report.go:253`) feeds the JSON. Neither is derived from what the indexer actually walks, so a new harness lands in `deja sources` and in neither of these. That is the same failure #999 fixed in the other direction, when cline and roo were in the text rows and in nothing else.

Found while running every parser against an empty store directory (BUG-HUNT 80): on a machine where all 21 stores exist and are empty, `deja index`, `deja search` and `doctor` all behave — 0 sessions everywhere, honest "nothing to index yet", no crash, no stray files.

vshulcz:
Fixed by #1739, on main since the collector merge in #1912.

PR #2916, opened 2026-09-01
fix(digest): a sentence with no stopword is not a listing dump

vshulcz:
`looksLikeListingDump` counted path-shaped fields and never read the count. What it actually applied was *eight or more fields, none of them a stopword* — which is an ordinary sentence in any language the stopword list does not cover, and the list is eighteen English words against twelve Russian.

`MessageText` runs every line through it, so this is upstream of every digest: recall, the session-start block, `share`, the conclusions.

Measured over 277375 prose lines of a real store, classifying a drop as a genuine listing when most of its fields are paths:

```
dropped as a listing   20294  ->  4048
of those, listings        55        55
```

The rule now asks what a sentence is made of rather than what it is missing: most of a sentence is plain words, most of a dump is not. That also fixes the shape the path count would have missed — `ls` inside a directory prints no slashes, and `ls -l` is three words and six fields that are not.

`plainWordField` is letters only, two or more, after stripping the punctuation a sentence puts around a word. A version number, a size, a timestamp, a mode string and a path are all not words.

**What it still drops, and I am leaving it.** The 4048 are markdown table rows and lines dense with backticked identifiers — `**80 \`[index-empty-store]\`** закрыт — #1738 → PR #1739 смержен`. Making those pass means treating code spans and dashes as neutral rather than as non-words, and at that point I am tuning a ratio against samples I picked. It wants an instrument that can judge whether a line is worth quoting, which is #2243's subject, not a third threshold here.

Eight mutations, all red — including the first version of this fix, which used a path count and let `ls -l` through; `TestPastedDumpsAreStillNotProse` caught it, which is the test doing exactly its job.

### DIFF
--- a/internal/digest/digest.go
+++ b/internal/digest/digest.go
@@ -271,12 +271,35 @@ func looksLikeListingDump(line string) bool {
 	if len(fields) < 8 {
 		return false
 	}
-	slashes := 0
+	words := 0
 	for _, f := range fields {
-		if strings.ContainsRune(f, '/') {
-			slashes++
+		if plainWordField(f) {
+			words++
 		}
 		if shareStopwords[strings.ToLower(strings.Trim(f, ".,!?:;"))] {
+			return false
+		}
+	}
+	// Most of a sentence is words; most of a dump is not. The path count this
+	// used to keep was never read, so the rule degenerated into "eight fields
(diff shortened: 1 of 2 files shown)


vshulcz:

PR #2918, opened 2026-09-01
fix(digest): a sentence that starts with a number is not a numbered dump

vshulcz:
Same sweep as #2916, next gate along. `shareLineNumRE` was `^\s*\d{1,6}\s` — the number alone — so `MessageText` dropped every sentence that opens with a count.

```
12 зелёных. Жду ревьюера перед мержем.
505 отказов записи шаблонов за 16 часов — это то, что я вчера «починил» ретраем.
21 стора, 1.26 мс и 0.39 мс. Локально дёшево — но проверяю утверждение про сетевой стор.
```

Measured on a real store rather than argued: of the assistant lines the rule matched, **916 are sentences and none is a listing**. The reason there are none is structural — a numbered listing arrives inside a code fence, and `MessageText` returns a fenced message whole without running any gate over it.

It does earn its place on tool output, where the same rule catches 111451 lines. So the rule stays and gets a discriminator, read off the store rather than guessed: `cat -n` and the file readers put a **tab** after the number, and the space-separated dumps — `3 files changed, 9 insertions(+)`, `300 index.js`, `1564 nonpass=0` — are fragments with no sentence in them. A space-separated line now survives only if it reads as one, by the same stop-plus-space rule `firstSentences` already uses, plus the CJK stops that take no space.

```
assistant lines:  916 sentences recovered, 616 still dropped
tool output:      111451 dumps still caught, 1386 through (was 1485)
```

A pipe opens a rendered table row, which is the shape most of the remainder had.

Two of the seven mutations were blind at first — both dump fixtures happened to have no sentence in them, so removing the tab and pipe checks changed nothing. They now carry one each: a numbered source line whose content is a sentence, and a table row ending in one. Seven mutations, all red.

### DIFF
--- a/internal/digest/digest.go
+++ b/internal/digest/digest.go
@@ -143,7 +143,7 @@ func MessageText(s string) string {
 }
 
 var (
-	shareLineNumRE = regexp.MustCompile(`^\s*\d{1,6}\s`)            // "1 diff --git", numbered dumps
+	shareLineNumRE = regexp.MustCompile(`^\s*\d{1,6}[ \t]`)         // "1\tdiff --git", numbered dumps
 	shareGrepRE    = regexp.MustCompile(`^\S+\.[a-z]{1,5}:\d+[:)]`) // path/file.go:18: grep output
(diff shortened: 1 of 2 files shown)


vshulcz:

Issue #2624, opened 2026-08-29
deja how recommends a command whose every recorded run failed


vshulcz:
`deja how` answers "how do I run this here" from what was actually run here. It strips the outcome codex and opencode append, because `$ make test  → exit 2` and `$ make test` are one command and counting them apart put one command on two rows (#2590). Having split the outcome off, nothing looks at it.

Measured on a 1,565-session store:

    command records                                   39,535
    carrying an exit suffix                              390
    distinct commands                                 23,693
    commands whose every recorded run failed             124
    commands with mixed outcomes, or suffixed and bare    65

So `deja how` will print

    $ git rebase origin/main
      ran once in 1 session · last 2026-05-21

for a command whose only recorded outcome on this machine is exit 1, in the same shape and with the same weight as one that has always worked. That is the sharpest miss this screen can make: the whole point is that these are the flags that worked here.

Not a ranking problem — 1% of command records carry an outcome at all, so ordering on it would be arbitrary. It is a truthfulness problem: where deja does know every recorded run failed, the line should say so and let the reader decide.

Nothing to change for a mixed history: a command that sometimes fails is ordinary.

vshulcz:
Fixed in #2625, and in main since 659f9158.

It stayed open because the bug hunt merges into a dated branch first and that branch reached main as one squashed commit, so nothing closed it at the time.

PR #2919, opened 2026-09-01
feat(cli): --json for how

### LINKED ISSUE #1931: deja how prints the command you want to pipe, and has no --json

vaibhav8a:
## What

`deja how <what> --json`, in the shape `docs/json-output.md` describes.

Fixes #1931.

## The envelope, and the two fields that are not obvious

```json
{ "schema_version": 2, "found": 13, "truncated": true, "commands": [ ... ] }
```

**`found` / `truncated`** are here because the cap note goes to **stderr**. A caller reading stdout alone could not otherwise tell eight ways to run the tests from thirteen — the same misread #1632 fixed for the human reader, which the machine reader would have inherited.

**`failed_every_time`** is the prose's failure note as a field. `how` offering a command that never once worked, in the shape of one that always has, is the sharpest miss this surface can make (#2624), so a machine reader has to be able to see it too.

`outcomes` and `failures` sit beside it deliberately, so a consumer can **weigh** that flag rather than trust it blind: only about one run record in a hundred carries an exit status at all, and `failed_every_time` is false whenever deja knows the outcome of no run. Without those two, a `false` reads as "this works" when it often means "nobody recorded". `exit_code` is present only when every recorded failure agreed on one, matching `MixedExit`.

The empty result keeps the envelope. The prose path answers with a sentence naming the terms, and separately with a policy note when a rule withheld everything — a script can branch on neither.

Same `search.SafeCommand` sanitiser as the prose path: a recorded command is untrusted text, and a JSON consumer pasting it into a shell is no safer than a human reading it.

## Tests

Five, including the two guards that matter for adding a flag to a command that takes free text:

- **`--json` alone is still a usage error** — it must not become the search term. That is the shape `fix` was refused over, and it would have been easy to reintroduce here.
- **an unknown flag is still refused** — answering one flag must not reopen the dropped-on-the-floor behaviour the unknown-flag guard closed.

## Gates

- [x] `go build ./...`
- [x] `go vet ./...` — clean
- [x] `gofmt -l` — clean
- [x] `go test ./cmd/deja/` — ok, 83s

Documented in `docs/json-output.md` beside the other surfaces, including why `outcomes` and `failures` accompany the flag.

Note: this is a sibling of #2895 (`--json` for `fix` and `friction`) and touches a different file, so they are independent — merge in either order.

### DIFF
--- a/cmd/deja/how.go
+++ b/cmd/deja/how.go
@@ -142,7 +142,7 @@ func runHow(dir string, args []string, stdout io.Writer) error {
 		fmt.Fprint(os.Stderr, note)
 	}
 	if asJSON {
-		return writeHowJSON(stdout, entries, limit)
+		return writeHowJSON(stdout, entries, limit, hidden, ignored)
 	}
 	if len(entries) == 0 {
 		if note := policyHiddenNote(policy.ActivationSearch, hidden); note != "" {
@@ -373,10 +373,19 @@ func isCommandWordRune(r rune) bool {
 // the cap note lives on stderr, so a caller reading stdout alone cannot otherwise tell
 // eight ways to run the tests from thirteen.
 type howJSON struct {
-	SchemaVersion int          `json:"schema_version"`
-	Found         int          `json:"found"`
-	Truncated     bool         `json:"truncated"`
-	Commands      []howRowJSON `json:"commands"`
+	SchemaVersion int  `json:"schema_version"`
+	Found         int  `json:"found"`
+	Truncated     bool `json:"truncated"`
+	// Withheld and Ignored are what the two rules took out before any of this
+	// was counted. howEntries carries them for a stated reason — filtering on
+	// its own turns a leak into a confident "no command mentions that" over
+	// records a rule hid — and the prose path says so in a sentence. Without
+	// them here the envelope reintroduces exactly what the counts exist to
+	// prevent: `commands: []` reads as "nothing matched" when it can mean
+	// "everything that matched was withheld".
+	Withheld int          `json:"withheld"`
+	Ignored  int          `json:"ignored"`
+	Commands []howRowJSON `json:"commands"`
 }
 
 type howRowJSON struct {
@@ -398,7 +407,7 @@ type howRowJSON struct {
 	ExitCode *int `json:"exit_code,omitempty"`
 }
 
-func writeHowJSON(stdout io.Writer, entries []howEntry, limit int) error {
+func writeHowJSON(stdout io.Writer, entries []howEntry, limit, withheld, ignored int) error {
 	found := len(entries)
 	if len(entries) > limit {
 		entries = entries[:limit]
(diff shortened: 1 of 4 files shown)


vshulcz:
read the whole diff. `outcomes` and `failures` sitting beside `failed_every_time` is the part i'd single out — a bare `false` reading as "this works" when it means "nobody recorded an exit status" is exactly the trap, and putting the denominator next to the flag is the right answer rather than a doc note. checked `SafeCommand` against the prose path this time and it is genuinely the same call, so that claim holds.

one blocker, and i want to be clear that this is **read, not reproduced** — i tried to build the case and my fixture never reached the policy path, so take the mechanism rather than a run.

**the JSON return sits above the empty-result branch, so the withheld count never reaches a machine reader.**

```go
if asJSON {
    return writeHowJSON(stdout, entries, limit)
}
if len(entries) == 0 {
    if note := policyHiddenNote(policy.ActivationSearch, hidden); note != "" {
        fmt.Fprint(stdout, note)
```

with a trust policy withholding everything, the prose reader is told a rule hid N matching sessions and the JSON reader gets `commands: []`, `found: 0`. `howEntries` carries that count for exactly this reason — its own comment says filtering alone "turns a leak into a confident 'no command mentions that' over records the policy" hid. the envelope reintroduces the thing the count exists to prevent.

your own argument for `min_sessions` in #2895 is the same one: a consumer cannot tell "nothing matched" from "the threshold ate it". a `withheld` field (and `ignored`, which is already in hand at the call site) would close it, and both numbers are right there in `entries, hidden, ignored, err := howEntries(...)`.

one note. `SafeCommand` clips at 300 runes, so a command longer than that arrives with an ellipsis. that is the sanitiser the prose path uses so it is consistent, and 300 is a different order from friction's 79 — not blocking, but if a `withheld` field brings you back here it is worth deciding whether the JSON row wants the unclipped command.

vshulcz:
Took the blocker myself rather than send you back for it — `19478dc` on this branch. The envelope carries `withheld` and `ignored` now, straight off the `entries, hidden, ignored` the call site already had, and the docs section gained a paragraph in the same voice as your `min_sessions` one.

The test needed an imported session and a policy that refuses imports; my first attempt used a local-only store and never reached the rule at all, which would have passed while measuring nothing. Both counts are pinned separately — the trust policy and the ignore rule are different questions — and three mutations go red, including swapping the two fields.

Everything else stands as you wrote it. `outcomes` next to `failed_every_time` is the part I'd keep pointing at.

vshulcz:
merged. thank you — two in a day, and the second one argued from the first.

what i'd name here: `outcomes` and `failures` sitting next to `failed_every_time`. a bare `false` reading as "this works" when it means "nobody recorded an exit status" is the kind of thing that becomes a wrong automated decision months later, and putting the denominator beside the flag is a better answer than a sentence in the docs.

i resolved a docs conflict against #2895 on your branch (both PRs added a section in the same place — both are there now) and pushed the withheld/ignored fields as `19478dc`. nothing else was touched.

#1930 and #1929 are still open if you want another.

Issue #2518, opened 2026-08-29
a conclusion one byte over the budget is answered with silence


vshulcz:
Measured on this machine's index — 120 real sessions, newest first, at the 200-byte budget the tool hook uses:

    91 yielded a conclusion
    29 did not — and 16 of those only because the line did not fit

The 16 are not sessions without a conclusion. They are sessions whose conclusion is a sentence longer than the budget, and `Conclusions` answers with nothing rather than with less:

    233B: VERDICT: PASS MISSING: - None SUMMARY: README and section docs cover onboarding, …
    414B: DONE: Restored `should_test()` precedence so exact-key FAILED retries are allowed …

Both are exactly what the line before an edit or a command exists to carry.

The rule is deliberate — "whole sentences or nothing", written so a cut conclusion cannot read as a finished thought while the decisive half fell off. The trade it makes is silence, and deja does not make that trade anywhere else: the environment block, the plan finding and the handoff all cut and mark the cut with `…`, so the reader can see that something followed.

Fix, narrowly: when a caller asks for one conclusion and nothing fits, give the newest candidate cut to the budget and marked. Only for `max == 1` — that is the shape where the line is the whole answer and nothing follows it, so the marker cannot collide with the "nothing follows a marked cut" rule the digest keeps.

The remaining 13 misses are honest: sessions whose assistant turns are all progress narration, and one whose only candidate is a code block.

vshulcz:
Fixed in #2519, and in main since 659f9158.

It stayed open because the bug hunt merges into a dated branch first and that branch reached main as one squashed commit, so nothing closed it at the time.

PR #2922, opened 2026-09-01
fix(digest): mark an overlong conclusion instead of dropping it

vshulcz:
Chasing a number I left behind in #2906: the block came back empty for a handful of sessions that had settled something, and I noted it without following it.

`Conclusions` falls back to a marked cut when one conclusion is longer than the budget, and #2518 restricted that to `max == 1` with the reason that a caller asking for several lines would get text after the marker. But the condition that actually matters is that nothing follows the cut, and an empty `out` is that whatever `max` was asked for — the loop breaks immediately after appending.

So a session whose only conclusion is one overlong sentence answered a caller asking for one line with the sentence and a caller asking for three with silence. The session-start block asks for three.

```
sessions that settled something, with no block at all:  9 of 452  ->  0
```

**All four mutations were blind twice over, and both times it was the fixture reaching a different code path than the one under test.** First the conclusion had no sentence stop, so `firstSentences` hit its own 240-byte cap and returned a marked line without the branch ever running. Then it had one stop at the very end, which `firstSentences` cannot count as two sentences — same cap, same false pass. It takes two sentences with the first alone over budget to reach the cut at all, and the test says so in a comment now. The fourth mutation needed its own case: a line that fitted, followed by one that did not, so that `if true` in place of the empty check is visible.

### DIFF
--- a/internal/digest/digest.go
+++ b/internal/digest/digest.go
@@ -1052,7 +1052,14 @@ func Conclusions(s model.Session, budget int, max int) []string {
 				// cut, so it can be marked the way every other surface marks
 				// one; a caller asking for several keeps the old rule, since
 				// there text would follow the marker.
-				if max == 1 && len(out) == 0 {
+				// The condition is that nothing follows the marker, and an
+				// empty out is that whatever max was asked for: this breaks
+				// immediately after, so the cut is the last line either way.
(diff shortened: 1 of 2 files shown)


vshulcz:

PR #2986, opened 2026-09-02
feat(sources): add Amp thread parser

### LINKED ISSUE #1714: parser: Amp (Sourcegraph) thread history
Amp (Sourcegraph) keeps its conversations as one JSON file per thread under `~/.local/share/amp/threads/`. It is the only harness in this class that another cross-agent recall tool reads and deja does not, so it is a real gap rather than a long-tail one.

What the format looks like, from a public reader ([megamind's `src/sources/amp.rs`](https://github.com/michaelneale/megamind/blob/main/src/sources/amp.rs)) rather than from a store on my own machine — so treat it as a starting point to confirm, not a spec:

- one `*.json` per thread, whole file is a single object
- `id`, `title`, `created` (Unix milliseconds)
- `messages[]` with `role` (`user` / `assistant`) and `content[]`, where text lives in blocks of `{"type": "text", "text": "..."}`
- the working directory is `env.initial.trees[0].uri`, a `file://` URL — fall back to `title` when it is missing
- **messages carry no individual timestamps**; only the thread has one. Our model timestamps per message, so decide what a message gets and write down the choice: the thread's `created` for all of them is the obvious answer and makes ordering within a thread meaningless.

What a PR needs:

- store location + a `DEJA_AMP_ROOT` read override, and whatever native relocation variable Amp has
- parser in `internal/sources/amp.go` (user + assistant text is enough for v1)
- synthetic fixture + `docs/registry/registry.json` entry — the conformance test drives it
- `docs/registry/amp.md` describing the format and its quirks
- resume command if `amp` has one, and its MCP config path if it has an MCP client

[docs/registry/kimi.md](https://github.com/vshulcz/deja-vu/blob/main/docs/registry/kimi.md) and #255 show what a complete spec and implementation look like. Happy to review incrementally, and a spec-only comment that confirms or corrects the field list above from a real Amp install is already worth a lot.

Tethys0:
## Summary

Adds production parsing and indexing support for Amp (Sourcegraph) thread history, so `deja` can recall conversations stored as one JSON object per thread under Amp's local data directory.

Closes #1714.

## User-visible behavior

- Discovers Amp threads under the Linux XDG data location (`$XDG_DATA_HOME/amp/threads/` or `~/.local/share/amp/threads/`) and supports the `DEJA_AMP_ROOT` override.
- Parses thread IDs, titles, `file://` project locations, user/assistant text content blocks, and Unix-millisecond creation times.
- Uses the thread creation time for each message because Amp does not provide per-message timestamps.
- Handles malformed or truncated files through the existing per-file ingest error path instead of aborting the scan.
- Registers Amp with the source registry and incremental index path, and adds synthetic format-registry coverage.

## Verification

- Focused Amp source tests: passed with Go 1.25.
- Focused Amp/index tests: passed with Go 1.25.
- `go vet ./...`: passed.
- `go build ./cmd/deja`: passed.
- `gofmt -s` and `git diff --check`: passed.
- Independent spec review: passed after tightening strict JSON EOF handling and normalizing custom root paths.
- Independent quality/security review: approved with no critical, important, or minor findings.

The broad unfiltered source/index suite was attempted separately and timed out in an unrelated existing environment-sensitive index test after the source package completed; no Amp-focused failure was exposed.


vshulcz:
Thanks — the parser and fixtures look right, and the two Windows-safe root tests are exactly what the goose PR needed last week.

CI is red on the consistency tests rather than on the parser: ten of them pin the harness count and the per-store lists across the docs, doctor and the capability registry, so a 22nd harness has to show up in each. The full list is in the failed job (`TestReadmeSpellsTheHarnessCountTheRegistryHas`, `TestPageTitlesCountTheRestOfTheHarnesses`, `TestDoctorNamesEveryStoreTheIndexerReads`, `TestCapabilityRegistryMatchesCode`, `TestRegistryPagesAreBuiltAndListed` and friends). That is maintainer chores, not parser work, so I'll push those commits onto your branch rather than ask you to chase forty prose counts; "allow edits by maintainers" is on. Shout if you'd rather do it yourself.

vshulcz:
Merged, with two commits of mine on top: the count chores across the docs, and one change in the parser itself — `AmpRoot` now honours `XDG_DATA_HOME` on every platform, which is what your own `TestAmpDiscoveryAndRegistry` expected (it failed on the macOS runner with the linux-only branch). Amp is the twenty-second harness; it ships with the next tag, and the release notes will carry your name on the line.

PR #3015, opened 2026-09-02
fix(install): cline, pi and hermes end the flags before the query too

vshulcz:
Finishing the sweep #2988 started for dsh and #3014 for opencode. The `/deja` command deja writes into Cline, pi and Hermes passed the user's words straight to `deja search`, so a query that names one of deja's own flags died in flag parsing and the reader was told their history holds nothing.

Measured on this machine's store, `deja search --json --limit 3 <query>`: `--json`, `--limit`, `--re`, `--since`, `--all-matches` and `--dry-run` all fail without the terminator and answer with it; `--no-verify` and `-race` pass either way. In a repository about a CLI those flag names are ordinary things to look for.

One test now holds the rule for all four generated plugins at once, so the next one added has to answer it. Each generated file was written out and parsed by its own runtime — node for the two JS, python for Hermes — rather than only matched as a string, because these live inside Go string literals where a syntax error is invisible until install time.

pi is TypeScript and node cannot check it, so that one is asserted on shape only. Grok, Kimi and Zed were checked and need nothing: they carry no query-building command of their own.

### DIFF
--- a/cmd/deja/generated_query_terminator_test.go
+++ b/cmd/deja/generated_query_terminator_test.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"strings"
+	"testing"
+)
+
+// Every plugin deja writes builds a search out of what the user typed, and a
+// query that names one of deja's own flags — `--json`, `--all-matches` — dies
+// in flag parsing. The caller turns the failed call into "", so the reader is
+// told their history holds nothing, which is worse than an error.
+//
+// Measured against a real store: of eight dash-leading queries, six failed
+// without `--` and all eight answered with it.
+func TestGeneratedPluginsEndTheFlagsBeforeTheQuery(t *testing.T) {
+	for _, tc := range []struct {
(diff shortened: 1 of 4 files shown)


vshulcz:
Correcting my own last paragraph before anyone reads it as checked work: Grok, Kimi and Zed do carry a query-building command, and all three have the bare-query form.

    extensions/zed/src/lib.rs   Command::new(binary).arg(query)
    extensions/grok/commands/recall.md   deja "$ARGUMENTS"
    extensions/kimi/commands/recall.md   deja "$ARGUMENTS"

That is the dispatch half of #2988 — `/deja version` runs deja's version command rather than searching for the word — on top of the flag collision this PR is about. I had grepped the two package directories for a `"search"` call and taken the absence as an answer; the Rust one spells it differently and the markdown ones are instructions to the model.

Fixing them in a follow-up rather than widening this PR.

PR #3042, opened 2026-09-03
feat(codex): let recall serve again what compaction threw away

vshulcz:
Codex compacts on its own and fires `PreCompact` with trigger `auto` first. deja was not listening, so the session kept its record of what it had been shown while losing the blocks themselves — recall then stays quiet about exactly what compaction discarded.

Measured on codex 0.149.0, same seeded store as #3041: after a compaction, the seen-ledger still held `seedfrobnicator01` and its block fingerprint without the hook, and held neither with it.

Stacked on #3041.


vshulcz:
Rebuilt on main as the successor PR; this one no longer merges cleanly.

PR #3043, opened 2026-09-04
fix(codex): the digest comes back after a compaction

vshulcz:
Codex fires `SessionStart` again with source `compact` every time it compacts — one `codex exec` run here compacted 12 times. Our matcher was `startup|resume`, so none of them reached the hook and the agent came out of every compaction with nothing. Claude's wiring has had no matcher on this event all along.

Measured on codex 0.149.0 with only SessionStart wired: 2 of 13 requests carried a recall block with the old matcher, 13 of 13 with none.

Install also rewrites the matcher on an entry it already owns. It only rewrote the command, so an entry written by an older deja would have kept the narrow pattern through every upgrade.

Stacked on #3042.


vshulcz:
Rebuilt on main as the successor PR.

PR #3052, opened 2026-09-04
feat(cursor): recall at the moment of the action, not only at the start

vshulcz:
Cursor wired two events: `sessionStart` and `beforeSubmitPrompt`. Its own bundle (2026.07.23) carries two things that make the rest worth adding:

- the list of events whose `additionalContext` it takes — `sessionStart`, `beforeSubmitPrompt`, `preToolUse`, `postToolUse`, `postToolUseFailure`; the client throws at startup if that list and its type union disagree, so it is the client's own statement of what works;
- a table mapping Claude's names onto its own — `PreToolUse`→`preToolUse`, `PostToolUse`→`postToolUse`, `PreCompact`→`preCompact`. Cursor reads `~/.claude/settings.json` too, so it has been running exactly these three deja hooks for anyone who also has Claude Code installed, and none of them for anyone who does not.

So this writes the same three into cursor's own hooks.json under cursor's names. The command strings are identical to the ones deja writes for Claude, which is what cursor dedupes on — a user with both still gets one hook per action.

Evidence, plainly: this is the client's own event map plus hooks cursor already runs, not a driven stand. I could not drive one here — `cursor-agent` reads its access token from the login keychain before it starts a session, and no supported override reaches that path, so a scratch home gets "Password not found" whatever the endpoint is. If you want it measured before merging, say so and I will run it on a logged-in machine.

### DIFF
--- a/cmd/deja/hook_context.go
+++ b/cmd/deja/hook_context.go
@@ -291,6 +291,8 @@ func runHookContext(dir string, plain bool) error {
 		Source    string `json:"source"`
 		SessionID string `json:"session_id"`
 		CWD       string `json:"cwd"`
+		// Cursor leaves cwd empty and names the project here instead.
+		WorkspaceRoots []string `json:"workspace_roots"`
 	}
 	// Best effort, as every hook is — but not silent about it. A payload deja
 	// cannot decode carries the session this injection went to, and losing it
@@ -302,7 +304,7 @@ func runHookContext(dir string, plain bool) error {
 	// environment, so a host that sends the payload without exporting
 	// CLAUDE_PROJECT_DIR got no memory at all — indistinguishable from having
 	// none (#759).
-	digest, sessions, raw, taskMatched, withheld, servedIDs, servedProjects := cachedHookDigestFor(dir, input.CWD)
+	digest, sessions, raw, taskMatched, withheld, servedIDs, servedProjects := cachedHookDigestFor(dir, hookProjectPath(input.CWD, input.WorkspaceRoots))
 	if digest == "" {
 		// No session from this project, which is the usual state in a new
 		// checkout — and exactly where knowing what this machine is missing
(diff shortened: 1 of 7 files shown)


vshulcz:
Measured on the current build now — cursor-agent 2026.09.02, a real turn on a real account, hooks on every event, with a shell command run and approved.

What the run settled:

- The hooks fire: `sessionStart`, `beforeSubmitPrompt`, `preToolUse`, `postToolUse`, `stop`. `postToolUseFailure` and `preCompact` did not, because nothing failed and nothing compacted.
- What a hook returns reaches the model. The transcript has the assistant opening with "deja-vu recalled this same request from earlier today" and closing with "deja-vu recalled: same `echo hi` check earlier today" — deja's own recall, quoted back.
- The set of events that take `additionalContext` is unchanged in this build: sessionStart, beforeSubmitPrompt, preToolUse, postToolUse, postToolUseFailure.

The run also found three things this PR needed before the wiring could say anything, fixed in `57754fc`:

- the shell tool is called `Shell`, which `isCommandTool` did not know;
- the output arrives under `tool_output`, as a JSON document inside a JSON string (`{"output":"…","exitCode":0}`), where deja read `tool_response`;
- `cwd` is empty and the project is named in `workspace_roots`, so recall was answering for whatever directory the hook happened to be spawned in.

One thing is still unproven: that `preToolUse` / `postToolUse` context specifically reaches the model. Two behavioural probes were inconclusive — the model ignored an injected "correction" about what a command printed, which is reasonable of it. The recall that did land arrived ahead of the turn, so what is demonstrated is `sessionStart` or `beforeSubmitPrompt`.

Worth knowing for anyone testing this: cursor refuses to load hook config when the project path contains a symlink, so a stand under `$TMPDIR` on macOS (`/var` → `/private/var`) sees no hooks at all and looks exactly like a build with no hook support.

PR #3071, opened 2026-09-04
feat(install): the proof opens with the questions this machine asked more than once

### LINKED ISSUE #3064: install proof opens with the error the machine keeps hitting, not with the questions it asked more than once
`deja install` ends with "deja already knows this machine:" and a lead line — the recurring error (#2967) — then three recent sessions. The figure that says most about what the index holds is not there: `deja stats` already computes `RepeatQuestions` (questions asked more than once across sessions, `internal/stats/stats.go`), 26 on this machine, and the install never mentions it.

Expected: the lead names the repeat count with the most-repeated question as the example —

```
deja already knows this machine:
  26 questions were asked more than once here — the last: "prepared statement errors behind pgbouncer" — next time the agent hears the earlier answer first
  this machine has hit `command not found: timeout` in 18 sessions (…)
  Jun 8 · codex · api — …
```

Both lines are from the store the install just built; none of it is new data. Needs `stats.RepeatQuestions` to also hand back an example, and the install to print it above the friction line when the count is above zero.

vshulcz:
Under "deja already knows this machine:" the install now says `26 questions asked more than once on this machine — one of them: "…" — next time the agent hears the earlier answer first`, then the recurring-error line from #2967, then the sessions. `stats.RepeatQuestionExample` is the existing repeat count plus the most-repeated question as last typed (ties resolved by stem order, so the example is stable); `RepeatQuestions` wraps it. The line reads the whole store once, after the build, under the same trust-policy filter as the listing.

Tests: `TestRepeatQuestionExampleNamesTheMostRepeatedQuestion`, `TestInstallLeadOpensWithTheRepeatCount`.

Closes #3064

### DIFF
--- a/cmd/deja/install.go
+++ b/cmd/deja/install.go
@@ -396,7 +396,7 @@ func installIndexWarmup(dir string, mcp, hooks, guidance int, summary bool) {
 // few real sessions deja already indexed from this machine's history, so the
 // value is visible before the first agent session ever runs.
 func printInstallProof(dir string) {
-	printMemoryProofOf(dir, "deja already knows this machine:", nil, recurringErrorLine(dir))
+	printMemoryProofOf(dir, "deja already knows this machine:", nil, installLead(dir))
 }
 
 // printMemoryProofOf is the proof narrowed to the rows a caller can honestly
--- a/cmd/deja/install_lead.go
+++ b/cmd/deja/install_lead.go
@@ -1,0 +1,62 @@
+package main
+
+import (
+	"fmt"
+	"io"
+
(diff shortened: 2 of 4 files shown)


vshulcz:
On this store (1,444 sessions) the lead reads:

```
27 questions asked more than once on this machine — one of them: "what did we settle about the checkout worker dropping connections" — next time the agent hears the earlier answer first
  this machine has hit `command not found: timeout` in 17 sessions (claude, opencode) — your agent will be told before it runs into it again
```

2.55 s for the pass over the store, once, after the build.

PR #3072, opened 2026-09-04
feat(antigravity): the repair arrives at the failing command

vshulcz:
The last of the three channels antigravity can carry, measured on antigravity-cli 1.1.13.

`PostToolUse` is handed the error — the payload has an `error` field — and then its contract allows no answer at all: it must return an empty object. So the moment a command fails is reachable only from the `PreInvocation` that follows it, and the failure is already in the transcript the payload names: a step reporting which code the command exited with, and the output under it.

Only the newest tool step is read. A failure two steps back has been answered or worked around, and injecting it again would be talking about something the agent has moved on from.

It drives the ordinary post-tool hook rather than reimplementing anything, so the budget, the dedupe and the "found something, waiting for a second sighting" wording are the ones every other harness gets. The failure takes precedence over the question when both are available: a command that just went wrong is the more urgent memory, and the question will still be there on the next call.

Before and after, live, read out of antigravity's own transcript — a project that fails to build, with the repair for that exact error in another project's history:

    before   digest 1 | fix pair 0
    after    digest 1 | fix pair 1

Two things the stand taught, both worth knowing for the next pass: a run without `--add-dir` executes commands in antigravity's own scratch directory, so a build in the project has to be asked for explicitly; and deja keys a Go compile error by its line and column, so `main.go:4:2` and `main.go:9:2` are different errors — a seeded repair has to carry the error the build actually prints.

### DIFF
--- a/cmd/deja/hook_antigravity.go
+++ b/cmd/deja/hook_antigravity.go
@@ -52,6 +52,9 @@ func runHookAntigravity(dir string, stdin io.Reader, stdout io.Writer) error {
 	if len(input.WorkspacePaths) > 0 {
 		workspace = input.WorkspacePaths[0]
 	}
+	if workspace == "" {
+		workspace = workspaceFromConversation(input.TranscriptPath, input.ConversationID)
+	}
 	// Past the first invocation the digest is already in the transcript, and
 	// the harness has no per-prompt event of its own — so this is where the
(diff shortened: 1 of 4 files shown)


vshulcz:
Rebuilt on current main as the successor PR — this branch grew from a commit that had already been squashed in, and CI never picked it up.

PR #3085, opened 2026-09-04
fix(grok): read grok's payload, and reach the agents it spawns

### LINKED ISSUE #3084: grok: the session hook never fires, and the payload is camelCase
Measured on Grok Build 1.0.5, against a stubbed chat proxy so the request the model actually receives can be read.

**The session hook never fires.** deja wires `SessionStart` with `startup|resume`, copied from Claude Code. Grok's docs name the same two sources, but 1.0.5 sends `source: "new"` for a fresh session and `"load"` for a resumed one, so the matcher matches neither. That is the hook that starts the index warming, so grok is the harness where the first recall of a session waits for a rebuild.

**The payload is camelCase.** Grok sends `sessionId`, `workspaceRoot`, `toolName`, `toolInput`, `transcriptPath`; only `cwd` and `prompt` are spelled the way every other harness spells them. deja reads none of the rest, so the recall looks right — correct project, correct question — and is filed under no session at all. Every grok session on the machine shares the empty key in the dedup ledger, and a compaction forgets nothing.

**The tool hook does not know grok's tools.** `toolHookLine` switches on `Bash` / `Edit` / `Write`. Grok's are `run_terminal_command`, `search_replace`, `write`, and its own alias table is what let the wiring match in the first place (`Bash` matches `run_terminal_command`), so the hook fires on every action and has nothing to say about any of them.

**What grok does with an answer.** Also measured, three reply shapes each: `SessionStart`, `UserPromptSubmit`, `PreToolUse` (allow) and `PostToolUse` are passive — whatever the hook prints is discarded. Two replies reach the model: a `PreToolUse` `deny`, which deja will not send because it does not block work, and `Stop`, which reaches it by keeping the agent working for up to eight more rounds.

The exception is the useful one: a `PreToolUse` reply carrying `updatedInput` is applied. That is the shape `hook-spawn` already answers in, so a spawned agent's prompt is the one place deja can still put memory in front of a model in grok — once the spawn is recognised under grok's name for it (`spawn_subagent`) and read out of grok's spelling of the input.

vshulcz:
Closes #3084.

Four fixes, each measured on Grok Build 1.0.5 against a stubbed proxy:

- `SessionStart` drops the `startup|resume` matcher, which grok's own sources (`new`, `load`) never matched. `PreCompact` drops its matcher for the same reason — every trigger it can name is a compaction.
- `updateClaudeHook` now brings an existing entry up to the matcher this build wires, so the correction reaches machines that already installed deja. Only when deja's hook is alone in the entry: a matcher covers every hook beside it.
- The hook inputs read grok's camelCase spelling, so a session is filed under its own name and a compaction forgets what it threw away.
- `spawn_subagent` is recognised as a spawn and its `toolInput` is read. A `PreToolUse` rewrite is the one reply grok acts on, so this is what puts memory in a spawned agent's prompt there — verified end to end by reading the subagent's own request off the stubbed proxy.

The tool hook also learns grok's shell and editor names; `run_terminal_command` joins the shell tools the post-tool hook already knows.

Every fix has a test that goes red when the fix is reverted.

### DIFF
--- a/cmd/deja/hook_context.go
+++ b/cmd/deja/hook_context.go
@@ -44,6 +44,15 @@ type precompactHookInput struct {
 	CWD            string `json:"cwd"`
 	HookEventName  string `json:"hook_event_name"`
 	Trigger        string `json:"trigger"`
+	// Grok spells all of this in camelCase. See hook_grok.go.
+	grokEnvelope
+}
+
+// adopt fills in what grok spells differently, so the session this compaction
+// belongs to is the one deja forgets.
+func (i *precompactHookInput) adopt() {
+	i.SessionID = adoptGrok(i.SessionID, i.grokEnvelope.SessionID)
+	i.TranscriptPath = adoptGrok(i.TranscriptPath, i.grokEnvelope.TranscriptPath)
 }
 
 // hookStdinWait bounds how long any hook waits for its payload.
@@ -122,6 +131,7 @@ func endsAValue(b []byte) bool {
 func runHookPrecompact(dir string) {
 	var input precompactHookInput
 	_ = json.Unmarshal(readHookStdin(), &input)
+	input.adopt()
 	// Compaction throws away the blocks this session was shown, and the list
 	// that stops them repeating outlives them — so the memory the agent just
 	// lost is exactly the memory recall refuses to send again. Forget what this
@@ -293,6 +303,8 @@ func runHookContext(dir string, plain bool) error {
 		CWD       string `json:"cwd"`
 		// Cursor leaves cwd empty and names the project here instead.
 		WorkspaceRoots []string `json:"workspace_roots"`
+		// Grok spells all of this in camelCase. See hook_grok.go.
+		grokEnvelope
 	}
 	// Best effort, as every hook is — but not silent about it. A payload deja
 	// cannot decode carries the session this injection went to, and losing it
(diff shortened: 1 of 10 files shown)


vshulcz:
Cost, mean of five runs against a copy of the real index, grok's own payloads, each run a fresh session id:

| channel | time | bytes |
| --- | --- | --- |
| session start | 100 ms | 2.4 KB, once per session |
| per prompt | 179 ms | 1.9 KB when it answers, 0 when it does not |
| ordinary tool call | 26 ms | 370 bytes when it answers, and it usually does not |
| spawn | 567 ms | 2.0 KB, only on a spawn |

Both spellings do the same work: the same prompt costs 242 ms read as `session_id` and 179 ms read as `sessionId`, and the spawn is 566 vs 567 ms. The spread is the store, not the reading of it. The spawn is dear because it runs a whole recall inside the hook — that is what it has always cost, in every harness with a subagent tool, and it fires only when one is spawned.

Proof it arrives, read off the stubbed proxy rather than off the hook's stdout: a subagent spawned in grok started with

```
This was asked here before (2026-08-25) — "why does the packer keep refusing our manifest".
  - Assistant: because glimwraxctl holds the manifest lock; --sync releases it
```

`glimwraxctl` exists only in the stand's seeded history, so it cannot be a guess.

PR #3088, opened 2026-09-05
feat(sources): index VS Code Copilot Chat sessions

### LINKED ISSUE #3061: parser: VS Code Copilot Chat sessions (workspaceStorage/*/chatSessions/*.json)
### Harness

GitHub Copilot Chat inside VS Code (the `github.copilot.editsAgent` / agent mode and the plain chat), as distinct from Copilot CLI, which deja already reads from `~/.copilot/session-state/`. VS Code Insiders and VSCodium keep the same layout under their own product directory.

### Session locations

- macOS: `~/Library/Application Support/Code/User/workspaceStorage/<workspace-hash>/chatSessions/<sessionId>.json`
- Linux: `~/.config/Code/User/workspaceStorage/<workspace-hash>/chatSessions/<sessionId>.json`
- Windows: `%APPDATA%\Code\User\workspaceStorage\<workspace-hash>\chatSessions\<sessionId>.json`

`<workspace-hash>/workspace.json` next to it holds `{"folder": "file:///…"}` — the project the session belongs to, which is what the project scope needs. On this machine: 34 session files across 27 workspaces, 292 requests, 82.7 MB, versions all `3`.

### Minimal redacted sample

One file is one session: `requests[]` in order, each with the user's text, the model's reply as a list of parts, a millisecond timestamp, the agent and model ids.

```json
{
  "version": 3,
  "sessionId": "3b7e1f1b-0000-4000-8000-000000000000",
  "creationDate": 1763727100000,
  "lastMessageDate": 1763727400000,
  "requesterUsername": "someone",
  "responderUsername": "GitHub Copilot",
  "customTitle": "fix the retry loop",
  "requests": [
    {
      "requestId": "request_1",
      "timestamp": 1763727104742,
      "agent": { "id": "github.copilot.editsAgent" },
      "modelId": "gpt-5",
      "message": { "text": "the retry loop drops the last attempt", "parts": [ { "text": "the retry loop drops the last attempt" } ] },
      "response": [
        { "value": "The loop compares with < where it needs <=. " },
        { "kind": "toolInvocationSerialized", "toolId": "copilot_readFile", "invocationMessage": "Read retry.go" },
        { "kind": "inlineReference", "inlineReference": { "path": "/w/retry.go" } },
        { "value": "Changed the bound and added a test for the last attempt." }
      ],
      "result": { "metadata": {}, "timings": {} },
      "isCanceled": false
    }
  ]
}
```

Plain reply text comes as `{"value": …}` entries; tool calls as `kind: "toolInvocationSerialized"` (with `toolId` and an `invocationMessage`), file mentions as `kind: "inlineReference"`. `message.text` is the user turn. `isImported` marks sessions restored from elsewhere.

### Resume command

None from a shell: VS Code reopens a session from the chat view's history (Chat: Show Chats…). `deja resume` would have to say so rather than print a command.

### Integration notes

VS Code has no session-start or per-prompt hook deja could install, so this is index-and-search plus the MCP server (`deja install` could write it into the user `mcp.json` VS Code reads). The value is the history: this is the most-used coding agent there is, and every chat it ever had is in these files.

### Data review

- [x] I reviewed the sample for secrets and private data.

Sora-bluesky:
## Summary

Adds a `copilot-chat` harness: the chat sessions GitHub Copilot Chat keeps inside VS Code (agent mode and plain chat), as distinct from Copilot CLI, which deja already reads.

Closes #3061.

The on-disk shape differs from the issue in one way that matters. Since VS Code 1.109 the store writes `chatSessions/<sessionId>.jsonl`, an append log from `objectMutationLog.ts`: the first line is `{"kind":0,"v":{…session…}}` with the whole object the issue shows, and later lines are mutations against it (`kind:1` set at a path, `kind:2` push onto an array with an optional truncate index, `kind:3` delete). The flat `.json` the issue describes is what older builds wrote, and what a current build still writes when `chat.useLogSessionStorage` is set to false. VS Code carries either kind along when a workspace moves, so one machine can hold both. A parser written to the issue's shape finds nothing on a current install.

## What it does

- Roots: `<VS Code User dir>/workspaceStorage/*/chatSessions/` and `globalStorage/emptyWindowChatSessions/` for Code, Code - Insiders and VSCodium on the three OS layouts, with `DEJA_COPILOT_CHAT_ROOTS` as a path-list override (same shape as `DEJA_ROO_ROOTS`).
- `.jsonl` is replayed the way VS Code's reader replays it; `.json` is decoded flat. When a session has both files, only the `.jsonl` is listed, which is the order VS Code reads them in. Any replay failure drops that session with a diagnostic rather than indexing partial state, again matching the store.
- Version 2 files take their title from `computedTitle`, version 3 from `customTitle`; files without a version key get no title and the index derives one.
- The project comes from the sibling `workspace.json` (`folder` or `workspace` URI), or from the session's `workingDirectory` for empty-window sessions. On Windows the `file:///c%3A/...` form is unescaped and the drive letter kept.
- User turns come from `message.text` (or the old plain string), assistant turns from the markdown parts of `response[]`. Tool invocations contribute file paths and terminal command lines under the existing `IndexToolPaths` / `IndexCommands` gates. Thinking parts and UI progress messages are skipped. `inputState` is never read (it carries the GitHub account label).
- No incremental (offset) parse: the log is not append-only in the byte sense, since VS Code rewrites the whole file on compaction and a push can truncate, so a changed file is re-read whole. Files are bounded by that compaction.
- Registry: no install artifacts in this PR (mcp/skill/command left as todo gaps), auto-recall impossible (no hooks), resume impossible (sessions reopen from Chat: Show Chats…), handoff by paste.

## Test plan

- `internal/sources` unit tests cover:
  - replay of kinds 0-3, a push past the end padding with holes the way `arr.length = i` does, a later `kind:0` resetting state
  - a mutation before any initial entry, a malformed line and an empty file, each dropping the session
  - flat `.json`, a string `response`, v2 `computedTitle`
  - `c%3A` and POSIX `workspace.json`, and `workingDirectory` on an empty-window session
  - the account label never reaching a message, thinking skipped, tool parts indexed with the gates at their defaults
  - the `.json` sibling dropped when the `.jsonl` exists, and the env override
- Registry fixture (`fixtures/registry/copilot-chat/`) passes the format conformance test; `go run ./scripts/genmatrix` and `./scripts/genregistry` regenerated the README table and docs pages, and the harness-count words/figures were bumped where those tests check them.
- `go build ./...`, `go vet` on the touched packages, `git diff --check`: clean.
- Full `go test ./internal/sources/ ./cmd/deja/ ./internal/index/` on Windows: the failure set is identical to a clean checkout of `upstream/main` on this machine (sqlite shim and a local stray file), no new failures.
- `deja doctor` on this machine reports the store: `copilot-chat found ~\AppData\Roaming\Code\User (3 files)`.

## Notes

The three sessions on this machine all have `requests: []`, so the request/response shapes are pinned to the VS Code type definitions (`ISerializableChatData3`, `SerializedChatResponsePart`, `IChatToolInvocationSerialized` at commit e4c7e7b1) and the sample in the issue, not to a captured non-empty file. The fixtures are synthetic. If a `.jsonl` with a few real requests is at hand, that would be a better fixture and I will swap it in.


vshulcz:
read the whole diff. the parser is the careful part and it holds up: kind:0 replaces state, 1/2/3 mutate, a later 0 resets, and the push writes the appended slice back into the parent so there's no slice-aliasing bug. malformed/unknown-kind/empty all drop the session the way vs code does rather than keeping partial state. reproduced locally — `go test ./internal/sources -run CopilotChat` green, build green, and the harness-count invariant + doctor testdata moved with it.

confirmed the redaction point from #3061 too: the parser never touches `inputState`, so the github account label doesn't reach the index.

one note, not a blocker: your three local sessions are all `requests: []`, so the request/response shapes rest on the vs code type defs and the issue sample. if 34 real flat files turn up a `.jsonl` with a couple of requests would make a better fixture than the synthetic one — fine to leave as is.

vshulcz:
thanks for this — replaying the mutation log instead of trusting append offsets is exactly right, and reading both the .jsonl and the flat .json so a machine with old and new sessions gets both is the part that would have been easy to miss. merging.

Sora-bluesky:
Thanks for reading the replay that closely, and for reproducing it.

The fixture offer stands: if a `.jsonl` with a couple of real requests turns up on any machine, I will swap it in for the synthetic one and re-pin the request and response shapes against it. The install side (writing deja into the VS Code user `mcp.json`) is another gap marked todo in the registry, and I can pick that up as a separate PR if it is wanted.

PR #3090, opened 2026-09-06
docs: document reproducible README GIF recipe

### LINKED ISSUE #1832: the README GIF recipe ends in an ellipsis, so nobody can rebuild it the same way
`scripts/demo/story.py` draws the frames for the README GIF, and its docstring ends the recipe with an ellipsis:

```
python3 scripts/demo/story.py --out /tmp/frames
ffmpeg -framerate 20 -i /tmp/frames/%04d.png ... demo.gif
```

Everything that decides how the GIF looks is inside those three dots. I regenerated `assets/demo.gif` in #1767 and had to recover the settings by comparing against the committed file: the first attempt came out 720 px wide with 320 frames against the original's 1000×467 and 192, and 833 KB against 620 KB.

What the committed GIF was actually built with, recovered from `ffprobe` and by trying the usual knobs:

```sh
python3 scripts/demo/story.py --out /tmp/frames

ffmpeg -y -framerate 20 -i /tmp/frames/%04d.png \
  -vf "fps=12,scale=1000:-1:flags=lanczos,palettegen=max_colors=64:stats_mode=diff" \
  /tmp/palette.png

ffmpeg -y -framerate 20 -i /tmp/frames/%04d.png -i /tmp/palette.png \
  -lavfi "fps=12,scale=1000:-1:flags=lanczos[x];[x][1:v]paletteuse=dither=none:diff_mode=rectangle" \
  -loop 0 assets/demo.gif
```

That lands at 1000×467, 192 frames, 592 KB — the same shape as the file in the repository, which is how I know it is close to the original recipe rather than merely a recipe that works.

## What a fix looks like

Put the real commands in the docstring instead of the ellipsis, and ideally behind a `make demo` target so the next person types one thing. A short note on why the numbers are what they are helps more than the numbers alone:

- `fps=12` over the source's 20 — the frames are drawn at 20 and dropped to 12 for size; the animation is slow enough that it does not show.
- `max_colors=64` with `dither=none` — the palette is six flat colours and a glow; dithering costs 140 KB and buys nothing on flat fills.
- `scale=1000` — the README renders it at 720 CSS pixels, so 1000 leaves room on a retina screen without paying for 2×.

Whoever does this should regenerate the GIF from their own run and check the result is byte-comparable in shape (`ffprobe -show_entries stream=width,height,nb_frames`), not just that a file appeared.

Worth stating in the source too: nothing in CI rebuilds this, and no test reads it, so the recipe is the only thing standing between the next editor and a mismatched GIF.

keii-2596:
### Summary

Documented the exact reproducible README GIF recipe and its visual tradeoffs.

## Summary
- replace the incomplete GIF recipe with exact two-pass ffmpeg commands
- document the output choices and verification command

## Testing
- regenerated the GIF in a temporary directory: 1000x467, 192 frames, 592199 bytes
- git diff --check

Fixes #1832

### Verification
- Full Pillow/ffmpeg generation produced 1000x467, 192 frames, 592199 bytes
- git diff --check passed

> Prepared by IssuePilot with Codex. This is a draft and requires human review.

### DIFF
--- a/scripts/demo/story.py
+++ b/scripts/demo/story.py
@@ -16,7 +16,28 @@ Four scenes, sixteen seconds:
   4. the line and the command
 
     python3 scripts/demo/story.py --out /tmp/frames
-    ffmpeg -framerate 20 -i /tmp/frames/%04d.png ... demo.gif
+
+    ffmpeg -y -framerate 20 -i /tmp/frames/%04d.png \
+      -vf "fps=12,scale=1000:-1:flags=lanczos,palettegen=max_colors=64:stats_mode=diff" \
+      /tmp/palette.png
+
+    ffmpeg -y -framerate 20 -i /tmp/frames/%04d.png -i /tmp/palette.png \
+      -lavfi "fps=12,scale=1000:-1:flags=lanczos[x];[x][1:v]paletteuse=dither=none:diff_mode=rectangle" \
+      -loop 0 assets/demo.gif
+
+The source frames are rendered at 20 fps, then reduced to 12 fps to keep the
+GIF small without making this slow animation look choppy. The six flat colours
+and glow fit comfortably in a 64-colour palette, so dithering only adds noise
+and bytes. A 1000 px output remains sharp at the README's 720 CSS px width
+without paying for a full 2x image.
+
+Nothing in CI rebuilds or checks this asset. After regenerating it, verify the
+shape of the result (tool versions can still cause small byte-size differences):
+
+    ffprobe -v error -show_entries stream=width,height,nb_frames \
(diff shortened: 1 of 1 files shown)


vshulcz:
read it. the two-pass palettegen/paletteuse recipe is the right way to get a clean gif at this size, and the verification step is what makes it reproducible. reproduced: the shipped assets/demo.gif is exactly 1000x467 with 192 frames, so the recipe matches the asset in the repo rather than a guess.

vshulcz:
thanks — ending the recipe on a concrete ffprobe check with the expected shape, instead of the old ellipsis, is the useful bit. merging.

PR #3091, opened 2026-09-06
fix(completion): suggest search limit flag

### LINKED ISSUE #1929: search --limit exists and the word never appears in any completion script

keii-2596:
### Summary

Added search --limit completion support for Bash, zsh, fish, and PowerShell with focused regression tests and changelog entry.

## Summary
- add --limit to search completions for Bash, zsh, fish, and PowerShell
- add focused scope and declaration tests
- record the fix in the changelog

## Testing
- go test ./... -race -count=1
- go vet ./...
- go build ./cmd/deja
- cmd/deja coverage: 90.0%

Fixes #1929

### Verification
- focused completion tests passed
- go test ./... -race passed
- go vet ./... passed
- go build passed
- cmd/deja coverage 90.0 percent; total 90.1 percent
- git diff --check passed

> Prepared by IssuePilot with Codex. This is a draft and requires human review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - OpenClaw: `@vshulcz/openclaw-deja`, the recall plugin as a package — `openclaw plugins install clawhub:@vshulcz/openclaw-deja` — with `before_prompt_build` recall and `deja_recall`, `deja_fix`, `deja_blame` as tools; it reads what `deja install openclaw-auto` wrote and adds only what is missing. (#3047)
 
+### Fixed
+- Search: offer `--limit` in bash, zsh, fish, and PowerShell completions. (#1929)
+
 ## [0.19.3] - 2026-09-04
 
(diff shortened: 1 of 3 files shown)


vshulcz:
read the whole diff. `--limit` is a real search flag bounded 1 to 100 (main.go: "--limit needs an integer from 1 to 100"), so the completion text and the range are accurate, and you added it to all four shells — bash, zsh, fish, powershell — not just one. reproduced: the completion test is green locally.

vshulcz:
thanks — catching that a shipped flag was missing from every completion surface, and covering all four shells in one go, is the kind of gap worth closing. merging.

PR #3102, opened 2026-09-06
feat(vscode): wire deja MCP into VS Code Copilot Chat

### LINKED ISSUE #3101: vscode: recall has no delivery channel into Copilot Chat
The VS Code Copilot Chat parser (#3088) indexes its sessions, but nothing delivers recall into a Copilot Chat turn — the registry entry says so: "MCP / skill / command: not wired. Index and search only."

Copilot Chat has no hook and no plugin an outside CLI can reach. The one channel it does have is MCP: since VS Code 1.102 agent mode reads MCP servers from the profile's `mcp.json`, and calls their tools. deja already ships an MCP server (recall/fix/how/blame/remember), so the gap is only the wiring — and VS Code's config shape is its own: a top-level `servers` map (not `mcpServers`), each entry with a `type`.

vshulcz:
Closes #3101.

`deja install vscode` (alias `copilot-chat`) writes the user `mcp.json` VS Code reads in agent mode — its own shape, a top-level `servers` map with `type: stdio`, one file per host present (Code, Insiders, VSCodium), falling back to the stable Code User folder so the config is ready before the editor is, the way cursor's and gemini's are. Merges into an existing file, keeps other servers and JSONC comments, and uninstall takes only deja's entry. Registry, doctor row, and the capability matrix move with it.

Proven on the GUI, VS Code 1.134.0 (Copilot Chat 0.62.0), read out of the MCP traffic between VS Code and deja: Copilot Chat in agent mode started the deja server (`initialize`, `tools/list`) and called it with `{name:"deja", arguments:{mode:"recall", query:"packer manifest glimwraxctl"}}`; deja returned the seeded decision "the packer refuses the manifest until glimwraxctl --sync has run". `glimwraxctl` exists only in the stand's seeded history, so recall genuinely reached the model's tool loop.

Cost: no per-action hook — the server starts once per session and recall is a tool the agent calls on demand, so a turn that does not ask costs nothing. Negative result recorded in the registry: Copilot Chat has no session-start or per-prompt hook and loads no shared skill directory, so MCP is the only channel; auto-recall stays impossible there. Installer behaviour (servers shape, type, other-server and comment preservation, uninstall) is pinned by tests, each red when reverted.


vshulcz:
Correcting a claim I made here. This PR's body, and the registry and guide text it shipped, said Copilot Chat "has no session-start or per-prompt hook and loads no shared skill directory, so there is nowhere to put a block that is simply there before the model reads the question". The first half is right; the conclusion is not.

VS Code applies a custom-instructions file to every chat when its frontmatter says `applyTo: "**"`. Measured on 1.134.0 by reading the MCP traffic: with such a file in the User folder and nothing else, Copilot Chat called deja on a question that had nothing to do with past work; with deja's own guidance installed and deja never mentioned in the question, it called `recall` on its own and answered from a session it had never seen.

So this harness does get recall unasked, and install should write that file. #3107 does it and fixes the wording in the registry, the guide and the matrix.

PR #3105, opened 2026-09-06
docs: a per-agent guide for VS Code Copilot Chat

vshulcz:
VS Code Copilot Chat got MCP wiring in #3102 and was the only wired harness without a per-agent guide. This adds one, modelled on the Zed page since both are MCP-only: where the sessions live (the JSONL mutation log and the older flat form), the `mcp.json` shape VS Code actually reads — top-level `servers`, `type: stdio`, not the common `mcpServers` — and the honest gap, that Copilot Chat has no session-start or per-prompt hook so recall arrives when the agent calls the tool.

The sidebar is copied into each guide page by hand, so the new link and the per-agent count move across all 43 of them, plus both sitemaps and llms.txt.

### DIFF
--- a/docs/guide/after-compaction.html
+++ b/docs/guide/after-compaction.html
@@ -51,7 +51,7 @@
 <a href="repeated-mistakes.html">Repeated mistakes</a>
 <a href="switching-agents.html">Switching agents</a>
 <a href="parallel-sessions.html">Parallel sessions</a>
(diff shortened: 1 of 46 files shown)


vshulcz:
The guide page this added carried a claim I have since disproved — that there was nowhere to put a block in front of Copilot Chat before it reads the question. There is: a custom-instructions file with `applyTo: "**"`, and with it the agent calls recall on its own. #3107 corrects the page and wires the file. Detail in the correction on #3102.

Issue #1628, opened 2026-08-24
deja files swallows a flag with no value, an empty --project, an empty topic and an unknown flag


vshulcz:
`deja files` has its own flag loop, and four ways of typing something that does nothing pass through it in silence. Every sibling command refuses all four.

Fixture: one claude session that mentions "retry" and records two edited files.

```
$ deja files retry --limit          → answers as if --limit were never typed
$ deja files retry --project        → same
$ deja files retry --unknown        → same, the flag vanishes
$ deja files retry --project ""     → the whole store, not a project
$ deja files ""                     → 0 sessions mention "", none of them recorded a file near it
```

Controls, same command and same store, showing none of this is a general policy:

```
$ deja files                        → deja: usage: deja files <topic> [--project name] [--limit n]   (exit 1)
$ deja files retry --limit 0        → deja: files: --limit wants a positive number, got "0"          (exit 1)
$ deja files retry --project nosuch → no sessions mention "retry" in project "nosuch"
```

And the siblings: `deja --harness` says `--harness needs value`; `deja retry --unknown` says `unknown flag "--unknown" — did you mean …`; `--project ""` is refused everywhere else since #1612; an empty query is `deja: query required`.

Expected: a flag with no value, an unknown flag, an empty `--project` and an empty topic are all mistakes, and each already has a sentence somewhere in deja.

Cause: `runFiles` (`cmd/deja/files.go:44`) guards its flag arguments with `if i+1 < len(args)` and has no else branch, so a trailing flag is dropped; the default branch keeps any non-dash token as a term, so `""` becomes the topic; and nothing rejects a dash-prefixed token it does not know.

Impact: the answer looks like an answer. `deja files retry --project api-gateway` typed as `--project` alone returns results from every project, which reads exactly like results from one.

This is the same defect class as #1612 and #1618. `files` was missed there because it does not share the parser or the wording — it says "wants a positive number" where the others say "needs value".

Found during the bug hunt, item `[files-topic]`.

vshulcz:
Fixed by #1629, on main since the collector merge in #1912.

PR #3106, opened 2026-09-06
feat(cli): --json for files

### LINKED ISSUE #1930: deja files has no --json while every sibling command does

vaibhav8a:
## Summary

`deja files <topic> --json`. Closes #1930.

The envelope follows `docs/json-output.md` and carries the counts a caller cannot recover from the list:

```json
{
  "schema_version": 2,
  "query": "singbox",
  "sessions_scanned": 34,
  "matched": 34,
  "read_capped": false,
  "truncated": true,
  "filtered": 2,
  "withheld": 0,
  "ignored": 0,
  "files": [
    { "path": "/srv/app/internal/singbox/render.go", "near": 18, "sessions": 6, "total": 21 }
  ]
}
```

Each row keeps the three numbers the ranking is built from — `near`, `sessions`, `total` — so a consumer can re-rank or threshold without re-deriving them. `total` is the denominator that makes a file specific to the topic rather than merely busy.

**All four exits emit the envelope, not just the one with rows.** Writing the tests is what found that: my first version handled the no-hits path and the ranked path and left `N sessions mention "x", none of them recorded a file near it` printing prose under `--json`, so a caller got JSON on a hit and a sentence on that miss. An empty `files` means four different things — nobody discussed it, a rule withheld every session, a tree is kept out of recall, or the paths recorded are not on this disk today — and the counts beside it are what tell those apart. That is the same argument `how --json` makes for carrying `withheld` and `ignored`.

**`truncated` and `read_capped` are kept separate on purpose.** One is the caller's `--limit` biting on the file list; the other is the read budget biting on the sessions behind it. Folding them into one flag would report the wrong cut, and the prose path already distinguishes them (`filesReadNote`).

## Test plan

```sh
go build ./cmd/deja
go vet ./...                      # clean
go test ./cmd/deja -race -count=1 # ok, 218s
gofmt -s                          # clean on both changed files
```

Per-package coverage, measured directly as CONTRIBUTING asks:

```
ok  github.com/vshulcz/deja-vu/cmd/deja  coverage: 90.0% of statements
```

90.0% against the `[cmd/deja]=88` floor in `ci.yml`. `runFiles` itself is 92.1%, `writeFilesJSON` 100%.

Three tests: the envelope on a hit (schema version, query, and that `total >= near` for every row, since near-touches are a subset), the envelope on a miss (same shape, empty list), and that `--json` did not open a hole in the unknown-flag refusal from #1628 — `--jsonn` and a bare `--json` with no topic both still error.

## Notes

`docs/json-output.md` gains a `deja files --json` section next to the others, including why the two truncation flags are distinct.

Verified by hand as the issue suggests: `deja files sometopic --json | jq .` parses.

### DIFF
--- a/cmd/deja/files.go
+++ b/cmd/deja/files.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"math"
@@ -13,6 +14,7 @@ import (
 	"time"
 
 	"github.com/vshulcz/deja-vu/internal/index"
+	"github.com/vshulcz/deja-vu/internal/jsonout"
 	"github.com/vshulcz/deja-vu/internal/model"
 	"github.com/vshulcz/deja-vu/internal/policy"
 	"github.com/vshulcz/deja-vu/internal/query"
@@ -46,8 +48,11 @@ func runFiles(dir string, args []string, stdout io.Writer) error {
 	var terms []string
 	limit := 10
 	project := ""
+	asJSON := false
 	for i := 0; i < len(args); i++ {
 		switch args[i] {
+		case "--json":
+			asJSON = true
 		case "--limit":
 			// A flag typed with nothing after it used to be dropped in silence,
 			// and so did an unknown one and an empty --project: the answer came
@@ -83,7 +88,7 @@ func runFiles(dir string, args []string, stdout io.Writer) error {
 		}
 	}
 	if len(terms) == 0 {
-		return fmt.Errorf("usage: deja files <topic> [--project name] [--limit n]")
+		return fmt.Errorf("usage: deja files <topic> [--project name] [--limit n] [--json]")
 	}
 	q := strings.Join(terms, " ")
 	o := search.Options{Query: q, All: true, Project: project}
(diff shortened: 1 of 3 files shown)


vshulcz:
read the whole diff and ran it: `go test ./cmd/deja` green, and drove `deja files --json` by hand against two scratch stores — the "sessions matched, nothing near" exit and the "nobody discussed it" exit both come back as the envelope with the counts that tell them apart, and the numbers match what the prose path prints for the same store. envelope shape matches `how --json` down to the encoder, so a consumer that handles one handles this.

two notes, neither blocking:

- `TestFilesJSONEnvelope` runs against `fixtures/synthetic/claude`, where the query returns no rows on this machine — i checked, `files: []` for `parser`, `index` and `search`. so the row assertions (`total >= near`, `sessions > 0`) never execute. a store seeded in the test with a path under a directory that has a `.git` in it would exercise the ranked branch for real; i'll push that test rather than send you back for it.
- `withheld` and `ignored` are filled on the no-hits branch only. on the ranked branch the fields are in the envelope and always zero, which for a consumer reads as "no rule touched this" rather than "unknown". the prose path has the same gap, so it isn't a regression from this PR — worth a follow-up either way.

`deja help` still prints `deja files <topic> [--project name] [--limit n]` while its siblings list `[--json]`; taking that too.

vshulcz:
thanks — the part that stands out is that you found the third exit by writing the tests, and said so in the PR body. that miss printing prose under `--json` is exactly the kind of thing that survives review and bites a script six months later. the `truncated` / `read_capped` split is the right call too: they cut in different places and folding them would report the wrong one.

vshulcz:
follow-up merged in #3124: the ranked-rows test (render.go in three sessions reports three, routes.go in one reports one — three mutations turn it red) and `[--json]` on the files line in `deja help`. nothing of yours changed.

PR #3136, opened 2026-09-07
test(files): seed the ranked-rows fixture as JSON, not as pasted text

vshulcz:
main has been red on windows-latest since #3124: `TestFilesJSONCarriesTheRankedRows` builds its transcript by pasting paths into JSON, and a Windows path is full of backslash escapes (`repo\render.go` carries a \r), so every seeded line failed to parse and the store indexed nothing.

The PR checks never saw it — the windows leg is skipped on an unlabelled pull request and runs on every push to main, so it only went red after the merge. This PR carries the `windows` label, so the leg runs here.

The records are marshalled now, and the test asserts its own seed is valid JSON, so the next hand-built fixture fails everywhere rather than on one runner.

### DIFF
--- a/cmd/deja/files_json_rows_test.go
+++ b/cmd/deja/files_json_rows_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/vshulcz/deja-vu/internal/index"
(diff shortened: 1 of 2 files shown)


vshulcz:
Windows leg green here (38m): both tests pass on the platform they were red on. Closes #3137, closes #3138.

PR #3158, opened 2026-09-07
feat(crush): read Crush history and wire recall into it

### LINKED ISSUE #2949: parser: Crush (charmbracelet) session history
Asked for by @Jelloeater on #2874, and it looks like an ordinary db-backed harness rather than a new shape.

[Crush](https://github.com/charmbracelet/crush) is Charm's terminal coding agent, 27.8k stars and active. It keeps sessions in SQLite, not in the JSON state file the README talks about — `internal/db` has a full schema with `sessions.sql`, `messages.sql` and five migrations.

The two tables deja would read, from `20250424200609_initial.sql`:

```sql
CREATE TABLE IF NOT EXISTS sessions (
    id TEXT PRIMARY KEY,
    parent_session_id TEXT,
    title TEXT NOT NULL,
    message_count INTEGER NOT NULL DEFAULT 0,
    ...
    updated_at INTEGER NOT NULL,
    created_at INTEGER NOT NULL
);

CREATE TABLE IF NOT EXISTS messages (
    id TEXT PRIMARY KEY,
    session_id TEXT NOT NULL,
    role TEXT NOT NULL,
    parts TEXT NOT NULL default '[]',
    model TEXT,
    created_at INTEGER NOT NULL,
    ...
);
```

So the shape is the opencode/cursor one: a session row with `updated_at`, message rows carrying a role and a JSON `parts` blob. That means it can take the same since-the-watermark path rather than re-reading the store every pass, and the registry entry gets a `ParseFrom` from the start.

Three things a first patch has to settle, and none of them can be guessed from the schema:

- **Where the database is.** The README documents `$HOME/.local/share/crush/crush.json` for JSON state and says nothing about the sqlite file; it needs finding on a real install, along with whether it is per-project (the `.crush/` directory suggests it might be) or one store per machine.
- **What `parts` holds.** The column is a JSON array; the parser needs the text shape and how tool calls and their output are represented, so `roleToolOutput` gets what it should and friction and `fix` work.
- **What the project is.** `sessions` has no `cwd` or path column here. If the project only exists as a directory the store sits under, that decides how sessions get attributed — the same question `omp` and `prime` each answered differently.

`parent_session_id` is worth noticing: Crush has subagent sessions, so the goose question from #2873 arrives here too — which of them are a person's work.

There is no ACP dependency for this. @Jelloeater was waiting for Crush to ship ACP support, which would let goose-style clients drive it; reading its history off disk needs nothing from either side.

vshulcz:
Crush (charmbracelet) is the 25th harness. It is the only one that keeps history beside the work — `<project>/.crush/crush.db` — so the stores are found through Crush's own registry in the data home rather than by walking anything.

Wiring, all measured on crush v0.92.0 against a recording endpoint:

- **MCP** under `mcp` (not `mcpServers`). `crush_info` reports `deja = connected (1 tools, 0 resources)`; a scripted call came back with the seeded decision in the bytes Crush sent next.
- **Auto-recall** as a `PreToolUse` hook, the only event Crush fires. It answers in Crush's flat `{"version":1,"context":…}` — a new `deja hook-tool --crush` — and carries no `decision`, since `"allow"` there is affirmative pre-approval and would skip the permission prompt on every call. With a scripted `go build ./...` against a store holding that command's history, the recall reached the model appended to the tool result.
- **Skill** via the shared `~/.agents/skills`, which Crush scans by default; **command** as `/user:deja`; **resume** as `crush --session <uuid>` run in the project.

Costs: deja's tool schema is 3,138 B of the 29,754 B tool block, the skill entry 384 B of the 22,487 B system prompt, a recall that answers 269 B on the turn that asked. With the binary moved aside the turn still completed — 28 tools instead of 29, hook failure logged and ignored.

Two things found along the way and fixed here: `genregistry`'s number-word table had `twenty-two` twice, so every registry page has been one short since it landed; and `itoa` overflowed its 8-byte buffer on the ninth digit.

Closes #2949.


Jelloeater:
Mentioned https://github.com/charmbracelet/crush/discussions/3739

Issue #3138, opened 2026-09-07
Windows: files --json scans 0 sessions in TestFilesJSONCarriesTheRankedRows since #3124


vshulcz:
Second Windows-only failure on main, since #3124 (today). Latest run: https://github.com/vshulcz/deja-vu/actions/runs/34142574155

```
--- FAIL: TestFilesJSONCarriesTheRankedRows (0.08s)
    files_json_rows_test.go:82: the ranked branch returned no rows, so nothing here is pinned:
        { "schema_version": 2, "query": "singbox", "sessions_scanned": 0, "matched": 0, …
```

`deja files --json singbox` scans 0 sessions on the Windows runner where ubuntu and macos scan the fixture and return rows — so the fixture the test writes is not found by the store on Windows (project/cwd matching or the path the fixture is written under), not a ranking difference. The test was added by #3124 and pinned a behaviour that never ran on Windows before.

Expected: the same rows on all three runners, `sessions_scanned > 0`. Related: the other Windows red on main from #3087 (deja_once), filed separately.

vshulcz:
Fixed in #3136: the fixture is built with json.Marshal and asserts its own seed is valid JSON, so a hand-built one fails everywhere rather than on one runner.

vshulcz:
Probe #3166 (the test with t.Logf around the two calls, nothing else) passed on windows-latest twice: runs 34167931441 and 34170249057. Not reproducible on demand on the runner; the next failure needs the runner's own log lines, which the probe branch `probe/windows-3137-3138` prints.

Issue #3137, opened 2026-09-07
Windows: deja_once lets the session-start digest in twice — test/windows-latest red on main since #3087


vshulcz:
`test / windows-latest` on main has been red on every run since #3087 landed (first red run: commit 5770986, 2026-09-04 15:54; last green: 3eafb1a, 15:08). ubuntu and macos pass. Latest run: https://github.com/vshulcz/deja-vu/actions/runs/34142574155

```
--- FAIL: TestDejaOnceKeepsTheDigestToTheFirstTurn (0.41s)
    hook_context_once_test.go:49: the digest went in a second time in the same session:
        "<deja-recall>\nRecalled history from prior sessions. …
    hook_context_once_test.go:76: --once let the digest in twice in one session:
```

On Windows the `deja_once` marker does not hold: the second `hook-context` call with the same session id and `"deja_once":true` returns the digest again. Same test, same fixture, passes on the other two runners, so the suspect is whatever the once key is derived from on Windows — `onceDigestKey(sessionID)` is plain, so more likely the marker file path or the project key from `cwd` (`filepath.Join(t.TempDir(), "tmp", "beta")` with a drive letter and backslashes) not matching between the two calls.

Expected: the second call returns nothing, on every runner. Control: `once-2`, a different session id, still gets the digest (the test's third assertion).

Also blocks: deepseek-plugin-store re-checks main on Windows before listing.

vshulcz:
Fixed in #3136: the payloads are marshalled now, and the test checks its own payload survives a decode. The windows leg ran on that PR and passed.

vshulcz:
Probe #3166 (the test with t.Logf around the two calls, nothing else) passed on windows-latest twice: runs 34167931441 and 34170249057. Not reproducible on demand on the runner; the next failure needs the runner's own log lines, which the probe branch `probe/windows-3137-3138` prints.

PR #3166, opened 2026-09-07
probe: Windows-only failures in deja_once and files --json (not for merge)

vshulcz:
Diagnostics only, to read the Windows runner's view of #3137 and #3138: the .hookseen file after the first once-call, and stats / search / fixture listing before files --json. Will be closed unmerged.


vshulcz:
Two runs of this probe on windows-latest pass both tests unchanged apart from the t.Logf lines (34167931441, 34170249057), so the failures on main are not reproducible on demand. Closing unmerged; the evidence goes on #3137 and #3138.

PR #3312, opened 2026-09-08
fix(aider): the banner and a multi-line output block are not the assistant speaking

### LINKED ISSUE #3311: aider's multi-line output (the --verbose dump, a long commit message) is indexed as the assistant speaking, and titles the session
aider prefixes its own output with `> `, but only the first line of it: a multi-line block — the `--verbose` configuration dump, a multi-line commit message — continues with no prefix, and deja's reader takes those lines as the assistant speaking. A session that opens with `aider --verbose` is then titled by the dump.

Real history written by aider 0.86.2 (live run against a mock model, `/private/tmp/qa-aider2/proj/.aider.chat.history.md`), lines 93–107:

```
93 > Command Line Args:   --verbose --model openai/mock-1 --no-stream …
94   Config File (/private/tmp/qa-aider2/home/.aider.conf.yml):
95     read:              ['/private/tmp/qa-aider2/home/.config/deja/aider-context.md']
96   Defaults:
97     --set-env:         []
…
$ deja last 10
[aider · proj · 2026-09-08 · aider-9b764be90907-3] agent: Config File (/private/tmp/qa-aider2/home/.aider.conf.yml): r…
```

Control: the model's own reply is separated from the output block by a blank line and follows a `#### ` turn — lines 15–25 of the same file index correctly as user then assistant.

Cause: `ParseAiderFile` (`internal/sources/aider.go`) ends the current block on a `> ` line and treats any other non-empty line as assistant text; nothing marks the lines that continue the block just ended. Expected: a line that follows a `> ` line with no blank line between them is the rest of that output, not speech.

vshulcz:
Closes #3311.

Two rules, both from what aider actually writes: nothing before a session's first `#### ` turn is speech (the banner and the `--verbose` dump come before anyone has asked anything, and that dump has blank lines inside it), and a line directly under a `> ` line is the rest of that output block (a long commit message).

Fail-before test: `TestAiderOutputContinuationsAreNotSpeech` (internal/sources), on the collector: the dump and the commit tail come back as assistant turns.

On the real history a live aider 0.86.2 wrote: 7 sessions / 19 messages → 6 / 18, and the session titled `agent: Config File (/private/tmp/qa-aider2/home/.aider.conf.yml): r…` is gone — it held nothing but the dump. The other six keep their titles and their turns.

### DIFF
--- a/internal/sources/aider.go
+++ b/internal/sources/aider.go
@@ -109,6 +109,16 @@ func ParseAiderFile(path string) ([]model.Session, error) {
 	var buf []string
 	inFence := false
 	idx := 0
+	// aider marks its own output with "> " on the first line only: the
+	// --verbose configuration dump and a multi-line commit message continue
+	// unprefixed, and those lines read as the assistant speaking — one of them
+	// went on to title the session (#3311). Two rules cover what aider writes:
+	// nothing before the session's first `#### ` turn is speech (the banner
+	// and the dump come before anyone has asked anything, and the dump has
+	// blank lines inside it), and a line directly under a "> " line is the
+	// rest of that block.
+	afterOutput := false
+	seenUser := false
 
 	flush := func() {
 		if cur == nil || len(buf) == 0 {
@@ -151,6 +161,7 @@ func ParseAiderFile(path string) ([]model.Session, error) {
 			id := aiderSessionID(path, idx)
 			cur = &model.Session{Harness: "aider", ID: id, Project: project, Path: path, Started: ts, Updated: ts}
 			inFence = false
+			afterOutput, seenUser = false, false
 			continue
 		}
 		if cur == nil {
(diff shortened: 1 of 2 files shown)


vshulcz:
## Review

Verified against aider 0.86.2 source installed locally:

- `seenUser` is safe: aider's `run_one()` always writes a `#### ` line via `io.user_input()` before any reply, for `--message` runs and interactive turns alike.
- `afterOutput` matches how `tool_output()` writes multi-line text as one blockquote call (prefix on the first line only); no call site emits a blank line inside a continuation block after the first turn.

Re-parsed the real aider history before (`e2773b8f`) and after (`6ca1cdfe`): 7 sessions / 19 messages became 6 / 18, and the session that went away is exactly the `--verbose` dump that was misfiled as one assistant message. The other six are byte-for-byte identical, session ids included.

Latent gap it flagged, not a live bug: `internal/sources/aider.go:170-175` sets the role to assistant when a fence opens without consulting `afterOutput`/`seenUser`. No aider path fences tool output today; filed as a queue item for a defensive test.

PR #3314, opened 2026-09-08
fix(doctor): name the target that adds the hook, not an install that cannot

### LINKED ISSUE #3313: doctor tells an MCP-only install to "reinstall": the hooks come from the -auto target, so the advice cannot fix it
After the MCP-only install (`deja install crush`, `deja install qwen` — the variants without `-auto`), `deja doctor` reports the harness under Hooks as `stale … reinstall`, and running the same install again changes nothing: the hooks come from the `-auto` target, which was never asked for.

Hermetic HOME, nothing else installed:

```
$ deja install crush
crush: command created ~/.config/crush/commands/deja.md
crush: created ~/.config/crush/crush.json
$ deja doctor | grep -i crush
  crush        stale       ~/.config/crush/crush.json  (no hook-tool call — reinstall)
$ deja install qwen && deja doctor | grep '^  qwen'
  qwen         stale       ~/.qwen/settings.json  (no hook-prompt call — reinstall)
```

Control: after `deja install crush-auto` the same row reads `wired`; on a machine where the file does not exist at all the row reads `missing`, which is right.

Cause: `doctorAutoRecall` (`cmd/deja/doctor_auto.go:123`) prints `stale … reinstall` for any file that exists without the hook marker, and the MCP install creates exactly such a file (the same config holds the MCP server). The advice cannot fix it — only `deja install <name>-auto` writes the hook.

Expected: the row names the command that adds the hook. Whether "stale" is still the right word for a machine that never asked for the hooks is a wording question for the PR.

vshulcz:
Closes #3313.

The row that says a harness's config carries no hook call now names `deja install <name>-auto` instead of "reinstall" — the MCP install writes that same file, so re-running it changes nothing.

Fail-before test: `TestDoctorNamesTheAutoTargetForAnMCPOnlyInstall` (cmd/deja), on the collector: `(no hook-tool call — reinstall)`.

On a hermetic HOME after `deja install crush` and `deja install qwen`:

```
  crush        stale       ~/.config/crush/crush.json  (no hook-tool call — `deja install crush-auto`)
  qwen         stale       ~/.qwen/settings.json  (no hook-prompt call — `deja install qwen-auto`)
```

Wording question: the state is still called `stale`, which reads as "it went bad" where the machine simply never asked for the hooks. If you want a second word for that case (`mcp only`? `not wired`?), say which and I will split the state; the mechanism to tell them apart is there — the file exists and holds deja's MCP entry but no hook marker.

### DIFF
--- a/cmd/deja/doctor_auto.go
+++ b/cmd/deja/doctor_auto.go
@@ -120,7 +120,10 @@ func doctorAutoRecall(w io.Writer) {
 		case err != nil:
 			fmt.Fprintf(w, "  %-12s %-11s %s%s\n", a.name, "missing", reportPath(path), note)
 		case a.marker != "" && !strings.Contains(string(b), a.marker):
-			fmt.Fprintf(w, "  %-12s %-11s %s  (no %s call — reinstall)\n", a.name, "stale", reportPath(path), a.marker)
+			// "reinstall" was the advice, and for the common way to get here
+			// it cannot work: the MCP install writes this same file, and only
+			// the -auto target writes the hook (#3313). Name that target.
+			fmt.Fprintf(w, "  %-12s %-11s %s  (no %s call — `deja install %s-auto`)\n", a.name, "stale", reportPath(path), a.marker, a.name)
 		default:
 			fmt.Fprintf(w, "  %-12s %-11s %s%s\n", a.name, "wired", reportPath(path), note)
 			// Only the rows that run the binary. aider's file is a digest of
--- a/cmd/deja/doctor_auto_advice_test.go
+++ b/cmd/deja/doctor_auto_advice_test.go
@@ -1,0 +1,41 @@
+package main
+
+import (
+	"bytes"
+	"os"
+	"path/filepath"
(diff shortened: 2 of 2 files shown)


vshulcz:
## Review

- `cmd/deja/doctor_auto.go:126`: every row in `autoWirings()` that can reach the stale branch has a matching `<name>-auto` target in `installTargetNames()` (opencode, cursor, gemini, qwen, kimi, antigravity, pi, hermes, openclaw, cline, omp, amp, prime, deepseek, goose, crush, grok all present). aider and roo carry an empty marker, so the guard keeps them off this line. claude-code and codex are not in `autoWirings()` at all, so the Hooks rows a different function prints are untouched.
- Nothing asserts the old "reinstall" wording; the remaining hits are unrelated prose.
- The advice is appended after the last fixed-width field, so no column can misalign even at the longest name.
- `go test ./cmd/deja -run Doctor` passes (33 tests, the new one included) and `go vet ./cmd/deja` is clean.

Issue #3274, opened 2026-09-08
Continue's "New Session" and "Untitled Session" placeholders become session titles, and a greeting title stays


vshulcz:
Continue's own placeholder titles — `New Session` (core/util/constants.ts `NEW_SESSION_TITLE`) and `Untitled Session` (extensions/cli/src/constants/session.ts `DEFAULT_SESSION_TITLE`) — are taken as the session's title, and a session whose title is a greeting stays titled by it; a non-empty source title bypasses the index's rule (#790, #3251).

Hermetic store written in Continue's own session shape (no real `~/.continue` on this machine), `deja last`:

```
[continue · goprojects/deja-vu · 2026-09-08 · cccccccc-…] Untitled Session
[continue · goprojects/deja-vu · 2026-09-07 · dddddddd-…] New Session
[continue · goprojects/deja-vu · 2026-09-07 · bbbbbbbb-…] hi
[continue · goprojects/deja-vu · 2026-09-07 · aaaaaaaa-…] Explaining how ContinueRoot picks its directory
```

Control: the same sessions with a blank `title` are titled by the index — `fix the flaky TestParseContinueFile in internal/sources, it…` for the "hi" one.

Expected: a placeholder is no title, and a title that is just the first user line (a greeting included) is left to the index, which already knows to look past a thin opener. Actual: `ParseContinueFile` (`internal/sources/continuedev.go:116`, `:131`) copies `doc.Title` / the list entry's title as-is.

vshulcz:
fixed on main by #3275, pinned by `TestContinuePlaceholderTitleIsLeftToTheIndex`. it landed on the collector branch, so the merge in #3420 brought the fix in without closing this.

PR #3316, opened 2026-09-08
fix(opencode): a session is named by opencode's own title, unless that says less than its first line

### LINKED ISSUE #3315: opencode names every session and deja ignores it: 922 subagent runs are titled by their whole task brief
opencode names every session — `session.title` is filled on all 1472 sessions of this store — and deja's reader never selects the column, so a session is titled by its first user line instead. For the 922 subagent runs that line is the whole task brief, and the listing reads as a wall of "Read-only review of /var/folders/…".

Real store (`~/.local/share/opencode/opencode.db`, copied read-only):

```
$ sqlite3 opencode.db "select count(*), sum(title is not null and title<>'') from session"
1472|1472
$ sqlite3 opencode.db "select substr(title,1,44), substr(<first user line>,1,44) from session order by time_created desc limit 3"
Review route overlap fix (@code-reviewer sub|Read-only review of the current uncommitted
Review mini deployment safety (@code-reviewe|Read-only review of /var/folders/jn/5q9cbdfn
Проверить оставшиеся риски Widgetd (@rese|Read-only targeted source audit for a deploy
```

Not every title is worth taking, which is why this is not a plain "read the column": 28 of the 1472 are eight characters or less — `done` ×12, `Greeting` ×11, `ok` ×3 — and for those the first user line is the better name (`done` vs "render the plots for the report"). 1427 of 1472 are longer than twenty characters.

Control: the Continue reader had the same shape and #3274 settled it — take the harness's title unless it is a placeholder or a thin one, and let the index derive the rest.

Cause: the projection in `ParseOpencodeDBWhere` (`internal/sources/opencode.go:99`) selects `s.id, s.directory, s.time_created, s.time_updated` and never `s.title`.

vshulcz:
Closes #3315.

`opencodeTitles` reads `select id, title from session` beside the rows (best effort, like the parent query — a store from before the column stamps nothing and the harness keeps working) and stamps the name unless it says less than the first line would: opencode's `New session - <timestamp>` placeholder (31 of 1472 on the real store) or a name of two words or fewer and short (`done` ×12, `Greeting` ×11, `ok` ×3). The same rule Continue's placeholders get (#3274).

Fail-before test: `TestOpencodeTakesTheSessionTitleUnlessItIsThin` (internal/sources), on the collector: the named session comes back with an empty title.

On the hermetic copy of the real store, `deja last 5 --harness opencode`:

```
before: Read-only review of the current uncommitted Widgetd route overlap change
after:  Review route overlap fix (@code-reviewer subagent)
```

and the session opencode named `New session - 2026-07-23T…` keeps `почему не поднимается локальный прокси`.

A first cut put `s.title` in the main projection; two tests caught it — a fixture whose `session` table has no `title` column failed the whole query with "the store schema may have changed", which would take the harness down on an older store. Hence the separate query.

### DIFF
--- a/internal/sources/opencode.go
+++ b/internal/sources/opencode.go
@@ -290,6 +290,18 @@ func ParseOpencodeDBWhere(db, where string, limit int) ([]model.Session, error) 
 			}
 		}
 	}
+	// opencode names every session, and for the 922 subagent runs of one real
+	// store that name is the only short thing about them — the first user line
+	// there is the whole brief (#3315). Read beside the rows, not in the main
+	// projection: a store from before the column would fail the whole query
+	// and take the harness with it.
+	if titles := opencodeTitles(db); len(titles) > 0 {
+		for i := range out {
+			if t := titles[out[i].ID]; t != "" {
+				out[i].Title = t
+			}
+		}
+	}
 	return out, nil
 }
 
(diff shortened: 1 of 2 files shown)


vshulcz:
## Review

- A store whose `session` table has no `title` column: verified against a fixture database (session/message/part, no column) — the parse still returns the session, the main query is untouched, `opencodeTitles` returns nil.
- sqlite3 missing or failing for the title query is swallowed at `internal/sources/opencode.go:424-427`, the same best-effort shape `opencodeParents` already had.
- Titles with newlines, unicode or excess length are not capped in the reader, deliberately: every consumer already sanitises, `boundSourceTitle`/`truncateTitle` on the indexed path and `SafeForDisplay`/`SafeNoteTitle` on the no-index listing, which is what kimi, deepseek and goose do too. All 1472 titles on the real store are 100 runes or fewer anyway.
- The thin rule is byte-for-byte the existing `thinTitle` (two fields or fewer, twelve runes or fewer) plus the placeholder. Of the 59 it classifies thin on the real store, all are placeholders, greetings or one-word; scanning the other titles for path-like or command-like noise turned up nothing that reads worse than a first line would.
- Parents and titles are independent passes over the same slice, no shared state and no ordering conflict.
- `go test ./internal/sources/... ./cmd/deja/...` and `go vet ./...` clean.

One efficiency note, not a defect and not new: the parse now does two unfiltered reads of the session table instead of one, on every call including the search-as-you-type path. Queued.

Issue #3317, opened 2026-09-08
doctor's OpenClaw row cannot say a transcript went unread — 13 files, and a planted one changes nothing


vshulcz:
`deja doctor`'s OpenClaw row prints a file count and nothing else, so a transcript deja cannot read in an OpenClaw store is invisible — the one thing the row exists to rule out (#701). Every other file-listing harness reports it.

Reproduction, on a hermetic HOME holding a copy of a real OpenClaw store (13 transcripts under `agents/main/sessions`), with one transcript moved one directory deeper so the reader no longer recognises it:

```
$ cp -R ~/.openclaw $H/.openclaw
$ cp $H/.openclaw/agents/main/sessions/<id>.jsonl $H/.openclaw/agents/main/sessions/2026-09/moved.jsonl
$ HOME=$H deja doctor
  openclaw     found     ~/.openclaw/agents  (13 files)
```

The count does not move and nothing is said. The same shape in a harness that uses `printFiles` does say it — the control, same run, same planted file under Kimi's store:

```
  kimi         found     ~/.kimi-code/sessions  (5 files, 1 not recognised here)
  openclaw     found     ~/.openclaw/agents  (13 files)
```

Expected: `(13 files, 1 not recognised here)`.

Cause: `cmd/deja/doctor.go:801` calls `printRow` with a bare count, where continue (#3297), copilot (#3303) and kimi (#3309) call `printFilesBeside`. This is the fourth harness of the same shape.

The plain `printFiles` is not enough here, because an OpenClaw store keeps files beside the transcripts that the reader skips on purpose and that would then all read as failures: on this machine's store that is 13 `<id>.trajectory-path.json` and one `sessions.json` (14 of the 27 json files under `agents/`), plus the `.checkpoint.<uuid>.jsonl` snapshots and the `.jsonl.reset`/`.deleted` archives whose live file still exists (`internal/sources/openclaw.go:39-73`). Those belong in the `beside` list, the same way Continue's `sessions.json` does.

Follow-on from #3297, #3303, #3309.

vshulcz:
fixed on main by #3318, pinned by `TestDoctorNamesAnUnreadOpenClawTranscriptButNotItsSidecars`. it landed on the collector branch, so the merge in #3420 brought the fix in without closing this.

Issue #3309, opened 2026-09-08
doctor calls Kimi's state.json "not recognised here" once per session: 8 of 8, and the reader opens that file itself


vshulcz:
`deja doctor` reports every Kimi session as a file it could not read: Kimi keeps `state.json` beside each session's `agents/main/wire.jsonl`, and the reader opens it itself — `internal/sources/kimi.go:92` reads it for the session's title and working directory. The note that flags a store layout deja failed to understand counts each of them.

Real store, hermetic copy of `~/.kimi-code/sessions` (8 sessions):

```
$ deja doctor | grep '^  kimi '
  kimi         found     ~/.kimi-code/sessions  (8 files, 8 not recognised here, 7 indexed sessions)
$ find sessions -name '*.json' -o -name '*.jsonl' | sed 's|.*/||' | sort | uniq -c
   8 state.json
   8 wire.jsonl
```

The row also contradicts itself: every file unrecognised, yet the sessions are indexed.

Control: the same store without the state.json files prints `(8 files, 7 indexed sessions)` and the sessions still index (the title falls back to the first user line).

Cause: `KimiSessionFiles` (`internal/sources/kimi.go:32`) takes `agents/main/wire.jsonl` only, as it should, and `unplacedFiles` (`cmd/deja/doctor.go`) counts every other `.json` under the root as unread. Same shape as Continue's `sessions.json` (#3297) and Copilot's `vscode.metadata.json` (#3303), both fixed with `printFilesBeside`.

vshulcz:
fixed on main by #3310, pinned by `TestDoctorDoesNotCallKimisStateUnrecognised`. it landed on the collector branch, so the merge in #3420 brought the fix in without closing this.

Issue #3303, opened 2026-09-08
doctor calls Copilot's vscode.metadata.json "not recognised here" once per session: 15 of 15 on this store


vshulcz:
`deja doctor` reports every Copilot CLI session as a file it could not read: Copilot keeps `vscode.metadata.json` beside each session's `events.jsonl` (the IDE's bookkeeping — workspace, editor state), and the note that flags a layout deja failed to understand counts each of them.

Real store (`~/.copilot/session-state`, 15 sessions, copied read-only):

```
$ deja doctor | grep '^  copilot '
  copilot      found     ~/.copilot/session-state  (15 files, 15 not recognised here, 15 indexed sessions)
```

Control: the same store without the metadata files prints `(15 files, 15 indexed sessions)`; the reader indexes all 15 sessions either way.

Cause: `CopilotSessionFiles` (`internal/sources/copilot.go:18`) takes `events.jsonl` only, as it should, and `unplacedFiles` (`cmd/deja/doctor.go`) counts every other `.json` under the root as unread. Same shape as Continue's `sessions.json` (#3297): a file the harness keeps beside its transcripts on purpose is not a transcript deja missed.

vshulcz:
fixed on main by #3304, pinned by `TestDoctorDoesNotCallCopilotsMetadataUnrecognised`. it landed on the collector branch, so the merge in #3420 brought the fix in without closing this.

Issue #3297, opened 2026-09-08
doctor says Continue's sessions.json is "not recognised here" on every store: the list deja reads on purpose counts as a missed transcript


vshulcz:
`deja doctor` reports every Continue store as having one file it could not read: `sessions.json`, the list Continue keeps beside the session documents, which the reader opens for titles and dates and skips as a transcript on purpose (`continueSessionFile`, `internal/sources/continuedev.go:45`).

Hermetic store in Continue's own shape (4 sessions + sessions.json):

```
$ deja doctor | grep continue
  continue     found     ~/.continue/sessions  (4 files, 1 not recognised here, 4 indexed sessions)
```

Control: the same store without sessions.json prints `(4 files, 4 indexed sessions)`; a Qwen store with every file placed prints no note (doctor_unplaced_test.go).

Cause: the continue row goes through `printFiles`, and `unplacedFiles` (`cmd/deja/doctor.go:422`) counts every `.json` under the root that the reader's file list does not contain; the list file is not in that list because it is not a session. Expected: a file the reader reads by design is not "not recognised here" — the note exists to flag a layout deja failed to understand (#701), and here it flags the one file deja understands best.

vshulcz:
fixed on main by #3298, pinned by `TestDoctorDoesNotCallContinuesListUnrecognised`. it landed on the collector branch, so the merge in #3420 brought the fix in without closing this.

PR #3320, opened 2026-09-08
fix(doctor): stop calling Grok's own bookkeeping transcripts deja could not read

### LINKED ISSUE #3319: doctor says 94 files not recognised in a Grok store with 11 transcripts — all 94 are Grok's own bookkeeping
`deja doctor` reports 94 files it could not recognise in a Grok store that holds 11 transcripts. All 94 are Grok's own bookkeeping, and the number is the loudest thing in the report — it reads as a parser that cannot cope with the store, which is the reading #701 warns about.

On this machine's real store, read-only:

```
$ deja doctor
  grok         found     ~/.grok  (11 files, 94 not recognised here, 11 indexed sessions)
```

Every one of the 94 is a file Grok writes itself and deja never wanted (`find ~/.grok -type f -name '*.json*' | xargs -n1 basename | sort | uniq -c`):

```
  13 summary.json          11 rewind_points.jsonl     11 announcement_state.json
  11 chat_history.jsonl    11 prompt_context.json      8 signals.json
  11 events.jsonl           6 resources_state.json     1 prompt_history.jsonl
```

plus eleven configuration files at the store root (`settings.json`, `user-settings.json`, `auth.json`, `registry.json`, `manifest.json`, `models_cache.json`, `template_taxonomy.json`, `active_sessions.json`, `deja.json`, `unified.jsonl`, one `memtrace/<n>-<n>.jsonl`). The reader takes `updates.jsonl` as the transcript and opens `summary.json` itself for the metadata (`internal/sources/grok.go:37-41`); nothing else in the store is a transcript.

Counter-case that rules out a mis-set root: the same run reports `kimi found ~/.kimi-code/sessions (5 files)` with no note, because Kimi's per-session `state.json` was given to `printFilesBeside` in #3309. Grok's row is a plain `printFiles` (`cmd/deja/doctor.go:766`), and its root is the whole `~/.grok` rather than the sessions directory, so the configuration files are counted as transcripts too.

Expected: `(11 files)` on an unchanged store, and a number only when there really is a transcript deja did not read.

Follow-on from #3297, #3303, #3309, #3317.

vshulcz:
Closes #3319.

`GrokSidecarFiles` names what a Grok store keeps beside the transcript — the `summary.json` the reader opens itself plus the per-session bookkeeping (`chat_history.jsonl`, `events.jsonl`, `rewind_points.jsonl`, `prompt_context.json`, `announcement_state.json`, `signals.json`, `resources_state.json`, `prompt_history.jsonl`) — and the row moves to `printFilesBeside`, as Continue (#3297), Copilot (#3303), Kimi (#3309) and OpenClaw (#3317) did.

The count also moves from the whole `~/.grok` to `~/.grok/sessions`: the root holds settings, credentials and caches, which are not transcripts and were eleven of the 94. That changes the location the row prints, the same way Kimi's says `~/.kimi-code/sessions` and Continue's `~/.continue/sessions` — flagging it in case you want the row to keep saying `~/.grok`.

Fail-before test: `TestDoctorDoesNotCallGroksOwnFilesUnrecognised` (cmd/deja), on the collector: `(1 file, 13 not recognised here)`, want no such note. Its second half plants an `updates.v2.jsonl` and requires the note to come back as `1 not recognised here`, so the row still does what it exists for.

Real store, read-only:

```
before:  grok  found  ~/.grok           (11 files, 94 not recognised here, 11 indexed sessions)
after:   grok  found  ~/.grok/sessions  (11 files, 11 indexed sessions)
```

### DIFF
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -762,8 +762,11 @@ func doctorHarnesses(w io.Writer, dir string) {
 
 	printRow("antigravity", doctorAntigravityLocation(), len(sources.AntigravityRoots()) > 0, doctorCount(len(sources.AntigravityTranscripts()), "file"))
 
-	grokRoot := sources.GrokRoot()
-	printFiles("grok", grokRoot, doctorExists(grokRoot), sources.GrokSessionFiles())
+	// The store root also holds Grok's settings, credentials and caches, which
+	// are not transcripts and never will be; the sessions directory is what the
+	// count is about (#3319).
+	grokRoot := filepath.Join(sources.GrokRoot(), "sessions")
+	printFilesBeside("grok", grokRoot, doctorExists(grokRoot), sources.GrokSessionFiles(), sources.GrokSidecarFiles()...)
 
 	qwenRoot := filepath.Join(sources.QwenRoot(), "projects")
 	printFiles("qwen", qwenRoot, doctorExists(qwenRoot), sources.QwenSessionFiles())
--- a/cmd/deja/doctor_grok_files_test.go
+++ b/cmd/deja/doctor_grok_files_test.go
@@ -72,3 +72,32 @@ func TestDoctorDoesNotCallGroksOwnFilesUnrecognised(t *testing.T) {
 		}
 	}
 }
+
+// The row and the JSON name the same place. doctorStoreChecks keeps its own
+// path per harness, and leaving grok's at the store root made `doctor --json`
+// say ~/.grok where the row says ~/.grok/sessions — and, on a store deja is
+// not allowed to read, turned "denied" into "missing" (review of #3319).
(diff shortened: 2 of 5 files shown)


vshulcz:
## Review

- A planted `updates.v2.jsonl` moves the row from `(11 files)` to `(11 files, 1 not recognised here)`, and a `prompt_context.json` two levels deeper is still matched by name, as every other harness's sidecar list matches.
- A fresh install with no sessions directory reads `missing ~/.grok/sessions`; the guides already name that path and nothing asserts the bare root.
- A sessions directory deja may not read now reports `denied` in both the row and `doctor --json`, which is what the third commit was for.
- Every json and jsonl basename under the real `~/.grok/sessions` is in the sidecar list, nothing left counted; the row's count agrees with `deja sources`.
- `go test ./cmd/deja/... ./internal/sources/...` and `go vet ./...` clean.

PR #3322, opened 2026-09-08
fix(doctor): count what Codex left unread in sessions, not its caches

### LINKED ISSUE #3321: doctor says 29 files not recognised in a Codex store — the plugin cache, auth.json and hooks.json
`deja doctor` tells a Codex user that 29 files in their store were not recognised. All 29 are Codex's own configuration and caches; none of them is a transcript.

On this machine's real store, read-only:

```
$ deja doctor
  codex        found     ~/.codex  (35 files, 29 not recognised here, 30 indexed sessions)
```

The 35 read are the 34 `sessions/**/rollout-*.jsonl` and `history.jsonl` (`internal/sources/codex.go:58-65`). The 29 are:

```
  23  plugins/cache/openai-curated-remote/**/*.json   the plugin cache
   3  cache/codex_app*                                 the app-server caches
   1  version.json     1  models_cache.json     1  hooks.json     1  auth.json
```

The row walks the whole `~/.codex`, so everything the CLI keeps beside the sessions is counted as a transcript deja failed to read. `.tmp/plugins` (452 files) is already excluded because `unplacedFiles` skips dot directories — that exclusion was added for exactly this reading (#701), and `plugins/cache` is the same thing without the dot.

Counter-case ruling out a mis-set root: the same run shows the transcripts are found and read — `30 indexed sessions` — and `kimi found ~/.kimi-code/sessions (5 files)` reports no such note because its row walks the sessions directory.

Expected: `(35 files)` on an unchanged store, and a number only when a rollout in `sessions/` really went unread. The row should keep naming `~/.codex`, since `history.jsonl` at the root is one of the files it counts.

Follow-on from #3297, #3303, #3309, #3317, #3319.

vshulcz:
Closes #3321.

`printFilesUnder` is `printFilesSkipping` with the walked subtree given separately from the path the row names, so Codex's row keeps saying `~/.codex` — `history.jsonl` at the root is one of the files it counts — while the unread number comes from `sessions` alone. Every other caller goes through the old signature unchanged.

Fail-before test: `TestDoctorDoesNotCallCodexCachesUnrecognised` (cmd/deja), on the collector: `(2 files, 5 not recognised here)`, want no such note. Its second half plants `rollout.v2.jsonl` inside the sessions tree and requires `1 not recognised here` to come back.

Real store, read-only:

```
before:  codex  found  ~/.codex  (35 files, 29 not recognised here, 30 indexed sessions)
after:   codex  found  ~/.codex  (35 files, 30 indexed sessions)
```

The 29 were the plugin cache (23), the app-server caches (3), `version.json`, `models_cache.json`, `hooks.json` and `auth.json`.

### DIFF
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -704,13 +704,9 @@ func doctorHarnesses(w io.Writer, dir string) {
 	// — the one thing `doctor` exists to rule out (#701).
 	// printFilesSkipping is printFiles for a harness that declines some of its
 	// own files by a rule, so the two can be told apart in the row.
-	// printFilesUnder is printFilesSkipping for a store whose transcripts live
-	// in one subtree while the row names the store: Codex keeps its plugin
-	// cache, model cache, hooks and credentials beside `sessions`, and walking
-	// the root reported all of it as transcripts deja could not read (#3321).
-	printFilesUnder := func(name, path, walk string, present bool, seen []string, skipped func(string) bool) {
+	printFilesSkipping := func(name, path string, present bool, seen []string, skipped func(string) bool) {
 		detail := doctorCount(len(seen), "file")
-		unread, byRule := unplacedFiles(walk, seen, skipped)
+		unread, byRule := unplacedFiles(path, seen, skipped)
 		if byRule > 0 {
 			// The variable named the way it was read as the cause of the
 			// skip — "skipped (DEJA_INCLUDE_SUBAGENTS=1)" — so somebody who
@@ -726,9 +722,6 @@ func doctorHarnesses(w io.Writer, dir string) {
 			detail += fmt.Sprintf(", %d not recognised here", unread)
 		}
 		printRow(name, path, present, detail)
-	}
-	printFilesSkipping := func(name, path string, present bool, seen []string, skipped func(string) bool) {
-		printFilesUnder(name, path, path, present, seen, skipped)
 	}
 	printFiles := func(name, path string, present bool, seen []string) {
 		printFilesSkipping(name, path, present, seen, nil)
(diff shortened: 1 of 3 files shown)


vshulcz:
## Review

It found that narrowing the walk to `sessions/` stopped the row from seeing a transcript in the wrong place: a rollout dropped at the store root read `(35 files, 1 not recognised here)` on the collector and `(35 files)` on the branch. Codex writes rollouts only under `sessions/` today, so nothing on a real store changes either way, but that file is exactly what the row exists to name.

So the walk stays over the whole store and `CodexSidecarFiles` names what is not a transcript: anything under `plugins/` or `cache/`, and the `.json` settings at the root. A rollout anywhere, or any file the reader does not take, is still counted. `printFilesUnder` is gone with it; the codex row is a `printFilesBeside` like Continue, Copilot, Kimi, OpenClaw and Grok.

The test now plants two strays — one inside the sessions tree, one at the store root — and requires both in the count.

Real store, read-only, unchanged from the first measurement:

```
before:  codex  found  ~/.codex  (35 files, 29 not recognised here, 30 indexed sessions)
after:   codex  found  ~/.codex  (35 files, 30 indexed sessions)
```

vshulcz:
## Review

cavecrew reviewer, adversarial pass — not refuted, and both design gaps it named are closed in the follow-up commit.

What it verified: the stray `rollout-*.jsonl` at the store root is counted again on the branch and is the only thing in the number; a misplaced rollout under `cache/` or `plugins/` is counted too; `history.jsonl` stays in the count and out of the unread number; the numbers agree with `deja sources` and a rebuild; `.tmp/` is neither resurrected nor double-counted.

The two gaps, both about the list being too wide:

- a non-rollout `.jsonl` under `cache/` or `plugins/` disappeared. Those subtrees hold json the CLI wrote for itself; a `.jsonl` there is not configuration, so the rule now keeps only `.json` under them.
- any `.json` at the store root disappeared. The four names deja knows are now named — `version.json`, `models_cache.json`, `hooks.json`, `auth.json`, plus `config.json` — and anything else arriving there is counted, because a `.json` under a name deja does not know may be a transcript in a format it cannot read yet.

The test plants four strays now: one inside the sessions tree, one at the root, a `queue.jsonl` under the cache and a `newformat-transcript.json` at the root, and requires all four in the number.

Real store, unchanged: `(35 files, 30 indexed sessions)` with no unread note.

Issue #3156, opened 2026-09-07
hook-prompt runs on the harness's own <task-notification> turns and says "you have been here" about them


vshulcz:
Seen in a Claude Code session on 2026-09-08. A background task finished and Claude Code delivered its `<task-notification>` block as the next user turn; the per-prompt hook ran on it and said, in the status line:

```
deja-vu — you have been here: "Summary: 1. Primary Request and Intent: - MOST R…" (Aug 11 · via: task-notification, task-id, br50mykp6)
```

Nobody asked anything. The "prompt" was the harness's own notification envelope, its identifying terms were the envelope's field names (`task-notification`, `task-id`) plus a task id that exists nowhere else, and the match was another notification in an August session. The block went into the model's context as recalled history, and the line taught the user that the hook fires on noise.

`digest.IsAgentArtifact` already knows `<task-notification`, `<teammate-message`, `<system-reminder`, `<local-command` and the rest — it is applied to session messages when a title is picked (`hook_prompt.go:691`, `:1225`), but never to `input.Prompt` itself. `promptTermsWorthAsking` only counts terms.

Expected: `hook-prompt` stands down (emits the nudge only) when the prompt is a harness artifact by the same test the digest uses, and the terms extractor treats a bare id token (`br50mykp6`, `toolu_…`, UUID fragments) as non-identifying. Control: a real question in the same session still gets its block.

Repro: `printf '{"session_id":"x","cwd":"%s","prompt":"<task-notification>\\n<task-id>br50mykp6</task-id>\\n<status>failed</status>\\n</task-notification>"}' "$PWD" | deja hook-prompt` — on this machine it returns a block and a "you have been here" line; it should return nothing.

PR #3325, opened 2026-09-08
fix(ctx): the context is what people said, not the host's own turns

### LINKED ISSUE #3323: deja ctx spends its 8 KB budget on teammate messages and idle notifications — 38% of ten real answers
`deja ctx` keeps every user turn regardless of what it holds, so a session whose transcript carries the harness's own envelopes — Claude Code's `<teammate-message>`, task notifications, idle-notification JSON — spends the 8 KB context budget on them and never reaches the turns that answer the query.

Reproduction, on a read-only copy of this machine's index (taken 12:07 UTC), ten questions worded from the ten newest merged PRs:

```
$ deja ctx "follow HERMES_HOME, the variable Hermes itself reads"
# deja context: claude · goprojects/deja-vu · 59eff8d3-… · updated 2026-09-08

## user

Another Claude session sent a message:
<teammate-message teammate_id="rev354" color="blue" summary="Review findings for the damaged-index JSON fix">
Review of c3d9c954 is complete. Answers to the five points:
…
<teammate-message teammate_id="rev354" color="blue">
{"type":"idle_notification","from":"rev354","timestamp":"2026-08-28T05:14:04.179Z","idleReason":"available"}
</teammate-message>
```

The answer is 8287 bytes, of which 7432 (89%) are inside harness envelopes, and none of it is about `HERMES_HOME`. Across the ten questions: 67649 bytes served, 26360 (38%) inside envelopes, three answers 89–90% envelope. The blocks served include twelve `idle_notification` payloads, which are one agent telling another it is free.

Cause: `internal/search/search.go:1531` keeps a turn when `carries || m.Role == "user"`, so a user turn that is nothing but the host talking to itself is kept ahead of the turns that carry the query. The tags are already known — `digest.StripHarnessBlocks` (`internal/digest/harness_blocks.go:18-28`) lists `teammate-message`, `task-notification` and the rest, and the prompt hook has used it since #3156. The fallback branch at `search.go:1553` keeps everything, so an overview has the same shape.

Counter-case ruling out "the session simply has nothing on the subject": the same query through `deja search` returns real work on HERMES_HOME, and the seven answers with no envelope in them read as ordinary transcript.

Expected: a user turn that is only the harness's envelope is not context; the budget goes to the turns that hold the person's words.

Follow-on from #3156.

vshulcz:
Closes #3323.

`withoutHarnessEnvelopes` takes the host's blocks off the user turns before the window is chosen, using the same list the prompt hook has used since #3156. A turn that was nothing but the envelope is dropped; a turn with the person's words beside it keeps the words, and the window then scores against what was said rather than against the envelope's field names.

Fail-before test: `TestContextSkipsTurnsThatAreOnlyTheHostTalking` and `TestContextOverviewSkipsTheHostsOwnTurns` (internal/search), on the collector: an idle notification and a teammate envelope come back in the context handed to another agent. `TestContextKeepsATurnThatHasWordsBesideTheEnvelope` passes before and after and holds the other side.

On a read-only copy of this machine's index, the ten questions from the issue:

```
before:  67649 bytes served, 26360 in envelopes (38%), 12 idle notifications
after:   69608 bytes served,     0 in envelopes  (0%),  0
```

The output grows slightly because the budget now buys transcript instead of envelopes.

Only user turns are stripped. An assistant turn or a tool record quoting one of these tags keeps it, since there the tag is something being discussed rather than the host talking.

### DIFF
--- a/internal/digest/harness_blocks.go
+++ b/internal/digest/harness_blocks.go
@@ -55,28 +55,36 @@ var harnessBlockRe = func() *regexp.Regexp {
 	return regexp.MustCompile(`(?is)<(` + strings.Join(alts, "|") + `)(?:\s[^>]*)?>(?:.*?</\s*` + "(?:" + strings.Join(alts, "|") + `)\s*>|.*$)`)
 }()
 
-// closedHarnessBlockRe is harnessBlockRe without the run-to-the-end arm: only
-// a block that is closed. An unclosed tag is not always the host — a person
-// asking "why does <system-reminder> show up in my prompt?" opens one and
-// never closes it, and taking everything after it would take the question
-// (review of #3323).
-var closedHarnessBlockRe = func() *regexp.Regexp {
-	alts := make([]string, 0, len(harnessBlockTags))
+// closedHarnessBlockREs is one regexp per tag: a block that opens and closes
+// with the same name. harnessBlockRe closes on any listed name, which is what
+// a nested Cursor block needs, but on a turn where two different envelopes
+// stand apart it swallowed the sentence between them (review of #3323).
+var closedHarnessBlockREs = func() []*regexp.Regexp {
+	out := make([]*regexp.Regexp, 0, len(harnessBlockTags))
 	for _, t := range harnessBlockTags {
-		alts = append(alts, regexp.QuoteMeta(t))
+		q := regexp.QuoteMeta(t)
+		out = append(out, regexp.MustCompile(`(?is)<`+q+`(?:\s[^>]*)?>.*?</\s*`+q+`\s*>`))
 	}
-	return regexp.MustCompile(`(?is)<(` + strings.Join(alts, "|") + `)(?:\s[^>]*)?>.*?</\s*(?:` + strings.Join(alts, "|") + `)\s*>`)
+	return out
 }()
 
-// StripClosedHarnessBlocks removes the envelopes a harness closed and leaves
-// everything else as it stands. The prompt hook wants the stricter rule — a
-// truncated notification is still not the person — but a surface that prints
-// the turn back must not lose words to a tag somebody typed themselves.
+// StripClosedHarnessBlocks removes the envelopes a harness opened and closed
+// under one name, and leaves everything else as it stands. The prompt hook
+// wants the stricter rule — a truncated notification is still not the person —
+// but a surface that prints the turn back must not lose words to a tag
+// somebody typed themselves.
 func StripClosedHarnessBlocks(text string) string {
 	if !strings.Contains(text, "<") {
 		return strings.TrimSpace(text)
 	}
-	return strings.TrimSpace(closedHarnessBlockRe.ReplaceAllString(text, ""))
+	for _, re := range closedHarnessBlockREs {
+		text = re.ReplaceAllString(text, "")
+	}
+	// What a nested block leaves behind, and Cursor's wrapper around the words
+	// a person typed: the tags go, the words stay.
+	text = orphanCloseRe.ReplaceAllString(text, "")
+	text = userQueryTagRe.ReplaceAllString(text, "")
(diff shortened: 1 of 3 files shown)


vshulcz:
## Review

What it found, both on the shared `harnessBlockRe`:

- An unclosed tag runs to the end of the text, so `"<environment_context>\n  <cwd>/repo</cwd>\nCan you fix the failing test in parser_test.go?"` stripped to nothing and the turn was dropped, question included.
- A person naming a tag in their own question opens one and never closes it: `"why does <system-reminder> show up in my prompt when I did not add it?"` stripped to `"why does"`.

Both are the prompt hook's rule, which is right there — a truncated notification is still not the person — and wrong for a surface that prints the turn back. `StripClosedHarnessBlocks` now takes only the blocks a harness closed and leaves everything else as it stands; `PrintContext` uses that one. `TestContextKeepsTheQuestionUnderAnUnclosedTag` covers both cases.

The measurement is unchanged on this machine's index: the same ten questions go from 38% envelope to none, twelve idle notifications to none.

Also checked and clean in the review: `deja show` does not go through `PrintContext`, no caller promises a verbatim transcript there, and nothing downstream hashes the printed text.

vshulcz:
## Review, second pass

Refuted again, and fixed in the follow-up commit.

- `<meta>…</deja-recall>` closed across two different names, so a sentence standing between two envelopes was eaten: `"why did the deploy fail last night?"` vanished. Each tag now has its own regexp and a block closes only under the name it opened with.
- The closed-block strip never ran the orphan-close and `<user_query>` cleanups, so `</additional_data>` leaked into the context and Cursor's wrapper printed its tags around the person's words. Both run now.

`TestContextKeepsWordsBetweenTwoEnvelopes` covers all three and fails on the previous commit with the bare tag in the output.

Measured again on a copy of the real index taken at 13:08 UTC, the same ten questions: 69621 bytes, no envelopes, no idle notifications, no orphan closing tags.

vshulcz:
## Review, third pass

Refuted, and the first finding is fixed.

- `<user_query>` was stripped as two bare tags wherever they appeared, so a sentence naming Cursor's wrappers came back gutted — the reviewer found it in real data, `"…wraps around a user message: `<user_query>`, `<additional_data>`…"` losing the first one. It is now unwrapped as a pair, which leaves a lone mention alone. `TestContextKeepsASentenceThatNamesTheWrapper` covers it and fails on the previous commit.

- The second finding, text between a nested block's close and the outer envelope's close: left as it is, deliberately. A sentence written inside `<additional_data>` is inside the host's envelope, and the whole point of the per-tag rule is that a block closes under its own name — the old cross-name behaviour kept that text by accident, because it stopped at whichever tag closed first. Two envelopes standing apart still keep the words between them, which is the case that occurs in real transcripts and is covered by `TestContextKeepsWordsBetweenTwoEnvelopes`.

- The third, that dropping envelope-only turns changes what the fallback window anchors on: yes, and that is the point — those turns were the anchor before, which is why ten real questions came back 38% envelope. Re-measured now on a copy of the real index: 69386 bytes, no envelopes, no idle notifications, no bare wrapper tags.

vshulcz:
## Review, fourth pass

Refuted, and fixed.

The reviewer confirmed the sentence naming the wrappers survives now, and agreed with the two answers. Its new finding: the wrapper was unwrapped by a regexp running from the first open to the first close, so a turn that opens one wrapper, never closes it, and closes a later one paired the wrong two — the stray tag ended up sitting against the question.

`unwrapUserQuery` now pairs each close with the nearest open before it. The closed pair unwraps; an open with no close stays where it is, which is the rule every other tag gets here — a person naming the tag keeps their sentence.

`TestContextPairsEachWrapperWithTheNearestOpen` covers it and fails on the previous commit with the tag against the question.

PR #3327, opened 2026-09-08
fix(antigravity): approving a plan is the IDE's record, not something the person said

### LINKED ISSUE #3326: an Antigravity plan approval is indexed as the person's words: "Comments on artifact URI … has approved this document"
When someone approves a plan in Antigravity, the IDE writes a `USER_INPUT` step whose `<USER_REQUEST>` is empty — the person typed nothing — and deja indexes the IDE's own text as their words, raw tag and absolute path included.

The record, from this machine's real store (`~/.gemini/antigravity/brain/f017fad5-…/.system_generated/logs/transcript.jsonl`, line 19):

```json
{"step_index":13,"source":"USER_EXPLICIT","type":"USER_INPUT","status":"DONE",
 "content":"Comments on artifact URI: file:///Users/shulcz/.gemini/antigravity/brain/f017fad5-…/implementation_plan.md\n\nThe user has approved this document.\n\n\n<USER_REQUEST>\n\n</USER_REQUEST>\n<ADDITIONAL_METADATA>\nThe current local time is: 2026-07-08T17:45:28+03:00.\n</ADDITIONAL_METADATA>"}
```

Indexed hermetically (a copy of the real store, 8 sessions, 580 messages) and searched:

```
$ deja "approved this document"
[antigravity] -  · Jul 8 · f017fad5-…3b7ce6844a — 2 matches
  Comments on artifact URI: file:///Users/shulcz/.gemini/antigravity/brain/f017fad5-…/implementation_plan.md The user has approved this document. <USER_REQUEST>
[antigravity] -  · Jun 23 · 6d5cee10-…5af794b21b — 2 matches
  Comments on artifact URI: file:///Users/shulcz/.gemini/antigravity/brain/6d5cee10-…/implementation_plan.md The user has approved this document. <USER_REQUEST>
```

Two of the 67 `USER_INPUT` records across the 50 real transcripts have this shape, and both have an empty `<USER_REQUEST>`.

Cause: `cleanAntigravityUserContent` (`internal/sources/antigravity.go:115-122`) strips a `<USER_REQUEST>` tag only when it opens the content — the regexps are anchored — so an approval turn keeps the preamble, keeps the opening tag, and is indexed as a user message. deja's own hook already reads this shape correctly: `userRequestText` (`cmd/deja/hook_antigravity_prompt.go:95-105`) takes what is inside the tag wherever it stands, which is why the hook does not answer an approval turn.

Counter-case: the other 65 `USER_INPUT` records open with `<USER_REQUEST>` and index correctly, so this is the shape, not the reader.

Expected: what is inside `<USER_REQUEST>` is the person's turn; an approval whose request is empty is not a turn at all and should not be indexed as one.

Follow-on from #3156.

vshulcz:
Closes #3326.

`cleanAntigravityUserContent` now takes what is inside `<USER_REQUEST>` wherever the tag stands, instead of only when it opens the content, and then removes the IDE's blocks — they appear both inside the tag and beside it, which the existing fixture covers. An approval whose request is empty leaves nothing and the caller already drops such a turn.

This is the rule deja's own Antigravity hook has always used (`userRequestText`, cmd/deja/hook_antigravity_prompt.go), which is why the hook never answered an approval turn while the index held it as a question.

Fail-before test: `TestAntigravitySkipsAPlanApprovalWithNoRequestInIt` (internal/sources), on the collector: `user turns = 3, want 2`, with the second turn being `Comments on artifact URI: … The user has approved this document. <USER_REQUEST>`.

Real store, hermetic copy, `deja index --rebuild`:

```
before: antigravity: 8 sessions, 580 messages   deja "approved this document" → two sessions whose top snippet is the IDE's comment
after:  antigravity: 8 sessions, 578 messages   → the only hit left is the agent's own sentence about the plan
```

The index version note names this reader change (and the two merged before it on this branch, #3311 and #3315); the version itself is already 35 on the collector, so the store is re-read once either way.

### DIFF
--- a/internal/index/index.go
+++ b/internal/index/index.go
@@ -124,7 +124,9 @@ import (
 // (#3281), Grok's tool input (#3285), Antigravity's edits (#3279), Zed's
 // tool results (#3291), Gemini's toolCalls (#3293), Roo and legacy Cline's
 // tool calls (#3295), opencode's synthetic parts (#3299) and subagent parents
-// (#3301), Copilot's injected skills (#3305). A store
+// (#3301), Copilot's injected skills (#3305), aider's banner and its
+// continuation lines (#3311), opencode's own session titles (#3315),
+// Antigravity's plan approvals (#3326). A store
 // built before holds skill bodies as the person's words and none of the new
 // tool output, and nothing re-reads a source without the bump (#3289).
 const version = 35
--- a/internal/sources/antigravity.go
+++ b/internal/sources/antigravity.go
@@ -12,9 +12,11 @@ import (
 	"github.com/vshulcz/deja-vu/internal/model"
 )
 
-const (
-	antigravityRequestOpen  = "<USER_REQUEST>"
-	antigravityRequestClose = "</USER_REQUEST>"
+var (
+	antigravityRequestTagRE = regexp.MustCompile(`[ \t]*\n?\s*</?USER_REQUEST>\s*`)
+	// "Comments on artifact URI: file:///…/implementation_plan.md" and the
+	// sentence under it, which is the IDE recording what the person clicked.
+	antigravityArtifactCommentRE = regexp.MustCompile(`(?m)^Comments on artifact URI:.*$|^The user has (?:approved|rejected) this document\.$`)
 )
 
 var antigravityUserBlockREs = []*regexp.Regexp{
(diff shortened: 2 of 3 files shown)


vshulcz:
## Review

cavecrew reviewer, adversarial pass — not refuted on the claim, two constructed edge cases found and fixed in the follow-up commit.

Verified on a hermetic copy of the real store, built from the collector and from the branch: user turns 67 → 65, and a diff of every user message text shows the only two removed are the approval records with the empty request tag. Nothing else changed. The `antigravity-cli` transcripts share the shape and the same reader path. The removed regexps have no references left. `go test ./...` and `go vet ./...` clean. On the index version: 35 has never shipped (the last tag carries 34), so nothing is skipping a re-read.

What it found, both hand-built and absent from the real store: text a person writes before the tag was dropped, and a turn with two request blocks kept only the first.

Both fixed: the tags now come off wherever they stand and whatever is around them stays, while the IDE's own approval comment — `Comments on artifact URI: …` and `The user has approved this document.` — is what gets removed. `TestAntigravityKeepsWhatIsWrittenAroundTheRequestTag` covers a line above the tag, two blocks with words between them, and the approval turn still leaving nothing.

PR #3330, opened 2026-09-08
fix(index): a one-word name from the store gives way to the question under it

### LINKED ISSUE #3328: 43 of 60 sessions from three real stores are titled "ok", "47" or "say hi" while the question is one line below
A title the reader takes from the store is used as it stands, so a harness that names a session with one word leaves that word in every listing while the person's question sits one line below. The greeting rule (#790) applies only when the reader supplies no title at all.

On this machine's real stores, indexed hermetically (cline 15 sessions, deepseek 43, goose 2), 43 of the 60 titles are two words or fewer and twelve runes or fewer:

```
$ deja last 100
[deepseek · dshproj · 2026-09-02 · 28341f5a-…] ok
[deepseek · pin-manifests · 2026-08-23 · 29902c0d-…] 47
…
 23  deepseek  'ok'
 15  deepseek  '47'
  2  goose     'say ok'
  1  deepseek  'say ok'
  1  cline     'say hello'
  1  cline     'say hi'
```

What those sessions actually hold:

```
$ deja show 28341f5a
2026-09-02 20:23 user:      what did we settle about the checkout worker dropping connections
2026-09-02 20:23 assistant: ok

$ deja show 29902c0d
2026-08-23 03:12 user:      what did we decide about longmemeval?
2026-08-23 03:12 assistant: 47
```

dsh names a session from its own title model (`internal/sources/deepseek.go:118-128`, the `session/title` event), and when that model answers with the reply rather than a summary the name is the reply. The person's question is right there and is what the reader would search for.

Counter-case: the goose row titled `say ok` is honest — that session's only user turn is `say ok`, so there is nothing better to name it, and it must keep the name it has.

Cause: `metaForSession` (`internal/index/ingest.go:1436-1464`) applies `sessionTitleFrom` — which drops a thin opening for the next substantial turn — only in the `title == ""` branch. A title the source authored skips the rule. The same judgement was made for one reader in #3315; every reader that sets `Title` itself has the shape: cline (`man.Prompt`, `m.Task`), roo (`item.Task`), continue (`entry.Title`), goose (`desc`), deepseek.

Expected: a source title too thin to name anything gives way to the first user turn that can, and stands as it is when there is none.

vshulcz:
Closes #3328.

`metaForSession` now applies the thin-title rule to a title the reader took from the store, not only to a derived one: a name of two words or fewer and twelve runes or fewer gives way to the first user turn that can name the session, and stands as it is when there is none. Notes keep their own names, as they do for the bound above.

Fail-before test: `TestAThinSourceTitleGivesWayToTheQuestion` (internal/index), on the collector: `title = "ok", want the person's question`. Two more hold the other side — a session whose only turn is `say ok` keeps that name, and a short title that does name the work (`cap the retries`) is not touched.

Three real stores on this machine, hermetic copy, `deja index --rebuild`:

```
before: 43 of 60 titles two words or fewer — 23 "ok", 15 "47", 4 "say ok", "say hello", "say hi"
after:   8 of 60, and every one of those eight is a session whose user turn really is that short
```

The eight are the honest ones: `deja show` on each shows a single turn saying `say ok` or `47`.

This is the shape #3315 fixed inside one reader; the rule now lives where the derived titles already had it, so cline, roo, continue, goose and deepseek get it without their own change.

### DIFF
--- a/internal/index/ingest.go
+++ b/internal/index/ingest.go
@@ -1460,18 +1460,7 @@ func metaForSession(s model.Session) SessionMeta {
 		// which is deja's own listing format (#1090 covers the escape bytes;
 		// this is the line break). Derived titles have been collapsed and cut
 		// since they existed.
-		title = boundSourceTitle(s.Harness, title)
-		// A name too short to name anything gives way to the question, the way
-		// a derived title has since #790: dsh's title model answered "ok" and
-		// "47" for sessions whose user turn is a whole sentence — 39 of the 43
-		// on this machine's store (#3328). Notes name themselves.
-		if s.Harness != "deja" && thinTitle(title) {
-			if next := nextSubstantialTitle(s.Messages, title); next != "" {
-				next, _ = redact.Text(next)
-				title = truncateTitle(next, 60)
-				agentTitle = false
-			}
-		}
+		title = widenThinSourceTitle(s, boundSourceTitle(s.Harness, title))
 	}
 	// The import fields travel with the session, not with the transcript: a
 	// rebuild reloads imported sessions out of the index itself, and rebuilding
@@ -2203,7 +2192,31 @@ func earliestTitle(ms []model.Message, role string) string {
 // tests" — are not. A length rule rather than a vocabulary is the only version
 // of this that works in every language the store holds.
 func thinTitle(t string) bool {
-	return len(strings.Fields(t)) <= 2 && len([]rune(t)) <= 12
+	return titleWords(t) <= 2 && len([]rune(t)) <= 12
+}
+
+// titleWords counts what a reader would call words. Whitespace separates them
+// in most scripts and in none of the CJK ones, where a whole sentence is one
+// field and a rune-length rule alone called it a greeting: 为什么测试失败了 is
+// eight runes and says why the test failed (review of #3328). Thai is spaced
+// the same way, which is why Unspaced covers both.
+func titleWords(t string) int {
+	words, inWord := 0, false
+	for _, r := range t {
+		switch {
+		case cjkfold.Unspaced(r):
+			words++
+			inWord = false
+		case unicode.IsSpace(r):
+			inWord = false
+		default:
+			if !inWord {
+				words++
+				inWord = true
+			}
+		}
+	}
+	return words
 }
 
 // nextSubstantialTitle is the first later user turn that can name the session.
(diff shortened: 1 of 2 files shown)


vshulcz:
## Review

- The thin rule counted whitespace-separated fields, so a title in a script that does not space its words was one field: `为什么测试失败了` — eight runes saying why the test failed — was called thin and replaced by a longer, worse English turn. Nothing on this machine hits it today (no CJK titles, 90 Cyrillic ones, none thin), but the rule was wrong. `titleWords` now counts an unspaced rune as a word, which covers Thai for the same reason `cjkfold.Unspaced` does. A two-character greeting is still thin.
- The widening ran only in `metaForSession`, so a session that gets its thin title later — dsh and opencode both retitle after the fact — kept it until a full rebuild. `widenThinSourceTitle` is now shared with the incremental path.
- `agentTitle = false` was dead: no reader ever sets `AgentTitle` true beside a source title. Removed with the extraction.
- Not a problem, checked: notes are exempt through the `deja` harness, and `nextSubstantialTitle` never picks the title's own text back.

`TestACJKTitleThatSaysSomethingIsNotThin` covers both halves of the first point. The measurement is unchanged on the three real stores: 43 of 60 thin titles before, 8 after.

Left as it is: `opencodeThinTitle` in the reader keeps its own copy of the rule, because `internal/sources` cannot import `internal/index`. It only decides whether to take opencode's title at all; this rule decides what the index does with one.

vshulcz:
## Review, second pass

Refuted again, and both findings fixed in the follow-up commit.

- Hangul sits in `cjkfold`'s CJK set because that is what folds, but Korean writes its words apart: counting each syllable as a word made `고마워` — one word, three syllables — look substantial, so exactly the case this PR targets was skipped. `titleWords` now counts a syllable as a word only for the scripts that really run their words together.
- `titleWorthy` had its own short list and had drifted from `notAsked`, so a session whose store title was thin could be renamed after a resume preamble: "This session is being continued from a previous conversation…". It now asks `notAsked`, which has rejected that text all along.

Two tests cover both, and both fail on the previous commit: `TestKoreanIsCountedByItsWordsNotItsSyllables` and `TestAWidenedTitleIsNeverTheHarnessOwnPreamble`.

The measurement on the three real stores is unchanged: 43 of 60 titles were two words or fewer, 8 after.

Left as it is: no test drives the incremental path end to end; the reviewer read it as correct in both directions. It is queued rather than faked here, because exercising it needs a store that retitles itself between two `deja index` runs.

vshulcz:
## Review, third pass

Both earlier findings verified fixed by the reviewer — Korean splits on its words now (`고마워`, `네`, `알겠습니다` thin; sentences and two-long-word titles kept), and a resume preamble can no longer be picked. It also drove the incremental path end to end against a hermetic opencode store: thin to thin keeps the widened title, thin to substantial takes the store's name immediately, substantial back to thin re-widens.

One new finding from reusing `notAsked` whole: it also matches "no visible output" anywhere in a turn, which is right for counting repeated plumbing and wrong for judging a name — "why does the button have no visible output when clicked" is a good title. The prefix list is now `harnessPreamble`, which is what a title candidate asks; `notAsked` keeps the phrase check for its own use. `TestAQuestionThatSaysNoVisibleOutputCanStillNameTheSession` covers it.

PR #3335, opened 2026-09-08
fix(zed): a thread's own order is its clock, so two identical turns both survive

### LINKED ISSUE #3333: 1385 of 6767 Zed messages never reach the index: no per-message clock, so identical turns collapse
deja parses 6767 messages from this machine's 30 Zed threads and stores 5382. The 1385 that never reach the index are genuinely separate turns: Zed writes no per-message timestamp, so every message in a thread carries the thread's start, and the ingest's duplicate check — session, role, timestamp, text — collapses any two that say the same thing.

Reproduction, hermetic HOME holding a copy of the real `threads.db`:

```
$ deja index
deja: zed: 30 sessions, 6767 messages
$ deja sources | grep zed
zed  …/Zed/threads  sessions=30 messages=6767 size=5.5 MB redacted=63

$ for id in $(deja last 40 --json | …ids…); do deja show $id --harness zed --json --limit 1 | jq .window.total; done | paste -sd+ | bc
5382
```

The gap is concentrated in the work records: reading the same file twice in a thread, or running the same command twice, writes the same text under the same role at the same stamp, and only the first survives. The fan-out measured 1519 file records parsed against 618 stored, up to 92 lost in a single thread.

Cause: `internal/sources/zed.go:305-312` stamps every message with `started` — "Zed stores no per-message timestamp in either thread format, so every message inherits the thread's start" — and `msgSeen.dup` (`internal/index/ingest.go:1425-1434`, keyed session + role + timestamp + text hash) exists to collapse the same session parsed twice from two files. With no clock in the transcript, the two rules meet and a real turn looks like a format twin.

Counter-case: a harness that stamps its messages does not lose them — the same run indexes 1283 Claude files and stores every message; the loss is Zed's alone among the stores here.

Expected: 6767 parsed, 6767 stored. Order is what Zed records, and the position in the thread is what tells two identical turns apart.

The same comment names aider as sharing the shape — filed separately once measured.

vshulcz:
Closes #3333.

A Zed thread carries no per-message clock, so every record took the thread's start and the ingest's duplicate check — session, role, timestamp, text — could not tell a repeated turn from the same session parsed twice. The place in the array is the only clock there is, so it becomes one: the start plus a millisecond per record, which keeps the order Zed recorded and stays inside the thread's own span.

Fail-before test: `TestZedTellsTwoIdenticalTurnsApartByTheirPlace` (internal/sources), on the collector: two file records and two command records share role, text and stamp, so the index stores one of each. `TestParseZedDBReadsBothStorageEncodings` asserted the old behaviour and now asserts the new stamps.

Real store, hermetic copy of `threads.db`, 30 threads:

```
before: parsed 6767, stored 5382   (the gap is mostly work records — the same file read twice in a thread)
after:  parsed 6767, stored 6767
```

The index version note names the reader change; the version is already 35 on this branch.

The comment this replaces said aider shares the shape. It does, and it is filed separately once measured — aider has six sessions here, so the loss there is small and the fix is not the same one line.

### DIFF
--- a/internal/index/index.go
+++ b/internal/index/index.go
@@ -126,7 +126,7 @@ import (
 // tool calls (#3295), opencode's synthetic parts (#3299) and subagent parents
 // (#3301), Copilot's injected skills (#3305), aider's banner and its
 // continuation lines (#3311), opencode's own session titles (#3315),
-// Antigravity's plan approvals (#3326). A store
+// Antigravity's plan approvals (#3326), Zed's per-message stamps (#3333). A store
 // built before holds skill bodies as the person's words and none of the new
 // tool output, and nothing re-reads a source without the bump (#3289).
 const version = 35
--- a/internal/sources/zed.go
+++ b/internal/sources/zed.go
@@ -302,15 +302,27 @@ func zedSession(db string, r zedRow) (model.Session, bool) {
 		Started: started,
 		Updated: updated,
 	}
-	// Zed stores no per-message timestamp in either thread format, so every
-	// message inherits the thread's start the way aider's do. Order is carried
-	// by the array itself, which is what recall actually reads.
+	// Zed stores no per-message timestamp in either thread format, so the place
+	// in the array is the only clock there is, and it becomes one: the thread's
+	// start plus a millisecond per record. Sharing one stamp made two turns
+	// that say the same thing — the same file read twice, the same command run
+	// twice — indistinguishable to the ingest's duplicate check, which dropped
+	// 1385 of this machine's 6767 Zed messages (#3333).
+	at := started
+	tick := func() time.Time {
+		t := at
+		at = at.Add(time.Millisecond)
+		return t
+	}
 	for _, raw := range th.Messages {
 		role, text := zedMessage(raw)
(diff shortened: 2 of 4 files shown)


vshulcz:
## Review

- Real measurement on a hermetic copy of the 30 real threads, binaries from both commits: 6767 parsed either way, 5382 stored before and 6767 after. Counted straight out of `records.bin`.
- All 30 threads now end with a record stamped past the thread's own `updated_at`, by at most 567 ms. It traced every reader of `Session.Updated`: the zed reader sets it from the database column, never through `Touch`, so ranking, freshness and ordering are untouched. The one place per-message times feed an `Updated` is a cross-machine import, where an imported Zed session used to collapse to its start; it is now the real update time plus under a second, which is more accurate rather than less.
- Two indexes over the same store, then an export and import between them: zero new records, because the stamps are deterministic — the thread's own start plus the array position — so the same source parses to the same content keys on both machines.
- Re-indexing without `--rebuild` stays at 6767 across 30 sessions, no drift, no duplication.
- Nothing else in the tree assumes Zed messages share a stamp.

Issue #2604, opened 2026-08-29
uninstall leaves an empty mcpServers block in a config it did not own


vshulcz:
A full install and uninstall leaves 22 backups in the tree. Comparing each with the file beside it says what they are:

    backups left: 22 — 9 identical to the live file, 13 differ, 0 orphaned

and the 13 differ by exactly one thing:

    live  { "mcpServers": {}, "mine": "MINE" }
    bak   { "mine": "MINE" }

So a config the reader owned comes back with an empty `mcpServers` block they never wrote. deja adds that block when a file has none — #676 already stops it from being added on the way *out* — and the uninstall removes only deja's entry inside it.

The block is deja's, so it should go with deja. The reader's own block stays whether it is empty or not, which needs provenance: the wiring record already knows which files deja created (#2583) and now records which containers it added.

The backups themselves stay. `TestUninstallLeavesNoFileOrDirItCreated` has said since #840 that "the user's own config and its snapshot are not ours to delete", and a snapshot of their file is theirs even when the live file has come back to exactly it.

vshulcz:
Fixed in #2605.

PR #3342, opened 2026-09-08
fix(uninstall): a snapshot holding only deja's own block is deja's to remove

### LINKED ISSUE #3340: uninstall keeps a .bak holding only deja's own block and calls it "configs you already had"
`deja uninstall deepseek-auto` leaves a `.bak` that holds nothing but deja's own block and tells the reader it is a config they already had.

Reproduction on a hermetic HOME, nothing pre-existing:

```
$ deja install deepseek            # writes ~/.dsh/cordis.patch.yml and plugins/deja/command.js
$ deja install deepseek-auto       # a second run: adds auto.js, so the write snapshots the file it finds
$ deja uninstall deepseek-auto
deepseek: removed ~/.dsh/cordis.patch.yml
deja: kept 1 snapshot of configs you already had — ~/.dsh/cordis.patch.yml.bak — yours to remove

$ cat ~/.dsh/cordis.patch.yml.bak
- insert:
    - id: mcp-deja
      name: '@deepseek-ai/dsh-mcp-client'
      …
# deja mcp:end
```

Every byte of that file is deja's, and the line says the opposite. The rule is already written down and already implemented — `writeIfChanged` in `cmd/deja/install.go:1300-1314` deletes a snapshot whose content `mentionsDeja`, "a snapshot that itself contains deja's wiring is deja's own and nothing the user asked to keep". It never runs here: the two `removed` fast paths above it (`cmd/deja/install.go:1245-1264`, the empty-file and structurally-empty cases) return before the `defer` is registered, and removing deja's whole block from a file deja created is exactly the path that takes them.

Counter-case: a config the reader already had keeps its snapshot, which is what #2604 and `TestUninstallLeavesNoFileOrDirItCreated` are about — that file does not mention deja and must stay.

Expected: no `.bak` and no "configs you already had" line when the snapshot is deja's own.

Found by the fan-out on the dsh harness; the same shape reaches any harness whose config deja writes whole and then edits a second time.

vshulcz:
Closes #3340.

The rule was already there — `writeIfChanged` drops a `.bak` whose content is deja's own wiring — but it lived in a `defer` registered below the two `removed` fast paths, and taking deja's whole block out of a file deja created is exactly the path that returns early. `dropOwnBackup` is that rule as a function, called from all three places.

Fail-before test: `TestUninstallDropsASnapshotThatIsOnlyDejas` (cmd/deja), on the collector: the snapshot survives the removal. `TestUninstallKeepsASnapshotThatIsTheReadersOwn` holds the other side, unchanged since #2604.

Hermetic HOME, the fan-out's exact sequence:

```
before:  deja: kept 1 snapshot of configs you already had — ~/.dsh/cordis.patch.yml.bak — yours to remove
         left behind: .dsh/cordis.patch.yml.bak, .dsh/plugins (empty), .agents/skills (empty)
after:   no such line, and nothing left under HOME at all
```

The empty directories went with it: the pruning that follows a removal only takes a directory that is empty, and the stray snapshot was what kept `.dsh` occupied.


vshulcz:
## Review

The finding: `mentionsDeja` matches loose markers, so a reader's own snapshot that merely says the word — a comment like `# my note: I looked at "deja" last year` — was deleted once the rule reached the paths that delete the file. The reviewer built it and checked the same case against the collector, where the snapshot survived, so the first commit really did widen that.

Now the two delete paths ask `wiringCreated(path)` first: deja drops a snapshot only beside a file it created itself, where nothing of the reader's can be sitting. The content rule stays where it was for the write path. `TestUninstallKeepsASnapshotBesideAFileDejaDidNotCreate` covers it and fails on the previous commit.

The dsh case the PR is about still comes out clean, because `install deepseek` created that file: the "kept 1 snapshot of configs you already had" line and the `.bak` are gone, and so are the empty directories under it.

Also checked and not refuted: the `.bak` lookup uses the path before symlink resolution, which is the same directory the removal used; and the "kept N snapshots" line stats fresh, so it drops out rather than miscounting.

vshulcz:
This and #3346 are the same fix — I had #3340 open in another session. Yours is the one to keep: the `wiringCreated` guard on the zero-length path is the case mine misses, since `mentionsDeja` also matches a config of the reader's that happens to name deja, and the third test is exactly that. Closing mine.

vshulcz:
## Review, second pass

Refuted again, and fixed by changing what ownership means.

The finding: `wiringCreated(path)` proves deja made the live file and says nothing about who made the snapshot beside it. The reviewer built the case — a file deja created, a `.bak` the reader wrote holding `# I disabled deja: here for now, keeping my own notes` — and deja deleted it, because `mentionsDeja` matches the word wherever it appears.

Ownership is now recorded rather than guessed: `backupOnce` remembers each `.bak` deja writes, the wiring record persists that list beside `created`, and the two delete paths ask `snapshotTaken` before dropping anything. A snapshot the reader put there is theirs whatever it holds, and one deja took is deja's whatever it holds.

`TestUninstallKeepsASnapshotDejaDidNotTake` is the reviewer's case; it fails on the previous commit. The dsh case the PR is about still comes out clean on a hermetic HOME: no "configs you already had" line, no `.bak`, no empty directories.

vshulcz:
## Review, third pass

Refuted again, and fixed.

The first finding is confirmed fixed: a snapshot deja never took is kept, and a wiring record without the field fails towards keeping.

The new one: `backupOnce` runs after the write path has followed a symlink, so the recorded name is the resolved one, while the two delete paths still hold the name they were given. On macOS every `/var` path resolves to `/private/var`, and a config in a dotfiles repository is the real case — so `snapshotTaken` never matched and deja kept its own snapshot forever.

`snapshotTaken` now asks both spellings, resolving the directory rather than the file, because by the time an uninstall asks, the file it is removing is already gone. `TestUninstallDropsItsOwnSnapshotThroughASymlink` covers it through a real symlink and fails on the previous commit.

vshulcz:
## Review, fourth pass

Refuted, and fixed.

The reviewer confirmed the cross-process case works, and found the reverse: a snapshot recorded under the link spelling — what an older deja wrote, or any write that did not resolve — was never found when a later run asked with the real path, because resolution only runs forwards and nothing recovers which link pointed at a directory. It failed safe, leaving deja's own `.bak` behind rather than touching anything of the reader's.

Fixed by canonicalising both sides instead of converting one into the other: `canonicalSnapshotPath` resolves the directory of the recorded name and of the asked name, and they meet in that form. The write also records the name it was given, not only the resolved one.

`TestUninstallDropsASnapshotRecordedUnderTheOtherSpelling` builds exactly that state — the record holding the link spelling, the uninstall arriving with the real path — and fails on the previous commit.

vshulcz:
CI caught a real thing the review did not: `TestDocCommentsNameWhatTheyDocument` failed on both runners because the snapshot list I added landed between `wiringCreated`'s doc comment and its declaration, so `go doc -u` would have shown that comment on the wrong symbol. Comment restored to its function; the new list has its own. Nothing else moved.

PR #3344, opened 2026-09-08
docs(deepseek): dsh lists one deja tool with a mode, not six tools

### LINKED ISSUE #3343: docs/registry/deepseek.md names six MCP tools; the server advertises one tool with six modes
`docs/registry/deepseek.md:43-44` says that after the MCP install "dsh lists `mcp__deja__recall`, `recall_context`, `remember`, `blame`, `fix` and `how` itself". The server advertises one tool.

Driven against the real server, initialize plus `tools/list` over stdio:

```
$ printf '%s\n%s\n%s\n' \
   '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"probe","version":"1"}}}' \
   '{"jsonrpc":"2.0","method":"notifications/initialized"}' \
   '{"jsonrpc":"2.0","id":2,"method":"tools/list"}' | deja mcp
tools advertised: ['deja']
  modes: ['recall', 'context', 'blame', 'fix', 'how', 'remember']
```

`cmd/deja/mcp.go:266` registers the single `deja` tool and `:315` dispatches on its `mode` argument. The six names in the page were the shape before the tool was collapsed into one, so a reader following that page looks for tools their client will never show and cannot tell whether the install worked.

Counter-case: no other registry page carries the sentence — `grep -rl "mcp__deja__" docs/registry/` returns deepseek.md alone — so this is one page, not a pattern.

Expected: the page names the one tool and its modes, the way the install itself does.

Found by the fan-out on the dsh harness.

vshulcz:
Closes #3343.

The page described the shape from before the six MCP tools were collapsed into one with a `mode` argument. Driven against the real server, `tools/list` returns `deja` alone with modes `recall`, `context`, `blame`, `fix`, `how`, `remember`.

Fail-before test: `TestNoRegistryPageNamesAToolTheServerDoesNotAdvertise` (internal/sources) reads every page under docs/registry and fails on a `mcp__deja__<name>` mention; on the collector it fails on deepseek.md. It is the guard for the next page that drifts the same way — this was the only one today.

### DIFF
--- a/docs/guide/memory-for-copilot-chat.html
+++ b/docs/guide/memory-for-copilot-chat.html
@@ -112,7 +112,7 @@
 <p>The same command writes <code>&lt;User&gt;/prompts/deja.instructions.md</code>. VS Code applies a custom-instructions file to every chat when its frontmatter says <code>applyTo: "**"</code>, and that is what makes recall arrive without being asked for: the block is in front of the model before it reads your question, and it says to search history first.</p>
 
 <h2>What arrives, and what does not</h2>
-<p>The six tools, in agent mode, and the instruction to use them. Measured on VS Code 1.134.0: with the instructions file installed and deja never mentioned in the question, "the packer keeps refusing our manifest — why?" made Copilot Chat call <code>recall</code> on its own, and answer from a session it had never seen.</p>
+<p>The <code>deja</code> tool with its six modes, in agent mode, and the instruction to use it. Measured on VS Code 1.134.0: with the instructions file installed and deja never mentioned in the question, "the packer keeps refusing our manifest — why?" made Copilot Chat call <code>recall</code> on its own, and answer from a session it had never seen.</p>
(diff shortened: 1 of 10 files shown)


vshulcz:
## Review

- The built `docs/registry/deepseek.html` is what a reader and a search engine open, and it still carried the sentence the markdown had lost: `go run ./scripts/genregistry` had not been run. It is now, which also rebuilt four pages that had drifted earlier (goose, hermes, kimi, prime).
- Three guide pages said "the six tools" as current fact: kimi, zed and copilot-chat. They now say the `deja` tool with its six modes. The wording is minimal and yours to overrule — the fact was wrong, not the sentence's shape.
- The guard read only `docs/registry/*.md`. It now reads the registry pages, the guide pages and README, in both markdown and html, and fails on a `mcp__deja__<name>` mention or the phrase "six tools".

Two things it checked and did not refute: the new sentence matches what the server actually advertises (it drove `tools/list` itself and got one tool with the modes in that order), and nothing else on the dsh page is stale.

Left alone: `TestRegistryPagesAreBuiltAndListed` compares only headings and the verified date between the markdown and the html, so it cannot catch prose drift. Widening it to the body is a bigger change than this PR; queued.

PR #3346, opened 2026-09-08
fix(uninstall): take back a snapshot that holds only deja's own block

### LINKED ISSUE #3340: uninstall keeps a .bak holding only deja's own block and calls it "configs you already had"

vshulcz:
`writeIfChanged` already deletes a `.bak` whose content is deja's own, but the two fast paths that delete a config deja wrote whole return before that runs — which is exactly the path a config deja created and then edited twice takes. `deja uninstall deepseek-auto` left a `.bak` containing nothing but deja's block and printed "kept 1 snapshot of configs you already had".

The rule is now one helper called from all three places. A snapshot of a config that was already there still stays (#2604), with a test for each half.

Closes #3340

### DIFF
--- a/cmd/deja/install.go
+++ b/cmd/deja/install.go
@@ -1041,6 +1041,29 @@ func isRealDir(p string) bool {
 	return err == nil && fi.IsDir()
 }
 
+// dropOurOwnBackup removes the snapshot beside path when it holds deja's
+// wiring: that copy is deja's own and nothing the user asked to keep, and
+// leaving it puts a file naming a binary they just removed back in their
+// config directory. Ownership is read from the content rather than from who
+// wrote it — installing a harness whose config deja edits twice takes the
+// snapshot on the second write, so the uninstall that meets it did not create
+// it and would otherwise leave it (goose).
+//
+// A snapshot of the reader's config stays even when the live file has come
+// back to exactly it: that copy is theirs, and
+// TestUninstallLeavesNoFileOrDirItCreated has said so since #840 — "the user's
+// own config and its snapshot are not ours to delete" (#2604).
+func dropOurOwnBackup(path string) {
+	bak := path + ".bak"
+	b, err := os.ReadFile(bak)
+	if err != nil {
+		return
+	}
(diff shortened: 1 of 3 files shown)


vshulcz:
Superseded by #3342, which has the same fix plus a guard mine lacks: it only drops the snapshot beside a file deja itself created, so a reader's own config that merely mentions deja keeps its .bak.

PR #3348, opened 2026-09-08
fix(search): --limit binds the ranked and error tiers too

### LINKED ISSUE #3345: --limit is ignored on the ranked and error tiers: 50 sessions for --limit 3
`deja "<question>" --limit 3` prints 50 sessions when the query falls through to the ranked tier, and so does a query with no flag at all, where the default is 15.

Measured on a read-only copy of this machine's index:

```
$ deja "follow HERMES_HOME, the variable Hermes itself reads" --limit 3   # ranked tier
50 sessions
$ deja "HERMES_HOME" --limit 3                                            # exact tier
3 sessions
$ deja "follow HERMES_HOME, the variable Hermes itself reads"             # no flag, default 15
50 sessions
```

The JSON envelope says the same: `--json --limit 1`, `--limit 3` and `--limit 10` all return `hits: 50` with `total: 973, capped: true`, so a consumer that asked for one session parses fifty.

Cause: the cap lives in `search.RunDetailed` (`internal/search/search.go:411-422`), which the exact path goes through. The two tiers that build their hits themselves — `search.TierRelevance` via `RelevanceHitsWeighted` and `search.TierError` via `ErrorHits` — never truncate: `cmd/deja/main.go:952-958` (plain) and `cmd/deja/main.go:1301-1327` (JSON). What arrives is the retrieval window, which is 50.

Counter-case that rules out a mis-read flag: the same binary, the same flag, one word instead of a sentence — the exact tier honours it exactly.

Expected: `--limit n` bounds every tier, and a query with no flag prints 15 rather than 50. The ranked tier is exactly where a reader is least sure of the answer, so the flood is worst there.

Found while comparing the MCP recall ranking with search's on ten real questions.

vshulcz:
Closes #3345.

The cap lives in `RunDetailed`, which only the exact path goes through. The ranked-relevance tier and the error-signature tier build their own hits, so `--limit` never reached them: on this machine's index, `--limit 3` printed 50 sessions where the same flag on the exact tier printed 3, and `--json --limit 1` handed a consumer 50 hits. `search.CapHits` is that cap as a function; `capTierHits` applies it to the two tiers.

Fail-before test: `TestLimitBindsTheRankedTier` (cmd/deja), on the collector: `hits = 50, want the 3 the reader asked for`. `TestCapHitsBoundsWhatEachTierHandsBack` (internal/search) pins the function itself.

Read-only copy of the real index:

```
before:  --limit 3 → 50 sessions      --json --limit 1 → 50 hits
after:   --limit 3 →  3 sessions      --json --limit 1 →  1 hit, total 973, capped true
```

A question rather than a change: with no flag at all these two tiers still print the whole window of 50, where every other tier defaults to 15. `TestRelevanceEnvelopeReportsThePreCapTotal` pins that 50 deliberately, so I left it — but 50 sessions is what a reader gets exactly when nothing matched their words, which is when they are least able to sift. Should the default be 15 there too?

### DIFF
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -1243,10 +1243,15 @@ func runBareSearch(dir string, args []string, sourceInstance string) error {
 // keep the window they have always served, which the JSON envelope's own test
 // pins; only what the reader asked for binds them.
 func capTierHits(hits []search.Hit, o search.Options) ([]search.Hit, bool) {
-	if o.Limit == 0 || o.All {
+	if o.Limit == 0 {
 		return hits, false
 	}
-	return search.CapHits(hits, o.Limit, o.All)
+	// A limit the reader typed binds even beside --all, which is how the exact
+	// tier has always read the pair: --all lifts the default, and the number
+	// asked for is still the number wanted. Letting --all win here made the
+	// same two flags mean opposite things depending on which tier answered
+	// (review of #3345).
+	return search.CapHits(hits, o.Limit, false)
 }
 
 func runSearch(dir string, args []string, sourceInstance string) error {
--- a/cmd/deja/main.go
+++ b/cmd/deja/main.go
@@ -951,10 +951,13 @@ func cmdCtx(dir string, rest []string) error {
 	var hits []search.Hit
 	if result.Tier == search.TierError {
 		fmt.Fprintln(os.Stderr, "deja: matched by error signature; showing the sessions that hit it")
-		hits = search.ErrorHits(ss)
+		// Both tiers below build their own hits and never went through the cap
+		// in RunDetailed, so `--limit 3` printed the whole window — fifty
+		// sessions — and so did a query with no flag (#3345).
+		hits, _ = capTierHits(search.ErrorHits(ss), o)
 	} else if result.Tier == search.TierRelevance {
 		fmt.Fprintln(os.Stderr, "deja: no exact match; showing sessions ranked by relevance to the whole query")
-		hits = search.RelevanceHitsWeighted(ss, index.RelevanceMatchTerms(o.Query), result.TermIDF)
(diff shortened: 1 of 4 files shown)


vshulcz:
## Review

The finding: `--all --limit 3` truncated to three on the exact tier and returned the whole window on the ranked one, because the new `capTierHits` let `--all` win while `CapHits` — the rule the exact tier has always used — lets the typed limit win. The same two flags meant opposite things depending on which tier answered. The reviewer reproduced it end to end and at the unit level.

Settled the other way: a limit the reader typed binds on every tier, and `--all` lifts only the default. That leaves the exact tier exactly as it was and makes the ranked tier agree with it. Two tests cover the pair, one per path: `TestCapHitsHonoursALimitBesideAll` and `TestAllWithALimitStillBindsTheRankedTier`, the second failing on the previous commit with `hits = 50, want the 3 the reader asked for`.

Also checked and clean in the review: `deja ctx`, `deja how`, `deja files`, `deja blame` and `deja last` do not route through these options, so none of them changed; `--limit 0`, a negative limit and a limit past 100 are rejected by the flag parser before any of this.

vshulcz:
## Review, second pass

Not refuted. The matrix, run end to end on both tiers:

```
exact tier   (10 matching, default 15):  --all --limit 3 → 3, total 10, capped   --all → 10   --limit 3 → 3   neither → 10
ranked tier  (61 matching, window 50):   --all --limit 3 → 3, total 61, capped   --all → 50   --limit 3 → 3   neither → 50
```

`--limit` binds identically everywhere now, `--all` lifts only the default, and the ranked tier's retrieval window of 50 is untouched by either flag, which is what `docs/json-output.md:66-71` already describes. `capped` and `total` still report the pool behind the window, not the window.

One pre-existing gap it names, not introduced here and not fixed here: the CLI help line reads `--all   return every match, no cap`, which read literally promises everything even beside a `--limit`. The docs and the code agree with each other; the help line is the odd one out, and its wording is yours. Queued rather than reworded here.

PR #3352, opened 2026-09-08
fix(cursor): a turn is dated by the stamp Cursor wrote into it

### LINKED ISSUE #3349: Cursor turns are dated by the file's mtime while the transcript stamps each one: a session spanning a month reads as one day
Cursor writes the time of each turn into the turn itself, and deja dates every message in a transcript by the file's modification time instead. A session whose turns are a month apart reads as if all of it happened on the last day.

On this machine's real store, a hermetic copy, one transcript:

```
$ grep -o "<timestamp>[^<]*" …/de875c53-….jsonl
Sunday, Jul 26, 2026, 1:06 PM (UTC+3)
Saturday, Aug 22, 2026, 9:35 AM (UTC+3)

$ deja show de875c53 --harness cursor
2026-08-22 09:35 user:
2026-08-22 09:35 assistant:
2026-08-22 09:35 user:
2026-08-22 09:35 assistant:
```

The July turn is dated four weeks late, and the session's `started` is the same stamp as its `updated`.

Cause: `internal/sources/cursor.go:341-343` takes `fi.ModTime()` with the comment "transcripts carry no timestamps", and every message is then stamped `s.Updated` (`:355`, `:362`, `:367`). The stamp is there, in band, wrapped in `<timestamp>…</timestamp>` ahead of the user's words — the same block the reader already strips before indexing the text.

What it costs: the recency half of ranking, `deja last`, the "changed since" line and the session's own span all read one date for a conversation that spans weeks; and a transcript copied or restored takes the copy's mtime.

Counter-case: a harness that stamps its turns does not have this — the Claude, Codex and Kimi readers all carry per-message times, and their sessions show a real span.

Expected: each turn carries the time Cursor wrote for it, and the session spans from the first to the last.

Found by the fan-out on the Cursor harness.

vshulcz:
Closes #3349.

Cursor writes `<timestamp>Sunday, Jul 26, 2026, 1:06 PM (UTC+3)</timestamp>` ahead of the person's words, in the same block the reader already strips before indexing. `cursorTurnTime` reads it, each turn carries it, and an assistant turn takes the question's stamp because it has none of its own. The file's modification time is the fallback for a transcript with no stamp deja can read — a copy or a restore moves that time and nothing else.

Fail-before test: `TestCursorTakesTheTimeFromTheTurnNotTheFile` (internal/sources), on the collector: `started = 2026-09-08 …, want the first turn's own stamp 2026-07-26 13:06`.

Real store, hermetic copy, one transcript:

```
before:  2026-08-22 09:35 user / assistant / user / assistant
after:   2026-07-26 13:06 user / assistant
         2026-08-22 09:35 user / assistant
```

The index version note names the reader change; the version is already 35 on this branch.

### DIFF
--- a/internal/index/index.go
+++ b/internal/index/index.go
@@ -126,7 +126,7 @@ import (
 // tool calls (#3295), opencode's synthetic parts (#3299) and subagent parents
 // (#3301), Copilot's injected skills (#3305), aider's banner and its
 // continuation lines (#3311), opencode's own session titles (#3315),
-// Antigravity's plan approvals (#3326). A store
+// Antigravity's plan approvals (#3326), Cursor's per-turn stamps (#3349). A store
 // built before holds skill bodies as the person's words and none of the new
 // tool output, and nothing re-reads a source without the bump (#3289).
 const version = 35
--- a/internal/sources/cursor.go
+++ b/internal/sources/cursor.go
@@ -6,8 +6,10 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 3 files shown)


vshulcz:
## Review

It built binaries from the collector and the branch, indexed a hermetic copy of the real store (19 transcripts, every stamp `(UTC+3)`), diffed the listings, and hand-built the awkward cases: stamps going backwards, tool records between two user turns, an incremental re-index with a turn appended, and other zones. `go test ./...` and `go vet ./...` clean.

Two latent gaps it names, neither reachable on the real corpus:

- the zone group only fires on the literal `UTC`, so `(GMT+3)` would parse successfully and drop the offset rather than falling back to the file time. A wrong-but-successful parse is worse than the intended failure, and it is worth narrowing — queued, since nothing Cursor writes today takes that path.
- seconds in the time, lowercase am/pm, and a non-English month or weekday all fall back to the file time with no signal. That is the existing silent-fallback design in this file rather than something new.

vshulcz:
CI caught what the review did not: `internal/sources` fell to 86.9% against its 87% floor, because the stamp reader added branches nothing exercised. Covered now — the shapes that parse (a zone with minutes, no zone, a month spelled in full) and the three that fall back to the file's time (a label other than UTC, seconds in the time, a month name deja does not know) — and `CursorTranscriptProjectDirBase`, which was exported with no test at all. The package is back at 87.8%.

PR #3358, opened 2026-09-08
fix(index): a stripped block leaves no blank line ahead of the question

### LINKED ISSUE #3357: every Cursor turn is indexed with a blank line ahead of the question
Every Cursor turn is indexed with a blank line ahead of the person's words. The text is right, the shape is not: what a reader sees in `deja show`, and what a snippet quotes, opens on nothing.

On a hermetic copy of this machine's real store:

```
$ deja show de875c53 --harness cursor --json | jq -r '.session.messages[].text' | head -3
"\n\nDo not use any tool. Count how many separate blocks in your context are tagged…"
"1\n\n[REDACTED]"
"\n\nодним словом: о чём была эта сессия?"
```

Cursor writes the turn as

```
<timestamp>Sunday, Jul 26, 2026, 1:06 PM (UTC+3)</timestamp>
<user_query>
why does the retry loop drop the last attempt
</user_query>
```

and the cleaning does two things in a row, each leaving a newline behind: `stripInjectedPrefixes` (`internal/index/selfrecall.go:159-179`) cuts the stamp and keeps what follows, including the line break under it, and `unwrapBlock` (`:229-243`) removes the tags and keeps the break under the opening one.

Counter-case: `stripBetween` — the path deja's own recall goes through — already trims exactly this (`:295-299`, "a message that now starts with a blank line is an artefact of the removal, not something anyone wrote"). The two other paths never got the same treatment.

Expected: the indexed text starts with the first word of the question. Paragraph breaks inside what someone wrote stay where they are.

Found by the fan-out on the Cursor harness; it affects every Cursor CLI session on the store, 19 here.

vshulcz:
Closes #3357.

`stripSelfRecall` trims the newlines at the ends of what it hands back, which `stripBetween` has done for deja's own recall since it was written. The two other paths — cutting a harness preamble, unwrapping a wrapper — each left the break that separated the block from the words, and Cursor's turns go through both.

Fail-before test: `TestStripSelfRecallLeavesNoBlankLineWhereABlockWas` (internal/index), on the collector: `"\n\nwhy does the retry loop drop the last attempt"`. Its second half holds the other side: a stamp in the middle of a message is prose, and the paragraph breaks around it stay.

Real store, hermetic copy, `deja index --rebuild`:

```
before: "\n\nDo not use any tool. Count how many separate blocks…"
after:  "Do not use any tool. Count how many separate blocks…"
```

19 Cursor sessions here, every turn of them.

### DIFF
--- a/internal/index/selfrecall.go
+++ b/internal/index/selfrecall.go
@@ -130,7 +130,12 @@ func stripSelfRecall(text string) string {
 		text = unwrapBlock(text, m[0], m[1])
 	}
 	text = stripInjectedPrefixes(text)
-	return stripInjectedLines(text)
+	text = stripInjectedLines(text)
+	// Removing a block or unwrapping a tag leaves the newline that separated it
+	// from the words: every Cursor turn was indexed with a blank line ahead of
+	// the question, because the stamp above it and the wrapper around it each
+	// left one (#3357). The paragraph breaks inside what a person wrote stay.
+	return strings.Trim(text, "\n")
 }
 
 // stripInjectedLines drops lines that are entirely a harness marker. The line
--- a/internal/index/selfrecall.go
+++ b/internal/index/selfrecall.go
@@ -117,7 +117,6 @@ var unwrapBlocks = [][2]string{
 // to a real user turn, so dropping the whole message would lose the question
 // that prompted the session.
 func stripSelfRecall(text string) string {
-	original := text
 	for _, m := range selfRecallMarkers {
 		text = stripBetween(text, m[0], m[1])
 	}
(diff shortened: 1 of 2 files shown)


vshulcz:
## Review

Both findings were the same mistake: the trim was unconditional.

- `TestAgentsHeadingNeedsItsBlockToFollow` failed on the branch — a person quoting `<INSTRUCTIONS>` with no injected block present must come back byte for byte, and the trailing newline was gone.
- A message with nothing to strip at all lost blank lines its author wrote: `"\n\n```\n\nfunc foo() {}\n```\n\n"` came back without either.

The trim now runs only when the cleaning actually changed the text, so it takes what a removal left behind and nothing else. `TestStripSelfRecallLeavesAnUntouchedMessageAlone` covers the second case; the existing test covers the first.

Its third finding stands as a known edge and is not fixed here: only newlines are in the cutset, so a leading `"  \n"` — trailing spaces on the stamp's own line — survives. No transcript on this store writes that shape, and widening the cutset to spaces would eat indentation a person meant.

The measurement is unchanged: all 19 Cursor sessions here now open on the first word of the question.

PR #3363, opened 2026-09-08
fix(doctor): the cline row says when a transcript went unread

### LINKED ISSUE #3360: doctor's cline row cannot say a transcript went unread — 15 files, and a planted one changes nothing
`deja doctor`'s cline row prints a file count and nothing else, so a transcript deja cannot read in a Cline store is invisible. Every other file-listing harness says so.

Reproduction on a hermetic HOME holding a copy of the real store (15 sessions), with one transcript copied under a name the reader does not take — the shape a format change has:

```
$ cp <id>.messages.json conversation.json
$ deja doctor
  cline        found     ~/.cline/data/sessions  (15 files)
```

The count does not move and nothing is said. The control, same run, the same planted file under Kimi's store:

```
  kimi         parsed-zero ~/.kimi-code/sessions  (1 file, 1 not recognised here)
  cline        found       ~/.cline/data/sessions  (15 files)
```

Expected: `(15 files, 1 not recognised here)`.

Cause: `cmd/deja/doctor.go:783-789` calls `printRow` with a bare count, where continue (#3297), copilot (#3303), kimi (#3309), openclaw (#3317), grok (#3319) and codex (#3321) call `printFilesBeside`.

Plain `printFiles` is not enough here: a Cline session directory keeps `<id>.json` beside `<id>.messages.json` — the manifest the reader opens itself for the title, the working directory and the timestamps (`internal/sources/cline.go:185`) — so all fifteen of them would read as transcripts deja failed to read, which is the miscount #3297 was about.

Follow-on from #3297, #3303, #3309, #3317, #3319, #3321. The same bare-count shape is still on aider, antigravity, roo and copilot-chat.

vshulcz:
Closes #3360.

The cline row moves to `printFilesBeside`, the pattern Continue (#3297), Copilot (#3303), Kimi (#3309), OpenClaw (#3317), Grok (#3319) and Codex (#3321) already use. `ClineSidecarFiles` names the per-session manifest — the file named after its own directory, which the reader opens for the title, the working directory and the timestamps — and nothing else, so a transcript under an unexpected name is still counted.

The row prints its modern store and its legacy roots on one line, and that string is not a directory, so `printFilesBesideIn` takes the walked path separately. Every other caller goes through the old signature unchanged.

Fail-before test: `TestDoctorNamesAnUnreadClineTranscriptButNotItsManifest` (cmd/deja), on the collector: `(1 file)`, want it to name the transcript deja did not read.

Real store, hermetic copy of 15 sessions, with one transcript copied to `conversation.json`:

```
before:  cline  found  ~/.cline/data/sessions  (15 files)
after:   cline  found  ~/.cline/data/sessions  (15 files, 1 not recognised here)
clean:   cline  found  ~/.cline/data/sessions  (15 files)
```

The bare-count shape is still on aider, antigravity, roo and copilot-chat; they are queued one at a time.

### DIFF
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -734,13 +734,19 @@ func doctorHarnesses(w io.Writer, dir string) {
 	// disagree with `deja sources` by one; counted as unread, it made every
 	// store report a file deja could not read (#3297). So it is neither: the
 	// count is the transcripts, and the note leaves the list alone.
-	printFilesBeside := func(name, path string, present bool, seen []string, beside ...string) {
+	// printFilesBesideIn is printFilesBeside for a row whose printed location is
+	// not one directory: cline names its modern store and its legacy roots on
+	// the same line, and that string cannot be walked (#3360).
+	printFilesBesideIn := func(name, loc, walk string, present bool, seen []string, beside ...string) {
 		detail := doctorCount(len(seen), "file")
 		placed := append(append([]string{}, seen...), beside...)
-		if unread, _ := unplacedFiles(path, placed, nil); unread > 0 {
+		if unread, _ := unplacedFiles(walk, placed, nil); unread > 0 {
 			detail += fmt.Sprintf(", %d not recognised here", unread)
 		}
-		printRow(name, path, present, detail)
+		printRow(name, loc, present, detail)
+	}
+	printFilesBeside := func(name, path string, present bool, seen []string, beside ...string) {
+		printFilesBesideIn(name, path, path, present, seen, beside...)
 	}
 
 	claudeRoot := sources.ClaudeRoot()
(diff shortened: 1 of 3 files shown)


vshulcz:
## Review

cavecrew reviewer, adversarial pass — not refuted, and the one finding it made is fixed in the follow-up commit.

The finding: the row prints the legacy VS Code roots on its own line, and the walk covered only the modern store, so a stray file under a legacy `tasks/<id>/` was invisible while the line advertised that root. Not a regression — the old row made no unread claim at all — but a promise the count could not keep. Every root the line names is now walked. `TestDoctorCountsAStrayFileUnderAClineLegacyRoot` covers it and fails on the previous commit.

What it verified: the manifest rule matches exactly what the reader opens, including a manifest whose name does not match its directory, which is still reported; `providers.json` and `cline_mcp_settings.json` live under `data/settings`, outside the walked tree either way; presence is right for a legacy-only store and for a fresh install; and a full `doctorHarnesses` diff between the merge-base and the branch shows every other row byte-identical.

PR #3371, opened 2026-09-08
fix(copilot): a command that exited non-zero says so

### LINKED ISSUE #3369: a failed Copilot command is recorded success:true, so deja fix answers nothing for that harness
Copilot CLI records a failed command with `"success": true`, and deja takes it at its word: the command that fixed the error is never mined, so `deja fix` answers nothing for this harness.

On a hermetic copy of this machine's real store:

```
$ deja fix "directory prefix . does not contain main module"
deja: no session on this machine ran a command after that error

$ deja search --role command "go vet"
  $ cd …/coprobe && go vet ./...
  $ cd …/coprobe && go mod init coprobe && go vet ./...
```

The pair is right there — the failing run, then the one that fixed it — and neither command carries an outcome, so `commandFailed` (`internal/index/fixpair.go:291` through `CommandExitOutcome`, `internal/index/commands.go:412`) never sees a failure to pair.

What the store actually holds for that run (`~/.copilot/session-state/<id>/events.jsonl`, `tool.execution_complete`):

```json
{"type":"tool.execution_complete","success":true,
 "toolCallId":"toolu_bdrk_011aFuMjrjoZ1mUi7pyFeSjX",
 "result":{"content":"pattern ./...: directory prefix . does not contain main module or its selected dependencies\n<shellId: 0 completed with exit code 1>"},
 "toolTelemetry":{"metrics":{"commandTimeout":30000,"exit_code":1}}}
```

`success` is true on all 13 `tool.execution_complete` records in the store, including the one that failed; the exit code lives in `toolTelemetry.metrics.exit_code` and in a `<shellId: N completed with exit code M>` trailer inside the result text. `internal/sources/copilot.go:147-159` reads neither, and unlike codex (`internal/sources/codex.go:295`) it never appends the `  → exit N` marker that the outcome reader looks for.

Counter-case: the same shell session under Codex indexes as `$ go vet ./...  → exit 1`, and `deja fix` answers with the command that followed.

Expected: a Copilot command that exited non-zero carries the marker, and `deja fix "<the error>"` names the repair.

Found by the fan-out on the Copilot CLI harness.

vshulcz:
Closes #3369.

Copilot writes `"success": true` on every `tool.execution_complete`, failed runs included — all 13 on this machine's store — and puts the outcome in `toolTelemetry.metrics.exit_code` and in a `<shellId: N completed with exit code M>` trailer under the output. `copilotExitCode` reads the telemetry first and falls back to the trailer, and a non-zero code appends the `  → exit N` marker to the command it belongs to, matched by `toolCallId`. That is the marker `CommandExitOutcome` already reads for codex and opencode.

Fail-before test: `TestCopilotMarksACommandThatExitedNonZero` (internal/sources), on the collector: `the failed run reads "$ go vet ./...", want it to end in the exit marker`. It also holds the other side — a run that exited zero gains nothing.

Real store, hermetic copy:

```
before:  $ … && go vet ./...
after:   $ … && go vet ./...  → exit 1
```

One thing this does not fix, found while checking it: `deja fix` still answers nothing for that particular error, because `FrictionLine` does not recognise `pattern ./...: directory prefix . does not contain main module or its selected dependencies` as an error line at all — a Go compile error in the same store pairs fine. That is a separate shape and is queued rather than widened here.

### DIFF
--- a/internal/index/index.go
+++ b/internal/index/index.go
@@ -127,7 +127,7 @@ import (
 // (#3301), Copilot's injected skills (#3305), aider's banner and its
 // continuation lines (#3311), opencode's own session titles (#3315),
 // Antigravity's plan approvals (#3326), Zed's inlined thread mentions (#3336)
-// and its per-message stamps (#3333). A store
+// and its per-message stamps (#3333), Copilot's exit codes (#3369). A store
 // built before holds skill bodies as the person's words and none of the new
 // tool output, and nothing re-reads a source without the bump (#3289).
 const version = 35
--- a/internal/sources/copilot.go
+++ b/internal/sources/copilot.go
@@ -1,6 +1,7 @@
 package sources
 
 import (
+	"encoding/json"
 	"fmt"
 	"path/filepath"
 	"regexp"
@@ -79,7 +80,7 @@ func parseCopilotFileFromOffset(path string, offset int64) ([]model.Session, err
 	}
 	// Which message holds each shell command, by the call id the completion
 	// event repeats: Copilot files the command and its outcome as two records.
-	commandAt := map[string]int{}
+	commandAt := map[string][]int{}
 	err := scanJSONLFromOffset(path, offset, func(m map[string]any) {
 		typ, _ := m["type"].(string)
 		data, _ := m["data"].(map[string]any)
@@ -139,10 +140,12 @@ func parseCopilotFileFromOffset(path string, offset int64) ([]model.Session, err
 					records = append(records, model.Message{Role: RoleEdit, Text: span, Time: t})
 				}
 			}
-			cmdAt := -1
+			// Every command of the call, not the last one: a dialect that
+			// hands back an array would otherwise mark one of them.
+			var cmdAt []int
 			if IndexCommands() {
 				for _, cmd := range commandsIn(part, copilotDialect) {
-					cmdAt = len(records)
+					cmdAt = append(cmdAt, len(records))
 					records = append(records, model.Message{Role: RoleCommand, Text: cmd, Time: t})
 				}
 			}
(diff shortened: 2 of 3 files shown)


vshulcz:
## Review

cavecrew reviewer, adversarial pass — refuted, and fixed.

The finding is exact: the scanner decodes every event with `UseNumber`, so a JSON number arrives as `json.Number` and never as `float64`. Asserting `float64` made the telemetry path dead code — every marker in the real store came from the trailer regexp instead, and a failed run whose output has no trailer got no marker at all. `jsonInt` now reads `json.Number`, `float64` and a string alike.

`TestCopilotReadsTheExitCodeFromTelemetryWithoutATrailer` is that case: a `go test` failure whose output carries no shellId line, with `exit_code: 1` in the telemetry. It fails on the previous commit with `command = "$ go test ./..."`.

Two smaller ones, both taken:

- a tool call that hands back more than one command marked only the last; all of them are marked now, which matters if the dialect's argument shape ever changes.
- the comment above `commandFailed` said codex and opencode append the marker; copilot does too, and it says so.

The reviewer also confirmed what it could not break: interleaved calls land on the right command in both completion orders, no non-shell tool in the real store carries an exit code, and no output text matches the trailer by accident.

vshulcz:
## Review, second pass

Refuted, and fixed.

The reviewer confirmed the number reading is right — `json.Number`, a string, a missing metrics object and a null all behave — and found the regression the multi-command change introduced: the map grew by append and nothing cleared it, so an id reused for a later call put the second outcome on the first run too. Their repro is now `TestCopilotDoesNotMarkAnEarlierRunWithALaterOutcome`, and it fails on the previous commit with `"$ go build ./...  → exit 1"` on the run that succeeded.

The list is assigned per call and deleted once its outcome is consumed. Their smaller note is taken too: a store writing the code as `1.0` is read through `Float64` rather than falling through to the trailer.

No real store has a repeated `toolCallId` — the ids are long random strings — so this was latent rather than live.

PR #3378, opened 2026-09-08
fix(doctor): the Antigravity row says when a transcript went unread

### LINKED ISSUE #3377: doctor's Antigravity row cannot report an unread transcript: the whole store lives under a dot directory
`deja doctor`'s Antigravity row prints a file count and nothing else, and it cannot do otherwise: every file in the store lives under a `.system_generated` directory, and `unplacedFiles` skips dot directories by design. So a transcript deja fails to read there is invisible twice over.

On this machine's real store:

```
$ deja doctor
  antigravity  found  ~/.gemini/antigravity:…:~/.gemini/antigravity-cli  (50 files, 50 indexed sessions)
```

What the store holds, by path shape:

```
  42  antigravity-cli/brain/<id>/.system_generated/logs/transcript.jsonl      ← read
  42  antigravity-cli/brain/<id>/.system_generated/logs/transcript_full.jsonl
  44  …/logs/chunks/transcript_full/<n>.jsonl
  43  …/logs/chunks/transcript/<n>.jsonl
  26  antigravity/brain/<id>/.system_generated/messages/<uuid>.json
   8  antigravity/brain/<id>/.system_generated/logs/transcript.jsonl          ← read
```

The dot-directory rule exists for a real reason — 452 of the 482 files codex reported as unread sat in `.tmp` (#701) — but here it is the store's own layout, so the rule silences the row completely rather than trimming its noise.

Cause: `cmd/deja/doctor.go:801-804` calls `printRow` with a bare count, and `unplacedFiles` (`cmd/deja/doctor.go:427-451`) returns early for anything under a dot directory.

Counter-case: kimi, continue, copilot, openclaw, grok, codex and cline all report an unread transcript today, and none of them keeps its transcripts under a dot directory.

Expected: the row says when a transcript in the Antigravity store went unread, and stays quiet about the store's own bookkeeping — the full transcripts, the chunk files, the message records and the metadata beside a plan.

vshulcz:
Closes #3377.

Two things had to meet for this row to say anything. The dot-directory rule that trims codex's `.tmp` noise silenced Antigravity completely, because its whole store is under `.system_generated`; and there was no list of what the store keeps beside a transcript, so lifting the rule alone would have reported 170 files as unread.

`unplacedFilesIn` takes that decision from the caller rather than guessing — codex writes in-progress rollouts under `.tmp` and must keep the rule, Antigravity must not — and `AntigravitySidecarFiles` names the rest: the full log, the chunk files it is written in, the message records, the metadata beside a plan, and everything outside `brain/`, which is settings, caches and the extension deja itself installs.

Fail-before test: `TestDoctorNamesAnUnreadAntigravityTranscriptButNotItsBookkeeping` (cmd/deja), on the collector: `(1 file)`, want it to name the transcript deja did not read.

Real store, four roots, 50 transcripts:

```
before:  antigravity  found  …  (50 files, 50 indexed sessions)
after:   antigravity  found  …  (50 files, 50 indexed sessions)   ← unchanged, nothing unread
```

and with one transcript copied to `transcript.v2.jsonl` the row reads `1 not recognised here`. The codex row is unchanged at `(35 files)`, which is the check that the rule still applies where it was written for.


vshulcz:
## Review

cavecrew reviewer, adversarial pass — refuted twice, both fixed.

- The list exempted every file outside `brain/`, so a transcript restored beside the conversations — `<root>/restored_backup/transcript.jsonl` — was absorbed as a sidecar and the row said nothing. That is the exact failure this PR exists to end. The list now names what it knows: `cache`, `mcp`, `plugins`, `updater`, `tmp`, and `settings.json`, `history.jsonl`, `import_manifest.json`. Anything else is counted.
- `transcript_full.jsonl` was exempt unconditionally, so a session whose only artefact is the full log had its content indexed nowhere and the row stayed quiet. It is a sidecar only when a `transcript.jsonl` sits beside it.

`TestDoctorNamesAnAntigravityTranscriptInTheWrongPlace` builds both and requires `2 not recognised here`; it fails on the previous commit with `(1 file)`.

Real store unchanged: `antigravity (50 files, 50 indexed sessions)`, codex still `(35 files)`. `internal/sources` coverage 87.2% against its 87% floor.

The reviewer ran out of budget before the remaining items — the byte-identical diff of the other rows, the missing-root and duplicate-id cases, and the timing. Those are still open on this PR.

vshulcz:
The ubuntu job failed the coverage floor again — `internal/sources: 86.8% below floor 87%` — because `AntigravitySidecarFiles` was exercised only from the doctor test in `cmd/deja`, and the floor is per package. It now has its own test in `internal/sources`, which pins the rule directly: the full log, the chunk files, the message records, the plan metadata, the caches and the settings are claimed; a transcript, one restored outside a conversation, and a session whose only log is the full one are not. Package back to 87.8%.

vshulcz:
## Review

cavecrew reviewer, adversarial pass — not refuted, after two earlier rounds it did refute.

- Every other doctor row is byte-identical between a collector build and the branch on the same populated hermetic HOME. The codex row reads `2 files, 1 not recognised here` on both, and its `.tmp/rollout-partial.jsonl` is not in that number — the dot-directory rule still applies where it was written for.
- The four Antigravity roots: a broken-symlink root is dropped, an empty one is inert, and two roots holding the same session id each contribute their own count, with a stray file in either one showing up.
- The rest of its list — the sidecar rule against a transcript restored outside a conversation, a session whose only log is the full one, timing, `go test`, `go vet` — was covered in the earlier rounds and reconfirmed here.

Issue #3202, opened 2026-09-08
the hermes-memory provider deja writes does not import on Hermes 0.17: no RecallStatus


vshulcz:
The memory provider `deja install hermes-auto` writes does not load on the Hermes on this machine (v0.17.0, 2026.6.19):

```
$ HERMES_HOME=… python -c 'import plugins.deja_memory'          # what Hermes' loader does
ImportError: cannot import name 'RecallStatus' from 'agent.memory_provider'
$ hermes …  discover_memory_providers()  → deja-memory: unavailable
$ hermes …  load_memory_provider("deja-memory") → None ("loaded but no provider instance found")
```

`plugins/deja-memory/__init__.py:18` (the file cmd/deja/install_hermes*.go writes) imports `MemoryProvider, RecallStatus` from `agent.memory_provider`; this Hermes has `MemoryProvider` and no `RecallStatus`. The hook, meanwhile, stands down when `memory.provider: deja-memory` is set and the plugin directory exists (checked: provider `deja-memory` → 0 deja calls from the hook), so a user who activates the provider as the docs say gets recall from neither.

Control: with `memory.provider` unset the hook calls deja on every turn (2 calls in the shim run). Evidence: /private/tmp/qa-hermesclaw/hermes-provider-vs-hook/provider-import-error.txt, provider-run.txt, hook-shim2-*.txt.

Expected: the provider imports what the installed Hermes has (a guarded import for `RecallStatus`, or none), and the hook stands down only once the provider actually loads.

PR #3391, opened 2026-09-08
fix(hermes): the memory provider loads on the build it finds

### LINKED ISSUE #3390: the Hermes memory provider does not load on 0.17, and the hook stands down for it anyway
The memory provider deja installs for Hermes cannot load on Hermes 0.17.0, and the hook that would otherwise carry the recall stands down anyway because it only reads the config.

The provider's generated `__init__.py` opens with

```python
from agent.memory_provider import MemoryProvider, RecallStatus
```

and that build's `agent/memory_provider.py` (315 lines, `pyproject.toml` says `version = "0.17.0"`) defines one class:

```
43:class MemoryProvider(ABC):
```

There is no `RecallStatus`. Hermes' own loader says so:

```
DEBUG:plugins.memory:Failed to exec_module _hermes_user_memory.deja-memory:
  cannot import name 'RecallStatus' from 'agent.memory_provider'
```

`load_memory_provider('deja-memory')` returns `None`, and `discover_memory_providers()` lists it as unavailable.

The second half is what makes it silent. The hook plugin's `_provider_active` (`cmd/deja/install_hermes.go:100`) decides whether to stand down by reading the configuration alone:

```python
if cfg_get(load_config(), "memory", "provider") != "deja-memory":
```

So on a machine where the reader followed the guide — `hermes config set memory.provider deja-memory` — the hook believes the provider is doing the work, and the provider is not there. Neither path recalls anything, and nothing says so.

deja uses `RecallStatus` in one place, `recall_status()` at `install_hermes_memory.go:230`, which is an optional reporting hook.

Two fixes, and they are separable:

- import it defensively, so a build without `RecallStatus` still loads the provider and only loses the status line;
- have `_provider_active` check that the provider actually loaded, not that the config names it.

Found by the fan-out on Hermes; follow-on from #3202, which is the import half of this.

vshulcz:
Closes #3390.

Two halves of one silence:

- the generated provider opened with `from agent.memory_provider import MemoryProvider, RecallStatus`, and Hermes 0.17's module defines only `MemoryProvider`. The import failed, so the file never loaded. `RecallStatus` is now optional and `recall_status` returns nothing when the build has no such class — the status line is all that is lost.
- `_provider_active` decided by reading `memory.provider` from the config, so on a machine that followed the guide the hook stood down for a provider that was not there. It now asks the loader whether the provider actually loaded, keeping the file check as the fallback for a build with no loader to ask.

Fail-before tests: `TestHermesProviderLoadsWithoutRecallStatus` and `TestHermesHookAsksTheLoaderNotTheConfig` (cmd/deja), both failing on the collector with the exact sentences above.

Against the real Hermes 0.17.0 on this machine, hermetic HOME, its own loader:

```
before:  DEBUG:plugins.memory:Failed to exec_module _hermes_user_memory.deja-memory:
           cannot import name 'RecallStatus' from 'agent.memory_provider'
after:   provider loaded: True
```

Found by the fan-out on Hermes; #3202 is the import half of the same thing.

### DIFF
--- a/cmd/deja/install_hermes.go
+++ b/cmd/deja/install_hermes.go
@@ -99,10 +99,23 @@ def _provider_active():
         from hermes_cli.config import cfg_get, load_config
         if cfg_get(load_config(), "memory", "provider") != "deja-memory":
             return False
-        # Config can outlive the provider (an uninstall, a deleted directory);
-        # then the hook is the memory again, or there would be none at all.
+        # Config can outlive the provider (an uninstall, a deleted directory),
+        # and a provider that is there can still fail to load on a Hermes build
+        # it was not written for — the hook then stood down for a provider that
+        # was answering nothing (#3390). Ask the loader, and fall back to the
+        # file when this build has no loader to ask.
         here = os.path.dirname(os.path.abspath(__file__))
-        return os.path.isfile(os.path.join(os.path.dirname(here), "deja-memory", "__init__.py"))
+        present = os.path.isfile(os.path.join(os.path.dirname(here), "deja-memory", "__init__.py"))
+        if not present:
+            return False
+        try:
+            from plugins.memory import load_memory_provider
+        except Exception:
+            return True
+        try:
+            return load_memory_provider("deja-memory") is not None
+        except Exception:
+            return False
     except Exception:
         return False
 
--- a/cmd/deja/install_hermes_memory.go
+++ b/cmd/deja/install_hermes_memory.go
@@ -85,7 +85,13 @@ import shutil
 import subprocess
 from typing import Any, Dict, List, Optional
 
-from agent.memory_provider import MemoryProvider, RecallStatus
+from agent.memory_provider import MemoryProvider
+
+try:  # Hermes 0.17 has no RecallStatus; the provider loads without it and
+    # loses only the status line (#3390).
+    from agent.memory_provider import RecallStatus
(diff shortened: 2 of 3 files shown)


vshulcz:
## Review

cavecrew reviewer, adversarial pass — not refuted, driven against the real Hermes 0.17 checkout on this machine (a copy, in a hermetic HOME).

- The provider loads and answers, not merely imports: `load_memory_provider("deja-memory")` returns a real provider and its prefetch came back with genuine recalled history from this machine's index. `recall_status()` returns nothing on this build, which is the only thing the missing class costs.
- Reverting to the unconditional import on the same build reproduces the failure exactly — module exec fails, the loader returns nothing — and with the old `_provider_active` the hook would still have stood down. With the fix it returns false in that state and the hook takes over.
- No recursion: the provider does not import the hook module. Timing, three consecutive real calls: 3.5 ms cold, about 0.5 ms after.
- The fallback with `plugins.memory` absent returns true from the file check, which is what older builds did before the loader existed.
- `go test ./cmd/deja/...` and `go vet ./...` clean.

One nit it flagged and I am leaving as it is: `load_memory_provider` constructs a fresh provider each call rather than handing back Hermes' active one, so the liveness check is "construct one and see" rather than "ask if loaded". The constructor is three assignments and no I/O, and the alternative is reaching into Hermes' internals.

Issue #3373, opened 2026-09-08
three real walls are not read as errors: go.mod not found, no main module, no such table


vshulcz:
Three walls this machine hits repeatedly are not read as errors at all, so nothing is mined from the sessions that hit them: `deja fix` answers "no session on this machine ran a command after that error" even when the very next command was the fix.

Counted in the real stores under `~/.claude/projects`, `~/.codex/sessions` and `~/.copilot`:

```
17×  go: go.mod file not found in current directory or any parent directory; see 'go help modules'
12×  pattern ./...: directory prefix . does not contain main module or its selected dependencies
 4×  Error: in prepare, no such table: part
```

The Copilot session at `7dad4ddc-…` is the whole shape in one place: `go vet ./...` exits 1 with the second line above, the next command is `go mod init coprobe && go vet ./...`, and it exits 0.

```
$ deja fix "pattern ./...: directory prefix . does not contain main module or its selected dependencies"
deja: no session on this machine ran a command after that error
```

Cause: `isFriction` (`internal/index/friction.go:355-390`) matches a phrase list, and none of these three is in it. `does not exist` is there but only behind a leading `error`, which the go lines do not have; `no such file or directory` is there, `no such table` is not.

Counter-case that keeps this narrow: a Go compile error from the same store — `queue/retry.go:12:2: undefined: backoffCap` — is recognised and pairs correctly, so the miner and the pairing work; it is the phrase list that has the hole.

Expected: each of those three lines reads as friction, while a line of source that quotes one — a test asserting on the text, a comment, an `echo` — still does not.

vshulcz:
fixed on main by #3375, pinned by `TestFrictionReadsTheGoModuleAndSqliteWalls`. it landed on the collector branch, so the merge in #3420 brought the fix in without closing this.

PR #3393, opened 2026-09-08
fix(friction): a missing table is the server's line, not a sentence about tables

vshulcz:
Follow-up to #3373, which merged before its review landed — my mistake on the gate, and the review found this.

`no such table` went into the phrase list bare, so it fired on prose: `FrictionLine("there is no such table in the spec, so I improvised one")` read as an error. It now needs the same shape `does not exist` has needed since #2431 — the line has to open with the server's own `Error`, which is how the real one arrives: `Error: in prepare, no such table: part`.

Fail-before test: `TestNoSuchTableNeedsTheServersOwnShape` (internal/index), on the collector both prose sentences read as friction.

The reviewer also measured what the merged change bought, against a hermetic copy of the real index: two new walls cross the three-session floor — `pattern ./...: directory prefix . does not contain main module or its selected dependencies` at 10 sessions and `go: go.mod file not found…` at 3 — and confirmed both are genuine compiler errors in the underlying text. The seven real sqlite sightings never reached the floor, which is why this narrowing costs nothing measurable.

### DIFF
--- a/internal/index/friction.go
+++ b/internal/index/friction.go
@@ -382,7 +382,7 @@ func isFriction(l string) bool {
 		// in the real stores, and sqlite saying a table is missing, 4 times.
 		// "no such file or directory" was already here; its sibling was not
 		// (#3373).
-		"go.mod file not found", "does not contain main module", "no such table",
+		"go.mod file not found", "does not contain main module",
 	} {
 		if strings.Contains(low, p) {
 			return true
@@ -392,7 +392,12 @@ func isFriction(l string) bool {
 	// line opens with the server's ERROR marker. The phrase on its own is a
 	// sentence people write — "the orders table does not exist yet" — so it is
 	// only a wall where the server said it.
-	if strings.HasPrefix(low, "error") && strings.Contains(low, "does not exist") {
+	// A database saying a thing is not there says it in its own shape, and
+	// "no such table" needs the same guard for the same reason: "there is no
+	// such table in the spec, so I improvised one" is a sentence someone
+	// writes (review of #3373).
+	if strings.HasPrefix(low, "error") &&
+		(strings.Contains(low, "does not exist") || strings.Contains(low, "no such table")) {
 		return true
 	}
 	return false
--- a/internal/index/friction.go
+++ b/internal/index/friction.go
@@ -392,16 +392,28 @@ func isFriction(l string) bool {
 	// line opens with the server's ERROR marker. The phrase on its own is a
 	// sentence people write — "the orders table does not exist yet" — so it is
 	// only a wall where the server said it.
-	// A database saying a thing is not there says it in its own shape, and
-	// "no such table" needs the same guard for the same reason: "there is no
-	// such table in the spec, so I improvised one" is a sentence someone
-	// writes (review of #3373).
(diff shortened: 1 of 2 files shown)


vshulcz:
## Review

cavecrew reviewer, adversarial pass — refuted, and fixed.

The prefix guard was the wrong axis. It dropped every shape whose runtime puts the marker somewhere other than the first word — `D1_ERROR: no such table: users`, `sqlite3.OperationalError: no such table: users`, `Parse error: no such table: x`, and the bare `no such table: users` a Go driver returns — and it still let prose through, because "Error handling for missing tables is not written yet…" opens with the word.

What separates a driver from a person is the table: the driver names it right after the phrase. The rule is now that colon-and-a-name, or an `error:` marker anywhere in the line. All five real shapes read as friction and all three prose sentences do not; the test carries every one of them.

Its third point stands and is not fixed here: `go.mod file not found` and `does not contain main module` are still bare matches. Neither has a prose form I can construct that reads naturally — both name a file and a concept that only a toolchain puts together — but that is an argument, not a measurement, and it is worth its own look.

vshulcz:
Third pass. The reviewer's four shapes (`D1_ERROR:`, `sqlite3.OperationalError:`, `Parse error:`, bare `no such table: users`) were already covered by the second commit and are in the test. What survived was a false positive the review did not name: a sentence can end a clause on a colon —

```
I checked and there is no such table: the schema only has runs
```

— and the colon-plus-token rule read it as a driver line. The rule is now: the phrase names a token after the colon, **and** the line either opens with the phrase (the bare driver line) or carries `error` somewhere in it (every runtime puts its marker in front). Both sentences are in the test now, and `SQL logic error: no such table: main.sessions` beside them.

`go test ./internal/index` passes; no row on this machine's index mentions the phrase either way.

vshulcz:
## Review

Adversarial pass by the author after the reviewer's refutation — the reviewer subagents on this machine stopped executing at 23:05, so the third pass is mine, with the cases written down.

Ran `FrictionLine` over walls and prose the tests did not have:

| line | read as friction |
|---|---|
| `Error: in prepare, no such table: part` | yes |
| `D1_ERROR: no such table: users` | yes |
| `sqlite3.OperationalError: no such table: users` | yes |
| `Parse error: no such table: x` | yes |
| `SQL logic error: no such table: main.sessions` | yes |
| `no such table: users` | yes |
| `there is no such table in the spec, so I improvised one` | no |
| `we have no such table yet — add a migration` | no |
| `Error handling for missing tables is not written yet, so no such table checks run` | no |
| `TODO: no such table handling anywhere in the loader` | no |
| `the docs say: no such table exists for this mapping` | no |
| `I checked and there is no such table: the schema only has runs` | no (was yes before this commit) |

The last one is what the second pass still let through and what the third commit closes: the colon is not the driver's shape on its own.

Narrower than before in one place worth naming — a sentence that carries the word `error` *and* names a token after `no such table:` would still be read as a wall. I could not build one that reads like something a person writes.

`go test ./internal/index` passes. No row on this machine's index moves either way: the phrase does not appear in its friction list before or after.

Verdict: CONFIRMED at 03d2a566.

PR #3400, opened 2026-09-08
fix(doctor): the machine report names the cline root a transcript came from

### LINKED ISSUE #3399: doctor --json names ~/.cline/data/sessions for a store whose transcripts are all in the VS Code tree
`doctor --json` names `~/.cline/data/sessions` for a store whose transcripts are all in the VS Code extension tree — the row on the same run names both.

Cline keeps two generations: the CLI/SDK store under `~/.cline/data/sessions` and the VS Code extension's `globalStorage/saoudrizwan.claude-dev/tasks`. `ClineSessionFiles()` reads both, the printed row names both, and `doctorStoreChecks()` in cmd/deja/doctor_report.go:449 lists one:

```go
{"cline", []string{sources.ClineSessionsDir()}, sources.ClineSessionFiles(), sources.ParseClineFile},
```

## Reproduction

A hermetic HOME with a legacy task and no CLI store:

```
$ export HOME=/private/tmp/qa169-home DEJA_INDEX_DIR=$HOME/idx
$ mkdir -p "$HOME/Library/Application Support/Code/User/globalStorage/saoudrizwan.claude-dev/tasks/1757000000000"
$ cat > .../tasks/1757000000000/api_conversation_history.json   # two turns

$ deja doctor | grep cline
  cline        found     ~/.cline/data/sessions, /private/tmp/qa169-home/Library/Application Support/Code/User/globalStorage/saoudrizwan.claude-dev  (1 file)

$ deja doctor --json | jq -c '.stores[] | select(.name=="cline")'
{"name":"cline","state":"ok","paths":["/private/tmp/qa169-home/.cline/data/sessions"],"files":1,"indexed_sessions":0}
```

The one file counted lives in a directory `paths` does not name, and the directory it does name does not exist.

## The half that costs recall

`paths` is also what `inspectDoctorStore` stats and what `firstDeniedDir` walks, so a locked legacy tree is never looked at. `chmod 000` on the same `tasks` directory:

```
$ deja doctor --json | jq -c '.stores[] | select(.name=="cline")'
{"name":"cline","state":"missing","paths":["/private/tmp/qa169-home/.cline/data/sessions"],"files":0,"indexed_sessions":0}
```

Expected `"state":"denied"` with `"denied"` naming the directory — the distinction #802 exists for. Actual: a harness nobody has used.

## Counter-case

With the same task readable and `DEJA_CLINE_ROOT` pointed at a real CLI store, `paths` is right, because it is the only root there is — the entry is correct on any machine that has never used the extension, which is why no test caught it.

## Fix

The row builds `clineWalks` (cmd/deja/doctor.go:818) from the modern directory plus `<legacy>/tasks` per root. The json entry should use the same list.

Follow-on from #3320 (the grok row and the json entry disagreeing on the same kind of path). The sweep of the other entries found no second case where json is narrower than the row: hermes and aider are wider in json than the row (`HermesHome`, `$HOME`), crush names the registry file in the row and the data directory in json.

vshulcz:
`doctorStoreChecks()` listed one root for cline while the walk and the printed row use two, so on a VS Code Cline machine `doctor --json` counted files under a directory it never named — and the permission walk, which reads the same list, never entered the legacy tasks tree: a locked one reported `"state":"missing"` instead of `denied`.

Both surfaces now take the list from `sources.ClineStoreRoots()`.

Fails before the change:

```
--- FAIL: TestDoctorJSONNamesTheClineRootTheFilesCameFrom
    paths = [.../cline/data/sessions], want the legacy tasks tree the transcript came from
--- FAIL: TestDoctorJSONSaysDeniedForALockedClineLegacyTree
    a locked legacy tree reports state "unplugged", want denied
```

On a hermetic HOME with one legacy task and no CLI store:

```
before {"name":"cline","state":"ok","paths":[".../.cline/data/sessions"],"files":1}
after  {"name":"cline","state":"ok","paths":[".../.cline/data/sessions",".../saoudrizwan.claude-dev/tasks"],"files":1}
after, tasks chmod 000: "state":"denied","denied":".../saoudrizwan.claude-dev/tasks"
```

This machine has only the CLI store, so its row and json are unchanged (15 files, 15 indexed sessions).

Closes #3399

### DIFF
--- a/cmd/deja/doctor.go
+++ b/cmd/deja/doctor.go
@@ -816,10 +816,7 @@ func doctorHarnesses(w io.Writer, dir string) {
 	// Every root the line names is walked, or the count would promise coverage
 	// the row does not have: a stray file under a legacy tasks tree was
 	// invisible while the line advertised that root (review of #3360).
-	clineWalks := []string{clineModern}
-	for _, root := range sources.ClineLegacyRoots() {
-		clineWalks = append(clineWalks, filepath.Join(root, "tasks"))
-	}
+	clineWalks := sources.ClineStoreRoots()
 	printFilesBesideIn("cline", clineLoc, clineWalks, false, clineFiles > 0 || doctorExists(clineModern),
 		sources.ClineSessionFiles(), sources.ClineSidecarFiles()...)
 
--- a/cmd/deja/doctor_cline_json_paths_test.go
+++ b/cmd/deja/doctor_cline_json_paths_test.go
@@ -1,0 +1,77 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// The machine form kept its own path per harness, and cline's named only the
+// CLI store: on a machine whose transcripts are all in the VS Code extension
+// tree the report counted files under a directory it never named, and the
+// directory it did name did not exist (#3399).
+func TestDoctorJSONNamesTheClineRootTheFilesCameFrom(t *testing.T) {
+	legacy := clineLegacyOnlyStore(t)
+
+	check := doctorCheckNamed(t, "cline")
+	if len(check.files) != 1 {
+		t.Fatalf("the walk found %d files, want the one legacy task", len(check.files))
+	}
+	tasks := filepath.Join(legacy, "tasks")
+	found := false
+	for _, p := range check.paths {
+		if p == tasks {
+			found = true
+		}
+	}
+	if !found {
+		t.Errorf("paths = %v, want the legacy tasks tree the transcript came from", check.paths)
+	}
+}
+
+// And because paths is what the permission walk covers, a legacy tree deja may
+// not read was reported as a harness nobody had used.
+func TestDoctorJSONSaysDeniedForALockedClineLegacyTree(t *testing.T) {
+	if os.Getuid() == 0 {
+		t.Skip("root reads everything")
+	}
+	legacy := clineLegacyOnlyStore(t)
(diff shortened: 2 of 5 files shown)


vshulcz:
## Review

Adversarial pass by the author — the reviewer subagents on this machine stopped executing at 23:05 and none has run a command since, so this is `.claude/rules/pr.md` step 2 rather than a second pair of eyes; the checks are recorded so they can be repeated.

- **Both layouts, hermetic HOME.** Legacy-only: `paths` names the tasks tree the transcript came from, `state` ok, `files` 1. CLI-only (this machine): `paths` unchanged at one entry, row unchanged at 15 files / 15 indexed sessions. Both generations present with the legacy tree `chmod 000`: `state` denied and `denied` names the locked directory, which is the behaviour the PR claims.
- **A path that does not exist** does not change `state`: the stat loop skips it, and `storeDiskGone` only fires when `files` is empty, which it is not in the mixed case.
- **`<root>/tasks` is the right root**: `ClineSessionFiles()` reads `api_conversation_history.json` under exactly that tree, and nothing else in the legacy layout is read.
- **Found and filed, not fixed here**: in the mixed case above the report says `denied` with no `partial`, though the CLI store read fine. That early return has never set `partial` — cursor, hermes, roo and antigravity reach it too — so it is pre-existing and now reachable for cline. #3407.
- `go vet ./cmd/deja ./internal/sources`, `go test ./cmd/deja ./internal/sources` pass; `internal/sources` coverage 87.6% against the 87% floor.

Verdict: CONFIRMED, with #3407 as the follow-on.

PR #3402, opened 2026-09-08
fix(friction): two failures that share a URL no longer print as one row twice

### LINKED ISSUE #3401: deja friction prints the same row twice — two different failures share their first 79 bytes
Two rows of `deja friction` on this machine's real index print the same text:

```
$ deja friction --limit 15
what this machine keeps tripping over — 1542 sessions read
…
   5 sessions  Get "http://toy-load-toy-load.default.svc.cluster.local/work?cpu_ms=20&jitte…
               opencode · last Apr 29
   5 sessions  Get "http://toy-load-toy-load.default.svc.cluster.local/work?cpu_ms=20&jitte…
               opencode · last Apr 29
```

They are two different failures. `--json`, which carries the whole line (#2419), says so:

```
{"error":"Get \"http://…/work?cpu_ms=20&jitter_ms=5\": dial tcp 0.0.0.0:0->10.43.138.73:80: connect: connection refused","sessions":5,…}
{"error":"Get \"http://…/work?cpu_ms=20&jitter_ms=5\": read tcp 10.42.0.35:39473->10.43.138.73:80: read: connection reset by peer","sessions":5,…}
```

A service that refuses the connection and one that resets it mid-read are different problems with different answers, and the surface whose whole job is to name what keeps going wrong shows one of them twice. `trimFriction` (cmd/deja/friction.go:303) cuts at 79 bytes from the front, and everything that tells these two apart is past the cut — the URL, which they share, fills the row.

## Counter-case

The rows are not duplicates of one signature: the signature is what folds a failure's varying ports and pids together (#2375), and these two have different signatures — `--json` prints them as separate rows with separate counts. Nothing about the grouping is wrong; only what the terminal shows.

## Expected

Two rows that are different errors read as different errors. When the cut would make two printed rows identical, the part that distinguishes them has to survive it.

## Impact

`deja friction` is 15 rows on a 1542-session store, so a wasted row is 7% of the answer, and a reader counting "5 sessions" twice for what looks like one error draws the wrong conclusion about what this machine trips over most.

Evidence: ~/.cache/deja-hunt/2026-09-08-harness/170/friction.txt, friction.json (a read-only copy of the real index).

vshulcz:
On this machine's index two rows of `deja friction` print identically — a connection refused and a connection reset on the same URL, 79 bytes of which is the URL. `--json` shows them as the two different errors they are.

`trimFriction` cuts from the front; where that cut would make two printed rows the same, the cut now moves into the middle so the end survives:

```
before:  5 sessions  Get "http://toy-load-toy-load.default.svc.cluster.local/work?cpu_ms=20&jitte…
         5 sessions  Get "http://toy-load-toy-load.default.svc.cluster.local/work?cpu_ms=20&jitte…
after:   5 sessions  Get "http://toy-load-toy-load.default.…138.73:80: connect: connection refused
         5 sessions  Get "http://toy-load-toy-load.default.….73:80: read: connection reset by peer
```

Rows that do not collide are untouched — the other 13 rows of the same run are byte-identical.

Fails before the change: `TestFrictionRowsThatShareTheirHeadStillDiffer` — `both failures print as "3 sessions  Get \"http://toy-load-…&jitte…" — the reader sees one error twice`.

Closes #3401

### DIFF
--- a/cmd/deja/friction.go
+++ b/cmd/deja/friction.go
@@ -9,12 +9,16 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/vshulcz/deja-vu/internal/index"
 	"github.com/vshulcz/deja-vu/internal/jsonout"
 	"github.com/vshulcz/deja-vu/internal/policy"
 	"github.com/vshulcz/deja-vu/internal/search"
 )
+
+// frictionRowBytes is what one row of the list holds.
+const frictionRowBytes = 79
 
 // `deja friction` names what this machine keeps tripping over.
 //
@@ -156,9 +160,10 @@ func runFriction(dir string, args []string, stdout io.Writer) error {
 		rows = rows[:limit]
 	}
 	fmt.Fprintf(stdout, "what this machine keeps tripping over — %d session%s read\n", sessions, pluralS(sessions))
-	for _, r := range rows {
+	lines := frictionRowLines(rows)
+	for i, r := range rows {
 		where := strings.Join(r.harnesses, ", ")
-		fmt.Fprintf(stdout, "  %2d sessions  %s\n", r.n, trimFriction(r.line))
+		fmt.Fprintf(stdout, "  %2d sessions  %s\n", r.n, lines[i])
 		fmt.Fprintf(stdout, "               %s", where)
 		if !r.when.IsZero() {
 			fmt.Fprintf(stdout, " · last %s", r.when.Local().Format("Jan 2"))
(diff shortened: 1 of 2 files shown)


vshulcz:
## Review

Adversarial pass by the author — reviewer subagents on this machine stopped executing at 23:05, so this is `.claude/rules/pr.md` step 2 with the checks written down.

- **Width and runes.** A pair of rows built from Cyrillic text past the bound: both come back ≤79 bytes, valid UTF-8 at both cuts, and distinct (`абвгдеёжзийабвгдеёж…оединение отклонено`). Same for a 200-byte ASCII line.
- **Three-way collision.** Three rows sharing 90 bytes of head render as three distinct rows; a plain row beside them is untouched. No collision survives the pass.
- **Two rows whose text is byte-identical** stay identical, and there is nothing to do about it — that would be one error under two signatures with the same newest occurrence, a different question from this one.
- **Other callers.** `trimFriction` has one caller outside the list: `recurringErrorLine` in install.go, which prints the single top row, so a collision cannot arise there. `deja brief` on a copy of this machine's index is byte-identical between base and branch.

Verdict: CONFIRMED.

PR #3404, opened 2026-09-08
fix(fix): an unconfirmed remedy from an excluded tree is withheld like a confirmed one

### LINKED ISSUE #3403: deja fix serves an unconfirmed remedy out of the tree the ignore rule excludes
`deja fix` serves a command out of a tree the ignore rule keeps out of recall, when that command is an unconfirmed sighting rather than a confirmed pair.

The default rule is `*/.claude/jobs/*` — a background agent's own runtime tree. A hermetic HOME with one session in it, an error, and one command after it:

```
$ deja "sprocket bus"
deja: no matches for "sprocket bus"
deja: the ignore rule withholds every indexed session from this path (*/.claude/jobs/*)

$ deja how /tmp/zarnak
no command on this machine mentions "/tmp/zarnak"

$ deja fix "zarnakctl: FATAL: could not bind sprocket bus at 127.0.0.1:7788"
zarnakctl: FATAL: could not bind sprocket bus at 127.0.0.1:7788 · 2026-08-01
  ran next, unconfirmed: $ launchctl unload ~/Library/LaunchAgents/io.example.plist
```

Search and `how` refuse the session; `fix` hands back what it ran. This is #2660 — "`deja fix` offered a remedy out of the tree the reader excluded while `deja how` refused to name the very same command" — for the half of the store that arrived after it.

## Cause

`FixesFor` in internal/index/fixpair.go:610. The confirmed-pair loop checks `ignored[p.Project]`; the candidates it holds back for a second sighting are appended to `held` before that check, and the loop that serves `held` when nothing was confirmed applies only `allow` (the trust policy):

```go
if p.Candidate {
    held = append(held, p)
    continue
}
...
if ignored[p.Project] { continue }
...
if len(out) == 0 {
    for _, p := range held {
        if allow != nil && !allow(p.Project) { continue }
        out = append(out, p)
```

## Counter-case, and why the sweep did not catch it

`TestNoSurfaceServesTheIgnoredTree` (cmd/deja/ignore_every_surface_test.go) does ask `fix` about an ignored tree and passes: its fixture writes the same error and the same command in three sessions, so the pair is corroborated (`repeats >= 2`) and goes down the branch that does check the rule. One session — the ordinary case, and 373 of 485 real sightings on the machine measured in #1301 — takes the other branch.

Three of the callers of `FixesFor` speak unasked (the session-start block, hook-plan, hook-tool-after), which is what made #2660 worth closing in the first place.

Evidence: ~/.cache/deja-hunt/2026-09-08-harness/172/repro.txt (hermetic HOME, real claude record shapes).

vshulcz:
`FixesFor` held unconfirmed sightings aside before the ignore-rule check and served them, when nothing was confirmed, past it. Search and `how` refuse those sessions; `fix` handed back the command they ran.

```
$ deja "sprocket bus"
deja: the ignore rule withholds every indexed session from this path (*/.claude/jobs/*)
$ deja fix "zarnakctl: FATAL: could not bind sprocket bus at 127.0.0.1:7788"
before:  ran next, unconfirmed: $ launchctl unload ~/Library/LaunchAgents/io.example.plist
after:   deja: no session on this machine ran a command after that error
```

`FixCandidateSeen` right below it already applies the rule, with a comment saying it is the same rule as this one — so this was the half that was missed, not a decision.

The existing sweep `TestNoSurfaceServesTheIgnoredTree` asks `fix` about an ignored tree and passes, because its fixture repeats the pair in three sessions and takes the confirmed branch. The new test uses one session, which is the ordinary case. Fails before the change with `a withheld session's command reached deja fix: … terraform apply -target module.klindra`.

Three real errors on a copy of this machine's index answer byte-identically before and after.

Closes #3403

### DIFF
--- a/cmd/deja/ignore_candidate_fix_test.go
+++ b/cmd/deja/ignore_candidate_fix_test.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// The sweep in ignore_every_surface_test.go writes the same error and remedy in
+// three sessions, so the pair is corroborated and takes the branch that checks
+// the ignore rule. One session — the ordinary case — is held back as an
+// unconfirmed sighting, and that branch checked only the trust policy: `deja
+// fix` handed back a command out of the tree search and `how` refuse to read
+// (#3403).
+func TestFixDoesNotServeAnUnconfirmedRemedyFromTheIgnoredTree(t *testing.T) {
+	tmp := hermeticEnv(t)
+	t.Setenv("DEJA_INDEX_DIR", filepath.Join(tmp, "index.db"))
+	policyFile := filepath.Join(tmp, "policy.json")
+	t.Setenv("DEJA_POLICY_FILE", policyFile)
+	if err := os.WriteFile(policyFile, []byte(`{"ignore":["*scratch*"]}`), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	root := os.Getenv("DEJA_CLAUDE_ROOT")
+	// One session only, so the remedy stays a sighting; and a remedy that
+	// names nothing the error names, so it is not promoted by sharing a term.
+	writeClaudeFixture(t, filepath.Join(root, "-w-scratch", "s1.jsonl"), "s1", []string{
+		`{"type":"user","sessionId":"s1","cwd":"/w/scratch","timestamp":"2026-07-01T10:00:00Z","message":{"role":"user","content":"the zonkobuffer keeps overflowing"}}`,
+		`{"type":"user","sessionId":"s1","cwd":"/w/scratch","timestamp":"2026-07-01T10:01:00Z","message":{"role":"user","content":[{"type":"tool_result","content":"panic: zonkobuffer overflow at shard 7"}]}}`,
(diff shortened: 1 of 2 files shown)


vshulcz:
## Review

Adversarial pass by the author — reviewer subagents on this machine stopped executing at 23:05, so this is `.claude/rules/pr.md` step 2 with the checks written down.

The sharpest evidence is the unasked caller, not `deja fix`. Same hermetic HOME under the default `*/.claude/jobs/*` rule, the post-tool hook fed the real failing command:

```
before: {"hookSpecificOutput":{"hookEventName":"PostToolUse","additionalContext":"<deja-recall>…
        deja: this error came up in tmp/zarnak before (2026-08-01) — one session ran this after it,
        and nothing confirms it worked: launchctl unload ~/Library/LaunchAgents/io.example.plist"}}
after:  (no output)
```

So the withheld tree was reaching an agent's context without anyone asking for it.

- **Every caller is covered**: all six go through `FixesFor` — `deja fix`, the MCP fix mode, the environment line, hook-plan, hook-tool-after, and `fix.go`'s candidate check. No caller applies the ignore rule itself.
- **Nothing else is withheld**: ten real error lines from this machine's own friction list answer byte-identically before and after (10/10).
- **Cost**: the manifest read is still lazy — `ignoredProject` loads on the first match of either kind. `deja fix` on a warm 1542-session index: 0.03 s before, 0.04–0.05 s after.
- **Agreement with the sibling**: `FixCandidateSeen` already applied the rule, so before this change `deja fix` could serve a sighting the function beside it would have denied holding. Now both refuse.
- `go vet ./internal/index ./cmd/deja`; `go test ./cmd/deja` and `./internal/index` pass, `internal/index` coverage 90.8%.

Verdict: CONFIRMED.

Issue #3395, opened 2026-09-08
a question that starts with a dash cannot be asked of ctx, how or blame


vshulcz:
A question that begins with a dash cannot be asked of `ctx`, `how` or `blame`, and the refusal does not say what to do instead.

```
$ deja ctx "--auto looks for the six harnesses it had targets for and never detected"
deja: ctx takes no flags, only a query or id-prefix — got "--auto looks for the six harnesses…"
$ deja how "--auto looks for the six harnesses"
deja: how: unknown flag "--auto looks for the six harnesses"
$ deja blame "--auto looks for the six harnesses"
deja: blame: unknown flag "--auto looks for the six harnesses"
```

The remedy already exists and works on all three:

```
$ deja ctx -- "--auto looks for the six harnesses"
# deja context: claude · goprojects/deja-vu · …
```

deja's own Amp plugin knows the trick — `install_amp.go:251` builds `["search", "--", query]` when the query starts with a dash — so the mechanism is settled; only the message is missing. Bare `deja "<query>"` and `deja fix "<query>"` take it either way, which is why the shape is easy to miss: it depends on the verb.

Found while running the ten recall questions of this night's measurement, one of which is a PR title beginning with `--auto`.

vshulcz:
fixed on main by #3396, pinned by `TestADashLeadingQueryIsToldAboutTheSeparator`. it landed on the collector branch, so the merge in #3420 brought the fix in without closing this.

PR #3406, opened 2026-09-08
fix(cli): the refusal of --limit names what last and blame take instead

### LINKED ISSUE #3405: deja blame and deja last refuse --limit without naming what they take instead
`deja blame` and `deja last` refuse `--limit` without saying what they take instead, while the two commands beside them accept that exact flag.

```
$ deja files deja --limit 5 --json | jq '.files|length'
5
$ deja how internal/index/friction.go --limit 5 --json | jq '.commands|length'
5
$ deja blame internal/index/friction.go --limit 2
deja: blame: unknown flag "--limit"
$ deja last --limit 3
deja: last: unknown flag "--limit"
```

`blame` bounds its answer with `--all`, and `last` takes the count as a bare argument (`deja last 3`) — both are deliberate, and neither refusal says so. The reader has just been told the flag works by the two commands they used before this one.

## Counter-case

The house already does this where the flags were easy to get wrong:

```
$ deja friction --limit x
deja: friction: unknown flag "-x" — it takes --limit n and --json
$ deja install --oops
deja: install: unknown flag "--oops" — it takes a target plus --no-guidance, --no-index or --force
```

So this is the same sentence for two more commands, not a new idea. Same shape as #3395, where `ctx` and `blame` learned to name `--` for a query that starts with a dash.

## Cause

cmd/deja/main.go:2365 (`parseLast`) and cmd/deja/main.go:2634 (`parseBlame`).

## Expected

The refusal names the spelling that works: the count for `last`, `--all` for `blame`.

Evidence: ~/.cache/deja-hunt/2026-09-08-harness/174/repro.txt, limit-lines.txt (a read-only copy of this machine's index).

vshulcz:
`files` and `how` take `--limit`, so a reader arrives at `blame` and `last` having just been told the flag works. Both refuse it — `blame` widens with `--all`, `last` takes the count as a bare argument — and neither said so:

```
before:  deja: blame: unknown flag "--limit"
         deja: last: unknown flag "--limit"
after:   deja: blame: unknown flag "--limit" — it takes --all to show every session
         deja: last: unknown flag "--limit" — the count is a bare argument, `deja last 3`
```

Only `--limit` gets the note; every other flag keeps the refusal it had, including blame's `--` sentence from #3395. `flagName` makes `--limit=3` the same spelling as `--limit 3`.

The wording follows `friction: unknown flag %q — it takes --limit n and --json` and the same sentence in `install`. Question for the PR: "the count is a bare argument" is my phrasing — `deja last 3` is what the help spells, and I kept it verbatim.

Fails before the change: `TestTheRefusalNamesWhatTheCommandTakes` — `last says "last: unknown flag \"--limit\"", without the count it takes`.

Closes #3405

### DIFF
--- a/cmd/deja/limit_refusal_test.go
+++ b/cmd/deja/limit_refusal_test.go
@@ -1,0 +1,29 @@
+package main
+
+import (
+	"strings"
+	"testing"
+)
+
+// `files` and `how` take --limit, so a reader arrives at `blame` and `last`
+// having just been told the flag works. Both refuse it, both for a reason, and
+// neither said what to type instead — the sentence `friction` and `install`
+// have printed all along (#3405).
+func TestTheRefusalNamesWhatTheCommandTakes(t *testing.T) {
+	if _, _, _, err := parseLast([]string{"--limit", "3"}); err == nil {
+		t.Fatal("last took --limit")
+	} else if !strings.Contains(err.Error(), "deja last 3") {
+		t.Errorf("last says %q, without the count it takes", err)
+	}
+	if _, _, _, err := parseBlame([]string{"x.go", "--limit", "3"}); err == nil {
+		t.Fatal("blame took --limit")
+	} else if !strings.Contains(err.Error(), "--all") {
+		t.Errorf("blame says %q, without the flag that widens it", err)
+	}
+	// Every other flag keeps the plain refusal: the note is about the one
+	// spelling a reader carries over from the commands beside these.
+	if _, _, _, err := parseLast([]string{"--nonesuch"}); err == nil ||
+		strings.Contains(err.Error(), "deja last 3") {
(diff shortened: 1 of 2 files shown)


vshulcz:
## Review

Adversarial pass by the author (the reviewer subagents on this machine are out of model credit — their transcripts read `You've reached your Fable limit`).

- `--limit=3` gets the same note as `--limit 3`, for both commands, via `flagName`.
- Every other flag keeps its old refusal, including blame's `--` sentence from #3395; `deja blame -- --weird.go` still routes the dash-leading path.
- The spellings the note points at both work: `deja last 2` prints two sessions, `deja blame <path> --all` prints every session.

One wording note for the record: "the count is a bare argument, `deja last 3`" is my phrasing, built from what the help already spells. Happy to take a different sentence.

Verdict: CONFIRMED.

Issue #816, opened 2026-08-01
a locked subdirectory silently removes sessions from the index, and doctor stays quiet


vshulcz:
One harness root, one subdirectory the process cannot read, sessions in both:

```
$ deja index                      # subdirectory readable
deja: claude: 2 sessions, 2 messages
$ chmod 000 ~/.claude/projects/locked && deja index
deja: claude: 1 session, 1 message
$ deja "anchor windlass"
deja: no matches in 1 indexed session — try fewer words or --re
$ deja doctor | grep -c warning
0
$ deja doctor | grep '^  claude'
  claude       found    /tmp/p12/.claude/projects  (2 files)     # was 3
```

#802 covers the case where nothing at all could be read; this is the partial one, which is the likelier shape — one project directory restored from a backup with wrong ownership, or a root-owned subtree — and it is quieter: sessions vanish from recall and the only trace is a file count nobody has memorised.

`walkFiles` drops `err` on the floor, so nothing downstream can know. The diagnostics side channel already exists for exactly this (`diagFileError`, and doctor's "ingest" line), and the walk is already happening.

PR #3408, opened 2026-09-08
fix(doctor): a store with one locked root is reported as partly readable

### LINKED ISSUE #3407: doctor --json says denied without partial when one of several store roots is locked
`doctor --json` reports `"state":"denied"` without `"partial":true` when the locked directory is one of several store roots and the rest of the store read fine.

`inspectDoctorStore` (cmd/deja/doctor_report.go:554) stats each path first and returns on the first refusal:

```go
fi, err := os.Stat(path)
if err != nil {
    if os.IsPermission(err) {
        store.State = "denied"
        store.Denied = path
        return store, time.Time{}   // Partial never set
    }
```

The walk below it, which is the other way a denial is found, does set it:

```go
store.Partial = len(check.files) > 0
```

## Reproduction

A hermetic HOME with both Cline generations, one transcript each, and the VS Code tasks tree locked:

```
$ chmod 000 "$HOME/Library/.../saoudrizwan.claude-dev/tasks"
$ deja doctor --json | jq -c '.stores[]|select(.name=="cline")'
{"name":"cline","state":"denied","paths":[".../.cline/data/sessions",".../saoudrizwan.claude-dev/tasks"],
 "files":1,"indexed_sessions":0,"denied":".../saoudrizwan.claude-dev/tasks"}
```

`files` is 1 — the CLI store read — so this is exactly the case `partial` exists to name: "sessions are missing from recall rather than the whole harness" (#816). The key is absent, which a reader takes as false.

## Counter-case

Lock a file rather than a directory and the same store reports `"partial":true`, through the `anotherFileOpens` branch further down — so the field works, and only this early return skips it.

Every harness with more than one root reaches the same return: cursor (`CursorUserRoot`, `CursorCLIRoot`), hermes, roo, antigravity, and cline since #3399.

Follow-on from #3399.

vshulcz:
`inspectDoctorStore` stats each root and returns on the first refusal, without setting `partial` — so a store whose other root read fine was reported as a wholly denied harness. The walk below it has set the field since #816.

```
before: {"name":"cline","state":"denied","paths":[…,".../saoudrizwan.claude-dev/tasks"],"files":1,
         "denied":".../saoudrizwan.claude-dev/tasks"}
after:  {…, "partial":true}
```

`files` is 1 there — the CLI store read — which is exactly what `partial` is for.

Reachable for every harness with more than one root: cursor, hermes, roo, antigravity, and cline since #3399. On this machine, with nothing locked, `doctor --json` is byte-identical before and after.

Fails before the change: `TestDoctorJSONSaysPartialWhenOnlyOneRootIsLocked` — `the CLI store read 1 file(s) and the report calls the harness wholly denied`.

Closes #3407

### DIFF
--- a/cmd/deja/doctor_denied_partial_test.go
+++ b/cmd/deja/doctor_denied_partial_test.go
@@ -1,0 +1,60 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// A store with more than one root can be half-readable: the CLI transcripts
+// open and the VS Code tree refuses. The walk below says so — "sessions are
+// missing from recall rather than the whole harness" (#816) — and the stat
+// loop above it returned before setting the field, so the machine form called
+// a half-readable store wholly denied (#3407).
+func TestDoctorJSONSaysPartialWhenOnlyOneRootIsLocked(t *testing.T) {
+	if os.Getuid() == 0 {
+		t.Skip("root reads everything")
+	}
+	tmp := hermeticEnv(t)
+	sessions := filepath.Join(tmp, "cline", "data", "sessions", "1757000001_aa11b")
+	legacy := filepath.Join(tmp, "vscode", "globalStorage", "saoudrizwan.claude-dev")
+	task := filepath.Join(legacy, "tasks", "1757000000000")
+	for _, d := range []string{sessions, task} {
+		if err := os.MkdirAll(d, 0o755); err != nil {
+			t.Fatal(err)
+		}
+	}
+	t.Setenv("DEJA_CLINE_ROOT", filepath.Dir(sessions))
+	t.Setenv("DEJA_CLINE_ROOTS", legacy)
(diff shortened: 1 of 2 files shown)


vshulcz:
## Review

- **A wholly locked store must not claim to be partly readable.** Locked the only root that holds anything, with the CLI store empty: `state` denied, `partial` absent. The guard is `len(check.files) > 0`, and the file collectors swallow EACCES, so files counts what was actually reachable.
- **A locked file rather than a locked directory** still goes down the `anotherFileOpens` branch and reports partial the way it did (#1747's test passes).
- **The stat loop's other two exits** — `os.Open` and `Readdirnames` — take the same path now, so all three refusals agree.
- On this machine with nothing locked, `doctor --json` is byte-identical before and after; `go vet`, `go test ./cmd/deja` pass.

Verdict: CONFIRMED.

PR #3410, opened 2026-09-08
fix(files): a quoted topic answers the same as the same words unquoted

### LINKED ISSUE #3409: deja files answers nothing for a phrase that deja answers with ten sessions
`deja files` answers "0 sessions mention …" for a phrase that `deja <phrase>` answers with ten sessions. The two commands read the same index; only `files` skips the tier ladder.

On this machine's index (a read-only copy):

```
$ deja "cursor turn timestamps" --limit 3
deja: no exact match; showing sessions ranked by relevance to the whole query
deja: showing 3 of 10 — add --all to see the rest
[claude] compare/fixes · Sep 7 · aed85e0adf6c5d723 — 13 matches
[claude] goprojects/deja-vu · Sep 8 · aqa-curso…40107c8c5b — 62 matches

$ deja files "cursor turn timestamps"
0 sessions mention "cursor turn timestamps", none of them recorded a file near it
```

A single word works — `deja files cursor` lists 158 sessions and the files they touched — so the surface is fine and the query shape is what kills it.

Measured over ten questions about work merged tonight, each one naming a file I know was changed: `deja files` found the file in 0 of 10. Every one of the ten answers was "0 sessions mention …" or "2 sessions mention …, none of them recorded a file near it".

## Cause

cmd/deja/files.go:105 calls `index.SearchWithRecovery`, which is the exact run. `deja <query>` (cmd/deja/main.go:951) and `deja ctx` (main.go:1322) both call `SearchWithRecoveryDetailed` and then answer from the tier retrieval chose — error signature, close, relevance — with the sentence that says which rung answered. `files` has no such branch, so a query the exact tier cannot satisfy ends as a miss.

## Counter-case

The sentence is not merely pessimistic wording: `--json` returns `files: []` for the same query, so a consumer sees nothing either.

## Expected

The surface most likely to be handed a sentence — "which files did we touch when we did X" — answers from the same ladder as its neighbours, and says which rung it answered from.

Evidence: ~/.cache/deja-hunt/2026-09-08-harness/180/recall180.txt (the ten questions), files-vs-search.txt (the pair above).

vshulcz:
The file list anchors on the words of the topic, and it took them from argv rather than from the query — so a quoted topic went in as one needle and `topicTimes` looked for the whole sentence inside a single message:

```
before:  $ deja files "cursor turn timestamps"
         0 sessions mention "cursor turn timestamps", none of them recorded a file near it
         $ deja files cursor turn timestamps
         files touched while working on "cursor turn timestamps" — 10 sessions

after:   both spellings print the same 10 sessions and the same file list
```

Retrieval was never the problem: probed on a fixture it returns the same session and the same close tier for both spellings.

Found while measuring `deja files` against ten questions about tonight's merged work — every one of them passed as a single argument, and every one answered "0 sessions mention …" (#3409).

Fails before the change: `TestFilesAnswersTheSameQuotedOrNot` — `quoting the same words empties the answer`.

Closes #3409

### DIFF
--- a/cmd/deja/files.go
+++ b/cmd/deja/files.go
@@ -186,7 +186,11 @@ func runFiles(dir string, args []string, stdout io.Writer) error {
 	// The words the person typed, not the expanded variant list: that one runs
 	// to eighteen stemmed forms for a two-word query, and requiring most of them
 	// in one message matches nothing.
-	needles := lowerAll(terms)
+	//
+	// The words, not the arguments: quoting a topic put the whole sentence in
+	// as one needle, which no single message carries, so `deja files "a b c"`
+	// answered nothing where `deja files a b c` answered (#3409).
+	needles := lowerAll(strings.Fields(q))
 	filtered := 0                    // recorded paths dropped by the repository filter
 	near := map[string]int{}         // path -> times touched near the topic
 	nearSessions := map[string]int{} // path -> how many sessions touched it near the topic
--- a/cmd/deja/files_tier_test.go
+++ b/cmd/deja/files_tier_test.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// The words a topic is asked in are the needles the file list anchors on, and
+// they were taken from argv rather than from the query: `deja files "a b c"`
+// looked for the whole sentence inside one message and found nothing, while
+// `deja files a b c` — the same words, different shell quoting — answered
+// (#3409).
+func TestFilesAnswersTheSameQuotedOrNot(t *testing.T) {
+	tmp := hermeticEnv(t)
+	t.Setenv("DEJA_INDEX_DIR", filepath.Join(tmp, "index.db"))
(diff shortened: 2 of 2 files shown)


vshulcz:
## Review

- **Nothing else moves.** `deja files cursor` and `deja files cursor timestamps` on a copy of this machine's index are byte-identical between base and branch — the only spelling that changes is the one that was broken.
- **The query is unchanged**, only the needles: `--json` still carries `"query": "cursor turn timestamps"` and now returns 10 rows instead of 0.
- **A topic with punctuation** (`deja files "no such table"`) answers 23 sessions rather than tripping on the split.
- **Is the split the right one?** `q` is the arguments joined with a single space, so `strings.Fields(q)` is exactly the words a reader typed; `terms` was only ever an accident of quoting. The anchor rule (half the words in one message) is untouched.
- **The neighbours do not have this split**: `deja how` answers the same quoted or not, `deja blame` takes one path and refuses several, `deja fix` answers the same either way. Checked on the real index; recorded as `[argv-vs-query-elsewhere]`, clean.

Verdict: CONFIRMED.

Issue #2537, opened 2026-08-29
a decision removed by forget is still served from the session-start cache


vshulcz:
`deja forget --session deja-note-…` removes a promoted note from `notes.jsonl` and tombstones its row. Search agrees immediately. The session-start block keeps handing the decision to agents:

    notes before: 1 record
    forget-note: removed 1 promoted note from …/notes.jsonl
    notes after:  0 records

    deja "retry budget"   no matches in 5 indexed sessions …
    session start         standing decisions in this project (promoted and still accepted — follow them unless the user overrides):
                            · the retry budget stays at 5; the pool change is what fixed the timeouts

The block is folded into the cached session-start digest, whose gate is the recall mode and the trust policy (`hookGate`) with a 60-second TTL. Nothing in that gate mentions the notes file, so a decision the user has just deleted is served to every session start until the TTL runs out and the refresh lands.

Sixty seconds of stale recall is what the cache is for and is fine for history. A removal is not history: `forget` is the privacy command, and the comment on its own error path says the one thing it must not do is leave a line standing after it reports success. Serving the text from a cache is that, one layer up.

Reproduces only when the block was served before the forget — which is the ordinary case, since it is served at every session start.

Fix: put the notes file's stamp in the cache gate, so any change to it — a promotion, a correction, a removal — retires the cached block at once. The file is stat'd, not read.

vshulcz:
Fixed in #2538, and in main since 659f9158.

It stayed open because the bug hunt merges into a dated branch first and that branch reached main as one squashed commit, so nothing closed it at the time.

PR #3413, opened 2026-09-08
fix(forget): the cached session-start digests go with the sessions

### LINKED ISSUE #3411: deja forget leaves the cached digest that quotes the forgotten session on disk
`deja forget` leaves the session-start digest that quotes the forgotten session sitting in the cache beside the index.

The per-prompt digest is cached one file per working directory, `<index>.hookcache-<hash of cwd>` (cmd/deja/hook_context.go:711). Nothing ever removes one: this machine has 40 of them, 160 KB, and the oldest is from 2026-07-24. They hold the block as prose:

```
$ python3 -c 'import json;d=json.load(open("~/.cache/deja/index.db.hookcache"));print(d["digest"][:120])'
✓ recalled from claude session · today
  - Session: **Users/shulcz** `2fc1d1ef-59f`
  - User: 1378-  }  1379-   return false
```

`index.Forget` (internal/index/privacy.go:393) writes a tombstone and rebuilds the records; the cache files are not in its path, and the CLI does not remove them either.

## What is and is not affected

Recall is not: an entry is served only inside its 60-second TTL, and the gate string keeps a stale-shaped one from being used at all. What remains is the content — after `deja forget --session <id>`, a grep of `~/.cache/deja` still finds the text of that session. That is the one thing the privacy command is for; #2537 closed the same shape for a promoted note that a cached digest kept quoting after `deja forget`.

## Reproduction

```
$ deja hook-context < payload.json > /dev/null    # writes <index>.hookcache-<hash>
$ grep -l zonkobuffer ~/.cache/deja/*.hookcache*
/…/index.db.hookcache-1a2b3c4d
$ deja forget --session s1
dropped: 1 session(s), 4 message(s)
$ grep -l zonkobuffer ~/.cache/deja/*.hookcache*
/…/index.db.hookcache-1a2b3c4d      ← still there
```

## Expected

A forget takes the cached digests with it. They cost nothing to rebuild — the next session start pays the ~120 ms the cache exists to save.

Evidence: ~/.cache/deja-hunt/2026-09-08-harness/183/hookcache.txt (the 40 files and their dates, read-only).

vshulcz:
The session-start digest is cached as prose beside the index — one file per working directory, `<index>.hookcache-<hash>` — and nothing ever removed one. After `deja forget`, the block quoting the forgotten session was still on disk: this machine carries 40 of those files, 160 KB, the oldest from 2026-07-24 with a user turn in it.

Recall was not serving them (an entry expires after a minute, and the gate string rejects a stale shape), so this is about the content, not the answer — the same distinction #2537 drew for a promoted note a cached digest kept quoting.

They are dropped on any real forget, not only a matching one: the cache is keyed by directory, not by session, so there is nothing to match against. The next session start pays the ~120 ms the cache exists to save.

Fails before the change: `TestForgetTakesTheCachedDigestWithIt` — `the forgotten session is still quoted in index.db.hookcache-b517da24`.

Closes #3411

### DIFF
--- a/cmd/deja/forget_hookcache_test.go
+++ b/cmd/deja/forget_hookcache_test.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+)
+
+// The session-start digest is cached as prose beside the index, one file per
+// working directory, and nothing removed one — so after `deja forget` the
+// block quoting the forgotten session was still on disk. Recall was not
+// serving it (the entry expires in a minute), but the privacy command is about
+// the content, not only the answer (#3411, the shape #2537 closed for a note).
+func TestForgetTakesTheCachedDigestWithIt(t *testing.T) {
+	tmp := hermeticEnv(t)
+	dir := filepath.Join(tmp, "index.db")
+	t.Setenv("DEJA_INDEX_DIR", dir)
+	root := os.Getenv("DEJA_CLAUDE_ROOT")
+	writeClaudeFixture(t, filepath.Join(root, "-w-api", "s1.jsonl"), "s1", []string{
+		`{"type":"user","sessionId":"s1","cwd":"/w/api","timestamp":"2026-07-01T10:00:00Z","message":{"role":"user","content":"the zonkobuffer overflows under load"}}`,
+	})
+	if _, err := captureRun(t, "index"); err != nil {
+		t.Fatal(err)
+	}
+
+	cwd := filepath.Join(tmp, "proj")
+	if err := os.MkdirAll(cwd, 0o755); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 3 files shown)


vshulcz:
## Review

- **Blast radius.** With `<index>.hookcache`, `<index>.hookcache-deadbeef`, its `.refreshing` sentinel, `<index>.dejavu`, `<index>.builtnote`, the index directory itself and another index's `other.db.hookcache-cafe` all in one directory: the three hook caches go, everything else stays.
- **A dry run keeps them** — the drop is inside `if !o.DryRun`, and a `forget --session … --dry-run` leaves the cached digest where it was.
- **A failed rebuild still drops them**: the call is before the error branch on purpose, because the tombstone is already written at that point and the cached prose is the thing this is about.
- **`--unforget` does not drop them**, which I left alone: a cache that omits a restored session is stale for at most its minute, and nothing is disclosed by that.
- **Cost** is one `os.Remove` per project a hook has run in (40 on this machine) and a ~120 ms rebuild at the next session start.

Verdict: CONFIRMED.

PR #3417, opened 2026-09-08
chore: benches, tests and comments stand on invented subjects

vshulcz:
The fixtures, bench arms and comments that illustrate the Russian-language rules were written from a real store, so they carry real session subjects, a real service name and a real project path. They are examples: nothing about them needs to be real.

Every one of them now stands on an invented subject with the same shape — same word forms, same hyphenated compounds, same sentence structure — so the rules they exercise are unchanged. Tests and benches pass as they are, no assertion moved.

Files: `cmd/deja/bench_prompt.go`, `cmd/deja/files.go`, `cmd/deja/statscard_privacy_test.go`, `cmd/deja/statscard_term_test.go`, `internal/bench/prompt.go`, `internal/digest/digest.go`, `internal/digest/decision_state_test.go`, `internal/digest/opening_now_test.go`, `internal/digest/plan_marker_test.go`, `internal/prompt/prompt.go`, `internal/sources/hermes_project_test.go`.


vshulcz:
Superseded: the same replacements went into main with the history rewrite, so this branch has nothing left to add.

PR #3419, opened 2026-09-09
fix(friction): treat undefined symbol as friction

### LINKED ISSUE #1551: "undefined symbol" is not friction, so linker failures never reach fix or the walls
Found while testing the failure-moment delivery (#1550): `ld: undefined symbol: pthread_setname_np` is not matched by `isFriction`.

The phrase list has `undefined:` (Go's compiler shape) but nothing for the linker's `undefined symbol:`, `undefined reference to`, or `symbol not found`. Those are among the most common build failures in C, C++, Rust with a C dependency, and cgo — exactly the errors people lose an afternoon to and would most want the remembered remedy for.

Consequence: no pair is mined for them, `deja friction` never lists them as a wall, and the new PostToolUse hook stays silent on the failure it would be most useful at.

Worth checking the same way the list grew last time: run the candidate phrases over a real store, count what they newly match, and confirm none of them are lines *about* errors rather than errors.

rainhuang0220:
Closes #1551.

`undefined symbol:` and `undefined reference to` now count as friction. The colon matches the linker diagnostic while avoiding ordinary prose that only mentions an "undefined symbol."

On a local store (1,389 sessions / 2.9M tool-output lines), the change introduced no incidental matches. Public linker diagnostics hit as expected, while the checked prose/source negatives stayed out.

### DIFF
--- a/internal/index/friction.go
+++ b/internal/index/friction.go
@@ -375,6 +375,7 @@ func isFriction(l string) bool {
 		// sentence someone writes about roles.
 		"duplicate key value", "deadlock detected", "access denied for",
 		"failed to connect to", "cannot import name", "symbol(s) not found",
+		"undefined symbol:", "undefined reference to",
 		"failed to push some refs", "acquiring the state lock",
 		"no space left on device",
 	} {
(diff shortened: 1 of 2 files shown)


vshulcz:
read the whole diff and ran the suite locally on top of current main — green. the part that makes this mergeable is the negatives: the echo, the comment and the quoted fixture, which is where the phrase list usually starts costing false positives. keeping the colon on `undefined symbol:` does the same job.

thanks — this closes #1551, and the dyld note is a good catch to leave in the test rather than in a comment nobody reads.

PR #3436, opened 2026-09-09
feat: recover working context automatically after compaction

chrisgeo:
Compaction drops the working evidence an agent needs to resume. This change automatically captures objectives, conclusions, recorded verification commands, explicit gaps/conflicts, provenance, and repository freshness from the current Claude Code or Codex transcript.

The existing local index retains bounded packets, and the next hook for the same session and workspace restores one within a 4 KiB limit. Redaction, exclusion rules, forgotten-session tombstones, rebuild preservation, and failed-capture handling apply throughout. Codex plugin wiring includes the recovery and editing hooks.

`deja stats` reports locally observed raw tool calls before the first edit, with median/p75 and incomplete observations kept separate. It uses the existing usage log and requires no agent checkpoint workflow.

Validation:

- Independent review and regression fixtures for capture/recovery, privacy, repository changes, malformed transcripts, and raw action counts.
- Full Go suite with race detection and coverage; all configured coverage floors met.
- `go vet ./...`, `go mod verify`, and golangci-lint 2.13.2.
- `sh scripts/e2e.sh` and a separate-process compaction CLI check.
- All six release-target builds; lexical recall@5 meets the CI floor.


vshulcz:
Read the whole diff. There is real work here — the cache has locking, budget trimming and its own tests, and the CLI surface is coherent. Two things stop me merging it as it stands, and one of them is not yours to fix.

Blocking, mine to decide with the maintainer, not in review: this changes what deja is on every surface. The skill on all harnesses now opens with `deja ctx resume` and demotes recall to "only when ctx needs historical evidence"; the MCP tool description says recall runs "after ctx_resume reports gaps". deja's whole claim is that it answers from history the agent already wrote, with no state of its own to maintain — a resume layer the agent checkpoints into is a different product, and a good one to argue for, but it needs to be argued for before it ships as the default instruction to every agent.

Blocking, yours: the PR is two subsystems. `internal/ctxcache` is the one the description covers; `internal/instructions` — a store, hooks, install, render, resolve — is not mentioned at all, and it is a second design with its own surface (`deja instructions …`). Please split them; each is easier to judge alone, and the second one I cannot review against a stated intent I do not have.

Smaller, all real:

- CI is red: `cmd/deja/ctx_lifecycle_instruction_test.go:26` calls `writeCLISkill` with one return value, and it returns two on current main since #3431. A rebase fixes it.
- Checkpoints are agent-written state deja stores itself, which is new for docs/SECURITY-MODEL.md — that file is what someone reads before trusting the tool with a work machine, and it does not know about `~/.cache/deja/ctx`.
- `deja ctx <query>` still works, which I checked for specifically. Good.

I will bring the framing question to the maintainer and come back with an answer rather than leave this sitting.

vshulcz:
Reviewed the new head and measured it. The framing objection is gone — skills untouched, recall keeps its place, everything behind `ctx --cache`, and `deja ctx <query>` still answers. Thank you for splitting it; that made this reviewable.

Numbers, all on this machine: a scratch git repo of 40 files for the cache, and a copy of a real 1400-session index for the comparison.

    ctx --cache checkpoint        130-140 ms
    ctx --cache resume (warm)      80-90 ms, 860-1080 bytes
    ctx --cache resume (miss)         70 ms,      837 bytes
    ctx --cache status                80 ms
    resume, 50 MB untracked file     120 ms
    20 checkpoints                 92 KB, 24 files

    what deja hands an agent today, same machine:
    hook-context (session start)     120 ms warm, 2620 bytes of history
    hook-prompt (per prompt)      220-870 ms,  1.4-2.2 KB, fires on real questions

    tools/list schema             3147 -> 3427 bytes (+280, ~70 tokens a session)

    bench prompt and recall: unchanged against main (cross_paired 5/51, r@1 0.20, mrr 0.457)

What that says. The cache is fast and its payload is bounded, and it does not touch ranking. What it returns, though, is only what the agent itself wrote: on a miss it is 837 bytes of empty envelope, and a hit is worth exactly the checkpoint discipline behind it. The session-start digest is 2.6 KB of history the agent never had to produce, for 120 ms. Those are different things, and the second one is what deja is for — so this reads to me as a complement for resume-after-restart, not a substitute, which is how your description now frames it.

Three findings that are mine to raise rather than taste:

1. **No redaction.** Everything deja stores today goes through `redact.Text` before it is written, and `DEJA_NO_REDACT` is the documented escape hatch. Checkpoints skip that pass entirely; SECURITY-MODEL.md says "no secrets scanner" and "do not checkpoint secrets", which documents the gap honestly but leaves deja with one store that redacts and one that does not. Running the same pass on checkpoint text costs almost nothing and keeps the boundary a single sentence.

2. **The history never shrinks.** `snapshots/` is append-only by design and `invalidate` does not delete — 20 checkpoints made 24 files here, which is nothing, but an agent told to checkpoint durable progress will make thousands, and there is no cap, no age, no `ctx --cache prune`. deja's own usage log rotates at 1 MiB for exactly this reason.

3. **Everyone pays the schema.** +280 bytes on `tools/list` per session, for a feature that is opt-in and experimental. Worth gating the ctx_* modes behind the cache having been used at least once, or an env switch, so agents that never checkpoint carry nothing.

Smaller: `worktreeDigest` reads every untracked file whole on each identity resolve — 50 MB of untracked data cost 40 ms here, which is fine, but it is unbounded by construction and a repo with a stray artifact directory pays it on every resume.

Nothing above is a merge blocker for me; 1 is the one I would want fixed before it ships.

vshulcz:
Thanks for carrying this through so many rounds, and for the parts that are plainly careful: the redaction boundary before anything is persisted, retention that validates history before it publishes a new pointer, and worktree hashing that streams and caps instead of reading artefacts.

The schema is the part I want to keep — conclusions, tests, gaps, conflicts, provenance, freshness checked against repository state. That is the right shape for "what was in flight".

What I measured before answering, on this machine's own store, over the 16 days the usage log covers:

| direction | events |
| --- | --- |
| deja → agent (per prompt, point of action, session start) | 3793 |
| agent asks (recall, recall_context, blame) | 572 |
| agent writes (`remember`, `handoff`) | 9 |

deja already has two places an agent can write, and both are nearly empty: 9 writes against 3793 injections, and 2 accepted notes after months of daily use. A cache the agent has to invoke is a third place of the same kind, and the numbers say it will not be filled — not because the API is wrong, but because writing is not something an agent does on its own initiative.

Most of the content is also derivable. Conclusions and tests are in the transcript already. Gaps and conflicts are the two fields only the agent can state, and they are the small part of the schema.

And there is a moment where working state is lost and deja is already standing there: `hook-precompact`. Today it clears the injected list and asks for a warmup, nothing else. Measured on real transcripts: a cold session takes a median of 10 actions before its first edit; after a compaction it takes 28, p75 61, over 86 compactions. That gap is what this PR is really aiming at, and it is the largest single cost I have been able to measure anywhere in the tool.

So the direction: your schema, filled at compaction from the transcript, with no second store and nothing asked of the agent. That is a much smaller change than this one, and it comes with a metric already taken — actions after a compaction before the next edit, 28 today. If you want it, it is yours; say so and I will leave it alone.

I am not merging this as it stands, and that is about the mechanism rather than the work.

chrisgeo:
The PR now captures working context automatically before compaction and restores it afterward, without requiring agent checkpoints.

- Uses the existing local index; removes the separate context-cache workflow.
- Preserves objectives, conclusions, tests, explicitly stated gaps/conflicts, provenance, and repository freshness.
- Limits storage to 32 packets of 24 KiB each, with redaction and removal through `deja forget`. Recovery output is capped at 4 KiB.
- Reports raw tool calls before the first edit, including median/p75 and incomplete observations. Real-world improvement remains unmeasured.
- Adds regression fixtures and fixes issues found during independent review.

Validation at `431f1cc0`: full Go race suite, coverage gates, lint, end-to-end tests, six release builds, and all 17 CI checks passed. The upstream merge conflict is resolved.

vshulcz:
two blockers, both small and both in your own code, then this is ready.

built 431f1cc0, `go test ./...` green, and i reproduced capture and recovery end to end on two of this machine's real transcripts: the 703 MB one that is being appended to right now (correctly refused — changing/mismatched, nothing stored) and a quiet 324 MB one (packet stored, delivered by the next hook, and the second hook returned only the environment block, so delivery-once holds). the 4 MiB tail + 64 KiB header bound does what it says: no full read of a 700 MB file.

**1. the objective takes a nudge in every language but english.** `trivialContinuation` (internal/digest/compaction_context.go:214) is an exact-match english list, so:

```
tail "продолжай"     -> objective "продолжай"
tail "давай"         -> objective "давай"
tail "давай дальше"  -> objective "давай дальше"
tail "ok"            -> objective "the retry budget keeps hammering the server on failure, cap it"
tail "continue"      -> objective "the retry budget keeps hammering the server on failure, cap it"
```

(three-turn session, `ExtractCompactionContext` direct.) `nudgeWords` and `worthAsAsk` in the same package — internal/digest/resume.go:99 — already cover both languages and the "nudge plus two words" shape, and `worthAsAsk` returns false for all three of those. on a real packet here the objective came out as a delegation rather than the task, which is the line a resuming agent reads first.

**2. conclusions are the first thing cut and recorded commands the last, and that costs the packet its value.** `RenderCompactionContext` writes objective → gaps → conflicts → tests → conclusions and `add` stops at the budget; `boundCompactionContext` drops `Conclusions[0]` first as well. on the 324 MB transcript that produced a 3787-byte packet holding eight recorded commands (~1.3 KB), every one of them `cd /Users/shulcz/.cache/deja-hunt/2026-09-08-harness/wt && …` — a temporary worktree that no longer exists — and exactly one conclusion, cut mid-sentence.

measured rather than asserted, on a local 9B model at temperature 0 with reasoning off, three resume questions whose answer is the recorded decision, budget held fixed and only the section order varied:

| packet | answered |
| --- | --- |
| none at all | 1/3 |
| commands fill the budget, conclusion cut | 1/3 |
| conclusion kept, commands cut | 3/3 |

a packet whose budget went to the command list performs like no packet. the two failures were not vague, either: the model answered "not stated" and "there is no decision on record", which is worse than silence because it reads as an answer.

two notes, not blocking:

- the freshness line spends 64 characters on `worktree=fed2c658…689e5`, which an agent cannot compare against anything. `clean` / `dirty` plus the number of changed files is the same fact in a form it can act on; head and branch are useful as they are.
- one recorded command came back as `git -c user.name=… -c user.email=…` verbatim. redaction covers credentials, not identity, so that goes into the packet and into manifest.gob. worth deciding on purpose rather than by default.

happy to push the nudge fix onto your branch if you would rather not round-trip it — it is one call to the helper that is already there.

chrisgeo:
@vshulcz By all means, if you have the fix you want ready, otherwise I can do it tomorrow.

vshulcz:
here it is — `maintainerCanModify` is false on this PR so i cannot push to the branch, and this applies cleanly on 431f1cc0. full suite green with it, and the two new tests fail without it.

one thing i changed my mind about while writing it: i first routed `trivialContinuation` through `worthAsAsk` and that was wrong. `worthAsAsk` also applies a minimum length, which is right for picking a handover ask and wrong here — it dropped "Repair the parser." (eighteen runes) and took your own fixture's objective with it. so this uses `nudgeWords` alone, plus the "nudge and a word or two" shape.

<details>
<summary>patch</summary>

```diff
diff --git a/internal/digest/compaction_context.go b/internal/digest/compaction_context.go
index 2eafd24f..89cb1b3a 100644
--- a/internal/digest/compaction_context.go
+++ b/internal/digest/compaction_context.go
@@ -5,6 +5,7 @@ import (
 	"regexp"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/vshulcz/deja-vu/internal/model"
 	"github.com/vshulcz/deja-vu/internal/redact"
@@ -211,12 +212,36 @@ func contextObjective(s model.Session, window []model.Message) model.ContextFact
 	return model.ContextFact{}
 }
 
+// trivialContinuation reports whether a user turn is a way of saying "carry on"
+// rather than the task being resumed.
+//
+// The list it held was English and exact, so a reader who types "продолжай" —
+// the commonest turn on the machine this was measured on — became the packet's
+// objective, which is the first line a resuming agent reads. nudgeWords covers
+// the same idea in the languages a real store holds.
+//
+// Only the nudge words themselves, not worthAsAsk: that helper also applies a
+// minimum length, which is right for picking a handover ask and wrong here —
+// "Repair the parser." is eighteen runes and is the task.
 func trivialContinuation(text string) bool {
-	text = strings.ToLower(strings.Trim(strings.Join(strings.Fields(text), " "), " .,!?:;"))
-	switch text {
-	case "yes", "y", "ok", "okay", "thanks", "continue", "go on", "proceed", "please continue", "continue please":
+	text = strings.TrimSpace(strings.Join(strings.Fields(text), " "))
+	if text == "" {
 		return true
 	}
+	low := strings.ToLower(strings.Trim(text, " .,!?:;"))
+	switch low {
+	case "y", "proceed", "please continue", "continue please":
+		return true
+	}
+	for _, nudge := range nudgeWords {
+		if low == nudge {
+			return true
+		}
+		// "давай дальше", "ok go on": a nudge and a word or two, nothing else.
+		if strings.HasPrefix(low, nudge+" ") && utf8.RuneCountInString(low) < askMinRunes+10 {
+			return true
+		}
+	}
 	return false
 }
 
diff --git a/internal/digest/compaction_nudge_test.go b/internal/digest/compaction_nudge_test.go
new file mode 100644
index 00000000..fa87edac
--- /dev/null
+++ b/internal/digest/compaction_nudge_test.go
@@ -0,0 +1,48 @@
+package digest
+
+import (
+	"testing"
+	"time"
+
+	"github.com/vshulcz/deja-vu/internal/model"
+)
+
+// The packet's objective is the first line a resuming agent reads, and the turn
+// before a compaction is as often "carry on" as it is the task. The English
+// exact-match list let every other language through: measured on a real store,
+// the objective came back as "продолжай".
+func TestTheObjectiveSkipsACarryOnInAnyLanguage(t *testing.T) {
+	task := "the exporter retries are hammering the server, cap them"
+	for _, nudge := range []string{
+		"продолжай", "давай", "давай дальше", "дальше", "ок", "спасибо",
+		"ok", "okay", "continue", "go on", "thanks", "proceed",
+	} {
+		got := ExtractCompactionContext(session(task, nudge), ExtractOptions{})
+		if got.Objective.Text != task {
+			t.Errorf("tail %q made the objective %q", nudge, got.Objective.Text)
+		}
+	}
+}
+
+// And a terse instruction is the task, not a nudge: the length bound that keeps
+// an acknowledgement out of a handover would have dropped it.
+func TestATerseInstructionIsStillTheObjective(t *testing.T) {
+	for _, short := range []string{"Repair the parser.", "cap the retries", "revert that"} {
+		got := ExtractCompactionContext(session("something earlier and longer than this", short), ExtractOptions{})
+		if got.Objective.Text != short {
+			t.Errorf("the objective dropped %q and took %q", short, got.Objective.Text)
+		}
+	}
+}
+
+func session(first, last string) model.Session {
+	now := time.Now().UTC()
+	return model.Session{
+		Harness: "claude", ID: "s1", Project: "app",
+		Messages: []model.Message{
+			{Role: "user", Text: first, Time: now},
+			{Role: "assistant", Text: "capped it at three attempts because retrying past that spread the outage", Time: now.Add(time.Minute)},
+			{Role: "user", Text: last, Time: now.Add(2 * time.Minute)},
+		},
+	}
+}
```

</details>

on the second blocker: that one is your call, not mine, so i have not touched it. the reorder is three lines in `RenderCompactionContext` and one case in `boundCompactionContext`, and the reason to do it is the table above — a packet whose budget went to the command list answered like no packet at all. if you would rather keep the current order, say so and i will merge as it stands; the objective fix is the one that cannot wait, because it lands on every packet this machine produces.

chrisgeo:
Done. Also updated to ensure the PR has maintainerCanModify to true in case there are any updates you need.

vshulcz:
the objective fix is right and it stays. one blocker in it, reproduced by running rather than reading: `strings.HasPrefix(low, nudge+" ") && runes < askMinRunes+10` swallows an instruction that opens with a nudge. asked of `ExtractCompactionContext`, seven tails came back as "carry on" and the objective stayed the older task — `давай починим экспортер` (23 runes), `давай вернём прежний таймаут` (28), `ok cap the retries at three` (27), `continue with the parser fix` (28), `yes revert it`, `no do not touch the cache`, `go fix the pool`. on this store `давай <do X>` is how an instruction is usually given, so the packet would keep naming the task the reader had just replaced. the fix is "nudges all the way down" (longest match first, punctuation normalised so `thanks, continue` and `ага, давай` still read as nudges) rather than a length bound.

i also took the ordering call rather than leaving it with you, and three more things the feature was leaving on the table. measured on the product's own output, not on hand-written packets — a hermetic stand with an empty index, a git workspace and a transcript of a session that did real work, then `hook-precompact` followed by the three delivery surfaces:

- **the budget dropped conclusions first.** on an over-budget packet that left 94 commands and zero conclusions. tests shrink first now, and the render prints conclusions before the command list because the render budget cuts from the bottom.
- **"what is left" never reached the packet.** `explicitOpenItems` takes a line opening with gap:/unverified:/missing:/blocked:. over 72137 assistant lines on a real store that shape appears **once**; the same line-initial shape with the words people use (`Осталось:`, `Что осталось:`, `**Открытый вопрос —`, `Still open:`, `TODO:`) appears **28** times, and all fourteen sampled were genuinely open items. the shape is what keeps it precise — "осталось" mid-sentence is usually "12 of 17 jobs are left", and a test pins those out.
- **every command was `[recorded]`.** the outcome is read from an exit marker in the command text, which codex and opencode append and claude does not. the output is already in the transcript one record later, and whether it is a failure is a question `index.FrictionLine` answers — so the pairing happens in `cmd/deja`, where both packages are in reach, instead of as a second copy of that rule. `[failed]` then `[passed]` on the same transcript now.
- **the freshness line printed fingerprints.** `head=<40 hex>, branch=master, worktree=<64 hex>, checked=<stamp>` — 150 bytes of a 1.4 KB packet, and the worktree hash is deja's own digest, unusable outside deja. it now states the verdict: `Repository unchanged since capture (branch master).`, or the changed sentence plus `Captured at <12 hex>.` so a diff is possible. the changed case also read as "Repository freshness unavailable: Repository changed since compaction" — the recovery path puts its verdict in the same field.

what that is worth, on the local 9B at temperature 0, arms being the packets the product prints (no packet / this pr at its head / the four fixes), scored on the fact itself:

| | no packet | this pr | with the fixes |
|---|---|---|---|
| six questions about the lost session | 1/6 | 5/6 | 6/6 |
| five questions where the state is only in the records | 1/5 | 3/5 | 4/5 |

the sharpest cell: "do I need to re-run the suite to know where it stands?" — without the outcomes the model answers "yes, re-run" both with and without a packet; with `[passed]` it answers "no, the recall shows the suite already passed". and "name the one thing left to do" is `not stated` without the open-item rule and named with it. small n, one model, so read 3/5 → 4/5 as direction and 1/5 → 3/5 as the feature working.

i could not push any of this to your branch — `maintainerCanModify` is set, but the push is refused with 403 because the fork is org-owned, which is where that setting stops applying. so this pr merges as yours, and the four commits land right after it as a follow-up; nothing here is waiting on you.

vshulcz:
thanks — two things specifically. you wired the objective to the shared `nudgeWords` instead of growing a second list next to it, which is the thing that keeps this codebase honest; and you left the ordering call to me rather than guessing at it, with a table showing why it mattered. the capture design is the part i did not have to touch at all: fingerprints, one delivery per compaction, a failure mark so an older packet cannot masquerade as this one.

merging. the four follow-up fixes are measured in the comment above and land right after.

Issue #3421, opened 2026-09-09
a deja built under another file name adopts the user's hooks and then piles up: 40 entries, 8 firing per prompt


vshulcz:
A machine here ended the day with **40 deja hook entries** in one Claude Code settings file — eight for each of `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `SessionStart` and `PreCompact` — every one of them pointing at a different self-built binary in a scratch directory, and the entry for the installed `deja` gone. Every prompt started eight hook processes; the usage sidecar shows the result:

```
{"t":"…T09:40:25.894585Z","kind":"dejavu","bytes":833,"sessions":1,"into":"46eb…"}
{"t":"…T09:40:25.894881Z","kind":"dejavu","bytes":833,"sessions":1,"into":"46eb…"}
{"t":"…T09:40:25.894637Z","kind":"dejavu","bytes":833,"sessions":1,"into":"46eb…"}
```

Three identical injections inside 300 microseconds, one prompt. Several of those binaries were older builds, so the same session was also served by versions predating that day's fixes — the reader sees deja contradicting itself.

Two defects compound:

**1. The temporary-binary guard covers only the system temp.** `exeIsTemporary` (`cmd/deja/wiring_state.go`) refuses to adopt the wiring when the running binary sits under `os.TempDir()`, `/tmp` or `/private/tmp` (#2684). A binary built anywhere else that is equally not-installed — a scratch directory under the home, a build tree, a checkout — passes the guard, and `refreshWiringAfterUpgrade` repoints every target at it on its first run.

**2. Our own entry is recognised by the binary's file name.** `isDejaBinaryToken` (`cmd/deja/install.go`) accepts only `deja` and `deja.exe`, so `entryHasCommand` does not see an existing entry whose command is `…/deja-arm hook-prompt` as ours. The comment above it says the trailing subcommand is matched "so an install from a new binary path replaces our old entry instead of leaving it to fire alongside the new one" — that holds for a new *path*, not for a new *name*, and the entries accumulate instead.

Reproduction, hermetic HOME: build the binary twice under two names (`deja-a`, `deja-b`), run `deja-a install claude-auto`, then run `./deja-b` once with that HOME. The settings file ends with two entries per event, both firing.

[...]

PR #3437, opened 2026-09-09
feat: add instruction memory resolver

chrisgeo:
## Summary

Adds experimental, opt-in instruction memory: approved rules are resolved separately from ranked conversation recall and delivered through Claude Code/Codex lifecycle hooks. This is the independent instruction subsystem split from https://github.com/vshulcz/deja-vu/pull/3436; it has no context-cache dependency.

- Independent `deja instructions` CLI, shell completion, and command reference.
- Versioned JSON registry with compare-and-swap snapshot replacement and revision archives; `apply` is not a merge.
- Repository, worktree, task, session, environment, tool and path scopes; mandatory accumulation, keyed conflicts, effective windows, expiry and revision-pinned exceptions.
- Explicit supersession with cycle/target validation and protected-rule checks; a successor cannot weaken absolute, mandatory, or owner-authored rules. Competing applicable successors remain an explicit conflict.
- Bounded instruction/runbook delivery through the existing `hook-context`, `hook-prompt`, and `hook-tool` process: one handler and one combined response per event. Opt-in installation migrates old separate handlers, preserves unrelated configuration and survives normal reinstall. `SubagentStart` retains its instruction-only handler.
- Usage documentation, storage assessment, reproducible SQLite/DuckDB microbenchmark probe, regression coverage, and a 90% instruction-package coverage floor.

The three follow-up documents describe proposed runtime work, not additional implemented capabilities: [runtime design](https://github.com/full-chaos/deja-vu/blob/feat/instructions-set/docs/instruction-runtime-design.md), [storage decision](https://github.com/full-chaos/deja-vu/blob/feat/instructions-set/docs/instruction-runtime-storage.md), and [implementation/acceptance plan](https://github.com/full-chaos/deja-vu/blob/feat/instructions-set/docs/instruction-runtime-plan.md).

## Scope and limits

This is advisory v0: it does not execute/certify runbooks or enforce actions. Approval fields are provenance, not authentication. The registry is separate from the history index, with no blanket SQLite-to-DuckDB migration. Proposed task state, safe capture, delivery deduplication and gates are not implemented. No new head-to-head database benchmark or live-agent delivery validation is claimed.

The timezone and SQLite fixture corrections address four existing portability failures so this PR can be tested independently of the context-cache PR.

## Validation

Fresh validation of the review fixes on macOS arm64 with Go 1.25.13 and the default temporary directory (no TMPDIR workaround):

- `go test ./internal/instructions -race -count=1 -cover` passed: **93.5%** statement coverage, above the 90% package floor.
- CLI shared-hook, installer, help-contract, plugin-parity and nudge regressions passed with `-race`. These cover both installation orders, ordinary versus opted-in matchers, old executable migration, shell-wrapper preservation, mandatory context budgets, and retrying recall/nudges/session starts omitted for budget.
- A built-executable smoke test exercised 16 event deliveries: Claude and Codex, both installation orders, and SessionStart/UserPromptSubmit/PreToolUse/SubagentStart. Each event had one Deja handler and delivered the synthetic approved rule exactly once.
- `go vet ./...`, `golangci-lint run ./...`, build, changed-Go formatting and `git diff --check` passed.
- All 15 GitHub Actions checks passed on `beedc77e`, including the full Linux and macOS race-test suites, Linux coverage thresholds, cross-compilation, lint, security checks and platform smoke tests. Live-agent/model-request delivery is not claimed by the synthetic hook smoke test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,7 +155,7 @@ jobs:
           declare -A floor=(
             [cmd/deja]=88 [internal/atomicfile]=77 [internal/bench]=100
             [internal/cjkfold]=100 [internal/digest]=90 [internal/embed]=93
-            [internal/index]=90 [internal/mark]=98 [internal/model]=100
+            [internal/index]=90 [internal/instructions]=90 [internal/mark]=98 [internal/model]=100
             [internal/nfcfold]=100 [internal/peers]=76 [internal/policy]=100
             [internal/prompt]=84 [internal/query]=100 [internal/redact]=94
             [internal/search]=93 [internal/sources]=87 [internal/stats]=91
--- a/cmd/deja/brief_test.go
+++ b/cmd/deja/brief_test.go
@@ -115,8 +115,18 @@ func TestBriefReplacesAQuietWeekWithTheSpanItHolds(t *testing.T) {
 	if err := runBrief(dir, &out); err != nil {
 		t.Fatal(err)
 	}
+	// The brief renders the span in the reader's local timezone. Derive the
+	// expected endpoints from the same servable overview it describes; a fixed
+	// UTC date made this assertion fail west of Greenwich despite the correct
+	// span being rendered.
+	overview, err := index.OverviewServable(dir)
+	if err != nil {
+		t.Fatal(err)
+	}
 	got := out.String()
-	if !strings.Contains(got, "covering") || !strings.Contains(got, "Apr 11 2026") {
+	if !strings.Contains(got, "covering") ||
+		!strings.Contains(got, overview.Oldest.Local().Format("Jan 2 2006")) ||
+		!strings.Contains(got, overview.Newest.Local().Format("Jan 2 2006")) {
 		t.Fatalf("want the span the index holds:\n%s", got)
 	}
 	for _, unwanted := range []string{"today      0", "this week  0"} {
--- a/cmd/deja/citation_safe_test.go
+++ b/cmd/deja/citation_safe_test.go
@@ -42,7 +42,10 @@ func TestRepeatLeadStripsDisplayControls(t *testing.T) {
 			t.Fatalf("opener carried a control rune %U: %q", r, opener)
 		}
 	}
-	if !strings.Contains(opener, "deja:8f2c19ab77d") || !strings.Contains(opener, "(claude, Jan 2") || !strings.Contains(opener, "'quoted'") {
+	// Citation dates are deliberately local (the surrounding user sees local
+	// dates too). This safety test must assert provenance, not a UTC calendar
+	// day that shifts for readers west of Greenwich.
+	if !strings.Contains(opener, "deja:8f2c19ab77d") || !strings.Contains(opener, "(claude") || !strings.Contains(opener, "'quoted'") {
 		t.Fatalf("opener lost its provenance or kept a quote that closes the line: %q", opener)
 	}
 }
--- a/cmd/deja/completion.go
+++ b/cmd/deja/completion.go
@@ -54,7 +54,7 @@ _deja_completion() {
     command="${COMP_WORDS[1]-}"
     action="${COMP_WORDS[2]-}"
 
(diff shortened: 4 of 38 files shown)


vshulcz:
Built it, ran it, measured it. Thank you for splitting this out of #3436 — it is a much easier thing to judge on its own, and the scope-and-limits section is the kind of honesty that makes a review possible.

Numbers, hermetic stand, macOS arm64:

    instructions example                369 bytes
    instructions resolve             < 10 ms, 360 bytes delivered
    instructions hook (SessionStart)   10 ms, 445 bytes

    the same job with what deja has today — `deja remember` plus the session-start digest:
    hook-context --plain               40 ms, 922 bytes, the rule in it twice

So the mechanism is cheap and the payload is small. Three things came out of the stand.

**1. The tests need a canonical TMPDIR, and macOS does not have one.** `internal/instructions` fails here on a plain run and passes with `TMPDIR=/private/tmp`:

    --- FAIL: TestWorkspace           hook_test.go:104
    --- FAIL: TestNewFileSymlinkTarget hook_test.go:156

`isolated()` (resolve_test.go:35) returns `t.TempDir()` uncanonicalised, and `Workspace`/`canonicalTarget` resolve symlinks, so `/var/folders/…` never equals `/private/var/folders/…`. One line — `d, err := filepath.EvalSymlinks(t.TempDir())` — fixes both. It is also what the macOS leg of CI is failing on right now.

**2. It installs a second deja hook per event.** After `deja install claude-auto` and then `deja instructions install --agent claude`, this machine's settings file holds:

    SessionStart       2  [deja hook-context, deja instructions hook]
    UserPromptSubmit   2  [deja hook-prompt,  deja instructions hook]
    PreToolUse         2  [deja hook-tool,    deja instructions hook]
    SubagentStart      1  [deja instructions hook]

Two deja processes on every prompt where there was one. That is the exact shape #3421 was about — a settings file that collects entries per event until every prompt pays for them — and it is why #3424 collapses deja's own duplicates and #3428 reports a session served twice in a second. The instructions block would cost nothing extra if the existing `hook-context`/`hook-prompt` carried it: one process, one block, one place for the budget to be enforced. Worth trying before adding a second chain.

**3. The overlap with promoted notes is the product question.** deja already stores rules the user settled — `deja remember`, with accepted/rejected/superseded and project scope — and surfaces them at session start, which is what the 922 bytes above are. What this adds over that is real and specific: path, tool, worktree, task and environment scopes; `must` semantics that accumulate rather than being ranked; effective windows and expiry; supersession with cycle checks. What it costs is a second store, a second CLI, a second hook chain and a second approval model. That trade is the maintainer's to make, not mine — I will bring it to him with these numbers rather than sit on the PR.

Smaller notes:

- The registry is JSON with compare-and-swap on a revision number; concurrent approvals from two agents will lose one with a clear error, which is the right failure. Worth saying in the doc that `apply` is not a merge.
- `instructions hook` reads the event on stdin and answers in the Claude/Codex shape; it fails open on a malformed payload, same contract as deja's own hooks. Good.
- Nothing here touches ranking, and the benches agree: prompt and recall arms unchanged.

vshulcz:
Thanks for splitting this out, and for being straight about the limit in the PR body — that live delivery to a model is not claimed by a synthetic hook smoke test is exactly the distinction that usually goes unsaid. The timezone and SQLite fixture fixes are separately useful; those I will take regardless of what happens to the rest.

On the registry itself, the number that decides it for me: deja holds 20 notes on this machine and 2 of them are accepted, after months of daily use, and agents wrote 9 things in the 16 days the usage log covers against 3793 injections. Scopes, supersession, effective windows, keyed conflicts and revision-pinned exceptions are a lot of machinery for a store that holds two rules — and the shape of the problem at two rules is not the shape it has at two hundred.

What I do want from this is the narrow part: a scope and a severity on the notes that already exist, so a rule can say where it holds and how hard, delivered through the hook that already carries standing decisions. No second registry, no new delivery chain. That is small enough to judge on its own, and if it turns out agents and people actually fill it, the rest of what you built here is the obvious next step with evidence behind it.

Not merging the registry as it stands. If you want the scope-and-severity change, take it.

PR #3452, opened 2026-09-10
fix(copilot-chat): a turn VS Code raised itself is not the person's words

Sora-bluesky:
## Summary

v0.19.5 took the harness's own text out of the user role across the parsers, but copilot-chat was
not in that pass. It still indexes every request as the person's words, including the ones VS Code
wrote itself.

Two markers say so, both on the request object and both preserved into the transcript:

- **`isSystemInitiated`** is set when the editor raises the turn rather than the person. The
  producer that ships today is the terminal tool's background-completion notification, and its
  message text is the terminal's own output. A search for what someone asked came back holding a
  command's stdout. The producer is described in microsoft/vscode#330561.
- **`confirmation`** is set when someone clicks a confirmation button. The text is the IDE's own
  wording, assembled from the button's label and the confirmation title, so a click was stored as a
  sentence someone typed.

Only the user text is dropped. The turn still times the session, the response is still indexed, and
the work records are untouched: a background completion did real work and that work is worth
finding. What goes is the claim about who wrote the text.

Both markers are ordinary fields of the request, so a `.jsonl` mutation log can add one with a later
`kind:1` Set. The guard runs after replay rather than only on the entry the log opened with, and a
test covers each marker arriving that way.

Same shape as the fixes this release already merged for the other harnesses: #3268 and #3334 for
Claude Code, #3306 for Copilot CLI, #3256 for Roo and Cline, #3300 for opencode, #3327 for
Antigravity, #3249 for aider. The guard mirrors `internal/sources/claude.go:195-199`.

## Test plan

New file `internal/sources/copilot_chat_hosttext_test.go`, ten cases:

- flat `.json` with `isSystemInitiated: true`, and with `confirmation: "Accept"`: no user message,
  and the assistant reply from the same request is still there
- an ordinary request, `false` with an empty confirmation, `null` markers, markers of the wrong JSON
  type, and a whitespace-only confirmation: the user message is kept. These are the controls, and
  they fail if the guard is written too broadly
- a `.jsonl` log where `isSystemInitiated` arrives as a later Set, and one where `confirmation` does
- a system-initiated turn whose response carries a terminal `toolInvocationSerialized`: the
  `RoleCommand` work record survives
- a session whose second request is host-authored, later, and carries no response: it still advances
  `Updated`. Moving the `Touch` inside the guard makes that case fail, which is how the decision to
  time the session from a dropped turn is pinned rather than just written down.

Measured here:

| | |
|---|---|
| the new tests against the parser at `main` | 3 fail, each reporting host text stored as a user message |
| after the change, `go test ./internal/sources/ -run CopilotChat` | 45 pass |
| full `internal/sources` package | failure set identical to `main` on this machine |
| `go vet ./...`, `git diff --check` | clean |

## Notes

The three Copilot Chat sessions on this machine all have `requests: []`, so I cannot put a
before-and-after count from a real store next to this. The fixtures are built from the serialized
shapes in the VS Code source rather than captured from a populated transcript, and the marker
semantics come from the same place.

`systemInitiatedLabel` is deliberately ignored. It supplies optional display text and falls back to
the message; `isSystemInitiated` is the field that says who raised the turn.

Two other defects in this file came out of the same reading and are deliberately left out of this
change, since they belong to the sidecar reader rather than the request loop. I will send them as
one follow-up: `chatResourcePath` never percent-decodes what `entries[].resource` writes as a URI
string, and the `workingSet` field is typed as a list of objects where the storage format writes a
list of pairs, so a non-empty one makes the whole sidecar read return early. Happy to file them as
an issue first if you would rather decide the legacy-format question before any code moves.

### DIFF
--- a/internal/sources/copilot_chat.go
+++ b/internal/sources/copilot_chat.go
@@ -577,7 +577,16 @@ func copilotChatAppendRequest(s *model.Session, req map[string]any) {
 	t := parseTimeAny(req["timestamp"])
 	if user := copilotChatUserText(req["message"]); strings.TrimSpace(user) != "" {
 		s.Touch(t)
-		s.Messages = append(s.Messages, model.Message{Role: "user", Text: user, Time: t})
+		system, _ := req["isSystemInitiated"].(bool)
+		confirmation, _ := req["confirmation"].(string)
+		// Some turns VS Code raises itself: a background terminal completion
+		// carries the terminal's output as the message, and a confirmation
+		// carries the button's own label. The turn happened, so it still
+		// times the session and its work is still recorded, but nobody typed
+		// the text and recall must not quote it back as though someone had.
+		if !system && strings.TrimSpace(confirmation) == "" {
+			s.Messages = append(s.Messages, model.Message{Role: "user", Text: user, Time: t})
+		}
 	}
 	at := parseTimeAny(req["responseTimestamp"])
 	if at.IsZero() {
--- a/internal/sources/copilot_chat_hosttext_test.go
+++ b/internal/sources/copilot_chat_hosttext_test.go
@@ -1,0 +1,180 @@
+package sources
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/vshulcz/deja-vu/internal/model"
+)
+
+func copilotChatHostTextSession(t *testing.T, body, ext string) model.Session {
+	t.Helper()
+	path := filepath.Join(t.TempDir(), "session"+ext)
+	if err := os.WriteFile(path, []byte(body), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	sessions, err := ParseCopilotChatFile(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(sessions) != 1 {
+		t.Fatalf("sessions = %d, want 1 (%#v)", len(sessions), sessions)
+	}
+	return sessions[0]
+}
+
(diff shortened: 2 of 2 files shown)


vshulcz:
Checked the whole diff and ran it here; no blockers, merging.

What I verified rather than took on trust:

- the guard runs after replay, as you say — `copilotChatSession` is handed the already-replayed `state`, so a marker that arrives as a later `kind:1` Set is in `req` by the time `copilotChatAppendRequest` reads it. Your `.jsonl` test covers it from the outside; the code path agrees.
- titles are unaffected: `copilotChatTitle` reads `customTitle`/`computedTitle` off the state, not the first user message, so dropping the text cannot rename a session.
- mutation: with the condition forced open, all three of your tests fail. They guard the thing they claim to.
- `go test ./internal/sources -count=1` green locally, and the type-safety rows (`"true"` as a string, `true` as a confirmation, null, whitespace) are the part I would have forgotten.

One consequence worth naming out loud, since it changes a count: a session whose only request was system-initiated, with no response and no tool work, now has no messages and is dropped. That is the right answer — nobody spoke and nothing happened — but it means the session total can move down by a few on a store with background completions in it.

Thanks for citing microsoft/vscode#330561 for the producer. "A search for what someone asked came back holding a command's stdout" is the sentence that made the case.

Sora-bluesky:
Thanks for merging, and for forcing the condition open rather than reading the tests.

The dropped session is a consequence I should have put in the PR body myself. Noted for the next one.

PR #3467, opened 2026-09-10
Xcode Coding Assistant sessions join their provider's history

### LINKED ISSUE #3466: Xcode Coding Assistant sessions are outside the stores deja reads
## Problem

Xcode Coding Assistant runs Codex and Claude with provider-specific state directories under `~/Library/Developer/Xcode/CodingAssistant`. Their conversation files use the same formats as the existing Codex and Claude harnesses, but `deja` launched outside Xcode does not discover those directories by default.

As a result, an Xcode conversation can be resumed in Xcode while remaining absent from `deja search`, `deja last`, `deja sources`, and `deja doctor`. Launching a provider from inside Xcode or manually relocating a root can make one store visible, but that does not give terminal runs a stable view of both the CLI and Xcode histories.

## Proposed solution

Discover these stores automatically on macOS and keep their sessions under the existing provider harnesses:

- Codex: `${HOME}/Library/Developer/Xcode/CodingAssistant/codex/sessions/**/rollout-*.jsonl`
- Claude: `${HOME}/Library/Developer/Xcode/CodingAssistant/ClaudeAgentConfig/${CLAUDE_CODE_PROJECT_DIR_NAME:-projects}/**/*.jsonl`

Preserve the ordinary CLI/config roots as the primary stores, deduplicate equal roots, and provide dedicated read-only overrides for hermetic tests and unusual installations. `deja sources` and both doctor formats should report every discovered root.

No new parser is required because Xcode uses the existing provider formats. Tests can use synthetic transcripts under `t.TempDir()`; no real conversation data is needed.

## Alternatives

- Point `DEJA_CODEX_ROOT` or `DEJA_CLAUDE_ROOT` at the Xcode store. This hides the ordinary store because those overrides intentionally define the complete answer.
- Run indexing from inside Xcode so its environment relocates the provider. This is host-dependent and still does not combine both stores reliably.

anupamchugh:
Xcode Coding Assistant writes Codex and Claude conversations outside their ordinary CLI stores, so a terminal `deja` run could miss work that remained resumable in Xcode.

This change discovers the Xcode stores under the existing `codex` and `claude` harnesses. It keeps the ordinary stores first, deduplicates equal configured roots, limits prompt history to the ordinary Codex store, and exposes `DEJA_XCODE_CODEX_ROOT` and `DEJA_XCODE_CLAUDE_ROOT` for hermetic or relocated setups. `deja sources`, text/JSON doctor output, the registry, and generated documentation now describe all roots.

All new transcript cases are synthetic and use temporary homes; no local conversations, paths, IDs, or machine output are included.

Validation:

- `go test ./... -race -count=1`
- `go vet ./...`
- `go build ./cmd/deja`
- `go test ./internal/sources -coverprofile=coverage.out -count=1` — 88.3% statements (CI floor: 87%)
- `git diff --check`

Closes #3466


vshulcz:
this reads well and the tests hold up — one question and two notes, none of them blocking.

what i checked, and the one thing i could not: this machine has `~/Library/Developer/Xcode` without a `CodingAssistant` directory, so every transcript in my runs was synthetic, written into the exact paths you named. the real layout is the one claim resting on your machine rather than on mine.

built the head, `go test ./...` green, `go vet` and `gofmt` clean, and i stood both stores up under a temporary home:

- both Xcode-side sessions index and answer: a codex rollout under `CodingAssistant/codex/sessions/…` and a claude transcript under `CodingAssistant/ClaudeAgentConfig/projects/…`, alongside the ordinary stores, which keep working.
- prompt history does stay with the ordinary Codex store, as the body says: a `history.jsonl` in the Xcode root is not read, and searching for its text finds nothing.
- equal roots are read once — `DEJA_XCODE_CODEX_ROOT` pointed at the ordinary root gives one root, one store.
- doctor names both roots in the text row and in `stores[].paths` in `--json`, with the file and indexed-session counts adding up.
- and removing `XcodeClaudeRoot()` from `ClaudeRoots` turns `TestClaudeReadsXcodeCCMirrorVariantsAndTranscripts` red, so the tests are doing work.

the quiet part of this change is the one i would have missed: replacing the `strings.HasPrefix` match with `filepath.Rel` containment. i put a rollout in `~/.codex-backup/sessions/…` and it is correctly left alone; before this it matched, because the backup's path starts with the root's. that fix is worth more than the feature it arrived with.

the question: folding both stores into the existing harnesses means `--harness codex` now mixes CLI work and Xcode work, and only the project name tells them apart. the rollout meta carries `"originator":"xcode"`, so a later split is possible if anyone asks for it — i am not asking for it now, just naming the consequence so it is a choice on the record.

two notes:

- on a machine with an Xcode-side `history.jsonl`, doctor says `1 not recognised here` for codex forever. it is accurate, and it reads like a defect; a word in that row — history comes from the CLI store only — would save somebody a question.
- `deja sources` prints the shared ancestor when a harness has two roots (`…/home`), which is thinner than the row doctor prints. not yours to fix, and now slightly more visible than it was.

fork runs needed approval, so the checks were sitting idle — approved, they are running now. merging once they are green.

vshulcz:
thank you — merging.

the thing i want to name is not the feature, it is the containment fix you folded into it. swapping the root match from `strings.HasPrefix` to `filepath.Rel` means a store sitting next to a root — `~/.codex-backup` against `~/.codex` — stops being read as the harness's own. i reproduced that on a stand and it behaves; nobody asked for it, and it was a real bug.

the rest earned its merge the boring way: synthetic fixtures in the exact paths, a test that goes red when the root is removed, the prompt-history limit honoured, equal roots read once, and both text and JSON doctor rows carrying the counts.

PR #3476, opened 2026-09-11
fix(recall): the conclusions list leaves out the session's other work

vshulcz:
`what this session concluded` is read off the whole session and sorted by age, so a session that worked on several things serves its newest three to any question about any of them. Over sixteen recall calls against a real store, 16 of the 38 conclusion lines served (42%) shared no word with the query or with the excerpts they sat under — and one long session handed the same three lines to a question about blame rows, one about the MCP dispatcher and one about a connection pool, because those happened to be the last things it concluded.

An agent reads those lines as the answer, so a line about other work is worse than no line: the excerpt stays either way, and the excerpt is the memory.

Each line now has to share a content word with the query or with the excerpts under it (substring too, so a question about the manifest matches a conclusion naming `ManifestBuiltAt` — word equality alone dropped that one). When nothing passes, the list still offers one line rather than none: the store is full of sessions whose words are in a different language than the query, and an unasked-for conclusion can be the one that helps. Three of them is wallpaper; one is an offer.

On the same sixteen calls: 38 conclusion lines → 27, and off-topic ones 16 → 4, all four being that single-line fallback. Page size is unchanged (63266 → 63172 bytes over the sixteen answers) because the page refills with excerpts — this buys content, not tokens. `bench context` (deja-recall 1096 tokens, coverage 1.00), `bench block`, `bench prompt` and `bench recall` are identical.

Two tests: one session that concluded both the thing asked about and two unrelated things, and one where nothing relates and a single line still has to come back. Reverting the gate turns both red.

### DIFF
--- a/cmd/deja/mcp.go
+++ b/cmd/deja/mcp.go
@@ -953,6 +953,87 @@ func attachAnswers(dir string, hits []search.Hit) {
 	}
 }
 
+// conclusionsAboutIt keeps the conclusions that have something to do with what
+// was asked.
+//
+// "what this session concluded" is read off the whole session, and a session
+// that ran for days concluded things about everything it touched. Measured on
+// sixteen recall calls against a real store, 42% of the conclusion lines served
+// shared no word with the query or with the excerpts they sat under — and a long
+// session served the same three lines to a question about blame rows, one about
+// the MCP dispatcher and one about a connection pool, because those were simply
+// the newest things it had concluded.
+//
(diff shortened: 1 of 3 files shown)


vshulcz:
Second commit: `TestTheFreedSlotIsFilled` pinned the old rule — its fixture is a session that concluded four unrelated things, which is exactly what the gate now drops, so it saw one line instead of three. The slot-filling property is still what it tests; every turn in the fixture is about the retry queue now.

Issue #2933, opened 2026-09-01
none of the three benches can see what the block says


vshulcz:
Following #2931, I ran the same test on the other two benches: break the thing the bench is supposed to be watching, and see whether the number moves. None of the three can see the content of the block deja hands an agent.

**`bench prompt` — blind to the block's content.** With `AutoRecallDigestForAsked` returning the empty string, every column is identical:

```
real questions     13/13  13   1.00
negative controls   0/12   —   false fires: 0
cross-paired        6/47   —   false fires: 6
reworded           15/20  15   1.00
```

That is not a bug — its columns say `fired` and `correct`, and it is a selection benchmark: does the hook fire, and is the session it picked the right one. Worth stating plainly because "prompt 13/13" reads like an answer-quality number and is not one.

**`bench recall` — blind to ranking, and `@10` is `@5` by construction.** Reversing the hit order before scoring leaves `recall@5` at 1.00. The reason is in the corpus: every one of the 50 queries returns **exactly 5 hits**, so "is the gold session in the top 5" is "was it returned at all", and the `@10` column can never differ from the `@5` column. Nothing about the order inside the page is measured.

**`bench context` — #2931**: coverage stays 1.00 with either half of the arm deleted.

Taken together: the three benches measure whether the hook fires, whether retrieval returns the right session, and how many tokens the answer costs. All three are worth having. None of them measures what the block says once the right session is in hand — which is #2243's subject, and the thing every digest change touches.

**Why this matters beyond bookkeeping.** Five digest PRs merged today (#2906, #2913, #2916, #2918, #2922) each report "three benches unchanged" as a no-harm check. That sentence was true, and it would have been true whatever those changes did to the content — including breaking it. The real evidence in those PRs is the before/after against a real store; the bench line was worth less than it looked. I have qualified each of them.

What would close the gap is one arm that scores the block rather than the choice: the #2243 instrument — replay a session, ask whether the block carries what the session settled — run over the seeded corpus where the settled fact is known by construction. That is a bench to write, not a line to change, and the corpus already plants the fact.

vshulcz:
All three findings addressed, two with fixes and one confirmed as by design.

**bench recall** (#2938): recall@1 and MRR added. `recall@5` and `recall@10` answered only "did the session come back" — every query returns exactly five hits — and now the order is measured: 0.20 and 0.457 today, 0.00 and 0.200 when the answering session is pushed to the end of every list.

[...]

PR #3481, opened 2026-09-11
fix(bench): each recall query names one session

vshulcz:
`bench recall` printed recall@1 0.20 and MRR 0.457 for a long time, and those were not measurements of the ranking — they were the ceiling. The corpus took the topic from `i%10` and the variant from `i%5`, so the variant is a function of the topic: each topic's five gold sessions got identical text, and the five queries built from them were *one string with five different single correct answers*. A perfect ranker scores 1/5 on that, and MRR (1 + 1/2 + 1/3 + 1/4 + 1/5)/5 = 0.4567 — the number printed, to three places.

Ten topics and five variants make fifty distinct pairs, so each query now names one session and each answer is the only session that says it. Same ranker, same seed: recall@1 0.20 → 0.74, MRR 0.457 → 0.857. The 13 queries that still miss are ranking work someone can now do and see.

recall@5 and recall@10 stay saturated for the reason #2933 recorded — five hits per query — and the comment on `benchMetric` now says which two columns can move and why they couldn't.

Test asserts what was missing: every query string distinct, one relevant session each, and that session the only one carrying its text. Reverting the corpus line turns it red with the five repeated queries named.

### DIFF
--- a/cmd/deja/bench.go
+++ b/cmd/deja/bench.go
@@ -24,6 +24,10 @@ type benchMetric struct {
 	// every query returns exactly five hits, so recall@5 says only whether the
 	// session came back at all and recall@10 cannot differ from it — reversing
 	// the ranking left both at 1.00 (#2933).
+	//
+	// They only started moving once each query named one session: while five
+	// queries shared a string and five sessions shared a text, a perfect ranker
+	// scored 0.20 and 0.457, which is what this printed for a long time.
 	RecallAt1             float64 `json:"recall_at_1"`
 	MRR                   float64 `json:"mrr"`
 	RecallAt5             float64 `json:"recall_at_5"`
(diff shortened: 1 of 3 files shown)


vshulcz:
Checked where the remaining misses are, so 0.74 is not read as headroom ranking work can close: all 13 are still ties, one level down. The topic phrase appears in several of its own variants — "cache invalidation" is in the exact, the typo ("the correct term is …"), the quoted and the code-review session — so the exact-phrase query has five legitimate matches and the label names one. Ten of the thirteen are that variant, three the quoted one.

So this PR removes the one-string-five-answers tie and the number moves; closing the rest means giving each of the fifty sessions a subject of its own rather than one subject in five dressings. Separate change, bigger than this one.

PR #3498, opened 2026-09-12
fix(copilot-chat): decode the edited-file URIs and read the older snapshot shape

Sora-bluesky:
Two problems in the sidecar reader that #3431 added to the VS Code Copilot Chat parser. Both were checked against the writer, `src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingSessionStorage.ts` in microsoft/vscode at `e4c7e7b1`.

The resource path was never percent-decoded. The reader accepts `recentSnapshot.entries[].resource` in two shapes. The current writer emits `URI.toString()` (line 176), which is percent-encoded. `chatResourcePath` took that string as-is, so a resource written as `file:///c%3A/work/my%20app/main.go` ended up in the files record as `/c%3A/work/my%20app/main.go`. This hits any path with a space or a non-ASCII character, so it is not limited to Windows drive letters. The string form is now parsed with `net/url`, which decodes once. The object shape (`UriComponents`, whose `path` is already decoded) is left as it was, so a literal `%25` in a path is not decoded twice. A file URI with an authority is kept as `//host/path`. Other schemes keep the path and drop the authority, as before.

The older snapshot shape failed to unmarshal. `recentSnapshot` is a union (line 283): v2 is `{stopId, entries}`, v1 is `{requestId, workingSet, entries}`, and `workingSet` in v1 is `ResourceMapDTO`, an array of `[uri, state]` pairs (line 275). The struct declared it as an array of objects, so `json.Unmarshal` failed on any v1 file whose working set was not empty, and the files record was silently dropped. The conversation itself was fine. Only the sidecar record went missing. VS Code's own reader takes `entries` from either shape and never reads `workingSet` (lines 66-68), so the struct now does the same. I have not measured how many v1 files exist on real installs. The state files on my machine are all v2.

A resource that decodes to a CR or LF is rejected before trimming, because the files record joins paths with newlines and a decoded `%0A` would otherwise split one resource into two lines. This mirrors the guard `copilotChatRefPath` already has.

Two side effects to be aware of:

- The sidecar file is not part of the transcript fingerprint, so sessions already indexed will keep the old paths until they are re-indexed for some other reason.
- A follow-up to the check on #3452, which said titles are unaffected. For a v1 transcript the parser has only `customTitle`, which as far as I can see the session storage writes only after a rename. Those sessions are named by the index from their user turns (a thin opener like "hi" gives way to the next substantial one). A session whose first turn was a background terminal completion used to be titled by that output. It is now titled by a turn the user wrote.

Tests: four new cases in `copilot_chat_edited_files_test.go` (percent-decoding, the v1 shape, decoded newlines including trailing ones, a file URI with an authority). All four fail against the previous `chatResourcePath` and struct, and pass with this change. `go vet` is clean.

### DIFF
--- a/internal/sources/copilot_chat.go
+++ b/internal/sources/copilot_chat.go
@@ -438,8 +438,7 @@ func copilotChatSession(path string, state map[string]any) ([]model.Session, err
 //
 // The transcript names a path only when the text does, so a chat that edited
 // eleven files left one `files` record. Measured on this machine: 48 of those
-// state files, and while `recentSnapshot.workingSet` is empty in every one of
-// them, their entries carry 145 resources between them (#3381).
+// state files, whose entries carry 145 resources between them (#3381).
 func appendChatEditedFiles(s *model.Session, path, id string) {
 	if !IndexToolPaths() || id == "" {
 		return
@@ -451,9 +450,6 @@ func appendChatEditedFiles(s *model.Session, path, id string) {
 	}
 	var state struct {
 		RecentSnapshot struct {
-			WorkingSet []struct {
-				Resource any `json:"resource"`
-			} `json:"workingSet"`
 			Entries []struct {
 				Resource any `json:"resource"`
 			} `json:"entries"`
@@ -472,9 +468,6 @@ func appendChatEditedFiles(s *model.Session, path, id string) {
 		seen[p] = true
 		paths = append(paths, p)
 	}
-	for _, e := range state.RecentSnapshot.WorkingSet {
-		add(e.Resource)
-	}
 	for _, e := range state.RecentSnapshot.Entries {
 		add(e.Resource)
 	}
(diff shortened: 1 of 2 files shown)


vshulcz:
both bugs reproduce, and the four tests are honest: i ran them against main's `chatResourcePath` and struct before reading the diff — all four fail, and the v1 case comes back as an empty files record rather than a wrong one, which is the half that was silent.

measured on this machine's store while reviewing: 48 `chatEditingSessions` state files, 145 entries between them, **every resource in the string form with scheme `file`**, 0 of them percent-encoded (no spaces or non-ASCII in these paths). so the branch you fixed is the only one that occurs in practice here, and the bug shows up on anyone whose paths have a space — which is most Windows installs and plenty of macOS ones. 47 of the 48 snapshots are v2 and one carries a `workingSet` key that is empty, so i cannot add a real v1 sample either; your "i have not measured how many v1 files exist" is the right way to have put it.

`copilotChatRefPath` takes zero string refs on this store (the transcript hands it `UriComponents` objects), so mirroring its newline guard rather than touching it is the right call and there is no second caller left with the same bug.

two notes, neither blocking:

- dropping the whole resource when it decodes with a trailing `%0A` rather than trimming it is the safer side and i would keep it, but it is a path we lose rather than a path we repair. fine as documented.
- `internal/sources` coverage lands at 87.9% against the 87 floor, so the four cases pay for themselves.

the working-set removal is also cleanly argued: the old struct could never parse `ResourceMapDTO`, so those resources were never in a files record — the field was costing us the whole unmarshal and buying nothing.

vshulcz:
merged. two things i want to name: you read the writer (`chatEditingSessionStorage.ts` at a pinned sha) instead of guessing at the shape, and you said plainly which part you had not measured rather than rounding it up to a claim. that is the difference between a report i have to re-verify and one i can act on.

the percent-decoding in particular is a bug nobody on this machine could have found — every path in my own store is ASCII with no spaces.

Sora-bluesky:
Thanks for merging, and for the measurement. 48 state files, 145 entries, all string-form `file` URIs and none percent-encoded matches what I would expect from an ASCII store, and it is useful to have it on record that the string branch is the only one that occurs in practice. The 47 v2 plus one v1-keyed snapshot with an empty `workingSet` also explains why the old struct never failed on either of our machines: an empty array unmarshals fine, so the failure needed a v1 file that actually had pairs in it.

On the trailing `%0A` note, agreed that it is a path lost rather than repaired. I went with dropping it because a filename that really ends in a newline and one that was written with a stray newline look the same after decoding, and guessing which one it was seemed worse than skipping the record. If a real store ever shows one, that would be the moment to revisit.

Good to have the `copilotChatRefPath` count too. Zero string refs means its string branch is simply not exercised on that store, not that it is clean: that branch does not decode, so a string ref ending in `%0A` would pass through as-is. With no caller producing one, I left it alone rather than change code nothing reaches.

PR #3508, opened 2026-09-13
fix(deepseek): read the session.v3.jsonl logs dsh writes now

huszarrichard:
## Summary

dsh now names a session log `session.v3.jsonl.zstd`. The records inside are the shapes `ParseDeepSeekFile` already reads: a `session` header with `cwd`, then `user/message`, `assistant/message`, `tool/result`, `session/title`, `step/end` and `turn/end`. deja accepted only `session.jsonl` and `session.jsonl.zstd`, in two places:

- `DeepSeekSessionFiles`, used by discovery and by a rebuild;
- the `deepseek` file kind's `Match`, used by the incremental index.

On a store written by dsh 0.1.1-rc.2 this indexed 4 of 13 sessions, only the ones that still had a log under the old name. `deja doctor` reported `(6 files, 4 indexed sessions)`. A session directory can hold a header-only `session.jsonl.zstd` beside its `session.v3.jsonl.zstd`, which is why some old names are still on disk.

Patching discovery alone is not enough, and that is how the second place turned up. A rebuild then reads the v3 logs, but a new or growing one comes through the incremental pass as `incremental index changed_files=1 removed_files=0 sessions=0`. It stays unsearchable until the next rebuild.

Both paths now match on one list of log names (`deepSeekLogNames`), so a future name cannot reach one path and miss the other. `docs/registry` lists the v3 name. `harnesses.html` and `deepseek.html` are regenerated from `registry.json` with `scripts/genmatrix` and `scripts/genregistry`.

## Test plan

- `TestDeepSeekAcceptsEveryLogName` checks that discovery and the registry kind accept all four names, and reject `session.lock` and a `.bak`. On main it fails for both v3 names, in both places.
- `TestParseDeepSeekFileReadsAV3Log` checks that a v3 log parses to the same session as the old name, raw and zstd-framed (the framed case runs when the CLI is present).
- `TestDeepSeekV3LogReachesTheIncrementalIndex` starts from a first build, adds a new `session.v3.jsonl`, and checks that it joins the index through the incremental path and is searchable. It then appends a turn and checks that the turn is searchable after the next pass. On main it fails with `after a v3 log arrived there are 1 deepseek sessions, want 2`.
- `go test ./... -race -count=1`, `go vet ./...`, `go build ./cmd/deja`, and `gofmt -s` on the changed files all pass, apart from the two tests in the notes.
- Coverage is 87.5% for `internal/sources` (floor 87) and 90.0% for `internal/index` (floor 90).
- On a real store (dsh 0.1.1-rc.2, macOS arm64), a v0.20.0 build with this change indexed 13 dsh sessions instead of 4. A session copied in afterwards and a zstd frame appended to an existing v3 log both reached search through plain `deja index`, without `--rebuild`.

## Notes

- `TestParsePiFile` and `TestParseQwenFile` fail on this machine on unmodified main too. `/tmp/deja-vu` exists here, so the project decoder resolves `-tmp-deja-vu` against the filesystem to `tmp/deja-vu`. This is unrelated to the change, and a CI runner will not have that directory.
- After switching to a build with this fix, plain `deja index` kept reporting `index is up to date` and skipped the v3 logs that were already on disk. Only `deja index --rebuild` picked them up. This may deserve a line in the release notes.
- The dsh auto-recall plugin has a related problem: it asks deja about the directory dsh was launched from rather than the session's workspace. That fix is a separate PR.


vshulcz:
both places reproduce, and finding the second one is the part that makes this a fix rather than a patch — discovery alone would have left a v3 log unsearchable until the next rebuild, which is the failure that looks like deja working.

ran your tests against main before reading the diff:

```
--- FAIL: TestDeepSeekAcceptsEveryLogName
    DeepSeekSessionFiles found session.v3.jsonl = false, want true
    deepseek kind matches session.v3.jsonl = false, want true
--- FAIL: TestDeepSeekV3LogReachesTheIncrementalIndex
    after a v3 log arrived there are 1 deepseek sessions, want 2
```

and `genmatrix`/`genregistry` on your branch produce no drift, so the generated pages match the registry rather than being hand-edited.

what i could not check here: this machine's dsh predates the rename — every session under `~/.dsh/sessions` is still `session.jsonl.zstd` — so the 4-of-13 number is yours and i am taking it as reported rather than confirming it.

one blocking thing, and it is the note you already wrote under "after switching to a build with this fix, plain `deja index` kept reporting `index is up to date`". that is not a release-note item, it is a missing line: `internal/index/index.go` carries a content `version` whose entire job is to make a parser change reach stores that were already indexed. bumping it to 39 turns your note into nothing anyone has to do — without it, a machine with v3 logs on disk stays at 4 of 13 until some unrelated bump happens to force a re-read.

i am not sending you back for it: i will land the bump and its version note as a follow-up right after this merges, the way the compaction work went in. saying so here rather than doing it silently.

two other things from your notes, both ours:

- `TestParsePiFile` and `TestParseQwenFile` do fail on unmodified main when `/tmp/deja-vu` exists — i made that directory and reproduced both. the project decoder resolves the encoded name against the filesystem, so a real directory of that name wins over the fixture's intent. filing it; it is a test that depends on the absence of something outside its own temp dir, which is our bug and not yours to carry.
- the plugin's launch-directory problem in #3509 is a good split. reviewing that one next.

vshulcz:
merged. the thing worth naming: you found the second place. a fix to discovery alone would have passed review, shipped, and left every new v3 log unsearchable until someone happened to rebuild — the failure mode that looks exactly like deja having nothing to say. putting both paths on one list so a future name cannot reach one and miss the other is the shape that keeps it fixed.

the version bump is landing now as promised, so nobody has to run `--rebuild` to get the logs that are already on disk.

Issue #3502, opened 2026-09-13
an upgrade leaves every hook pointing at a binary that is gone, and doctor says nothing


vshulcz:
An install written before 0.19.5 names the binary it was installed from. `brew upgrade` moves that binary, the path stops existing, and every hook entry pointing at it exits 127 — session-start recall, the per-prompt block, the point-of-action line and the compaction capture all stop, silently.

`deja doctor` does not catch it. `doctorLauncherNote` returns early unless the config already names `~/.config/deja/bin/deja-hook`:

```go
b, err := os.ReadFile(path)
if err != nil || !strings.Contains(string(b), launcher) {
    return ""
}
```

So exactly the configs that need the warning — the ones still naming a Cellar path — are the ones it says nothing about. Reported from the field in #3500, with `/opt/homebrew/Cellar/deja-vu/0.19.3/bin/deja` in `~/.codex/hooks.json`, `~/.codex/config.toml` and Claude's hook entries while the terminal ran 0.20.0.

Two parts:

- **doctor**: check that whatever an entry names is there and executable, launcher or not, and say which command rewrites it.
- **install**: rewriting an entry whose target is gone is not a migration anyone should have to ask for. The launcher exists so a moved binary costs one file (#3426); an entry written before it should be lifted onto it on the next `deja install`, and the first-run notice should say so when it finds one.

Worth a regression case per surface: a config naming a path that no longer exists, a config naming the launcher with no deja behind it, and a config already correct.

vshulcz:
Correcting the headline of this issue: **doctor does report a dead hook binary**, and I should have reproduced that before writing that it says nothing. I read `doctorLauncherNote`, which only speaks about the launcher, and missed `hookExeNote` beside it, which is the check that actually fires.

Built the state from the report — hook entries and an MCP entry naming a Cellar path that no longer exists — on Claude Code, which has no approval step in the way:

```
claude-code  wired       ~/.claude/settings.json
             runs /…/Cellar/deja-vu/0.19.3/bin/deja, which is not there — `deja install claude-auto` rewrites it for this binary
```

and on Codex, where the MCP row says the same thing about `config.toml`.

[...]

PR #3509, opened 2026-09-13
fix(dsh): recall the session's workspace, and load under a CommonJS home

huszarrichard:
## Summary

This fixes two problems in the dsh auto-recall plugin. Both are in the file `deja install deepseek-auto` generates (`dshAutoJS`), and the first is also in the published `extensions/dsh` package.

**The plugin asks deja about the launch directory, not the session's workspace.** Both `hook-prompt` and `hook-context` were sent `cwd: process.cwd()`. dsh never changes directory, and one `dsh web` process serves sessions from every workspace, so `process.cwd()` is wherever dsh was launched. When dsh was started from `~` and used in `~/work/shop`, every recall and the session-start digest were scoped to `~`, and the project's own history never arrived.

The session's directory is in its header, which the agent exposes as `agent.session.header.cwd`. That is `Session.header: SessionHeader` in `@deepseek-ai/dsh-session`, and `readonly session: Session` in dsh-agent's runtime types. Both plugins now send that value and fall back to `process.cwd()`.

The recall cache had the same blind spot. It was keyed on the prompt alone, so a process that got the same question in two workspaces gave the second workspace the first one's recall. It now compares the workspace as well.

**The generated plugins fail to load under a CommonJS `package.json`.** `auto.js` and `command.js` are ES modules, written into `$DSH_HOME/plugins/deja/` with no `package.json` of their own. Node takes a file's module type from the nearest `package.json` above it. A home directory whose `package.json` declares `"type": "commonjs"` makes dsh print `Failed to load the ES module ... Make sure to set "type": "module"` and skip both plugins. The installer now writes `{"name": "deja", "type": "module"}` into the plugin directory before the plugins. The name is what `mentionsDeja` recognises, so an uninstall treats the file's snapshot as deja's own, and uninstall already removes the directory. The npm package declares `"type": "module"` itself, so it is not affected.

## Test plan

- `TestDeepSeekAutoPluginAsksAboutTheSessionsWorkspace` installs the plugin under a home directory that has a CommonJS `package.json`, and loads the generated `auto.js` in Node the way dsh does. It drives the digest and the recall for two sessions in two workspaces, with a stand-in deja that records its stdin, and checks that deja was asked about `/work/alpha`, `/work/alpha`, `/work/beta`, `/work/beta`. It is skipped on Windows and when node is missing. On main it fails at load with `SyntaxError: Cannot use import statement outside a module`.
- `TestInstallDeepSeekDeclaresThePluginsAsModules` checks that the marker is written, that a second install reports unchanged, and that uninstall removes the directory. It fails on main.
- A new case in `extensions/dsh/test/pure.test.js` pins the header lookup, the absence of a `cwd: process.cwd()` payload, and the workspace-aware cache guard. It fails on main.
- The existing DeepSeek installer tests pass unchanged, including the `prompt !== asked` guard check. `node --test test/*.test.js` in `extensions/dsh` passes (18 tests).
- `go test ./... -race -count=1`, `go vet ./...`, `go build ./cmd/deja`, and `gofmt -s` on the changed files all pass, apart from the two tests in the notes. `TestEveryInstalledConfigIsRecognisableAsDejas` covers the new `package.json`. Coverage for `cmd/deja` is 90.0% (floor 88).
- Against a real index, I loaded the patched plugin directly in Node and asked the same prompt from two sessions. The session in a project workspace got that project's recall, and the session in the home directory got the home directory's.

## Notes

- Existing installs keep their old `auto.js` until `deja install deepseek-auto` is run again.
- `TestParsePiFile` and `TestParseQwenFile` fail on this machine on unmodified main too, because `/tmp/deja-vu` exists here. They are unrelated to this change.
- The v3 session log names are a separate PR.


vshulcz:
both bugs reproduce, and the load failure is the worse of the two: a plugin that never loads is indistinguishable from deja having nothing to say.

your tests against main, before the diff:

```
--- FAIL: TestInstallDeepSeekDeclaresThePluginsAsModules
    no package.json beside the plugins
--- FAIL: TestDeepSeekAutoPluginAsksAboutTheSessionsWorkspace
    node could not run the plugin: exit status 1
    Failed to load the ES module: …/.dsh/plugins/deja/auto.js
    import { execFileSync } from "node:child_process";
```

the workspace half i checked against a real session on this machine rather than the types: the header dsh writes carries it —

```
header keys: ['type', 'version', 'id', 'createdAt', 'cwd', 'delegationDepth']
cwd: /private/tmp/dshproj
```

— and `process.cwd()` under `dsh web` is wherever the server was started, which is exactly the mismatch you describe. one process serving several workspaces makes that the common case rather than the edge one.

the cache key is the same bug one layer down and i would have missed it. keying on the prompt alone means the second workspace is served the first one's answer, which is worse than no recall: it is confidently the wrong project.

nothing blocking. two notes:

- writing `{"name": "deja", "type": "module"}` rather than a bare `{"type": "module"}` is the right call for the reason you give — `mentionsDeja` has to recognise the file as ours — and it is worth the line of comment it has.
- "existing installs keep their old `auto.js` until `deja install deepseek-auto` is run again" is the same shape as #3502, where hook entries keep naming a binary that was moved by an upgrade. #3511 makes a bare `deja` say that out loud for the dead-binary case; a plugin that is merely out of date is quieter than that and still worth a way to notice. not this pull request's problem.

vshulcz:
merged. two things: you tested the generated plugin by loading it in node the way dsh does, rather than asserting on the string deja writes — which is the only way the CommonJS failure could have been found at all. and the cache key. a recall keyed on the prompt alone hands the second workspace the first one's answer, and that is worse than silence: it is another project's history, offered confidently. i would have missed it.

Issue #2970, opened 2026-09-02
Claude Code deletes transcripts after 30 days, and the next reindex deletes them from deja too


vshulcz:
Claude Code keeps session transcripts under `~/.claude/projects/` for **30 days by default** and then deletes them (`cleanupPeriodDays`, [data-usage docs](https://code.claude.com/docs/en/data-usage#data-retention)). People find out afterwards: anthropics/claude-code#62476 "silently deletes conversation transcripts after 30 days" is open with 24 reactions; #64999 calls it "silently and irreversibly".

deja indexes those files, so it holds the content. But the index follows the file: when the source is gone, the next rebuild drops the session.

**Reproduction** (hermetic test, `cmd/deja`):

```
index a claude transcript containing "zorblax"     → deja "zorblax" finds it
os.Remove(transcript); index.Ensure(dir, "", true) → deja "zorblax" finds nothing
```

```
before delete: found=true | after delete+reindex: found=false
```

**Cause:** `internal/index/ingest.go` around line 2829 — for per-file harnesses "a removed or changed file's old records are already dropped by the two clauses above". `DeepVerify` already recognises the state as `orphan-file` (`deepverify_test.go`, `TestDeepVerifyFlagsDeletedSource`) but reports it rather than keeping the records. Mount-point disappearance is already treated as "not a deletion" (`missingTree.mount`), so there is precedent for retaining records whose source is absent.

**Expected:** the sessions deja indexed stay searchable after the harness's own cleanup removes the file. That is the one thing a memory tool can offer against a 30-day default, and it is what "starts full" implies: the history you had is not lost because a client's housekeeping ran.

**Tension to decide, not to assume:**
- A user who deletes a transcript on purpose expects it gone from recall too. `deja forget` exists for the deliberate case; the question is which default the *implicit* case gets.
- If retained: the record's `Path` points at a file that no longer exists — `show`, `resume`, `export` need to say so rather than fail.
- Retention could be scoped: keep records whose file vanished but whose sibling store still exists (cleanup), drop when the whole tree is gone (uninstall / moved away) — `missingTrees` already draws that line.

Impact if done: the pitch gains a sentence no competitor that reads transcripts can make, and it answers the open Anthropic issue directly. Related: #2966 (install proof), the `session-files-on-disk` guide, which does not yet mention the 30-day cleanup at all.

vshulcz:
[...]

Issue #3505, opened 2026-09-13
what the suite does not ask: update cost, other people's paths, an upgraded machine, windows before main


vshulcz:
Five bugs were found in two days by two contributors and by reading the code, and none of them needed a clever test to find. What they have in common is the part of deja nobody measures. Dates first, because "we shipped it in 0.20.0" turned out to be wrong for most of them:

| what | introduced | first released | found by |
|---|---|---|---|
| a new transcript rewrites the whole store (#3500, fixed in #3503) | 2026-07-14, the first incremental ingest | 0.16.x | a contributor's measurement |
| the freshness walk runs to be discarded (#3500, fixed in #3501) | 2026-08-24 (#1752) | 0.19.1 | the same report |
| Copilot Chat resources never percent-decoded (fixed in #3498) | 2026-09-09 (#3431) | 0.19.5 | a contributor on a path with a space |
| an upgraded machine's hooks point at a binary that is gone, and doctor is silent (#3502) | 2026-09-09 (#3426) | 0.19.5 | the same report's environment section |
| the windows leg deadlocks on a captured pipe (fixed in #3493) | the shape is old; the table crossed 4096 bytes when Xcode's paths landed (#3467) | 0.20.0 | main went red |

So one of the five was caused by 0.20.0. The rest were already shipping, and 0.20.0 is only where someone finally measured. That is the actual problem: four of them are invisible to everything we run.

## What we do not measure

**What an update costs.** Four benchmarks (`recall`, `context`, `block`, `prompt`) and all four are about the quality of what comes back. The event that happens most — a new conversation appears on disk — had no number anywhere until #3500 produced one, which is why a path that re-tokenized the entire store on every new session survived two months and twenty releases. #3503 adds a test that pins the path and the bytes; what is still missing is a number in CI, with a budget, for each update class: unchanged, one appended turn, one new transcript, a rewind, a version bump.

**Anything but this machine's data.** The Copilot Chat sidecar reader was written against the 48 state files on one laptop: every one of them v2, every path ASCII with no spaces. The decode bug could not be seen here and was obvious to a contributor with `C:\Users\me\my app`. Every parser that reads a path wants a fixture set that is hostile on purpose — a space, a non-ASCII character, a percent-encoded octet, a UNC path, a drive letter — and the registry is the place to keep it.

[...]

PR #3530, opened 2026-09-13
fix: a full rebuild keeps sessions whose transcript the client deleted

### LINKED ISSUE #3529: a full rebuild drops every session whose transcript the client deleted
Claude Code deletes a transcript after 30 days and deja keeps the session — that is what #2970 built, and the incremental pass holds the line. A full rebuild does not: it reads the sources, writes what it read, and everything whose file is gone goes with it.

Measured on a store of 8 sessions, 6 of the transcripts deleted:

```
indexed:                       8 of 8
after the harness deleted six: 8 of 8   (records.bin 1288 bytes)
after `deja index --rebuild`:  2 of 8   (records.bin 322 bytes)
```

`deja last` lists two, `doctor` reports "2 files, 2 indexed sessions", and nothing is said about the six.

A full rebuild is not a rare path: a content-version bump runs one (39 and 40 both shipped this week), so does a changed exclude list, a damaged index, and `deja index --rebuild`.

`rebuildWithTombstones` already carries the one other class of records that exists only in the index — `importedSessions` rescues `deja-sync-import` records for exactly this reason (internal/index/ingest.go:658). Sessions whose source file is gone need the same carry, under the same rule the incremental pass uses: a file gone while its store directory is still there is the client's cleanup, a tree gone whole is an uninstall.

vshulcz:
Claude Code deletes a transcript after 30 days; #2970 made the incremental pass keep the session. A full rebuild did not — it reads the sources and writes what it read:

```
indexed:                       8 of 8
after the harness deleted six: 8 of 8   (records.bin 1288 bytes)
after `deja index --rebuild`:  2 of 8   (records.bin 322 bytes)
```

An index-version bump runs a rebuild (39 and 40 both shipped this week), so does a changed exclude list and a damaged index.

`orphanedSessions` carries those sessions out of the old records, the way `importedSessions` already rescues `deja-sync-import` records, under the incremental pass's rule: a file gone while its store directory is still there is the client's cleanup; a tree gone whole is an uninstall or an unmounted disk and is dropped as before. A row is carried only if records came back with it, so `deja forget` is not undone and leaves no phantom "still searchable" count.

Two tests used "delete the transcript and rebuild" as shorthand for making a session go away; they now forget it, which is the deliberate path.

Three mutations checked: no carry, carrying rows with no records, and dropping the store-directory rule each fail a test. Closes #3529.


vshulcz:
Measured the cost on a bigger store: 1200 sessions, 8 turns each, 300 of the transcripts deleted before the rebuild. Fresh corpus per arm.

```
main, nothing deleted          0.23s  records.bin 1,574,112 B
with carry, nothing deleted    0.23s  records.bin 1,574,112 B
main, 300 deleted              0.23s  records.bin 1,181,392 B
with carry, 300 deleted        0.23s  records.bin 1,574,112 B
```

25% of the store gone on main, and the carry costs nothing on the clock.

vshulcz:
Ran the migration matrix #3505 asks for, with the released binaries rather than a faked manifest version: build a store with each published deja, delete one transcript the way a client's cleanup does, then let this build read it.

```
version   content built with   answers after upgrade   the deleted-transcript session
v0.19.0   (pre keep-back)      yes                     already gone before the upgrade
v0.19.2   (pre keep-back)      yes                     already gone before the upgrade
v0.19.4   35                   yes                     lost at the upgrade
v0.19.5   35                   yes                     lost at the upgrade
v0.20.0   40                   yes                     kept, with this branch
```

Keep-back (#2970) shipped in 0.19.4 — before it the old binary itself dropped the session on the next pass, so those two rows are not about the upgrade.

0.19.4 and 0.19.5 are the interesting ones: they keep the session, and the upgrade cannot. Their records are content version 35, and version 38 moved the role out of the compressed body, so this build cannot decode what they wrote; a transcript that is gone from disk has nowhere else to come from. The carry recovers a same-layout store (0.20.0 → now, the row above) and nothing older.

The last commit makes that loss audible instead of silent:

```
deja: this build reads a newer index than the one on disk (40, was 35) — re-reading your sources once
deja: 1 transcript no longer on disk and written by an older deja — this build cannot read those records, so they go with this rebuild
```

with a test on both branches of it: the store that can be carried says "still searchable", the one that cannot says why.

vshulcz:
Ran `scripts/migrate-matrix.sh` against main and against this branch. The last column is the session whose transcript was deleted before the upgrade.

main:

```
version   built-with   after-upgrade upgrade-said               deleted-transcript
v0.19.0   s3           s3            re-reading your sources    none
v0.19.2   s3           s3            re-reading your sources    none
v0.19.4   s3           s3            re-reading your sources    none
v0.19.5   s3           s3            re-reading your sources    none
v0.20.0   s3           s3            re-reading your sources    none
```

this branch:

```
v0.19.0   s3           s3            re-reading your sources    none
v0.19.2   s3           s3            re-reading your sources    none
v0.19.4   s3           s3            re-reading your sources+cannot read those records none
v0.19.5   s3           s3            re-reading your sources+cannot read those records none
v0.20.0   s3           s3            re-reading your sources    s2
```

So: upgrading from the current release keeps the session whose transcript is gone, which is the fix. Upgrading from 0.19.4 or 0.19.5 still loses it — those stores predate the record-layout bump, so the records cannot be decoded to carry — and the run now says so instead of dropping it quietly. 0.19.0 and 0.19.2 have nothing to carry: the keep-back they would need landed later.

PR #3545, opened 2026-09-13
perf: quote the hits a search serves, not every session that matched

### LINKED ISSUE #3544: perf: every candidate session is quoted before the cap, so a search costs the same at --limit 1 and --limit 100
Every candidate session gets its quotes rendered during scoring, and the result cap trims them afterwards. On a synthetic store of 1,500 sessions (31 MB of `records.bin`), the same query costs the same no matter how many hits are asked for:

```
--limit 1    807.8 ms
--limit 5    806.9 ms
--limit 15   805.8 ms
--limit 50   805.8 ms
--limit 100  804.1 ms
```

One search allocates 127.8 MB, and the quoting is a third of it:

```
15.5 MB flat / 43.6 MB cum (34%)  search.snippet
22.0 MB                           strings.Fields   (under it)
```

with `internal/stringslite.Index` taking 24.4% of the CPU time.

The loop that does it is internal/search/search.go:356 — `snipCands` is sorted and up to three passages are rendered per session, inside the per-session scoring loop, before anything has been ranked. Fifteen hits are served from work done for every session that matched a word.

Deferring the render until after the ranked cap would leave the work proportional to what is returned. Two ways: keep each session's chosen passage offsets and render the top N afterwards, or re-read the N records post-cap, which is at most a hundred reads.

Sibling of #3491 rather than a duplicate: that one is about how a record is read, this one is about how many records are read at all. The caveat there applies here too — the corpus repeats, so nearly every session matches; the flat line across `--limit` is the part that does not depend on the corpus.

vshulcz:
Quotes were rendered inside the scoring loop, for every session that matched a word, and the cap trimmed them afterwards — so a search cost the same whatever it was asked for:

```
              before    after
--limit 1     807.8 ms  746.6 ms
--limit 15    805.8 ms  743.5 ms
--limit 100   804.1 ms  745.8 ms
allocated     127.8 MB   88.9 MB   (one query)
```

`search.snippet` and the `strings.Fields` under it were 34% of those bytes and are out of the top ten entirely now.

Choosing the passages stays where it was — a session is ranked partly on how tightly its words come together. What moves is the render: the top 50 are quoted before `markEarlierAttempts`, which compares their text, and the rest wait until after the cap.

Measured on a synthetic store of 1,500 sessions, 31 MB of `records.bin`, median of 5. The three benches are unchanged: recall@1 1.00 / MRR 1.000, every prompt arm identical (13/13, 1/1, 1/1, 0/12 false fires, 5/51, 15/20, 6/6), context medians 1096 / 735 / 360 with the same coverage.

Two mutations checked: quoting in the loop again fails the test that 70 hits below the window were quoted; dropping the post-cap render fails the test that an uncapped answer quotes its tail.

Closes #3544.


vshulcz:
Measured on a copy of a real 194 MB store afterwards, and there is no difference: 22–23 ms before and after, on narrow queries and broad ones alike, at `--limit 15` and `--limit 1`.

```
query              before     after
deja               22.7 ms    23.3 ms
index              22.1 ms    23.0 ms
error              22.5 ms    22.6 ms
the file           22.4 ms    22.7 ms
deja --limit 1     23.1 ms    22.2 ms
```

The synthetic corpus every number in the pull request came from has every session matching every query, so "quote what matched" and "quote what is served" were 1,500 sessions against 15. A real store's exact tier hands the scoring loop a handful, and the two are then the same work.

So the change is not the speed-up the numbers above it read as. It is still right — the cost now follows what was asked for rather than what the store happens to hold — and the worst case it removes is the one a large project with a common term walks into. But the honest summary is: no regression, no visible win on this machine's store, 30% fewer bytes allocated where the candidate set is large.

vshulcz:
Retracting the numbers in the comment above: that arm measured an empty index.

The copy was read with a temp HOME, so none of the manifest's source paths existed, the search answered `tier=exact total=0` for every query, and 22–23 ms was the cost of finding nothing. Re-running it with the real HOME makes each search walk the whole source tree for its freshness check, which swamps the thing being measured — so a clean before/after on a real store still needs a harness that freezes the sources, the way the ingest bench does.

What stands is what the pull request measured: on a synthetic store where every session matches, 806 ms → 745 ms and 127.8 MB → 88.9 MB, and the three benches unchanged. What the real store says about it is not yet known.

PR #3547, opened 2026-09-13
fix: a renamed transcript is not indexed again

### LINKED ISSUE #3546: a renamed transcript is indexed again: twenty renames, twenty-one copies
A transcript renamed inside a store deja already reads is ingested as a new file: its records are appended again, and the row under the old name stays as "no longer on disk". Twenty renames of one 4 KB transcript:

```
after the first build   records.bin  4,150 B   phantoms 0
after rename 1          records.bin  8,300 B   phantoms 1
after rename 5          records.bin 24,900 B   phantoms 5
after rename 10         records.bin 45,650 B   phantoms 10
after rename 20         records.bin 87,150 B   phantoms 20
```

Twenty-one copies of the same forty turns. Search still answers once — the query-side dedupe hides it — and `doctor` reports "1 file, 1 indexed session", so nothing on any screen says the store is twenty times the size of what it holds.

Renaming the project directory is fine: the whole tree going away is recognised, the replacement path runs and `records.bin` does not move (measured: 50 sessions, 39,900 B before and after). This is the single-file case, where the directory stays.

It is not hypothetical: dsh renamed its log from `session.jsonl` to `session.v3.jsonl` (#3508), so every dsh user's store carries one duplicate copy of those logs. Any harness that writes a resumed session under a new name does the same, on every resume.

#3503 already fixes the manifest side of this — "a rename reaches this path as a new file holding a session deja already has", and the row keeps the live path rather than the dead one. What is missing is the other half: the records. A file whose size and leading bytes match a row whose path has just gone is the same file under a new name, and can carry the old row's watermark instead of being read from zero.

vshulcz:
A transcript that only changed its name was read from the first byte and written again. Twenty renames of one 4 KB log:

```
                    before              after
first build         records.bin  4,150 B    4,150 B
after rename 1                   8,300 B    4,150 B
after rename 5                  24,900 B    4,150 B
after rename 10                 45,650 B    4,150 B
after rename 20                 87,150 B    4,150 B
dead rows                             20          0
```

Both sides already carry a prefix fingerprint — the scan computes one for every `.jsonl` it has not seen before — so the pairing is on content: same size, same last-complete-line offset, same sample, same harness. Records do not move at all: their source path is interned, so one entry in that table is every record's path at once.

The pass says what it did (`1 transcript renamed — the index followed the new name`) instead of reporting the old name as a transcript deja still holds.

Deliberately narrow, with tests: a different conversation of the same size is not a rename, a row with no fingerprint is left alone (a store built before the sample existed), two harnesses are never paired, and a plain deletion still keeps its session searchable. A file that was renamed *and* appended to in the same pass is not paired — its fingerprint no longer matches — and falls back to today's behaviour.

Not hypothetical: dsh renamed its log from `session.jsonl` to `session.v3.jsonl` (#3508), so every dsh user's store holds one duplicate copy of those logs; any harness that writes a resumed session under a new name does it on every resume.

Two mutations checked: disabling the detector doubles `records.bin` in the test, and pairing on size alone pairs a different conversation.

`deja bench ingest` is unchanged — unchanged 10.4 ms, appended turn 36.9 ms, new transcript 40.8 ms, rewritten 232 ms, and only the rewrite replaces the store.

Closes #3546.


vshulcz:
Second commit covers the other half of the same shape: a session renamed **and** written to since — a resume under a new name. The fingerprint covers the bytes deja had already read, so the pairing still holds and the pass reads the tail rather than the whole log:

```
deja: 1 transcript renamed — the index followed the new name
deja: updated 1 file (1 new message)
```

Before it, that case fell back to re-reading the file whole and keeping the old name as a transcript deja still holds.

vshulcz:
Third commit gives the class an arm in `deja bench ingest`, so it cannot regress in silence the way it arrived:

```
update                       wall      records      added rewrote the store
unchanged                   11.4ms        0.3MB       0.0KB no
appended turn               41.9ms        0.3MB       0.1KB no
new transcript              50.3ms        0.3MB       0.1KB no
renamed transcript          12.4ms        0.3MB       0.0KB no
rewritten transcript       222.8ms        0.3MB       0.0KB YES
```

The test asserts the rename adds zero bytes — the bytes are already on file under the old name — and still only a rewrite replaces the store.

vshulcz:
One more thing the pairing buys, unplanned: moving a whole project directory. Fifty transcripts, renamed as a tree —

```
before   69 ms, the pass re-read all fifty and rewrote records.bin
after    37 ms, "50 transcripts renamed — the index followed the new name", records.bin unchanged
```

The limit to know about: for a harness that has no `cwd` in its records, the project name on those rows is derived from the directory at ingest, so after a directory move it stays what it was until something re-reads the file. Claude, Codex and the rest carry the working directory in the transcript, so they are unaffected.

vshulcz:
The windows leg caught one of mine from #3541, now fixed here: the dead-hook fixture wrote `/nowhere/deja`, which is not an absolute path on windows, so the check that reads those files skipped it and the row came back healthy — on the one platform the test exists to cover. It now builds the path under the test's own home and passes it through `filepath.ToSlash`, and the live arm writes a real file named `deja` rather than pointing at the test binary, whose name that check does not recognise.

The other failure in that run was the encoded-project fixture #3542 fixed; this branch has main merged in now.

PR #3563, opened 2026-09-14
test: every parser reads its fixture from somebody else's path

vshulcz:
One of the boxes on #3505: adversarial path fixtures per parser.

Every registry fixture is now also read from under a directory called `проект (v2.1) %20 name` — a space, a percent octet that is not an escape, parentheses, a dot and a non-ASCII name. The whole fixture tree moves rather than the one file, because several parsers read a sibling (kimi's `state.json`, goose's header) and copying the file alone would test the copy instead of the path. All 26 harnesses pass as they are; the point is that the next reader written against this laptop's ASCII paths does not.

Beside it, the case that actually bit: a Copilot Chat workspace folder arrives as a URI, and a folder with a space in it arrives percent-encoded. Reading it raw named the project `my%20app` — a project nobody searches for and no project filter matches (#3498). Four shapes through the helper and one through the file the reader opens; against the raw path they all fail.

`parseRegistryFixture` now takes the directory it works in rather than making one, which is what lets the hostile test hand it a path of its own.


vshulcz:

PR #3564, opened 2026-09-14
test: a captured pipe is drained while the call writes, not after it

vshulcz:
The last box on #3505: "no capture in a test reads a pipe after the call — the rule is the point."

#3493 and #3504 fixed the sites that were known. This checks the rule instead, and found five more: `main()`'s version output, the session-start injection, the stats JSON, and both share-count captures all still read the pipe after the call that fills it. A pipe holds one buffer — 4 KB on Windows — so each of them is a deadlock waiting for its output to grow, which is exactly how the windows leg hung on main rather than on the pull request that grew it.

The check is an AST walk over every test file: where the write end of a pipe is handed to the code under test (`os.Stdout = w`, or passed as an argument), the read of the other end has to be off the calling path — in a goroutine, or through a helper that starts one. A pipe the test fills and closes itself is standing in for a file and is left alone.

Five captures fixed, one small `drainPipe` helper for the three that had no helper of their own. Put any of them back the old way and the check names the file and the line.


vshulcz:

PR #3569, opened 2026-09-14
fix: a mistyped command exits non-zero, and doctor stops calling a missing root found

### LINKED ISSUE #3567: a mistyped command exits 0 after saying it is not a command
`deja unforget nope` prints

```
deja: no matches in 2 indexed sessions — try fewer words or --re (query "unforget nope")
deja: "unforget" is not a command — `deja forget --unforget <id>` is, and `deja forget --list` names the ids
```

and exits **0**. So does `deja doctro`, which is told it meant `deja doctor`.

deja knew enough to name the command that does exist, and then reported success to anything reading the exit code. A script that runs `deja unforget "$id"` and checks `$?` is told the work was done; nothing was done.

A search that finds nothing is a different thing and should keep exiting 0 — that is the ordinary answer to a question the history cannot settle. The distinction is already drawn by `commandHint`, which stays quiet unless a real command name is within one edit of what was typed, skips words shorter than four letters, skips anything that does not look like an invocation, and skips exact command names used as ordinary words.

For comparison, the commands that already exit non-zero on a wrong shape: `deja show nope`, `deja share nope`, `deja forget nope-nothing`, `deja sync` with no peers.

### LINKED ISSUE #3568: doctor calls an antigravity store found when its root is not on disk
On a machine where `DEJA_ANTIGRAVITY_ROOT` points at a directory that does not exist, `deja doctor` says the store is there:

```
  antigravity  found     ~/sources/deja_antigravity_root  (0 files)
  gemini       missing   ~/sources/deja_gemini_root  (0 files)
  grok         missing   ~/sources/deja_grok_root/sessions  (0 files)
```

Every other row on that machine says `missing`, and the paths are equally absent. Antigravity is the outlier because `AntigravityRoots()` returns the env override as given — the glob branch beside it filters to directories that exist — and doctor's presence check is `len(agyRoots) > 0`.

Found while walking a fresh machine for 0.20.1. It is the same shape as the states doctor exists to tell apart: a store that is there with nothing in it, and a path that is not there at all. A typo in the variable, or an install that has been removed, reads as healthy.

vshulcz:
Two fixes found by walking a fresh machine for 0.20.1. Fixes #3567, fixes #3568.

**A mistyped command exits non-zero.** `deja unforget x` named the command that does exist — `deja forget --unforget <id>` — and then exited 0, so a script that ran it and checked the code was told the work had been done. A search that simply found nothing still exits 0, which is the ordinary answer to a question the history cannot settle; the distinction is the hint's own, and it stays quiet unless a real command is within one edit of what was typed. The sentence is still printed once, by the command rather than by main.

**doctor calls an antigravity store missing when its root is not on disk.** `AntigravityRoots()` hands back `DEJA_ANTIGRAVITY_ROOT` as given — the glob branch beside it filters to directories that exist — and the row read `found (0 files)` where every other harness on the same machine said `missing`.

One existing test changed: `TestAnUnknownHookNameIsNotAQuery` asserted that a mistyped command returns no error, which was the contract this changes. It still asserts what it was really about — the search ran and built the store to answer.


vshulcz:

PR #3571, opened 2026-09-14
fix(mcp): blame says a file has no history rather than answering []

### LINKED ISSUE #3570: blame answers an agent with a bare [] when the file has no history
`blame` is the tool an agent calls before it edits a file. On a file nothing has touched, it answers:

```
[]
```

Every other mode answers in a sentence — `how` says "No command on this machine mentions …", `fix` says the text does not read like an error line, `recall` says the store is empty when it is. An empty array reads as a tool that failed, and an agent that reads it that way edits without the history it asked for.

#2862 drew exactly this distinction for the empty-store case, and blame got that half: a store with nothing indexed answers with a note. The other half — a store full of history where nothing touched this file — still answers `[]`.

Measured on a stand with three indexed sessions:

```
blame, a file with history in the store   []
blame, a file with none                   []
```

The CLI says it properly: `deja: no sessions mention main.go — searched 1 indexed session`.

What the note needs to carry is the count: "nobody touched this file, and I looked through 400 sessions" is a different fact from "I have three sessions here".

vshulcz:
Fixes #3570.

`blame` is the tool an agent calls before it edits a file, and on a file nothing has touched it answered a bare `[]`. Every other mode answers in a sentence, and the CLI has all along — an empty array reads as a tool that failed, and an agent that reads it that way edits without the history it asked for.

#2862 drew this distinction for a store with nothing indexed, and blame got that half. This is the other half: a store full of history where nothing touched this file. The note carries the count, because "nobody touched it and I looked through 400 sessions" is a different fact from "I have three sessions here".


vshulcz:

Issue #3535, opened 2026-09-13
a password passed as a command argument is indexed in the clear


vshulcz:
Twelve planted secret shapes through a store, then every surface that hands text to an agent. Ten were redacted at ingest. Two were not:

```
sshpass -p 'ZZexampleSSHpassword' ssh deploy@10.0.0.1
curl -u admin:ZZexampleBASICpass https://api.example.com
```

Both reach `deja search`, `deja recall`, `deja show` and — the part that leaves the machine — `deja sync export`.

Nothing in internal/redact/redact.go can see them: there is no key word beside the value for `-p`, and `admin:pass` is a pair `connURLRE` only matches inside a URL. The key-value patterns also need 16 characters, and a password rarely has them.

The uid:gid form (`docker run -u 1000:1000`) and `ssh -p 2222` have the same shape and are not secrets, so whatever matches has to name the command or keep the value's class apart.

vshulcz:
Swept a second batch of twelve shapes on top of the first twelve. Six more were in the clear: `docker login -p`, `az login -p`, `redis-cli -a`, `gh auth login --with-token`, a `.netrc` line (`machine … login … password hunter2`), a `Cookie:` header, and a Russian key word with words between it and the colon (`пароль от стейджа: …`). All in #3536.

PR #3573, opened 2026-09-14
fix(redact): a password given as a flag is masked whatever its length

### LINKED ISSUE #3572: a password given as --password=… is stored in the clear below sixteen characters
A password handed to a program as a long flag is stored in the clear unless it is sixteen characters or longer.

Measured through a real index pass — plant the lines, `deja index`, then `deja show`:

```
form01 mysql --password=MyRootPass2026 -u root          ← 14 characters, in the clear
form04 psql --password=PsqlPass2026xx -U app            ← 14 characters, in the clear
form05 app --password=Pass2026Short                     ← 13 characters, in the clear
form06 app --password=[redacted:credential]             ← 19 characters, masked
form07 DB_PASSWORD=[redacted:credential] ./run.sh
form08 password: [redacted:credential]
```

It reaches `deja show` and `deja share`, and it is in `records.bin`.

The floor comes from `genericKVRE`, which takes any value of sixteen characters or more. That is the right floor for `token = …`, where a short value is as likely to be a word — and the wrong one after `--password`, where the flag says what the value is. `mysql -pSecret` is caught (the attached form has its own rule); the long-flag spelling every other tool uses is not.

The bare `password: hunter2` is a separate decision and stays as it is: `escaped_json_test.go` lists it as a near miss on purpose, because the bare word opens more prose than credentials.

Redaction runs at ingest, so a fix reaches only the next conversation without a content-version bump — the reasoning #3535 recorded for version 41.

vshulcz:
Fixes #3572.

Found by planting fifteen secret shapes in a transcript, indexing, and looking for the raw value on every surface that hands text to somebody else. Fourteen were masked. This one was not:

```
form01 mysql --password=MyRootPass2026 -u root     ← 14 characters, reached `deja show` and `deja share`
form06 app --password=[redacted:credential]        ← 19 characters, masked
```

`genericKVRE` takes any value of sixteen characters or more, which is the right floor for `token = …` where a short value is as likely to be a word, and the wrong one after `--password`, where the flag says what the value is. `mysql -pSecret` already had its own rule; the long-flag spelling every other tool uses had none.

The new rule is the flag form only. `password: hunter2` stays a near miss — `escaped_json_test.go` records that decision and this does not touch it. A placeholder after the flag is left alone too (`$DB_PASSWORD`, `${…}`, `<your-password>`, `changeme`, `none`), because redacting those costs the reader the sentence and hides nothing.

Index format version 42, for the reason #3535 recorded for 41: redaction runs at ingest, so without the bump the fix holds for the next conversation and leaves every earlier one quotable, including through `deja sync export`. This is also the first bump to use `redactionFloor` — a store below it does not answer while it is re-read, which is what that floor is for.

After the fix the same sweep comes back clean on search, show, share, `show --json`, `last`, `stats --json`, the export bundle and `records.bin` itself.


vshulcz:

PR #3575, opened 2026-09-14
fix: the first search on a new machine says one thing, not two

### LINKED ISSUE #3574: the first search on a new machine says two things that cannot both be true
A search that lands while the first index build is running prints both of these:

```
deja: answering from the index as it was — refreshing in the background
deja: the index is being rebuilt right now — run this again in a moment
```

and exits 1 with no results. On a first build there is no index to answer from, so the first line is false — and it is a new machine's first question, which is the worst run to contradict anything on.

Reproduced by seeding 150 transcripts, starting `deja index` in the background and searching immediately: two attempts in five printed the pair, the others waited for the build and answered normally. The exit code follows the same interleaving, so a script that searches during the first build fails at random.

vshulcz:
Fixes #3574.

A search that lands while the first build is running printed both `answering from the index as it was` and `the index is being rebuilt right now`, then exited 1 with nothing. There is no index to answer from on a first build, so the first line is false — on a new machine's first question.

The line is now printed only when there is a readable snapshot behind it, which is the case it was written for: a store that answers while the refresh runs behind it.

The test takes the index lock the way another deja would, so both paths are exercised without a race: an empty store must not claim to be serving one, and a store that can answer still says it is serving the older view.


vshulcz:

PR #3577, opened 2026-09-14
fix(install): refuse a TOML config that is already broken

### LINKED ISSUE #3576: install splices its block into a TOML config that is already broken
Every JSON target refuses a config that does not parse and says where. The TOML path does not look:

```
$ cat ~/.codex/config.toml
model = "gpt-5"
[mcp_servers.theirs        ← the bracket is missing
command = "x"

$ deja install codex
skill: unchanged ~/.agents/skills/deja-search/SKILL.md      (exit 0)

$ cat ~/.codex/config.toml
model = "gpt-5"
[mcp_servers.theirs
command = "x"

[mcp_servers.deja]
type = "stdio"
command = "…/deja"
args = ["mcp"]
```

Nothing is lost — the splice appends and a `.bak` is kept — but deja's entry is now in a file the harness cannot load, and the harness's parse error names deja's lines. The reader ends up debugging deja for a bracket they left open.

For comparison, on the same run: claude with truncated JSON, claude with a JSON array, opencode and gemini with half-written objects all refuse with `1 target refused` and leave the file untouched.

deja ships no TOML parser on purpose — the writer splices text so a hand-written config keeps its comments and its order. The smallest check that catches what actually breaks is that every table header opens and closes on its own line.

vshulcz:
Fixes #3576.

Found while installing into configs a user has already broken by hand. Claude, opencode and gemini all refuse a file that does not parse and leave it untouched; the TOML path spliced its block in and exited 0, so deja's entry ended up in a file the harness cannot load — and the harness's parse error then named deja's lines.

deja ships no TOML parser on purpose: the writer splices text so a hand-written config keeps its comments and its order. The check here is the smallest one that catches what actually breaks — every table header opens and closes on its own line — and it leaves ordinary configs alone: headers with trailing comments, arrays of tables, quoted keys, commented-out headers.


vshulcz:

PR #3579, opened 2026-09-14
fix: the last slice of a session says so instead of offering a next one

### LINKED ISSUE #3578: the last page of a session offers a next page that is past the end
`deja show` closes every slice with the offset that reads the next one, including the slice that ends the session:

```
$ deja show sw0 --offset 13 --limit 3
deja: showing messages 14-16 of 16 — `--offset 16` reads the next slice

$ deja show sw0 --offset 16 --limit 3
deja: --offset 16 is past the end — the session has 16 messages
```

Following deja's own advice gets deja's own refusal. The note already knows where it is — it prints `of 16` — so the last slice can say so instead.

vshulcz:
Fixes #3578.

`deja show` closed every slice with the offset that reads the next one, including the slice that ends the session — and following that advice answered `--offset 16 is past the end — the session has 16 messages`. The note already knows where it is, so the last slice says the end of the session and the ones before it keep the offset they had.


vshulcz:

PR #3581, opened 2026-09-14
fix(mcp): fix stops telling an agent its error is not an error

### LINKED ISSUE #3580: fix tells an agent its error is not an error, for half the errors a tool prints
`fix` answers text it does not recognise with:

```
That text does not read like an error line - pass the failing output itself.
```

The agent did pass the failing output. Measured against twenty error lines a tool actually prints, the MCP mode refused ten of them:

```
REFUSED  connection refused
REFUSED  permission denied
REFUSED  segmentation fault
REFUSED  OOMKilled
REFUSED  context deadline exceeded
REFUSED  undefined: foo
REFUSED  Exit code 137
REFUSED  SyntaxError: unexpected token '<'
REFUSED  docker: Error response from daemon: pull access denied
REFUSED  cannot use x (variable of type int) as string value
```

The CLI took all twenty on the same store, so the two surfaces disagree about the same input.

The gate is `index.LooksLikeError`, which is the heuristic that mines fix pairs out of a transcript: a phrase list plus a twenty-character floor (`frictionLineMin`). That is the right shape for mining, where a false positive costs a bad pair. Used on the caller it costs an accusation, and the accusation is wrong more often than not — five of the ten refusals are under twenty characters, and the rest miss the phrase list.

Both branches answer "nothing found" anyway; only the wording differs. The advice is worth keeping for prose ("it did not work"), but it can ride along with the honest sentence instead of replacing it.

vshulcz:
Fixes #3580.

`fix` answered text it did not recognise with `That text does not read like an error line - pass the failing output itself`. The agent did pass the failing output. Measured against twenty error lines a tool actually prints, the MCP mode refused ten — `connection refused`, `permission denied`, `segmentation fault`, `OOMKilled`, `Exit code 137` among them — while the CLI took all twenty on the same store.

The gate is the heuristic that mines fix pairs out of a transcript: a phrase list and a twenty-character floor. That is the right shape for mining, where a false positive costs a bad pair, and the wrong one for judging the caller. Both branches answer "nothing found" anyway; only the wording differed.

The advice survives for the case it was written for — prose about a failure rather than the failure — and now rides along with the honest sentence instead of replacing it.


vshulcz:

Issue #3001, opened 2026-09-02
the pre-tool hook costs 172 ms a call and returns a decision 0 times in 48 on a real store


vshulcz:
The point-of-action hook has two registers: content — `… — last time: <what was settled>` — and a pointer — `run in 5 sessions, last 2026-08-19`, `main.go has been worked on in 8 sessions`. It costs the same either way, and it now runs on every Bash, Edit, Write and Task.

Measured on this machine (1409 sessions, 196,926 messages), sampling commands out of `deja how` and files out of the repository:

    48 probes (24 commands, 24 files)
      decision  0
      pointer  13
      silent   35

    cost per call, mean of five: 172 ms

The mechanism is not broken — on a seeded stand it produces the content register correctly, so the gate is what real sessions clear. Worth finding which one binds: `commandDecisionLine` wants a ranked session that ran the command *and* `digest.Conclusions` to yield a line from its tail; `promotedCommandDecision` wants a promoted note whose session ran it.

Filing the number rather than a fix: 172 ms on every tool call is a real tax, and "you have run this before" is the line the code's own comment calls the less useful of the two.

vshulcz:
Correcting my own measurement: the first run reused one `session_id` across every probe, and the hook refuses to repeat itself inside a session — so an unknown number of those 35 silences were the dedupe rather than the store.

Redone with a fresh session id per probe, five command families, 80 probes (40 commands out of `deja how`, 40 files from the repository):

    decision   0
    pointer   18
    silent    62

    by family (pointer/silent, 16 probes each)
      git   7/9    go 4/12    npm 7/9    deja 0/16    gh 0/16

The decision count is what the issue is about and it is unchanged. One thing the re-run did surface: `toolHookDecisionScan = 3` is a binding gate — a build with it raised to 25 produces the content register on commands where the shipped one is silent. Whether that is worth its latency is a separate question; at 25 the lines it found here were failure fragments rather than decisions.

vshulcz:
The other half of the pair, measured the same way. `hook-tool-after` works end to end — on this store it answers a real failure with

    deja: this error came up here before (2026-04-22) — what followed it: kubectl -n …

466 bytes, from the harness's own PostToolUse payload. What limits it is the second-sighting bar. Across the ten errors `deja friction` lists here:

    confirmed repair   2
    candidate withheld 6   ("waits for a second sighting")
    nothing at all     2

[...]

PR #3604, opened 2026-09-14
fix(hook): a repeated fact yields to the line behind it

### LINKED ISSUE #3603: the pre-tool hook says three lines in ninety-one actions: a deduped line swallows the decision behind it
#3001 measured the pre-tool hook returning a decision 0 times in 48 and asked which gate was stopping it. It is not a gate. `commandDecisionLine` has a line ready for **67 of 93** command/project pairs on this store — `go run ./scripts/commandgate` says so. The line never reaches the agent because a different producer is computed first and then deduped.

`toolHookLine` returns the first non-empty of three producers: the missing-program sentence, the failure line, then the history block that carries the decision. `runHookToolMode` takes that one line, hashes it, and when the same fact has already gone into this agent session:

```go
if alreadyInjected(dir, input.SessionID)[token] {
    return nil
}
```

It returns. It does not ask the next producer. So a fact that is said once suppresses every later action whose highest-priority producer yields the same fact — including all the actions that had something else to say.

Measured against the binary, over the 91 command/project pairs the store's own history says an agent runs and that do something (the hook's own rule already skips 79 inspection commands):

**One command at a time, no session id, so nothing is deduped:**

```
missing-program sentence   51
failure line                9
pointer only                0
pointer + decision          0
silent                     31
```

All 51 are about **one** program, genuinely not on this machine, and in none of the 51 is it the command's own program — it sits inside a compound. Cross-tabulated against what `commandDecisionLine` had for the same command, **34 of those 51 had a decision behind them**.

**The same 91 actions as one agent session, dedupe on:**

```
silent           88
something else    2
missing program   1
```

Three lines in ninety-one actions. The missing-program sentence is said once, correctly, and then silences the surface for the rest of the session — for every command that happens to mention that program, which here is 56% of them.

The sentence is not wrong and its own comment already records that it is not heard: "nine of the ten sessions told about a missing command ran it anyway." What it costs is the line behind it, which is the one thing this surface exists to say.

The fix is the fallthrough: when the chosen line is already injected, ask the next producer rather than returning. The producers are already ordered by certainty, so the order does not have to change — only the early return.

Cost is no longer the problem it was: median 14 ms, p90 31 ms per call, against the 172 ms #3001 measured.

Method: `scripts/commandgate` for the per-pair verdict, `deja hook-tool` for what the binary actually says, both against a **copy** of a real index. The commands and projects are one machine's own history and stay out of this issue; only counts are reported.

vshulcz:
Fixes #3603.

`toolHookLine` returns one line, chosen from four producers ordered by certainty. `runHookToolMode` then hashed that line, found the fact already injected into this agent session, and returned — without asking the next producer. So a fact said once silenced every later action whose top producer yielded the same fact, including all the ones that had something else to say.

Measured on a copy of a real index, over the 91 command/project pairs its own history says an agent runs and that do something:

- `commandDecisionLine` had a line ready for 67 of 93 pairs.
- Asked one at a time with no session id, the hook answered with the missing-program sentence 51 times — always about the same absent program, never the command's own — and 34 of those had a decision behind them.
- Walked as one agent session, with dedupe on: **three lines in ninety-one actions**.

## The retry is bounded, and why

The first version retried every producer. It works and it is too expensive to ship: the first two producers are a sidecar lookup each, while the last two rank and then load whole sessions. Same store, 42 actions across the three projects whose working directory still exists:

| | decision | other | missing program | silent | lines | median cost |
|---|---|---|---|---|---|---|
| before | 4 | 2 | 3 | 33 | 9 | 21 ms |
| retry everything | 6 | 6 | 3 | 27 | 15 | 133 ms |
| **retry the cheap ones** | 4 | 6 | 3 | 29 | **13** | **31 ms** |

133 ms is the 172 ms #3001 filed as the problem, coming back. And it would be paid on essentially every action, not occasionally: counted over sessions with twenty or more command actions, the median share that run a command the session had not run before is 100% (p10 98%), so the expensive producers are reached fresh each time.

So the walk retries among the two cheap producers and stops before the expensive ones once it has skipped something. That is +44% lines for +10 ms, and it leaves the two extra decisions on the table — reaching those needs the decision to become cheap, which is #3605.

Nothing new is invented: every line was already computed and thrown away.

The per-prompt hook already works this way, and the comment above its own dedupe records the same lesson being learned there — it drops the candidates it has already shown and keeps ranking, rather than giving up on the first repeat. The tool hook never got that treatment.

The test pins both ends — the most certain fact still goes first when nothing has been said, a repeat yields to the line behind it, and a walk with nothing left to say is still silent.

The file side has the same shape and must not be changed with it; the reasoning is in a comment on #3603.


vshulcz:
Rebound lifted in 72e3278: past the two cheap producers the retry pays lookups only — 7.5 ms median against 111 ms when the decision falls back to a search, measured over 60 command shapes on a real store whose table predates the session key #3605 added. Third claim in the fall-through test covers it; it fails with the bound put back.

Issue #3500, opened 2026-09-13
index: new transcript files rewrite existing history; redundant scans and background rebuild CPU


chrisgeo:
## What happened?

Background indexing can perform work proportional to the existing history when only one new conversation has appeared. On current main / 0.20.0, a synthetic reproduction confirmed that adding a one-message transcript rewrites the old records and postings; an ordinary changed-index command also discovers all transcript stores three times.

Separately, a forced rebuild of a small synthetic index averaged **460.5% CPU**. This matters when compaction launches the same index command in the background during active work. The user-visible problem is needing to accommodate expensive maintenance despite almost all history already being indexed and having no convenient idle window for an upgrade rebuild.

Related: #3491 covers profiling and allocation costs; this report adds measurements of when indexing runs and how a new transcript updates an existing corpus. #3422 covers stable launchers; the local upgrade wiring findings below are related to that closed issue. This investigation followed #3436, but does **not** attribute the pre-existing compaction warmup trigger to that PR.

### Measured behavior

Built exact commit `2f6ef66b6b0e6d26ab8e8354d01f89432d567f7a`. The comparison used `GOMAXPROCS=2`, synthetic Codex JSONL, 64 messages per existing session, and a fresh subprocess for each `index` command. One run per changed case; the unchanged range is two consecutive runs. The index/configuration outputs and overrideable source roots were isolated in temporary directories; the driver verified that Aider's non-overrideable default history file was absent.

| Operation | 128 sessions / 10.5 MiB records: wall | CPU seconds | 512 sessions / 42.2 MiB records: wall | CPU seconds |
|---|---:|---:|---:|---:|
| Initial build | 0.9300 s | 0.7986 | 2.4517 s | 4.0236 |
| Unchanged, first run | 0.0076 s | 0.0063 | 0.0129 s | 0.0123 |
| Unchanged, second run | 0.0070 s | 0.0058 | 0.0116 s | 0.0114 |
| Append one message to existing transcript | 0.0580 s | 0.0287 | 0.0936 s | 0.0767 |
| Add one new transcript containing one message | 0.3739 s | 0.3189 | 1.2415 s | 1.2190 |
| Unchanged after new transcript | 0.0091 s | 0.0080 | 0.0119 s | 0.0113 |

[...]

Issue #3590, opened 2026-09-14
the grok linearity test measures the runner: 8.6x on windows, a steady 4.0x here


vshulcz:
`TestGrokIndexesTheRunWithoutGoingQuadratic` failed the windows leg of a pull request that touches neither grok nor parsing:

```
--- FAIL: TestGrokIndexesTheRunWithoutGoingQuadratic (1.04s)
    500 lines in 62.3962ms, 2000 lines in 538.1151ms
    4x the lines cost 8.6x the time, which is not linear
```

Eight runs of the same test here, under `-race`:

```
500 lines in 373ms, 2000 lines in 1.496s     → 4.00x
500 lines in 367ms, 2000 lines in 1.477s     → 4.01x
500 lines in 390ms, 2000 lines in 1.499s     → 3.84x
…all eight between 3.84x and 4.02x
```

The parse is linear. What moved is the baseline: 62 ms on the runner against ~370 ms here, six times faster, which is not the machine being six times faster — it is a measurement too short to be stable. A pause landing in the large run then swings the ratio, and the ratio is what the bound reads.

The test already knows an absolute bound would measure the machine (its comment says so). A ratio of two short timings measures it too, just less obviously.

Best of three pairs fixes it without weakening the bar: a timing that is too slow is always noise, never a false pass, and a parse that has actually gone quadratic reads ~16x on every attempt.

PR #3609, opened 2026-09-14
ci: ingest cost gate; doctor --json gains the claude and codex hook rows; a file at the index path is no longer deleted

### LINKED ISSUE #3608: doctor --json: no auto_recall row for the Claude Code or codex hooks
`deja doctor --json` has no row for the Claude Code or codex hook wiring. Both print in the text report and neither reaches `auto_recall`, because both predate the `autoWirings()` table the other nineteen harnesses share.

Measured on a fresh stand with `install claude-auto`, which writes `~/.claude/settings.json`:

```
text:   claude-code  wired       ~/.claude/settings.json
auto_recall names: [opencode, cursor, gemini, qwen, kimi, antigravity, pi, hermes,
                    openclaw, cline, omp, amp, prime, deepseek, goose, crush, grok,
                    aider, roo]
```

So a script watching its own install sees nothing about the harness most people run — including the state an upgrade leaves, where the entries are wired, name a deja that is gone, and every hook exits 127. That is the state #3502 reported as healthy, and #3510 added `binary_missing` for exactly this, on the table these two are not in.

Fix is to decide each row once and let both surfaces read it, rather than the text lines computing their own state: `claude-code` and `codex-hook` in `auto_recall`, with `out of date` and `unreadable` for a hooks file that is partly wired or will not parse, and codex's `untrusted`/`disabled` for what its trust store says.

Found while writing the migration matrix's config column for #3505 — the check asked `doctor --json` about the wiring an older release wrote and got `absent`, which reads the same as a machine where deja was never installed.

### LINKED ISSUE #3610: a file at the index path is deleted silently by the next build
A build against an index path that holds a regular file deletes the file. Exit code 0, nothing printed.

Repro, on 0.20.0 and on main:

```
$ echo "MY IMPORTANT NOTES" > /tmp/stand/index.db
$ DEJA_INDEX_DIR=/tmp/stand/index.db deja index
deja: indexing sessions into ~/index.db ...
deja: claude: 1 session, 2 messages
$ ls -ld /tmp/stand/index.db
drwx------  … /tmp/stand/index.db
```

The file is gone, and not parked either — `swapIndexDir` renames whatever is at the path to `<dir>.old`, renames the new index into place, and then `os.RemoveAll(old)` on success. `deja search` reaches it too, through `tryLockDir` rather than `lockDir`.

Two ways in that need no mistake at the keyboard: `DEJA_INDEX_DIR` pointed at a database or an archive rather than a directory, and a sync client leaving a file where the directory belongs.

Deleting somebody's file is not a call to make silently. A build should refuse and name the path; `doctor` should say which state it is in rather than reporting a `missing` index, because "run `deja warmup`" is advice that cannot be followed here.

vshulcz:
Three things, all from #3505's checklist.

CI now gates what an update costs per class of change. `deja bench ingest` has reported it since #3507 and nothing looked at the number, which is how a path that re-tokenised the whole store on every new session survived two months (#3500). The gate is the shape rather than the clock — only a rewritten transcript may replace the records already on file, and every other class stays cheaper than that one — because a wall-clock budget on a shared runner measures the runner. Verified by putting #3500's shape back: two of the three assertions fail on it.

`doctor --json` gains the `claude-code` and `codex-hook` rows it never had (#3608). Both print in the text report; neither reached `auto_recall`, because both predate the table the other nineteen harnesses share. Each row is decided once now and read by both surfaces.

And the nightly migration matrix checks older config shapes, not only older stores: each release writes its own wiring, and this build has to read it as wiring rather than as nothing, and has to call it dead once what the entries name is gone. That last column is what found #3608 — the check asked and got `absent`, which reads the same as a machine where deja was never installed.

And one found by the pre-release stand rather than by the checklist: a build against an index path that holds a regular file deleted the file, silently, exit 0 (#3610). The swap parks whatever is at the path as `<dir>.old` and removes it on success, and `search` reached the same code through `tryLockDir`. Both refuse now, and `doctor` names the state — `index.state` gains `path-is-a-file` — because the advice attached to `missing` is a build that will not run there.

Fixes #3608, fixes #3610


vshulcz:
The clock assertion in the new gate failed on the runner and it was right to: `new transcript` came in at 149.8 ms against the rewrite's 148.3 ms, a 1% inversion, where the local margin was five-fold. The two shape checks both passed. Dropped it for the class list — the five classes have to be there and only one of them may rewrite — which is the claim a shared machine can hold, and the same lesson the step's comment cites from #3590.

Issue #3491, opened 2026-09-11
perf: where the weight is, and the three cheap levers that are not worth pulling


vshulcz:
A profiling pass over the whole tool on a real store (2,416 sessions, 1.8 GB of transcripts, 196 MB index). Three cheap levers measured and rejected; what is left is architectural, so it is written down here with the numbers rather than attempted in a hurry.

## Where the weight is

| | |
|---|---|
| index | 196 MB for 1.8 GB of sources (9:1) |
| `records.bin` | 135 MB (69%) |
| `buckets/` | 47 MB (24%) |
| everything else | ~4 MB |
| rebuild | 34.4s wall, 87s CPU |
| one `search` (relevance tier, cold process) | 865 ms wall, 2.55s CPU, **701 MB allocated**, 482 MB peak RSS |
| binary | 18.6 MB dev, 12.8 MB release (already `-s -w`) |

## Rejected, with numbers

**`compressFloor` 8192 → 2048** (deflate more records): index 196 → 163 MB (−17%), rebuild 34.4 → 36.0s, but the read side pays for it — `search` 865 → 1061 ms (+23%), `ctx` 891 → 1143 (+28%), `blame` 399 → 488 (+22%). `how` and `fix` unchanged, because they read the role outside the payload. Not worth it; the floor's own comment already said so and now there are numbers for today's code.

**`flate.BestCompression` at the same floor**: `records.bin` 135.2 → 134.9 MB (−0.2%) for +1.9s of rebuild. Pointless.

**Allocation-free guards in `redact.Text`**: the guards lowercase the whole message first, which is 1.8 GB of allocation over a rebuild and put `memclrNoHeapPointers` at 17% of the profile. Replacing the copy with a fold-aware byte scan made redaction *slower* — 740 ms → 824 ms over 13.8 MB of real text — because `strings.Contains` over a lowered copy is assembly and a hand-written Go scan is not. Reverted.

**Per-rune folding in `tokenHits`** (behaviour-identical, removes one copy per scored message, and `strings.ToLower` was 41% of that function): `search` 868 → 857 ms, `ctx` 875 → 863 ms, peak RSS unchanged. Within noise on wall time because the path is already parallel — CPU share is not wall share. Reverted under the rule that a fix which does not move the measurement does not land.

## What a rebuild actually spends

From `DEJA_CPUPROFILE` on `index --rebuild` (87s CPU):

[...]

PR #3615, opened 2026-09-14
fix: three from the redaction sweep — api-key spellings, --json, and show below the floor

### LINKED ISSUE #3614: a secret written api-key or api key is stored in the clear
Which spelling of "api key" a tool happened to print decided whether the value was stored in the clear.

```
api_key "s3cretvalue123"      → api_key "[redacted:quoted-secret]"
apikey "s3cretvalue123"       → apikey "[redacted:quoted-secret]"
api-key "s3cretvalue123"      → api-key "s3cretvalue123"
api key "s3cretvalue123"      → api key "s3cretvalue123"
API-KEY: "s3cretvalue123"     → API-KEY: "s3cretvalue123"
x-api-key: "s3cretvalue123"   → x-api-key: "s3cretvalue123"
```

Two causes, pulling opposite ways. The gate in front of `quotedSecretRE` tests `api key`, `api_key` and `apikey`, while the pattern itself accepts `api[_-]?key` — so the hyphen spelling matched a pattern the gate never let run. And the space spelling is the mirror image: the gate admits it, the pattern has no space in it.

The colon form is a third gap. `x-api-key: "…"` is a header rather than prose, so the quoted rule — which wants whitespace after the name — does not reach it, and `genericKVRE`, which owns assignments, starts at sixteen characters of value. A fourteen-character key falls between them, which is exactly the floor the quoted rule was added to get under (#2372's shape).

Same family as #3535, #3572, #3587 and #3596: redaction runs at ingest, so text written before the fix is text this build would not write, and the redaction floor has to move with it.

### LINKED ISSUE #3616: --json carries the bidi and invisible-tag characters the printed surfaces strip
`--json` hands out the display attacks every printed surface strips. Planted in one session and read back:

```
search --json   bidi override, invisible tag block, zero-width space
last --json     bidi override, invisible tag block
show --json     bidi override, invisible tag block, zero-width space
blame --json    bidi override, invisible tag block, zero-width space
```

The text path is clean — `SafeText` has stripped U+202E and the tag block since #1090 — and so is the MCP listing, which #1080's neighbour fixed for the same reason. The JSON paths encode `model.Session` as stored.

A control byte is why this went unnoticed: `encoding/json` escapes those, so the obvious half of the attack cannot reach a terminal through JSON. U+202E and U+E0000–U+E007F are ordinary characters to the encoder, and they are the half that survives — the first reorders a rendered line, the second is an invisible ASCII alphabet, which is exactly what #1090 was about.

Who renders it: a dashboard, `jq -r` into a terminal, anything that decodes and prints. `docs/json-output.md` says a raw control byte cannot reach a terminal through JSON, which is true and is not the whole set.

Fix is the filter the printer already uses, applied to the session, its title, project, the snippets and the lifecycle note before encoding — newlines and tabs kept, since a JSON string can hold them and the shape of a message is worth something.

### LINKED ISSUE #3617: show serves a store the redaction floor withholds
`deja show` prints a session from a store written before deja learned to mask something it now masks.

Indexed by 0.20.0, read by a build whose redaction floor has moved, with no rebuild in between:

```
search           secret in output: no
recall           no
last             no
blame            no
how              no
hook-context     no
show             YES
```

Every ranked surface reaches the plan gate, which rebuilds before answering when the store is below the floor. `show` does not rank: with `--harness` it goes straight to `index.FindByIdentity`, which reads the manifest and loads the records. So the one command whose whole job is to print a session verbatim is the one that printed the text this build would not write — the shape #3598 fixed for the session-start block, on the surface that shows the most at once.

The same loader answers the MCP tools, so an agent asking for a session by identity gets the same bytes.

The fix belongs in the loader rather than in `show`: refuse there, name the state, and let the commands that can rebuild do so first.

vshulcz:
Which spelling of "api key" a tool printed decided whether the value was stored in the clear: the gate in front of the quoted-secret rule tested three spellings while the rule's own pattern accepted a fourth, and the colon form — `x-api-key: "s3cretvalue"` — fell between that rule and the key-value one, whose value class starts at sixteen characters.

The gate also picks up #3491's one remaining cheap narrowing: the pattern cannot match without a quote character, so `ContainsAny` in front of it is strictly weaker than the pattern and only stops runs that could not have matched.

Index format version 46, and the redaction floor moves with it, because redaction runs at ingest. Both halves are mutation-checked: dropping the hyphen from the gate leaves `api-key "…"` in the clear, and removing the colon rule leaves four of the eight spellings.

Second fix from the same stand: `--json` handed out the bidi overrides and the invisible tag block that the printed surfaces have stripped since #1090 (#3616). The encoder escapes control bytes, which is the half that was covered; U+202E and U+E0000–U+E007F are ordinary characters to it, and they reach whatever renders the text. `search`, `last`, `show` and `blame` now filter what they encode, newlines and tabs kept.

Third from the same stand, and the one that matters most: `deja show` printed a session from a store the redaction floor withholds (#3617). Every ranked surface reaches the plan gate and rebuilds first; `show` loads by identity, and so do the MCP tools that share that loader. It refuses there now and names the state, and `show` runs the same pass the prefix form already ran, so it answers with the text masked rather than as stored.

Verified with the upgrade stand: index with 0.20.0, read with this build and no rebuild in between — `search`, `recall`, `last`, `blame`, `how` and the hooks were already silent about the planted secret, `show` was not, and now neither is.

Fixes #3614, fixes #3616, fixes #3617


vshulcz:
Self-review caught one of mine in the new colon rule: it ran over text the key-value rule had already masked and counted the match it then declined, so `{"token": "ghp_…"}` reported both `credential` and `quoted-secret` for one masked value — and those counts are the `redacted=N` that `sources` and `doctor` print. Moved to `replaceGroup`, which counts nothing for a match it skips; the test now pins the count as well as the marker, and fails both ways with the skip removed.

Issue #2606, opened 2026-08-29
two YAML configs do not come back as the reader left them


vshulcz:
#2604 gave the JSON writers the rule: a container deja added goes with deja, one the reader wrote stays. Sweeping the rest of the family — seed a config of the reader's own for codex, grok, goose, hermes, zed, cline, copilot, openclaw, qwen and kimi, install, uninstall, diff against the original — two come back changed:

    ~/.hermes/config.yaml
      1a2,3
      > 
      > mcp_servers:

    ~/.config/goose/config.yaml
      1c1
      < GOOSE_MODEL: gpt-5
      ---
      > GOOSE_MODEL: gpt-5
      \ No newline at end of file

Hermes is #2604 in YAML: deja adds the `mcp_servers:` block when the file has none and removes only its own entry, so the reader keeps a block they never wrote.

Goose is smaller and stranger: the file comes back without its trailing newline. `dropEmptyYAMLKey` removes the key and the blank lines after it, and the last of those is the file's own final newline — a diff in someone's dotfiles for nothing.

The other eight come back byte-identical.

vshulcz:
Fixed in #2607.

PR #3704, opened 2026-09-17
fix: a guidance block round trip leaves the reader's file as it was

### LINKED ISSUE #3703: an install and uninstall leave the reader's AGENTS.md one blank line longer
grok is the one harness left whose guidance is a marked block inside a file the reader owns (`~/.grok/AGENTS.md`). An install and an uninstall give that file back one blank line longer than it was:

```
$ printf '# My own notes\n\nAlways run `make check` before pushing.\n' > ~/.grok/AGENTS.md
$ deja install grok && deja uninstall grok
$ diff before.md ~/.grok/AGENTS.md
6a7
>
```

56 bytes in, 57 bytes out. It does not accumulate — five round trips leave it at 57 — so it is one line in someone's dotfiles, once, which is the same cost #2606 fixed on the goose side from the other direction (their file came back *without* its final newline).

The cause is in the pair: install does `TrimRight(old, "\r\n")` and then appends `newline + newline + block`, and the uninstall cuts the block alone — so the separator install added stays behind. Cutting one preceding blank line with the block gives the file back exactly.

The rest of that writer holds up: their text survives both ways, a hand-edited block is still recognised and removed, and a block whose end marker someone deleted is refused with the file named and their text untouched (`deja's guidance block has no end marker — put it back, or delete the block entirely`).

vshulcz:
Grok Build's guidance is a marked block inside a file its owner writes, and an install followed by an uninstall gave that file back one blank line longer:

```
$ printf '# My own notes\n\nAlways run `make check` before pushing.\n' > ~/.grok/AGENTS.md
$ deja install grok && deja uninstall grok
$ diff before.md ~/.grok/AGENTS.md
6a7
>
```

56 bytes in, 57 out. It does not accumulate — five round trips stay at 57 — so it is one line in somebody's dotfiles, once, which is the same cost #2606 fixed on the goose side from the other direction. Install does `TrimRight(old, "\r\n")` and appends `newline + newline + block`; the uninstall cut the block alone and left the separator. Cutting one preceding blank line with it gives the file back exactly, for LF and CRLF alike.

The rest of that writer holds up and the test says so: their text survives both ways, a hand-edited block is still recognised and removed, a block whose end marker someone deleted is refused with the file named and their text untouched, and on a bare home both files deja created (`GROK.md` and `AGENTS.md`) go on the way out. The trailing run of blank lines still collapses to one at install time — that is the file's state while deja is wired, so it is pinned as the contract rather than left to be found later.

Also here, because it is how tonight's measurement was made: `scripts/longmemeval -dump-misses <path> -dump-all` writes every question rather than only the ones ranked worse than 1. A reordering cannot be judged without what it would demote — the relevance tier answers 373 of 470 questions and 318 of those at rank 1, which the miss-only dump cannot show.

Fixes #3703.


vshulcz:

PR #3708, opened 2026-09-17
ci: decide the windows leg in the run that posts the check

### LINKED ISSUE #3707: a labelled PR can get a windows check that ran nothing and still says success
A pull request labelled `windows` can still get a windows check that ran nothing, and it reports the same green as a real pass. That is how `main` went red in #3699.

#3693 carried the label. `gh pr checks` said `test / windows-latest SUCCESS`. The job took four seconds:

```
Set up job                                    success
Run actions/checkout@…                        skipped
Run actions/setup-go@…                        skipped
Run go test -timeout 25m ./...                skipped
Run go build ./cmd/deja                       skipped
Complete job                                  success
```

Then the same three tests failed on `main` immediately after the merge.

The cause is where `FULL` comes from:

```yaml
FULL: ${{ matrix.os != 'windows-latest' || github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'windows') }}
```

`github.event.pull_request.labels` is the payload of the event that started **that** run. Opening a PR and labelling it a second later starts two runs — one without the label, one with — and the concurrency group cancels one. When the survivor is the unlabelled run, every step skips and the check still posts success. The workflow comment above that line says `labeled` is in the trigger set for exactly this case; what it cannot fix is which of the two racing runs lives.

The label is a fact about the pull request, not about the event, so read it as one: resolve it at job start from the API and use that for the step conditions.

```yaml
- name: Does this leg do real work
  id: leg
  run: |
    if [ "${{ matrix.os }}" != "windows-latest" ] || [ "${{ github.event_name }}" != "pull_request" ]; then
      echo "full=true" >> "$GITHUB_OUTPUT"; exit 0
    fi
    labels=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}" --jq '[.labels[].name] | join(",")')
    case ",$labels," in
      *,windows,*) echo "full=true" >> "$GITHUB_OUTPUT" ;;
      *)           echo "full=false" >> "$GITHUB_OUTPUT" ;;
    esac
```

and say in the job summary which of the two it was, so the four-second run is legible in the run list rather than only in its step list.

This does not make an unlabelled PR run the leg — that opt-in is deliberate, the runner is ten times slower — it makes a labelled one stop lying about having run it.

vshulcz:
#3693 carried the `windows` label, its `test / windows-latest` check said SUCCESS, and that job ran for four seconds with every step skipped:

```
Set up job                       success
Run actions/checkout@…           skipped
Run go test -timeout 25m ./...   skipped
Run go build ./cmd/deja          skipped
Complete job                     success
```

The three tests that leg would have caught turned `main` red on the next push (#3699). Nothing distinguished that green from a real pass.

The decision came from `contains(github.event.pull_request.labels.*.name, 'windows')` in the matrix job's `env` — the payload of the event that started *that* run. A PR opened and labelled a second later has one run without the label and, when the `labeled` event produces a run at all, one with it; for #3693 only the unlabelled one existed.

So the decision moves into a job of its own, in the same run that posts the checks:

- `windows leg` resolves it once — every push to main, the canary and a manual run take the suite; on a pull request it reads the label from the API, and failing that the diff, using the same path rules `windows-label.yml` had (a parser change is what broke this leg last, and nobody labels one `windows`). It still adds the label when the paths ask, so the pull request says why.
- `test` waits on it and reads `needs.windows-leg.outputs.full`.
- **And a skipped leg still builds.** `go build ./cmd/deja` on Windows is about two minutes, and it makes the green mean something the platform's own compiler said about this commit, instead of meaning nothing. `go mod verify`, `go vet` and the suite stay behind the decision.
- The run's summary says which of the two happened, so a four-second job is legible in the run list rather than only in its step list.

`windows-label.yml` is folded in and deleted; its reasoning moved into the comment above the new job. It was not a required check, and `test / windows-latest` still is.

This pull request is its own test: it touches `.github/workflows/ci.yml`, which is one of the paths that asks for the leg, so the new job should label it and run the whole suite.

Fixes #3707.


vshulcz:
It happened again while this was open, on #3709 — a second occurrence within the hour, and this time the label was added by the labeller rather than by hand:

```
$ gh pr view 3709 --json labels --jq '[.labels[].name]'
["windows"]

$ gh api …/runs/35283574844/jobs --jq '.jobs[]|select(.name|test("windows-latest"))'
{"conclusion":"success","started_at":"22:45:12Z","completed_at":"22:45:16Z","ran":2,"skipped":12}
```

Four seconds, twelve steps skipped, two "ran" (Set up job, Complete job), and `test / windows-latest` green. One CI run for that branch, from the `opened` event; the `labeled` event produced none, so nothing ever re-evaluated the label — the same shape as #3693.

This pull request's own leg is the counter-example: the `windows leg` job resolved the decision, labelled this PR itself, and `test / windows-latest` is running checkout, setup-go and the suite rather than skipping them.

One consequence worth stating for the merge order: a `pull_request` run uses the workflow from the PR's head, so #3704 and #3709 keep the old behaviour until they are updated onto a main that has this. Their windows legs are worth nothing before that.

Issue #3167, opened 2026-09-07
every search stats 11,000 files to find out nothing changed: the store walk costs 452 ms


vshulcz:
Every CLI search walks every harness's roots before it can decide whether anything changed. On this machine that is the largest fixed cost left in the path.

Traced on the live store (`DEJA_TRACE=1 deja <query>`):

```
trace lock                0.1ms
trace manifest            3.0ms
trace walk stores       451.7ms
trace hand to warmup      2.0ms
```

452 ms, paid whether or not a single byte changed. The same trace on a synthetic store of 1,531 sessions in one root costs **7.5 ms**, so it is not the session count — it is the number of files under the roots:

```
~/.codex                                              5,541 files
~/.claude/projects                                    2,322
~/…/Code/User/workspaceStorage                        2,298
~/.gemini                                               890
~/.grok/sessions                                        169
~/.cursor/projects                                       51
```

About 11,000 stats, on a store where 26 harness rows are configured and most of them hold single digits of sessions. `currentFilesReusing` already reuses the prior manifest's state for unchanged files, so the parse work is avoided — the walk that finds them is not.

Worth measuring before choosing a fix, roughly in this order:

1. Which roots dominate. `~/.codex` at 5,541 files is the largest and holds 34 sessions; if the walk is descending into things that are not transcripts, a tighter glob is the whole fix.
2. Whether a root's directory mtime can skip its subtree. A root nothing has written to since the manifest was built cannot have changed a file inside it on most filesystems, and that is one stat instead of thousands.
3. Whether the walk can be deferred like the ingest already is: answer from the snapshot, walk in the detached warmup. The receipt for that already exists ("answering from the index as it was").

Found while trying to reproduce #3021, which does not reproduce at its stated shape — a 40 MB live file with a 13-message tail now costs 55 ms inline. This is what is left.

vshulcz:
Correcting the file counts above: they came from a `find` over each whole root, and deja does not walk whole roots — `CodexFiles` only walks `~/.codex/sessions`, so the 5,541 figure for `~/.codex` describes plugins and logs deja never touches. Sorry for the noise.

Instrumented the walk per harness instead. On the live store, three runs, same to within a millisecond:

```
walk copilot-chat     72.3ms    42 files
walk claude            3.3ms  1177 files
walk antigravity       4.0ms    50 files
walk cursor            1.3ms    19 files
walk copilot           1.4ms    15 files
walk deepseek          1.2ms    49 files
[...]

PR #3709, opened 2026-09-17
feat(copilot-chat): read the extension's own transcripts

### LINKED ISSUE #3637: copilot-chat: VS Code 1.137 writes chats to extension transcripts + session-store.db, so the source finds 0 files
## Summary

On VS Code **1.137.0** the `copilot-chat` source finds **0 files**: the
`<User>/workspaceStorage/<hash>/chatSessions/` directory the reader looks in does not exist here at
all, and neither does `globalStorage/emptyWindowChatSessions/`. The chat history this machine has was
written to the stores below instead. (`registry/copilot-chat.md` documents the `chatSessions` layout
and was last verified on 1.134.0.)

| store | what it holds | on this machine |
| --- | --- | --- |
| `<User>/workspaceStorage/<hash>/GitHub.copilot-chat/transcripts/<id>.jsonl` | the extension's own transcript, one file per chat | **47 files — 36 with at least one user message, 52 MB, 12 workspaces, 2026-07-02 → 2026-09-16** |
| `<User>/globalStorage/github.copilot-chat/session-store.db` | SQLite "chat sessions" store | **6 sessions / 22 turns, all created 2026-09-14 or later** — the file itself dates from 2026-07-09, so it never held the older sessions. Every id in it also exists as a transcript: it adds `cwd`, not history |
| `<User>/workspaceStorage/<hash>/GitHub.copilot-chat/debug-logs/<id>/main.jsonl` | OTel spans only (`session_start`, …), no conversation | 83 dirs; not a candidate source |

There is **no** `chatSessions` directory and **no** `emptyWindowChatSessions` directory anywhere
under the User folder (checked with `find <User> -type d -name chatSessions` → 0).

The transcripts are the only complete history here: the SQLite store would recover the last two days,
the transcripts the last eleven weeks. A fix that reads only the store is therefore not enough.

## Environment

- deja **0.20.1** (0.20.0 behaves identically), Linux, VS Code **Server 1.137.0** (`645f29cc31`), reached over WSL (`wsl+ubuntu`).
- Transcripts are written by `copilot-agent` 0.54–0.65 (`producer` / `copilotVersion` in `session.start`).
- User folder here is the **Server** layout, `~/.vscode-server/data/User`. The desktop `~/.config/Code/User` exists but holds no chat data on this machine.

## Reproduction

```sh
deja doctor                                                   # copilot-chat missing  ~/.config/Code/User  (0 files)
DEJA_COPILOT_CHAT_ROOTS=~/.vscode-server/data/User deja doctor # says the same for the Server User folder
find <User> -type d -name chatSessions | wc -l                 # 0
find <User> -path '*GitHub.copilot-chat/transcripts/*.jsonl' | wc -l   # 47
```

So the read override works and the User folder is correct — what is missing is the format. deja's
own matrix says this harness needs no external tool, which matches a pure-JSON reader.

## Transcript record shape as written on 1.137.0

This is a `type`-discriminated event log, not the `kind` 0/1/2 incremental delta log the parser
handles today: different directory, different records, so it needs its own reader rather than a
branch inside the replay function.

One JSON object per line: `{"type": …, "data": …, "id": …, "timestamp": …, "parentId": …}`.

```jsonc
{"type":"session.start","data":{"sessionId":"…","version":1,"producer":"copilot-agent","copilotVersion":"0.54.0","vscodeVersion":"1.137.0","startTime":"2026-09-14T05:46:01.086Z"},"id":"…","timestamp":"…","parentId":null}
{"type":"user.message","data":{"content":"…","attachments":[]},"id":"…","timestamp":"…","parentId":"…"}
{"type":"assistant.message","data":{"messageId":"…","content":"…","toolRequests":[{"toolCallId":"…","name":"read_file","arguments":"{\"filePath\":\"/home/user/code/work/projA/…\"}","type":"function"}]},"id":"…","timestamp":"…","parentId":"…"}
{"type":"tool.execution_start","data":{"toolCallId":"…","toolName":"manage_todo_list","arguments":{…}},"id":"…","timestamp":"…"}
{"type":"tool.execution_complete","data":{"toolCallId":"…","success":true},"id":"…","timestamp":"…"}
```

`assistant.turn_start` / `assistant.turn_end` come in pairs, one per assistant turn. Type histogram
for one 8-turn session (366 KB): `tool.execution_complete` 132, `tool.execution_start` 132,
`assistant.message` 63, `assistant.turn_end` 63, `assistant.turn_start` 63, `user.message` 7,
`session.start` 1.

Reading notes that may save you time:

- user turns are the `user.message` events; the reply text is `assistant.message.data.content`
  (an `assistant.message` can carry `toolRequests` and empty `content`);
- files are append-only and keep growing while a session is live (a byte-offset resume would
  restart on the last partial line — the same reason the current reader full-parses per pass);
- 11 of the 47 files have **no** user turn at all (agent-only runs) — those are the ones a reader
  should skip rather than emit as empty sessions;
- `debug-logs/<id>/main.jsonl` looks like a per-session store by name but is spans only.

## session-store.db schema

```sql
sessions(id, cwd, repository, host_type, branch, summary, agent_name, agent_description, created_at, updated_at)  -- 6 rows
turns(id, session_id, turn_index, user_message, assistant_response, timestamp)                                    -- 22 rows
session_files(id, session_id, file_path, tool_name, turn_index, first_seen_at)                                    -- 75 rows
checkpoints(id, session_id, checkpoint_number, title, overview, history, work_done, technical_details, important_files, next_steps, created_at)  -- 0 rows
```

All 6 `sessions.id` values also exist as transcript ids, so on this machine the DB adds nothing the
transcripts do not already have — except `cwd`, which the transcripts only hint at through the
absolute paths their tool calls touch.

## Workspace attribution (for whichever reader you pick)

The storage directory name is the **md5 of the workspace URI**. Here:
`md5("vscode-remote://wsl%2Bubuntu" + "/home/user/code/work/projA")` reproduces the
`workspaceStorage/<hash>` name exactly — verified against the two hashes whose `cwd` the SQLite
store records, then used to resolve all 12. A reader that has `cwd`, or that enumerates the
ancestors of the paths a session touched, can therefore name the project instead of emitting a hash.

## Local workaround (context, not a request)

I materialize the transcripts into the JSON deja already parses (`requests[]` with `message.text`
and `response[].value`) under `<User>/workspaceStorage/<own-hash>/chatSessions/<id>.json` plus a
`workspace.json`, and deja indexes that fine: `copilot-chat sessions=36 messages=2814 size=13.3 MB`,
searchable, with correct project names. It works, but it is a copy that has to be refreshed and it
duplicates data — it exists only because deja cannot read the real store.

## What would fix it

Either of these would remove the workaround:

1. **The fix:** read `<User>/workspaceStorage/*/GitHub.copilot-chat/transcripts/*.jsonl` as a second
   layout for `copilot-chat` (records above; user turns from `user.message`, replies from
   `assistant.message.data.content`, tool paths from `toolRequests[].arguments`). This is the only
   store holding the full history.
2. **Optional, and not a substitute:** read `globalStorage/github.copilot-chat/session-store.db`
   through the `sqlite3` CLI the way opencode / Cursor / Zed already are. It covers the newest
   sessions only, but it is the one place that records each session's `cwd`, so it is the cheapest
   way to get project attribution right for those.

Related: the same drift class for DeepSeek Harness (`session.jsonl.zstd` → `session.v3.jsonl.zstd`)
was fixed in 0.20.1 (#3508). This looks like the copilot-chat equivalent.

## Notes for a fixer

- Where this lands: `internal/sources/copilot_chat.go` — `CopilotChatRoots` / `copilotChatStoreDirs`
  (discovery), `copilotChatFile` / `copilotChatMatch` (matching), `ParseCopilotChatFile` /
  `copilotChatReplay` (parsing). Selection is per file, so the existing `kind`-based path is untouched.
- Do not match on the parent directory name alone: `transcripts` is a generic name — this repository's
  own `internal/sources/claude.go` reads Claude Code's `<config>/transcripts`. Match
  `<hash>/GitHub.copilot-chat/transcripts/<id>.jsonl`.
- No `<hash>/workspace.json` exists in any of the 12 workspace dirs here, so
  `copilotChatProjectFromWorkspace` cannot name the project; the md5 reverse-lookup above, or the
  `cwd` column in `session-store.db`, are what work.
- The transcript carries no tool results: `tool.execution_complete` is `{toolCallId, success}` only
  (3,000 records sampled, no other key), and `assistant.message.reasoningText` is the model thinking
  out loud. Index `user.message.content` and `assistant.message.content`, take tool names and paths
  from `toolRequests[]`, skip `reasoningText`.
- Discovery cost: `copilotChatStoreDirs` documents (#3167) that walking 2,643 workspace directories to
  reach 28 costs 72 ms of a 95 ms walk, which is why it asks each one for `chatSessions` directly.
  Adding a second directory per workspace directory doubles that term, so whatever the fix does there
  deserves a sentence in the PR.
- Fixtures and tests: `fixtures/registry/copilot-chat/workspaceStorage/<hash>/chatSessions/<id>.jsonl`
  is the existing shape; tests are hermetic (`t.TempDir()`, `t.Setenv` for `HOME` and the `DEJA_*`
  overrides), `internal/sources` has an 87% coverage floor in CI, and `docs/registry/copilot-chat.md`
  plus `docs/registry/registry.json` are checked by tests, so they move in the same commit.

vshulcz:
Reads `workspaceStorage/<hash>/GitHub.copilot-chat/transcripts/<id>.jsonl`, which on a newer Copilot Chat is the only store with the history in it.

@yizhixiaokong's report did the part that is hard to redo — the layout, the record histogram, both stores and what each one actually holds — and it reproduced here on a different host: desktop VS Code 1.136.1 has 28 `chatSessions` directories that deja read and five transcripts it did not. So this is neither Server-only nor 1.137-only.

**What it does.** A `type`-discriminated event log rather than the `kind` 0/1/2 delta log beside it, so it is its own reader rather than a branch in the replay: user turns from `user.message`, replies from `assistant.message.data.content`, the file a tool call names and the command it ran (a command is what `deja how` and `deja fix` are built on, so it is filed as one), and the project from the `workspace.json` two levels above a transcript — one more than for a `chatSessions` file — falling back to the absolute paths the session touched.

**A transcript with no user turn is indexed.** 11 of the reporter's 47 and all five here have none; they are agent runs, and what the agent said and the files it touched are the only record of that work. The delta log is a different session store, not a copy of them.

**Measured on this machine:** `copilot-chat 17 sessions, 1822 messages` → `22 sessions, 1832 messages`, and a search for a phrase from one of the five returns it with its project resolved. Discovery costs one more failed open per workspace directory — the term #3167 replaced the walk with — which came to about 20 ms across 179 of them on `deja sources` (422–431 ms before, 429–475 ms after).

**Not read:** `globalStorage/github.copilot-chat/session-store.db`. Every id in it also exists as a transcript, so it adds each session's `cwd` rather than history, and on the reporter's machine it covered two days against the transcripts' eleven weeks. The registry page says so rather than leaving it as an unexplained gap.

Tests: the two shapes (with and without user turns), a `transcripts` directory belonging to another extension (the trap in the report — this repository reads a `transcripts` of its own for another harness), and a half-written last line, which is the ordinary state of a live chat.

Two things ride along:

- the registry pages for commandcode, gjc and kilocode were regenerated: #3701 changed their `.md` (the skill moved to `skills/deja-history/`) and the `.html` beside them still said `deja-search`;
- `TestDoctorNamesASessionServedTwiceInASecond` stamped three events 300µs apart at `time.Now()`, so around a second boundary they landed in two seconds and the row read 2 where the test wrote 3 — a 0.06% flake in a test added yesterday, now stamped inside one second on purpose.

Fixes #3637.


vshulcz:

Issue #3706, opened 2026-09-17
`deja wip` lists words as files in flight


vshulcz:
`deja wip` listed words as files:

```
$ deja wip
working on: …
settled: …
files in flight: /Users/shulcz/…/tmp/scope_audit.sh, was., it, as, exactly
last command: …
```

`was.`, `it`, `as` and `exactly` are not files. The collection takes every whitespace-separated token of a files or edit record as a path:

```go
case sources.RoleEdit, sources.RoleFiles:
	for _, f := range strings.Fields(text) {
		if f = strings.TrimSpace(f); f != "" {
			files = append(files, f)
		}
	}
```

So any such record whose text carries a sentence — a tool result phrased as prose, a reader that files a description under the files role — contributes one "file" per word, and they crowd out the real paths, which are capped at a handful (`resumeFileCap`).

This is the line an agent reads to pick up its own work after a compaction, so the cost is the real paths being pushed out by words, and an agent opening `it` or `exactly`.

A token is worth naming as a file when it looks like one: it carries a path separator, or an extension, or begins with `/`, `~` or `.`. Everything else is prose that reached the wrong role.

Issue #3705, opened 2026-09-17
`deja how` answers from the whole machine: in one repo it returns another repo's wrapper command


vshulcz:
`deja how` answers from the whole machine, so an agent asking how to run the tests *here* gets another repository's wrapper.

Run inside `~/coding/goprojects/deja-vu`:

```
$ deja how "test"
$ rtk go test ./...                        ran 40 times in 24 sessions · last 2026-09-02
$ rtk go test ./internal/domain/partner    ran 28 times in 22 sessions · last 2026-05-26
$ rtk go test -v ./internal/domain/partner ran 11 times in 10 sessions · last 2026-05-26
$ go test ./...                            ran 24 times in 8 sessions  · last 2026-09-04
```

`rtk` is a wrapper another project on this machine uses; the top three answers are that project's package paths. This repository's own test command — `go test ./cmd/... -count=1`, run dozens of times today — is not in the eight. Pass the project and the answers are right:

```
$ deja how "test" --project deja-vu
$ cd …/wt && go test ./cmd/deja …
$ go test ./cmd/deja/ 2>&1 | tail -5
```

The cause is one line: `runHow` starts with `project := ""` and only fills it from `--project`. Nothing derives the project from the working directory, and the ranking is by run count across every project on the machine, so the busiest repository wins every question asked in any other.

It is the one surface that does this. `deja wip` is "what the last session **in this directory** was doing", recall's ranking carries an in-project signal, and the hooks pass the session's cwd. `how` is also the surface whose answer gets *run*: following it costs a wrong command in the wrong repository, or a `command not found` for anyone without that wrapper.

What it should do: default to the project the working directory belongs to, using the same derivation the hooks and `wip` use; keep `--project` for asking about another one; add a flag to widen back to the machine, and say in the text output which of the two it answered; fall back to the machine when the current project has nothing, rather than "no command on this machine mentions …"; and put the project in `--json` rows, which today cannot be told apart.

vshulcz:
`deja files` has the same default and the same symptom. Asked inside `~/coding/goprojects/deja-vu`:

```
$ deja files install
files touched while working on "install" — 218 sessions (of 390 matching sessions, the 250 most recent were read)
  …/internal/infrastructure/routes/routes.go               22
  …/internal/infrastructure/macos/topology.go              16
  …/routepilot/internal/application/vpn.go                 75
```

[...]

Issue #3681, opened 2026-09-17
every build that ever installed leaves its own hook entry — six copies per event on one machine


vshulcz:
Reported as "the hook seems to fire several times", and it does. On this machine `~/.claude/settings.json` had **six** deja entries per event:

```
UserPromptSubmit   <job dir>/tmp/deja-cont hook-prompt
UserPromptSubmit   <job dir>/tmp/deja-sl hook-prompt
UserPromptSubmit   <job dir>/tmp/deja-prog2 hook-prompt
UserPromptSubmit   <job dir>/tmp/deja-shapes hook-prompt
UserPromptSubmit   <job dir>/tmp/deja-arm hook-prompt
UserPromptSubmit   <job dir>/tmp/deja-prog hook-prompt
```

Claude Code runs every matching hook, so that is six processes per prompt and the recall block arriving several times in one turn. The same stacking, counted across the other config-based harnesses on the same machine: codex `hooks.json` 30 deja commands, cursor 35, qwen 32, grok 68.

**Cause.** An entry is recognised as deja's own by `hookCommandKindOf`, which accepts either the exact command string being written or a binary whose *name* is `deja`/`deja.exe`/`deja-hook`, or a path in `wiring.json`'s `exes` record. A build called `deja-cont` (what `go build -o` and every probe run leave) satisfies none of those: the record is capped at ten and on this machine held three paths, none of them the six that wrote the entries. So each install read the existing lines as somebody else's hook, left them alone, and added its own. The collapse logic from #3421 is right there and never fires.

**Why it was invisible.** doctor's own repeat note (#3421, `runs deja more than once per event …`) asks the same question through the same predicate, so six entries under six names counted as one and the row said `wired`. The only surface that did tell the truth was the injections detector — `one agent session was served 12 dejavu injections inside a second … the hooks are wired more than once` — which is in the Index section, not beside the hook row.

**Fix.** Recognise the pair rather than the name: a binary whose basename starts with `deja` **and** which is running one of deja's own hook subcommands is deja's entry. That is the same reasoning #3659 used for doctor's binary checks. A wrapper line that merely contains deja's hook stays `hookWrapsDejas` and is left alone, and a foreign `other hook-prompt` stays foreign.

[...]

Issue #3659, opened 2026-09-17
a config naming a deja build under another basename is invisible to every check


vshulcz:
A config that names a deja build under a basename other than `deja` is invisible to every check doctor has. On the machine this was found on:

```yaml
# ~/.hermes/config.yaml
  deja:
    command: "/Users/…/.claude/jobs/<id>/tmp/deja-cont"
```

`deja doctor` prints `hermes  wired  guidance written  ~/.hermes/config.yaml` and nothing else, because:

- `isDejaBinaryToken` accepts a basename of `deja`, `deja.exe` or `deja-hook`, or a path recorded in `dejaWrittenExes()`. `deja-cont` is none of those.
- so `dejaCommandIn` returns nothing, and both the missing-binary check (#2216) and the stranger-binary check (#3656) stay silent.
- the row still says `wired`, because the server key is `deja` and that is what the wiring check reads.

The same reason makes the entry survive an uninstall: removal matches entries by what they name.

Where this comes from is not exotic. Anyone who builds deja to try something — `go build -o /tmp/deja-probe ./cmd/deja` — and runs `install` from it leaves an entry no later deja recognises, unless the path is still in the written-exe record. Which is the part worth noticing: the `deja-shapes` entries in #3656 are visible *only* because that path is in the record, so the record is load-bearing, and there is no test that says what happens when it loses one.

Options, cheapest first:

1. Widen the basename rule to any file whose name starts with `deja` — catches `deja-probe`, `deja-cont`, `deja-shapes`, and misses a binary called something else entirely.
2. Treat a `command` under a server key deja owns (`deja`) as deja's whatever it names: the key is the claim, and deja wrote it. Then the row can say "runs X" for any X and the uninstall can take it out.
3. Ask the binary: run `<path> --version` and read it. Honest, and executing a path out of a config file to produce a health report is a worse idea than the report being incomplete.

(2) looks right and (1) is a cheap safety net beside it. Either way the check that matters is the one that does not exist yet: a test that an entry written from a differently-named build is still recognised, reported and removable.

vshulcz:
Fixed in #3658 by the second option: the server key is the claim. An entry under the `deja` key is deja's whatever binary it names — read from JSON, and from the YAML and TOML spellings for the formats doctor parses as text (Hermes' `deja:` block is the case this came from). Tests cover all three shapes plus the negative one, where another server's command stays out of it; mutation-checked by removing each half in turn.

[...]

PR #3710, opened 2026-09-17
fix: answer about the project deja was asked in

vshulcz:
Three bugs in the answers deja gives an agent, rather than in the wiring.

**#3705 — `deja how` answered about the wrong project.** It read the whole machine and ranked by how often a command had run anywhere, so the busiest repository answered every question asked in any other: inside this one, `deja how "test"` returned another project's `rtk go test ./...` wrapper (40 runs) three times before this repository's own command, which was not in the eight at all. The scope is now the project of the working directory, derived the way `deja wip` and the hooks derive it (worktrees included). `--project` still asks about another one, `--all-projects` asks the machine, the text says which of the two answered, `--json` carries it as `project`, and the MCP tool takes the same default. A project with nothing to say still falls back to the machine's answer — an agent told nothing exists invents one.

**#3713 — `deja files` had it too.** Ranked over the machine by how much a file was touched, a generic topic answered with the busiest project's tree: asked in this repository, `deja files "timeout"` named eight paths — SQL migrations and a Python service from two unrelated projects — and not one was here, against the four this repository has on the topic. Same default, flags and widening as `how`. One thing was in the way: `search.Options` carried a single `Project` substring and a project has several name forms, so the filter takes a set now and a worktree is still the project it belongs to.

**#3706 — `deja wip` named words as files.** Every whitespace-separated token of a `files`/edit record counted as a path, so a record carrying a sentence contributed one "file" per word: `files in flight: …/scope_audit.sh, was., it, as, exactly`, with the words crowding the real paths out of the handful that get named. A token has to look like a path now — a separator, an extension, or a leading `/`, `~` or `.`. A bare `Makefile` drops out, which is the price of not printing prose.

Also carries the invariant for #3659, which was already fixed by #3681's predicate but never pinned: install every auto target with the binary under a different name, and no writer may read deja's own entries as a stranger's.

Labelled `windows` — paths and cwd-derived project names.


vshulcz:

PR #3712, opened 2026-09-17
fix: asked-twice ignores one conversation under several session ids

### LINKED ISSUE #3711: asked-twice shows the reader their own resumed conversation
On a 2,300-session store (a year of history, 19 harnesses, stands and benchmark corpora excluded) every repeat `FindAskedTwice` can see is one conversation written under more than one session key:

```
repeat groups >=48h apart: 9      one conversation under several keys: 9
winner as shipped: "ты точно весь репозиторий привел в порядок? …"
   ses_1b636323a…  opencode  2026-06-16  asked=8  mpc-autoscaler
   ses_24e62619d…  opencode  2026-05-29  asked=8  mpc-autoscaler
   ses_1b636aadb…  opencode  2026-05-21  asked=8  mpc-autoscaler
   ses_1b636af66…  opencode  2026-05-21  asked=8  mpc-autoscaler
```

Four session ids, the same eight asked hashes in each, one conversation continued over four weeks. The brief's line — "a question you asked in more than one session" — is telling the reader their own resume back to them as a repeat, and on this store that is the only thing it has ever had to show.

A resume, a fork or a share is what produces it: the harness writes the prior turns into the new session file, so every question in it repeats verbatim under a second key. opencode is where it shows up most here (92 sessions on this store share their whole user side with another), but nothing about it is opencode-specific.

The signal is already in the manifest, so the fix costs no record read: the two sessions' `Asked` sets, minus the hash they matched on. Two sessions that share only the question itself are the case the surface exists for and must survive — which is why the matched hash is excluded rather than counted. Measured on the same store, a third of the rest in common drops 9 of the 9, and the verdict does not move anywhere between a fifth and a half.

After it, this store's asked-twice line says nothing, which is the honest answer here.

`internal/index/retrieval.go:4458` (`FindAskedTwice`), and `internal/stats/stats.go` reads the same thing.

vshulcz:
`FindAskedTwice` counted a resumed conversation as two askings. The harness writes the prior turns into the new session file, so every question in it repeats verbatim under a second key — and on a 2,300-session store (a year, 19 harnesses, stands and bench corpora excluded) that was all nine of the repeats at least 48 hours apart. The winner was one opencode conversation under four ids with the same eight asked hashes in each, presented as "you asked this in four sessions" across four weeks.

The test is manifest-only, so it costs no record read: the two sessions' `Asked` sets minus the hash they matched on, a third of the rest in common. Excluding the matched hash is the point — two sessions that share only the question are exactly what the line exists for, and counting it would drop them. Measured on that store the verdict is the same anywhere between a fifth and a half, and it drops nine of the nine.

After it, that store's asked-twice line says nothing. That is the honest answer for it.

Test covers both directions, and fails with the filter removed.

Fixes #3711.


vshulcz:

Issue #1622, opened 2026-08-24
the MCP tools cannot open a session by id: recall_context says nothing matched about a session deja holds


vshulcz:
`deja ctx <id>` opens a session by its id or a prefix of it. The MCP `recall_context` and `recall` tools take the same string as words only, so an agent that has an id — the id deja itself printed — is told the session does not exist.

Fixture: one claude session, `hhhh0001-1111-1111-1111-111111111111`.

```
$ deja ctx hhhh0001
# deja context: claude · tmp/huge · hhhh0001-1111-1111-1111-111111111111 · updated 2026-07-11

MCP recall_context {"query":"hhhh0001"}
→ No prior deja sessions matched "hhhh0001".

MCP recall_context {"query":"hhhh0001-1111-1111-1111-111111111111"}
→ No prior deja sessions matched "hhhh0001-…-111111111111".

MCP recall {"query":"hhhh0001"}
→ No prior deja sessions matched "hhhh0001".
```

The control is the same tool on the same store with words: `recall_context {"query":"retry backoff"}` returns 8248 bytes of context. So the tool works, the store is reachable, and the transport is fine — it is the id vocabulary that stops at the CLI.

Expected: an id that `deja ctx` resolves should resolve through the tool an agent is told to call. The hook's own lead line names `recall_context`, and every result line deja prints — in the terminal, in the recall block, in `--json` — carries an id, so "look this session up" is the obvious next move for an agent.

Cause: `cmdCtx` resolves an id through `findByPrefix` before searching (`cmd/deja/main.go`, and after #1614 again after the search fails). The MCP handlers go straight to retrieval with the string as a query; nothing on that path consults `FindByPrefix`.

Impact: the agent-facing surface is the one that cannot follow up on what deja shows it. A person typing the same string gets the session.

Found during the bug hunt, item `[ctx-budget]`, while checking what `ctx` returns on a huge session. That part is clean and worth recording: a 4.4 MB session of 4000 messages gives 8097 bytes over 36 lines, and a session whose single message is 2.3 MB gives 8098 bytes — the budget holds in both directions, against `deja show`, which prints all 3.8 MB by design.

vshulcz:
Fixed by #1623, on main since the collector merge in #1912.

Issue #1614, opened 2026-08-24
deja ctx says no session matches an id-prefix that deja show opens


vshulcz:
`ctx` only treats an argument as an id-prefix when it is at least six characters long. A shorter one falls through to the text search, finds nothing, and comes back with a sentence that is false about the store.

Fixture: three claude sessions, ids `abc12aaa-…`, `abc12bbb-…`, `ffff3333-…`.

```
$ deja ctx ffff
deja: no session matches "ffff"          exit 1

$ deja show ffff
# claude · tmp/hunt · ffff3333-3333-3333-3333-333333333333    exit 0
```

The control is the same prefix one character longer, which proves the session is reachable and that the id path works:

```
$ deja ctx ffff3333
# deja context: claude · tmp/hunt · ffff3333-3333-3333-3333-333333333333
```

An ambiguous short prefix behaves the same way — `deja ctx abc12` says nothing matches, while `show`, `resume` and `promote` all resolve it, take the most recent and say they chose:

```
$ deja show abc12
deja: 2 sessions match "abc12" — showing the most recent; use a longer prefix for another
```

Expected: the same id vocabulary everywhere. If a prefix names a session that `show` can open, `ctx` should open it too — and if it genuinely names nothing, the message is fine.

Cause: `cmdCtx` (`cmd/deja/main.go:595`) gates the prefix lookup on `len(q) >= 6`. The threshold is defensible on its own — a three-letter token is far more likely a query than an id, and ctx is the command an agent is told to call with a sentence — but nothing catches the case where the query path then finds nothing and a real session was sitting behind the prefix.

Impact: it is the one command the hook names to an agent (`recall_context`). An agent that copies a short id out of a listing is told the session does not exist.

Found during the bug hunt, item `[search-session-flag]`. The flag itself is clean: `--session abc12` returns both sessions that share the prefix, which is right for a filter, and `--session nosuch` says so plainly.

vshulcz:
Fixed by #1615, on main since the collector merge in #1912.

PR #3718, opened 2026-09-18
fix: recall_context resolves an id as an id, before the words

### LINKED ISSUE #3717: recall_context asked for the id the block prints answers with another session 80% of the time
The session-start block prints a session id beside every session and tells the agent to follow up with `recall_context`. Asked with that id, `recall_context` answers with a **different** session 80% of the time.

Measured over the 120 most recent sessions on a real store — each one asked for by its own 8-character id prefix:

```
the id the block prints: 24 of 120 bring their own session back

asked "2915986c" -> sessions=1 first="141e8ef3-c2d4-4642-a67a-a64aa6d1800a"
asked "adf51b62" -> sessions=1 first="ses_01cbcf4b7ffeM8dABUAkeJ3H70"
asked "46eb5315" -> sessions=1 first="b322ba5c-f949-4b95-adc5-1187d3a5cc27"
```

Not an empty answer — a confident wrong one. `contextByID` is tried only when the word search finds nothing (`cmd/deja/mcp.go:1765`, the fallback #1622 added), and an id is a searchable string: any transcript that mentions a session id matches it lexically, so the search answers first and the id never resolves. `deja ctx` has the opposite order — it resolves an id prefix before searching (`cmd/deja/main.go:947`) — so the CLI opens the right session and the tool an agent is told to call does not.

The other half of the same loop is the instruction itself. The block says "call recall_context with **a term** from it", and a term is the weakest selector available. Over the same sample, taking the distinctive words out of the block's own quoted line and asking with them:

| selector | session comes back at rank 1 | in the top 3 | not in the answer |
|---|---:|---:|---:|
| one distinctive word | 3% | 26% | 74% |
| three distinctive words | 39% | 76% | 24% |
| the id printed beside it | — | — | 80% wrong session (above) |

So the advertised follow-up path is the least reliable one deja has, and the id — which is exact by construction — is right there in the block.

Two fixes, both small:

1. Resolve an id-shaped query as an id **before** the search, the way `deja ctx` does. Id-shaped rather than merely wordless: hex and dashes, or a store's own prefix (`ses_`), so an ordinary word cannot be swallowed.
2. Have the block name the id it prints, rather than "a term from it".

vshulcz:
The session-start block prints a session id beside every session and tells the agent to follow up. Handed that id, both recall tools answer about a **different** session — not an empty answer, a confident wrong one.

Measured on a real store, each of the 120 most recent sessions asked for by its own 8-character id prefix:

| tool | names the session | answers about others | says nothing |
|---|---:|---:|---:|
| `recall_context`, before | 24 | 96 | 0 |
| `recall`, before | 37 | 77 | 6 |
| either, after | 120 | 0 | 0 |

An id is a searchable string: any transcript that mentions a session id matches it lexically, so the search always answered first. `recall_context` had an id route but only as a fallback for when the words found nothing (#1622); `recall` had none at all. `deja ctx` has resolved a selector before searching since #1614 — this gives the two tools an agent actually calls the same order.

Id-shaped rather than merely wordless, so an ordinary query cannot be swallowed: a digit, a dash or an underscore, which every store's id carries — a uuid, `ses_01cbcf…`, cline's `1784846110172_0h0xw`, the `arev-cline-json2-…` deja writes for a subagent run — and which a word does not. `wiring` and `timeout` stay searches.

The instruction the block gives changed with it, because the same measurement covers it: taking the distinctive words out of the block's own quoted line and asking with them brings the session back at rank 1 in 3% of cases with one word and 39% with three. The block names the id it prints now.

Test covers the shadowing case — one session whose transcript mentions another's id — for both tools, and fails with the reordering removed.

Fixes #3717.


vshulcz:

PR #3719, opened 2026-09-18
fix: the answers an agent reads, measured and corrected

### LINKED ISSUE #3637: copilot-chat: VS Code 1.137 writes chats to extension transcripts + session-store.db, so the source finds 0 files

### LINKED ISSUE #3659: a config naming a deja build under another basename is invisible to every check

### LINKED ISSUE #3703: an install and uninstall leave the reader's AGENTS.md one blank line longer

### LINKED ISSUE #3705: `deja how` answers from the whole machine: in one repo it returns another repo's wrapper command

### LINKED ISSUE #3706: `deja wip` lists words as files in flight

### LINKED ISSUE #3711: asked-twice shows the reader their own resumed conversation

### LINKED ISSUE #3713: deja files answers from the whole machine: in one repo it names another project's files
`deja files <topic>` reads the whole machine, so in a repository with 39 sessions on the topic it answers with another project's files. Asked inside this repository, on a store with 84 projects:

```
$ deja files "timeout" --limit 8
files touched while working on "timeout" — 145 sessions (of 273 matching sessions, the 250 most recent were read)
  …/migrations/000064_…sql        12
  …/migrations/000066_…sql         9
  …/mysqlrepo/helpers.go          12
  …/mysqlrepo/…_repository.go      9
  …/mysqlrepo/shared_test.go      13
  …/src/…/service.py             206
  …/routes/routes.go              16
  …/domain/…/repository.go        15
```

Eight paths, none of them in the repository the question was asked in — SQL migrations and a Python service from two unrelated projects, ranked above it by how much they were touched. The same question scoped to this repository:

```
$ deja files "timeout" --project deja-vu
files touched while working on "timeout" — 39 sessions
  …/deja-vu/cmd/deja/hook_context.go   10
  …/deja-vu/cmd/deja/hook_tool.go       2
  …/deja-vu/internal/sources/notes.go   2
  …/deja-vu/cmd/deja/hook_conventions_test.go  2
```

Same shape as #3705, which fixed it for `deja how`: an agent asking "which files are involved in X" is asking about the repository it is working in, and it cannot tell that the answer is somebody else's tree — the paths are plausible, they are just not here. A hint pointing at four files in another project is worse than no hint, because it gets acted on.

The default should be the project of the working directory, derived the way `deja wip` and the hooks derive it (worktree roots included), with `--project` still asking about another one, an `--all-projects` for the machine, and the scope named in the text and in `--json`. `deja how` widens to the machine when the project has nothing to say, for the reason stated there — an agent told nothing exists invents one — and the same holds here.

One thing in the way: `search.Options` carries a single `Project` substring, and a project has several name forms (a worktree records a different one). The scope needs to be a set, or the surface will miss its own worktrees.

`cmd/deja/files.go:94`.

### LINKED ISSUE #3716: deja wip spends its file slots on probe scripts and scratch paths
The `files in flight:` line of `deja wip` spends its five slots on throwaway paths. Asked in this repository, all five were probe scripts under a temporary directory — nothing an agent picking the work back up should open.

Measured over the 400 most recent sessions with file records on a real store, before any filter:

```
handovers naming files:                400
  every file named is a throwaway one: 238 (60%)
  names at least one repository file:  162 (41%)
  a repository file lost its slot:      29
```

The last row is the one that costs: the session did touch a repository file and the cap of five went to scratch paths instead. What counts as throwaway here is what `deja files` already refuses — the harness's own scratchpad and task files, `/.claude/`, `/.cache/`, a log, `node_modules` — plus the temporary directories, which is where a probe script or a generated fixture lives.

`digest.ResumeFrom` collects every path it sees (`internal/digest/resume.go:73`) and `deja files` filters two layers above it (`isScratch`, `inRepository` in `cmd/deja/files.go`), so the two surfaces disagree about what a file worth naming is. The digest package is a leaf and cannot stat, but it does not need to: a path-shape rule catches the whole list in 189 of the 238 cases.

With one in, over the same 400 sessions:

```
handovers naming files:                220
  names at least one repository file:  171
  every file named is a throwaway one:  49 (the disk check's half: paths no longer under a repository)
  a repository file lost its slot:      20
```

180 handovers stop naming files at all, which is correct — they touched nothing else — and the line is right more often when it speaks. Same argument as #3706, one layer further in: what the compaction handover names is what the next turn acts on.

### LINKED ISSUE #3717: recall_context asked for the id the block prints answers with another session 80% of the time

### LINKED ISSUE #3720: deja fix can answer an error with the command that produced it
`deja fix` can answer an error with the command that produced it. On a real store:

```
$ deja fix "fatal: not a git repository (or any of the parent directories): .git"
fatal: not a git repository (or any of the parent directories): .git · 2026-09-02
  ran next: $ rtk git status --short && rtk git diff --stat && rtk git diff --name-only
  after this failed: $ rtk git status --short && rtk git diff --stat && rtk git diff --name-only  → exit 128
```

The two lines are the same command. The pair is honest about what happened — a session did re-run it — but as a remedy it is a loop, and this is the surface the post-error hook speaks from, so an agent is being told to repeat what just failed.

The two strings are not equal, which is why nothing caught it: a stored command carries its exit status (`  → exit 128`), so the comparison has to be on the command alone. With that stripped, 1 of the 7 pairs served on this store that carry a failing command is this shape — 1 in 74 pairs overall. Small, and the wrong kind of wrong for a surface that speaks unasked.

`internal/index/fixpair.go:734` (`FixesFor`), which is where all seven callers come through.

vshulcz:
One batch, since each of these touches the CHANGELOG and every separate pull request costs a windows leg. Everything here was measured on a real store before it was written; the numbers are in the entries and in the issues.

**What deja tells an agent**

- **#3717 — the recall tools answered about the wrong session when handed an id.** Every session-start block prints a session id and tells the agent to follow it up, and an id is a searchable string: any transcript that mentions one matches it lexically, so the search answered first. Asked for the 120 most recent sessions by their own id prefix: `recall_context` named the right session 24 times, `recall` 37; both now 120. Id-shaped rather than merely wordless — a digit, a dash or an underscore, which every store's id carries and a word does not — so `wiring` and `timeout` stay searches. The block's instruction names the id too: a single distinctive word from its own quoted line brings the session back at rank 1 in 3% of cases, three words in 39%.
- **#3705 / #3713 — `how` and `files` answered about the whole machine.** Inside this repository `deja how "test"` returned another project's wrapper (40 runs) three times before this repository's own command, and `deja files "timeout"` named eight paths from two unrelated projects and none from here. The scope is the project of the working directory now, derived the way `deja wip` and the hooks derive it; `--project` still asks about another, `--all-projects` asks the machine, the text says which answered and `--json` carries it. The project filter takes a set, so a worktree is still the project it belongs to.
- **#3706 / #3716 — the compaction handover named words and scratch files.** Every whitespace-separated token counted as a path, so a sentence contributed one "file" per word; and of the 400 most recent sessions with file records, 238 handovers named only throwaway paths — probe scripts under a temporary directory, the harness's own scratchpad — with a repository file crowded out in 29. A token has to look like a path and not be a throwaway one; 180 handovers now say nothing about files, which is correct, and the line names a repository file in 171 against 162 before. On compacted sessions, the population the line exists for, the four lines are unchanged: asked 100%, decision 100%, command 100%, files 90%.
- **#3711 — asked-twice told the reader their own resumed conversation back.** A resume or a fork writes the prior turns under a second key, so every question repeats verbatim: all nine repeats at least 48 hours apart on a 2,300-session store were that, the winner being one conversation under four ids. The test is manifest-only — the two sessions' asked hashes minus the one they matched on — so it costs no record read, and two sessions sharing only the question still fire.
- **#3721 — `deja blame` quoted a tool's write confirmation as the reason a file looks the way it does.** Over 1,439 paths sessions actually touched: 7,417 snippets, 1,567 of them a tool's own sentence, 3,027 of 4,835 rows carrying at least one. `digest.IsAgentArtifact` — what recall has used since #2068 — recognises 1,565 of them and blame never called it. Tool output only, since a command record starts with `$ ` and is an artifact by that predicate while what a session ran is evidence.
- **#3720 — `deja fix` answered an error with the command that produced it.** `ran next: X` above `after this failed: X`, on the surface the post-error hook speaks from. The strings are never equal because a stored command carries its exit status, so the comparison is on the command alone.

**Harness and install**

- **#3637 — Copilot Chat's second store.** `workspaceStorage/<hash>/GitHub.copilot-chat/transcripts/*.jsonl`, which on a newer extension is the only store with history in it: 17 sessions → 22 and 1822 → 1832 messages here. Reported by @yizhixiaokong, whose layout and record histogram is what made this cheap to write; reproduced on a different host. `session-store.db` is deliberately not read and the registry page says why.
- **#3703 — an install and an uninstall give a reader's `AGENTS.md` back byte for byte.** 56 bytes in, 57 out before.
- **#3659** gets the invariant it asked for: install every auto target with the binary under another name, and no writer may read deja's own entries as a stranger's.

Two small things ride along: the registry pages for commandcode, gjc and kilocode were regenerated (#3701 changed the `.md` and the `.html` beside them still said `deja-search`), and `TestDoctorNamesASessionServedTwiceInASecond` stamped three events 300µs apart at `time.Now()`, so around a second boundary they landed in two — a 0.06% flake, now stamped inside one second on purpose.

Benches unchanged: recall@1 1.00 and mrr 1.000, context coverage 1.00 at the same token medians, prompt precision 1.00 with the same 5/51 cross-paired false fires.

Replaces #3704, #3709, #3710, #3712 and #3718 — same commits, one CI run.

Fixes #3637, fixes #3659, fixes #3703, fixes #3705, fixes #3706, fixes #3711, fixes #3713, fixes #3716, fixes #3717, fixes #3720, fixes #3721.


vshulcz:
End to end through the MCP server an agent actually talks to, rather than through the function: `tools/call recall_context {"query":"<8-char id>"}` against a copy of a real index.

Before: the id `2915986c` came back as session `141e8ef3-…`. After: `# deja context: claude · <project> · 2915986c-9f79-4f59-aed1-6cb929ca62b8`.

Same call, same store, same binary flags — the only difference is the order the selector and the search are tried in.

vshulcz:
One check on the cost of the throwaway filter, since it makes 180 handovers stop naming files: on the population the line exists for — sessions that were actually compacted — the four lines are unchanged. Over the compacted sessions in the same 400-session sample:

```
asked 100%  decision 100%  command 100%  files 90%  all four 90%  nothing 0%
```

The sessions that go quiet are short agent runs that touched nothing but a probe script, which is what the filter is for. `deja wip`'s own documented rates (94/94/94/76 over 96 compactions) still hold.

vshulcz:
Two more rode in while this was open:

- **#3720 — `deja fix` answered an error with the command that produced it.** `ran next: X` above `after this failed: X`, on the surface the post-error hook speaks from. The strings are never equal — a stored command carries its exit status — so the comparison is on the command alone; 1 of the 7 pairs served on a real store that carry a failing command is this shape. Dropped in `FixesFor`, where all seven callers come through.
- The `internal/query` coverage floor: the project filter taking a set added two functions to a package pinned at 100%, and the ubuntu leg caught it. Covered now, and the package is back at 100%.

Also fixes #3720.

vshulcz:
And #3721, which is the same class one surface over: `deja blame` answers "why does this file look like this" and quoted a tool's write confirmation as the reason.

Over 1,439 paths that sessions on a real store actually touched, asked through the CLI's own path: 7,417 snippets, of which **1,567** were a tool's own sentence and 3,027 of the 4,835 rows carried at least one. `digest.IsAgentArtifact` — the predicate recall has used since #2068 — recognises 1,565 of the 1,567; blame reads the same transcripts and never called it. Tool output only, since a command record starts with `$ ` and is an artifact by that predicate while what a session ran after touching a file is evidence.

The other class the measurement found, 2,524 grep-style dumps, is deliberately left: a tool-output record that quotes the file's own code around a match is sometimes the best evidence there is, and every rule for "a dump rather than a reason" also threw away the good ones. The number is on the issue.

Also fixes #3721.

PR #3722, opened 2026-09-18
feat: deja blame <path>:<line> names the session that wrote the line

### LINKED ISSUE #1181: git blame says who changed a line; nothing says why the agent did it that way
Code is increasingly written by agents. `git log` gives "fix: handle edge case"; the session where the agent hit the case, tried three things and explained why the workaround was unavoidable is on disk, unread and unaddressable from the code.

deja already holds those transcripts — including for commits made before deja was installed. Attaching reasoning at commit time requires a checkpoint the agent never made; reading it back afterwards only needs the history, which is already there.

Proposal: correlate a commit with the session that produced it, and expose it where the code is read — `deja blame <path>:<line>`, and optionally written to `refs/notes/deja` so `git log --notes` and a teammate's clone can see it too.

## What I measured first

The naive correlation — a session whose [started, updated] span contains the commit time and that touched one of the changed files — on this repo, 40 commits from the last 14 days:

```
exactly 1 candidate session   22 / 40
several candidates            16 / 40
none                           2 / 40
```

Then the control, which is the part that matters. Dependabot bumps have no human session behind them, so a correct matcher returns nothing for them:

```
bot/bump commits matched to a session   4 / 4   (should be 0)
```

All four. A long human session that also touched the workflow files overlaps every bump in its window, so the bump inherits an author it never had. Naive time ∩ file attribution is not merely imprecise, it is confidently wrong, and a wrong "why" is worse than no "why" — it invents a rationale for a line nobody reasoned about.

So the work here is attribution, not plumbing.

## The tension

Sessions are long and commits are squashed. The top session for `internal/search/search.go` spans 2026-08-09 to 2026-08-11 and covers dozens of commits; the commit timestamp is the merge time, not when the work happened. Time containment alone cannot separate them.

Signals worth weighing, roughly in order of how much they should count:

- the edit span the session actually wrote against the commit's diff hunks, not just the file name
- the session naming the issue or PR number the commit closes
- the error string or symbol in the diff appearing in the session
- proximity of the last edit to the commit, not containment in the whole span

And a matcher must be allowed to answer "nothing" — silence is the correct output for a bump, a merge, or a hand-edit deja never saw.

## Done when

- `deja blame <path>:<line>` names the session behind that line, or says it doesn't know
- the control holds: bot and dependency commits attribute to nothing, with a regression test that fails if a bump gets an author
- attribution precision is measured on this repo and reported as a number, with the naive 22/40 + 4/4 above as the baseline to beat
- commits predating the deja install attribute as well as later ones
- `refs/notes/deja` write is opt-in and never rewrites existing notes

vshulcz:
`deja blame <path>:<line>` now answers about the line. The line number was already parsed and thrown away; this uses it.

```
$ deja blame internal/digest/resume.go:75
resume.go:75 last changed in 33ec9160 · 2026-09-18 · fix: the handover names the work, not the scratch
  written in claude · 2915986c-9f7 · deja-vu
  replaced: if f = strings.TrimSpace(f); looksLikeAPath(f) {
  asked: <what that session was asked to do>
  why, in full: deja ctx 2915986c-9f7
```

**The rule, and why it is this one.** Measured over 494 human commits and the 4 dependency bumps from a month of this repository, against a copy of a real store:

| rule | one session | several | none | of 4 bumps attributed |
|---|---:|---:|---:|---:|
| time ∩ file (the naive baseline) | 378 | 7 | 109 | **2** |
| the session wrote a line the commit removed | 116 | 0 | 378 | **0** |
| the session names the number the commit closes | 355 | 0 | 139 | **2** |
| either signal | 360 | 0 | 134 | **2** |

Only the span rule holds the control the issue asks for, and it is never ambiguous — not one commit had two candidates. So it answers for about a quarter of lines and says, for the rest, which silence it is: `no indexed session wrote the lines this commit replaced`. The number-closed signal fails for a reason that only shows up in the control: the maintainer's own session names the PR number while merging it, and that session edited the same files.

**No "why" line, also measured.** The obvious next step is to quote what the session settled, and over 81 attributed lines neither candidate is about the change: the session's own conclusion overlapped it (the file name, or two content words of the commit subject) 0 times, and the first conclusion the session reached after writing the line 2 times. A session runs for hours and settles a dozen things, so a single line lifted out of it reads as the reason for a change it has nothing to do with — the failure this surface exists to avoid. What ships is what deja can stand behind: the commit, the session, the text it replaced, and `deja ctx <id>`.

Quiet where it has nothing to say: no git, no repository, an uncommitted line, a file no session mentions. `--json` is unchanged this pass — the envelope is an array and a line-level answer needs a shape of its own.

`refs/notes/deja` is not in this pass either.

Tests: the rule with a session that wrote the line, one that only read the file and one that edited it after the commit; the control, where the removed lines are a bump's; the same file reached through a worktree checkout; `path:line` and `path:line:col` parsing; and an end-to-end run against a real temporary repository, both directions. The time guard and the rule itself each fail the tests when removed.

Fixes #1181.


vshulcz:
Two follow-ups while the windows leg runs:

- The fixture paid this repository's most frequent windows-only debt: a path pasted into a JSON fixture. On Windows it carries backslashes, so the fixture was invalid JSON, the session never reached the index and the positive case read as "no session wrote this line". Encoded with `encoding/json` now.
- Blame quoted **its own line answer** back. A transcript keeps what deja printed and every line of the answer names the file, which is the trap #1330 closed for the shapes that existed then; the line answer is a new one. Recognised now, with a test that fails when the recogniser is removed.

The second one is worth stating as a rule rather than a fix: any new deja output shape that names a file becomes, on the next run, evidence about that file.

vshulcz:
Correction to the table in the description: "never ambiguous" holds for this repository, not for the rule. Run against two others on the same store — routepilot, which commits directly, and a smaller python project — the same rule gives 17 of 142 (3 of them with two candidate sessions) and 4 of 93 (none). So the tie-break this PR uses, the latest edit before the commit, is load-bearing rather than decoration.

The same three runs also kill the reason I gave for the 23%: coverage tracks how much of a repository's work went through a harness that records what an edit replaced — 2,259 spans here against 513 and 27 — and not the merge strategy. Numbers on #3723.

vshulcz:
One correction to how I described the rule, because the wording claimed something the data does not say.

An edit record holds **what an edit replaced**. So a match against a commit's deleted lines means that session *performed that very replacement* — it is the session that made the change, and therefore the author of the line being read. What I wrote earlier ("the session wrote a line the commit removed") says the opposite of the mechanism, even though the attribution it produces is the one intended. Fixed in the code comments, the test and the changelog; the output was already right (`written in <session>` for the line, `replaced:` for the text that stood there).

The changelog also no longer claims the rule is never ambiguous — that was this repository. Three of routepilot's seventeen attributed commits have two candidates, and the tie-break is the last edit before the commit.

Issue #3698, opened 2026-09-17
uninstall leaves 33 empty directories it created


vshulcz:
Install every target into an empty home, uninstall every target, and count what is left: no files (#3684, #3685, #3690 took care of those) and **33 directories**, every one of them created by deja and left empty.

```
/.dsh/plugins
/.gemini/config/plugins
/.gemini/extensions
/.gjc/agent/skills/deja-search
/.grok/hooks
/.hermes/plugins
/.kilocode/skills/deja-search
/.kiro/steering
/.omp/agent/extensions
/.openclaw/extensions
/.openclaw/hooks
/.pi/agent/extensions
/.prime/agent/extensions
/.senpi/agent/extensions
/.zcode/cli
/Library/Application Support/Code/User/prompts
…and the harness roots above them
```

Two halves to it.

`writeIfChanged` records only the file's immediate parent as "deja made this", so a writer that creates a tree — `~/.pi/agent/extensions/…` — leaves everything above the last level unrecorded, and the prune from #3239 stops at the first directory it cannot vouch for. Recording every missing ancestor is one change in one place.

Ten writers remove their directory themselves with `os.RemoveAll` and never look up. They also report `removed` whether or not anything was there:

```
$ deja uninstall openclaw-auto     # for the second time
openclaw-auto: removed ~/.openclaw/extensions/deja
               also removed ~/.openclaw/hooks/deja-recall
```

Neither existed. A line about work that did not happen is the same fault #3689 fixed in goose's report.

So: record every level a writer creates, prune upward while each is empty and recorded, and say `unchanged` when there was nothing to take. The invariant belongs with the file one — after a full install and a full uninstall, no directory deja recorded creating may still be there.

PR #3724, opened 2026-09-18
feat: the first-screen suggestion comes from what was asked

### LINKED ISSUE #3714: the first-screen suggestion optimises rarity, so it suggests a fragment of a sentence
`suggestFirstQuery` maximises the summed IDF of two adjacent words, so on a store with real prose in it the first screen suggests a fragment of a sentence rather than a subject. Measured on a 2,421-session store (`cmd/deja/suggest.go:84`):

```
bigram required in >=1 sessions:  idf=14.20  the pair appears in 2 sessions   search hits=10
bigram required in >=2 sessions:  idf=14.20  the pair appears in 2 sessions   search hits=10
bigram required in >=3 sessions:  idf=13.39  the pair appears in 3 sessions   search hits=18
bigram required in >=4 sessions:  idf=12.81  the pair appears in 4 sessions   search hits=23
candidate bigrams: 102,558 appear in one session, 10,301 in two or more
```

The rule today requires each *word* to appear in two sessions and says nothing about the pair, so the pick was a two-word fragment lifted out of one sentence — grammatically a verb and its object, naming nothing anyone would search for. It was the subject of neither of the two sessions that contain it; it is simply the rarest surviving pair.

Raising the bar to the pair itself recurring improves it monotonically on this store — 10 hits, then 18, then 23 — and the phrase changes each time, from a sentence fragment to an idiom to an actual noun phrase at four sessions. So the bar is the lever, and its current value (none) is the worst one available.

What no bar can do is tell a subject from a phrase that happens to recur. This line is the first thing a new store prints (the greeting uses the same function) and the invitation to use the tool at all, so what it should be picked from — a session title, the subject of the friction line, a recurring noun phrase — is a product question rather than a threshold.

Numbers only above: the phrases themselves come from this machine's own history.

vshulcz:
The other half of #3714: what the line should be picked *from*. The threshold half shipped in #3715; this answers the source, with the measurement on the issue.

Four candidate sources, scored on a 2,000-session store by what a reader running the suggestion gets back, and by whether the phrase is one somebody wrote as a task (proxy: it appears in a hit's own title):

| source | search hits | of the top five, in their own title |
|---|---:|---:|
| a bigram from any message (what shipped) | 18 | 0 |
| the subject of the friction line | 66 | 0 |
| the newest session's title | 50 | 1 |
| a phrase recurring across session **titles** | 50 | 2 |

Titles win for a structural reason rather than a statistical one: a title is what somebody typed when they asked for the work, while a mid-message bigram is a fragment of whatever was being said at the time. On this store the old rule picked an idiom; the new one picks `"structured data"`, which returns 49 sessions.

The friction subject scores well on hits and is deliberately not used: the brief prints that wall one line above the suggestion, so it would say the same thing twice.

**Borrowed titles are skipped**, and finding out why took the measurement: the ingest borrows a title from whatever the harness or a tool wrote when a session opens with plumbing, so a `<teammate-message>` envelope becomes `harness output: …` and no longer looks like an envelope. On an agent-heavy store that family is most of the titles — without the skip the first screen suggests a fleet's paperwork. `index.TitleIsBorrowed` names that rule where the prefixes are defined rather than duplicating the strings.

The message scan stays as the fallback, so a store whose sessions carry no title of their own still gets a suggestion — with a test for that direction too.

One thing the tests pin that is worth stating: a phrase in *every* title has no rarity left and the floor rejects it, which is the right answer for a store that has only ever been asked one thing.

Fixes #3714.


vshulcz:
#3725 rides along, because it is one file and the windows leg is the scarce part: a full uninstall left `~/.config/deja/bin` standing empty. It was the only directory left on a stand where every target had been installed and then removed, and #3698's prune could not see it — that walk removes what the wiring record names, and the launcher's directory was never recorded.

Also fixes #3725.

Four other sweeps on the same stand came back clean, worth recording so nobody re-runs them:

- **every `--json` surface against its own documented example** (13 sections in `docs/json-output.md`, 12 runnable): the six differences are all fields the doc states are omitted when empty — `capped`, `sidecar_size`, `embed`/`ingest_health`, `project`, and the whole of `wip` when the transcript carries nothing.
- **MCP argument edges**, 24 cases: missing mode, unknown mode, wrong types, `limit: -5`, `limit: 1e9`, `offset: 1e9`, a 1 MB query, a NUL byte, an unknown tool. Zero panics, zero hangs, typed errors, and the odd numbers behave as the default with the answer bounded by the budget (3.8 KB).
- **partial uninstall**: 18 auto targets installed, then removed one at a time — no case where removing one unwired another that shares its file.
- **two builds at two paths** (the homebrew-plus-go-install machine): a session start from the second build does not rewrite the launcher the first one wrote, so the wiring does not flap.

vshulcz:
And #3726, found by running the edges against the feature that landed in #3722 the same day: `deja blame <path>:<line>` printed nothing about the line whenever it could not answer, so the reader could not tell "this line has no history" from "your `:120` was ignored".

```
blame_line.go:999999 — the file has 335 lines
deja:12              — that is a directory, not a file
hosts:1              — this file is not in a repository
nosuchfile.go:4      — there is no such file here
new.go:1             — this file is not tracked, so no commit wrote the line
```

Also fixes #3726.

Two more sweeps came back clean on the way:

- **every hook against 17 payloads** — empty stdin, `{}`, `[]`, truncated JSON, prose, no `cwd`, a `cwd` that does not exist, a 1 MB prompt, control characters, a NUL byte, invalid UTF-8, deja's own block echoed back, tool payloads with and without input, a failing tool result, a plan: no panic, no hang, no broken envelope, no nonzero exit. The sidecars beside the index have to be cleared between cases or the project cooldown silences everything after the first answer and every later "quiet" reads as behaviour.
- **deja's own block pasted into the next prompt does not steer the recall**: the same question with an unrelated block in front (kafka, singbox, routepilot) still answered about the connection pool.
- **25 processes at once** — four of each hook, four searches, four `how`, four `wip` and a rebuild against one index: no panic, no corruption, the same search total before and after, and the hooks did not wait for the rebuild while the CLI did, which is the split #1306 asks for.

Issue #1829, opened 2026-08-25
only search suggests a flag when you mistype one; the other 22 sites just refuse


vshulcz:
`deja search` is the only command that helps with a mistyped flag. Every other one just says no:

```
$ deja search --nope x
deja: unknown flag "--nope" — did you mean --role?

$ deja show --nope
deja: show: unknown flag "--nope"

$ deja index --quiet
deja: index: unknown flag "--quiet"

$ deja stats --nope
deja: stats: unknown flag "--nope"
```

The suggestion in the first one exists because folding a near-miss flag into the query silently turned a working search into "you have no such memory" (#755) — the comment above `nearestSearchFlag` in `cmd/deja/main.go` says so. The same reasoning applies to a person who typed `--limt` at `deja show`; they just get less help.

There are **23** `unknown flag` sites across `cmd/deja/`, and one of them knows about `nearestSearchFlag`.

Two inconsistencies, both small:

- **Suggestions.** Only search offers one.
- **The prefix.** Search says `deja: unknown flag …`; everything else says `deja: <command>: unknown flag …`. The command name is the more useful form — it tells you which parser refused — so search is the odd one out, not the others.

## What a fix looks like

Lift the near-match into a small helper both can use — something like `unknownFlag(command string, arg string, known []string) error` — that returns `<command>: unknown flag "x" — did you mean --y?` when a candidate is within one or two edits, and the plain refusal otherwise. Then have the flag parsers pass their own known-flag list.

Doing all 23 in one pull request is a lot of surface for a first change; a few commands, with the helper and its tests, is a better shape. Say which ones you converted, and the rest can follow.

`nearestSearchFlag` already exists and already picks the candidate; the useful part of this change is the shared error, not new matching logic.

A test belongs with it: an unknown flag on a converted command returns an error naming that command, and a near-miss carries the suggestion. Both fail on `main` for any command other than search.

PR #3727, opened 2026-09-18
feat(index): --quiet for the runs nobody is watching (#1827)

### LINKED ISSUE #1827: deja index has no --quiet, so a scheduled run either prints a line or hides its errors
`deja index` always prints a line, and there is no way to ask it not to:

```
$ deja index
deja: incremental index changed_files=5 removed_files=1 sessions=5

$ deja index --quiet
deja: index: unknown flag "--quiet"
```

The workaround is `deja index >/dev/null`, which also swallows the errors you would want to see.

It matters wherever the command is not run by a person watching. Nothing in our docs tells people to schedule `deja index` — getting-started says the opposite, that no explicit index step is needed — but people do run it from shell profiles and hooks anyway, and a line printed over the prompt on every new shell is the kind of thing that gets a tool uninstalled rather than reported.

## What a fix looks like

A `--quiet` flag on `index` that suppresses the progress line and leaves errors on stderr. The refusal above comes from the same place the flag would be parsed, so both live in one function:

- `cmd/deja/main.go` — where `index` reads its arguments and rejects unknown ones
- the line itself is printed further down the ingest path; the flag has to reach it

`deja index --rebuild` prints its own progress too, and a reader would expect `--quiet` to cover both. Say in the PR which one you made it do.

## What "quiet" must not mean

Silent about failure. A run that cannot read a store, or finds one missing, still has to say so on stderr — quiet is about the success line, not about hiding the reason a scheduled run stopped working.

A test belongs with it: drive the command with and without the flag against a `t.TempDir()` index and assert on what reaches stdout, the way the existing command tests in `cmd/deja` do.

vaibhav8a:
Closes #1827.

## What this changes

`deja index` takes `--quiet`.

```
$ deja index --quiet
$ echo $?
0
```

Before, the command always printed a line and there was no way to ask it not to. Nothing in the docs tells people to schedule it — getting-started says the opposite — but they run it from shell profiles and session hooks anyway, and `deja: index is up to date (…)` over the prompt on every new shell is the kind of thing that gets a tool uninstalled rather than reported. The workaround, `deja index >/dev/null`, also swallows the errors you would want to see.

## Which of the two it covers

Both, as the issue asks: the incremental run and `--rebuild`. They print through the same sink, and a reader would not expect one of them to stay noisy. `--quiet` also skips the live progress display and the first-index greeting, which are the same report in a different shape.

## What quiet does not mean

Silent about failure. The success reporting goes to a sink the flag can discard; everything that explains why a scheduled run stopped working stays on stderr:

| still printed under `--quiet` | why |
|---|---|
| any build failure | `ensureError` explains what to change — losing it is the whole hazard |
| a store whose paths could not be read | sessions missing from recall |
| the exclude list changed since the build | a pattern that looks applied and is not |
| an index that came out empty for a reason | no history found, or a store behind a permission wall |
| notes regrouped from another time zone | the command doing more than it was asked |

Suppressed: the up-to-date line, the outcome line, the session/message counts, the "held no message" and merged-count explanations of those counts, the greeting, the progress bar.

## Tests

`cmd/deja/index_quiet_test.go`, four cases against a `t.TempDir()` index and the synthetic claude fixture, in the shape the existing command tests use:

- `--quiet` prints nothing on the ordinary already-fresh path
- without it, the same run still reports
- `--quiet --rebuild` is quiet too
- `--quiet` into an unwritable index directory still fails **and** still says why
- unknown flags are still refused (`--quiett` does not slip through the new `switch`)

`go test ./cmd/deja/ -race -count=1` green, `go vet ./...` clean, `gofmt -s` clean. Package coverage 89.6%.

## Not in scope

The refusal `deja index --quiet` used to give is one of the 23 sites in #1829; this change does not touch the other 22.


vshulcz:
Read the whole diff and the reasoning holds — merging.

What I checked beyond the tests: the flag reaches the help line, the failure paths stay on stderr (`ensureError`, the collision and exclude-list warnings, `emptyIndexReason`), `countingWriter` still counts when its sink is `io.Discard` so the outcome line is still skipped when the build narrated, and `prepareFirstIndexGreeting`/`maybeFirstIndexGreeting` are both guarded rather than one of them. The greeting suppression is process-local, so a later interactive run is unaffected.

Two things I liked: quiet being defined as *the success reporting and nothing else*, with the table of what stays — that is the distinction a scheduled job actually needs — and the test that pins the unwritable-directory case, which is the one a "just add a flag" change usually drops.

Two leftovers are mine rather than yours, and I am landing them separately: `deja completion` lists `--rebuild` for this command in both the bash and zsh halves and needs `--quiet` beside it, and `docs/guide/commands.html` describes the command with `--rebuild` alone.

vshulcz:
Thank you — and specifically for the table of what `--quiet` does not silence. That is the part this feature lives or dies on: a flag that quiets a scheduled run has to keep saying why the run stopped working, and you wrote the rule down before writing the code. The unwritable-index test is the other half of it.

Issue #755, opened 2026-08-01
a typo in a search flag silently becomes part of the query


vshulcz:
```
$ deja pool --limti 5
deja: no matches in 3 indexed sessions — try fewer words or --re (query "pool --limti 5")
$ deja pool --limit 5
[claude] proj/p · Jul 20 · caveat — 1 matches
```

The misspelled flag is folded into the query, which then matches nothing — so a typo turns a working search into "you have no such memory". This is the command people run most.

It cannot simply reject anything starting with a dash: a query may legitimately contain one (`deja "--retry budget"` is a real search, and #753 documents `deja search` for exactly that). But a token one edit away from a real flag is a typo, not a query — the same rule #674 used for mistyped subcommands.

Fourth instance of this class after #721 (ctx), #745 (sync export) and #747 (sources).

PR #3728, opened 2026-09-18
feat(cli): one refusal for a mistyped flag, from eleven more commands (#1829)

### LINKED ISSUE #1829: only search suggests a flag when you mistype one; the other 22 sites just refuse

vaibhav8a:
Closes #1829.

## What this changes

The near-match lives in one helper now, and eleven more commands answer through it.

```
$ deja show --limt 5
deja: show: unknown flag "--limt" — did you mean --limit?

$ deja stats --jsson
deja: stats: unknown flag "--jsson" — did you mean --json?

$ deja doctor --nope
deja: doctor: unknown flag "--nope"
```

Both inconsistencies the issue names:

- **Suggestions.** `index`, `show`, `forget`, `stats`, `doctor`, `view`, `log`, `remember`, `restore`, `promote` and `handoff` now offer one.
- **The prefix.** Search gains its own name. The issue's reading is the right one — the command name tells you which parser refused — so search was the odd one out, not the other 22.

`unknownFlag(command, arg, known)`, as sketched in the issue. `nearestSearchFlag` becomes a one-line search-shaped name for `nearestKnownFlag`, because search is the one command where that answer also decides whether a dashed token is a query term.

## The matching rule is unchanged

Deliberately: it is the rule #755 arrived at, and widening it here would be a second change hiding inside this one. A token still has to look like a long flag and be within a prefix or two edits of a real one, so a bare word, a single-dash argument, and a dashed query term all behave exactly as before — `deja search "--retry budget"` is still a search, pinned by a test.

## The part that needs a guard

A suggestion is only as good as the list it is drawn from, and the lists sit away from the parsers that read the flags. A flag added to a parser and not to its list produces a *wrong* suggestion — "did you mean --limit?" for a flag the command now takes — which is worse than no suggestion at all, and nothing else would catch it.

So `unknown_flag_lists_test.go` reads the parsers with `go/ast`: for every function that calls `unknownFlag`, every flag-shaped literal it mentions must appear in the list it passes. Verified to fail on drift by dropping `--limit` from `showFlags`:

```
parseShow takes --limit but showFlags does not list it — a near miss would be told to use something else
```

It also fails if the call sites are renamed away or drop below eleven, so it cannot quietly end up asserting over nothing.

## What I left alone

The other eleven sites refuse with guidance of their own — `sync export` names the three flags it takes, `blame` and `how` explain that a dashed argument goes after `--`, `install` lists its targets — and that is worth more than uniformity. Threading suggestions into those without losing the prose is a separate change; happy to do it if you want the set closed.

## Checks

`go test ./cmd/deja/ -race -count=1` green, `go vet ./...` clean, `gofmt -s` clean.


vshulcz:
Read the whole diff. The shape is right and the guard is the part I would have asked for: a suggestion drawn from a list that sits away from the parser is a wrong answer waiting to happen, and `unknown_flag_lists_test.go` holds the two together — including the count, so it cannot end up asserting over nothing. Leaving the 11 sites that carry their own guidance alone is the right call too; `blame` explaining `--` is more useful than a near miss.

You beat me to the deprecation: I had a `ParseFile` rewrite in hand when your rebase landed with one.

Three things were mine and are pushed to this branch:

- **the rebase onto main**, where #3727 had just landed `deja index --quiet` in the same parser this branch converts — `cmdIndex` conflicted;
- **`--quiet` in `indexFlags`**, which your own guard demands the moment those two meet: without it the list would offer `--rebuild` to somebody who typed `--quiett`;
- **the two places #3727 left behind**: `deja completion` lists `--rebuild` for the command in both halves, and `docs/guide/commands.html` described it with `--rebuild` alone.

Suite green locally on the merged branch. Merging once CI agrees.

vaibhav8a:
The `windows-latest` failure here is not from this PR — it is `TestIndexQuietStillReportsFailure`, which came in with #3727 and is red on `main`. A mode bit denies nothing on Windows, so the build succeeded there and the test read that as `--quiet` swallowing the error. Fix in #3734; this branch goes green once that lands and it is rebased. Nothing in this change touches that test.

vshulcz:
Thanks — one refusal shared across twelve commands, and you took the near-miss guard with it rather than leaving the flag lists to drift apart again. #1829 closed.

PR #3730, opened 2026-09-18
fix: mask a password stated in prose

### LINKED ISSUE #3729: a password stated in prose is stored in the clear
A password stated in prose is stored in the clear. Every rule in `internal/redact` wants a delimiter — a colon, an equals sign, a flag — and a person telling an agent a password writes none:

```
$ deja show <id>
…
also the admin password is hunter2-not-a-real-password, do not put it in code
```

Found by seeding four obviously fake credentials into a transcript, indexing, and asking every surface for them. Three of the four were masked on the way in — `postgres://appuser:[redacted:url-credentials]@…`, `[redacted:aws-access-key]`, `OPENAI_API_KEY=[redacted:credential]`, and the key named mid-sentence as `[redacted:openai-key]` — and the fourth went through because the sentence had no separator in it. `deja show` and `deja ctx` then read it back; every other surface stayed clean because none of them quotes that record.

The shape is not exotic. "the admin password is X", "пароль это X", "the passphrase was X" is how somebody unblocks an agent, and it is exactly the line they forget inside a minute.

What keeps a rule like this away from prose is already established here: the digit or the symbol in the value, which is what `worthRedactingIntl` rests on. "the password is wrong", "the password is correct", "the password is the same as staging" carry neither, and a value class that stops at the first space cannot swallow a sentence.

Redaction runs at ingest, so a store built before the fix keeps the text until it re-reads its sources — the index format has to move with it, the way 41 and 47 did for the same reason.

vshulcz:
Found by seeding four obviously fake credentials into a transcript, indexing, and asking every surface for them. Three were masked on the way in — the connection url as `[redacted:url-credentials]`, the AWS key, the `.env` assignment as `[redacted:credential]`, and the key named mid-sentence as `[redacted:openai-key]`. The fourth went through:

```
$ deja show <id>
…
also the admin password is hunter2-not-a-real-password, do not put it in code
```

Every rule in `internal/redact` wants a delimiter — a colon, an equals sign, a flag — and a person telling an agent a password writes none. `deja show` and `deja ctx` then read it back; the other surfaces stayed clean only because none of them quotes that record.

What keeps a rule like this away from prose is already established here, in `worthRedactingIntl`: the value has to look chosen — a digit or a symbol — and the value class stops at the first space, so a sentence cannot be swallowed. `the password is wrong`, `the password is correct`, `the password is the same as staging`, `пароль это неправильный` and `the passphrase is stored in 1Password` are all untouched, with a test for each.

Index format 49: redaction runs at ingest, so a store built before this keeps the text until it re-reads its sources — the same reason 41 and 47 moved.

Fixes #3729.


vshulcz:
False-positive count on a real store, since a rule that reads prose can only be judged that way: 312,405 records, 2,504 of them mention a password word, the prose shape matched 121 times and the gate kept 9 of those across 4 sessions. The other 112 are the sentences you would not want touched — the value was a word.

Also in this PR: the windows leg was red on the test #3727 left on main, not on this change. `os.Mkdir(ro, 0o500)` does not deny writes on Windows or under root, so `index --quiet` correctly reported success and the assertion failed. Skipped there the way the other permission tests in this package are (#3731).

Issue #2224, opened 2026-08-27
the status line reads the usage log twice on every prompt


vshulcz:
The status line renders on every prompt and reads the usage log twice: `TodayDemand` always, then `TodayRaw` when there were recalls today, or `Week` when there were none. Each is a full read and parse.

    2000 events (  202 KB): TodayDemand  2.820ms, Week 2.502ms, TodayRaw 1.934ms
   12000 events (  1.2 MB): TodayDemand  8.176ms, Week 7.655ms, TodayRaw 7.129ms
   40000 events (  4.1 MB): TodayDemand 24.852ms, Week 24.144ms, TodayRaw 23.814ms

So 16 ms per render on a 1.2 MB log, 49 ms on a 4 MB one, for numbers one pass could produce.

`TodayDemand`'s own doc makes the argument this misses:

    Injections come back from the same pass rather than from a second call: the
    statusline renders on every prompt, and two passes over the log can also
    straddle a write and report two numbers that were never true together.

That is exactly what the second call does — today's bytes and the week's counters can come from either side of an append, and the line prints them together.

(`Totals` and `Impact` behind `deja stats` cost the same 7 ms and 24 ms, but once, when someone asks. Different question.)

vshulcz:
Fixed by #2225 (perf(statusline): read the usage log once per render).

That PR merged into `hunt/2026-08-23` rather than into the default branch, so nothing here acted on its `Fixes` line. The branch went into main as #2351, so the fix is shipped and this can close.

PR #3732, opened 2026-09-18
perf(usage): the brief reads the log once, not once per figure (#1576)

### LINKED ISSUE #1576: the statusline walks the usage log three times to print one line
`internal/usage` has no reader that returns more than one figure at a time, so every surface showing two numbers walks the whole log more than once. Counted by matching the helpers that call `read(Path(...))` against their call sites:

| surface | full passes | helpers |
|---|---|---|
| `cmd/deja/statusline.go` | 3 | `TodayDemand`, `Week`, `TodayRaw` |
| `cmd/deja/brief.go` | 4 | `TodayDemand`, `Week`, `DejaVuWeek`, `TodayRaw` |
| `cmd/deja/hook_context.go` | 2 | `TodayWithInjections`, `TodayRaw` |

The statusline renders on every prompt and the hook context runs before every session, so those two are the hot ones. On this machine the log is 324 KB across 3,162 lines — rotation keeps 14 days or 1 MB, whichever comes first — and each pass opens the file, scans it line by line and JSON-decodes every line, including the ones outside the window that pass is going to keep.

There is a second reason beyond speed: separate passes are separate snapshots. An event recorded between two of them lands in the second and not the first, so a line can print a "today" figure and an "injected" figure that were never true at the same moment. #1569 hit exactly that between two of these helpers and folded them into one pass; the rest were left as they were.

A reasonable shape: one walk that fills what the caller asked for — a `Scan(indexDir)` returning the counts, or a small accumulator — with the current helpers kept as thin wrappers so nothing outside the package has to change at once. The statusline then builds its line from one pass, the brief from one.

Please measure it rather than assume: a benchmark over a log of a few thousand lines, before and after, in the same PR. If the difference turns out to be noise, that is a useful answer too and the snapshot argument still stands on its own.

Not a correctness bug — that part was #1569 — so this is a cleanup with a number attached.

vaibhav8a:
Closes #1576.

## What was left

Two of the three surfaces in the issue's table are already fixed — `statusline.go` and `hook_context.go` both call `usage.StatusCounters`, folded in #2224. `brief.go` still called a reader per figure, and `hook_week_note.go` (not in the table, but written from a session-start hook) called two.

So this is the remainder: the brief goes from **four passes to one**, the week note from **two to one**.

## Measured, as asked

A benchmark of both shapes ships in the same package, over a log of 3,200 lines — the size the issue measured on a real machine, and near what rotation keeps for a busy fortnight:

```
BenchmarkBriefFiguresSeparatePasses-8   200   12159167 ns/op   14900060 B/op   76880 allocs/op
BenchmarkBriefFiguresOnePass-8          200    3076806 ns/op    3724998 B/op   19220 allocs/op
```

**12.2 ms → 3.1 ms** and **76,880 → 19,220 allocations** per brief render. Stable across `-count=3` (12.06/12.26/12.39 against 3.10/3.12/3.10). Not noise, and it is the ratio you would predict: four walks become one.

## The snapshot argument, which stands on its own

Separate passes are separate snapshots. An event recorded between two of them lands in the second and not the first, so the brief could print a "today" figure and a "this week" figure that were never true at the same moment — exactly the class of thing #1569 fixed between two helpers. One pass makes that unrepresentable rather than unlikely.

## Shape

`StatusCounters` grows the two counters it did not carry — `DejaVuWeek`, and `WeekInjections`/`WeekInjectedBytes` — and each is kept **on its own terms** rather than on the nearest existing one:

- the déjà vu count sits *above* the empty rule and opens **strictly after** the cut, because that is what `DejaVuWeek` has always done (`After(cut)`, not `!Before(cut)`);
- an arrival is counted whether or not it carried bytes, which is why the fixture's three déjà vu moments are five arrivals with 1,200 bytes between them — `Week()` counted it that way and the field keeps counting it that way.

`TodayDemand`, `Week`, `DejaVuWeek` and `TodayRaw` become wrappers around the single pass, which is the shape the issue suggests: nothing outside the package changes, and **the existing tests for each of those four still assert the numbers they always asserted** — that is the regression evidence for the fold, more than any new test is.

Two new tests beside them: one pins the new counters against a fixture by value (so they are right, not merely self-consistent), and one checks each wrapper returns *its own* fields — a wrapper wired to the wrong field would compile and still look consistent with itself.

## Checks

`go test ./internal/usage/ ./cmd/deja/ -race -count=1` green, `go vet ./...` clean, `staticcheck` clean on both packages, `gofmt -s` clean.


vshulcz:
Checked the part a benchmark cannot: that one pass answers what four readers answered. Reimplemented the four old bodies verbatim beside the new ones and compared them over 40 random logs of 60 events each, with stamps landing exactly on the week cut, one nanosecond past it, on midnight, one nanosecond before it, and in the future — no disagreement. The two boundaries that differ by design, `After(cut)` for déjà vu and the empty rule sitting above the raw sum, survive that.

Then mutated `TodayDemand` to return `n.Injections` instead of `n.Injected` — count where bytes belong — and three tests failed, including your `TestTheWrappersReturnTheirOwnFigures`, so the wrappers are pinned to their own fields rather than to each other.

The ratio reproduces on this machine: 8.2 ms against 2.0 ms, four to one.

vshulcz:
Thanks — the snapshot argument is the half that makes this worth merging on its own, and you found it in the issue's shadow rather than in the issue.

PR #3733, opened 2026-09-18
feat(doctor): report a stale Grok plugin, as it already does for Kimi (#1828)

### LINKED ISSUE #1828: doctor reports a stale Kimi plugin but says nothing about a stale Grok one
`deja doctor` tells a Kimi user when their installed plugin is behind the one this deja ships:

```go
// cmd/deja/kimi_plugin.go
const kimiPluginVersion = "0.1.0"
...
return "v" + got + " installed, v" + kimiPluginVersion + " ships with this deja — reinstall it in Kimi to update"
```

Grok Build has the same problem and no such report. Its plugin is installed from a marketplace entry that pins a commit SHA, so an installed copy stays at whatever the pin was until someone moves it — and nothing on the machine says the copy is old. The bundle went to 0.2.0 in #1721 while every installed copy is still 0.1.0, and no user can tell.

## What a fix looks like

Mirror what Kimi already does, in the same shape:

- a `grokPluginVersion` constant beside the existing grok code in `cmd/deja/`, holding the version in `extensions/grok/.grok-plugin/plugin.json`
- read the installed copy's version from `~/.grok/installed-plugins/*/`—the directory name is generated, so this means globbing rather than a fixed path, which is the one part that differs from Kimi
- a line in `deja doctor` when the two disagree, worded like Kimi's
- a test that the constant and the manifest agree — `TestKimiManifestsAgree` in `cmd/deja/kimi_plugin_test.go` is the model, and the reason it exists is that a constant nobody checks drifts silently

## Two things to check while you are in there

The plugin can also be installed from a local path (`grok plugin install ./…`), where "behind" means something different — the source is a working copy, not a pinned release. Deciding what doctor should say in that case is part of the issue; saying nothing is a defensible answer if you state why.

And `deja doctor` must not fail when Grok is not installed at all. The Kimi path returns an empty string when the plugin directory is not there, and this should do the same rather than reporting a missing plugin as an error.

vaibhav8a:
Closes #1828.

## What this changes

```
  grok         plugin      ~/.grok/hooks/deja.json  (v0.1.0 installed, v0.2.0 ships with this deja — reinstall it in Grok to update)
```

Before, that row said `missing`, which is wrong twice over: the plugin carries recall on that machine, and the copy carrying it is two releases behind with nothing to say so. `deja doctor` has told Kimi users since #1721; this is the same report for Grok.

Both lines, not one: the plugin declares the MCP server *and* the per-prompt hook, and stands down where the installer already wrote each, so `doctorMCP` reported `not wired` for the same reason `doctorAutoRecall` reported `missing`.

## The two places the harness forces a difference from Kimi

**The directory name is generated**, as the issue notes, so there is no path to join onto. The copy is found by the `name` in its own manifest under `~/.grok/installed-plugins/*/.grok-plugin/plugin.json` — which also means a directory holding *another* plugin is somebody else's rather than a misread, pinned by a test.

**Behind, never merely different.** This is the local-path question the issue asks about, and I did not go with "say nothing". `grok plugin install ./…` installs a working copy whose version may legitimately be *ahead* of the bundle, and deja cannot tell a local copy from a marketplace one by the files it can see — the manifest is identical either way. So the two versions are **compared** rather than diffed, reusing `compareUpdateVersions`:

| installed | doctor says |
|---|---|
| older than the bundle | `v0.1.0 installed, v0.2.0 ships with this deja — reinstall it in Grok to update` |
| equal | `v0.2.0` |
| newer (a working copy) | `v9.9.9` — the number, and no claim about it |
| unparseable | the same: reported, nothing claimed |

Telling someone their newer plugin is stale is a wrong answer; printing its version is a quiet one. Kimi's `got != kimiPluginVersion` would have got this wrong, which is why it is not copied verbatim.

## Not an error when Grok is not there

Most machines have no Grok at all and doctor runs on them, so: no `~/.grok`, an `installed-plugins` holding only other plugins, and a `plugin.json` that does not parse are all silent — three tests.

## Tests

`cmd/deja/grok_plugin_test.go`, seven cases: the constant agrees with `extensions/grok/.grok-plugin/plugin.json` (the `TestKimiManifestsAgree` reason — a constant nobody checks drifts silently), the manifest carries the name the lookup needs, the note names both versions when behind, the three quiet cases above, and the two doctor rows themselves.

The doctor tests were verified to bite: with `grokPluginInstalled` stubbed to `false` they fail with `grok row = "  grok         missing     ~/.grok/hooks/deja.json"`, which is the old behaviour.

## Checks

`go test ./cmd/deja/ -race -count=1` green, `go vet ./...` clean, `staticcheck` clean, `gofmt -s` clean.


vshulcz:
The layout claim holds on a live Grok: `grok plugin install <dir> --trust` into an isolated `GROK_HOME` puts the copy at `installed-plugins/plugin-<hash>/.grok-plugin/plugin.json` with `"name": "deja"`, and the registry beside it keys the same directory. Built your branch, set the installed manifest to 0.1.0, ran `deja doctor`:

```
grok  plugin  …/.grok/hooks/deja.json  (v0.1.0 installed, v0.2.0 ships with this deja — reinstall it in Grok to update)
```

Your "behind, never merely different" rule gets support from something I did not expect: a local install is a symlink — `grok plugin update` answers `local symlink, already live` and the installed manifest tracks the source — so the copy that can legitimately be ahead is exactly the one whose number moves on its own.

One note, not blocking: Grok has `grok plugin update [<name>]`, and its own plugin guide says session start scans marketplaces and updates plugins unless `plugin_auto_update = false` is pinned. So "reinstall it in Grok to update" is weaker than what the harness offers. The Kimi line reads the same way, which is why I am not asking you to change it here — I will take both lines in a follow-up and name the command.

Merging once the windows leg is back; I updated the branch onto main, since the test that was failing there was #3727's and is fixed in #3734.

vshulcz:
Thanks — finding the copy by the name in its own manifest instead of a path is the part that makes this survive Grok's generated directory names, and the unreadable-manifest case being silent rather than an error is what keeps doctor usable on the machines that have no Grok at all.

Issue #3731, opened 2026-09-18
index --quiet failure test assumes a mode bit denies writes


vshulcz:
`cmd/deja/index_quiet_test.go:92` fails on the windows leg:

```
index --quiet into an unwritable directory reported success
```

The test builds an index into a directory made with `os.Mkdir(ro, 0o500)` and expects the run to fail. Windows ignores the mode bit on a directory, so the write goes through and the command correctly reports success — the unwritable premise the assertion rests on does not exist there. Same under a root uid.

The other permission tests in this package already skip for that reason; this one, added in #3727, does not.

vshulcz:
Fixed by #3734 — a regular file where the parent directory should be, so the case keeps running on Windows instead of skipping.

PR #3734, opened 2026-09-18
fix(tests): make the index --quiet failure case fail on every platform

vaibhav8a:
`windows-latest` is red on `main`, and it is mine — from #3727.

```
--- FAIL: TestIndexQuietStillReportsFailure (0.69s)
    index_quiet_test.go:92: index --quiet into an unwritable directory reported success
```

The test makes the build fail by pointing the index at a directory created `0o500`. A mode bit denies nothing on Windows, so the build *succeeded* there, and the test read that as `--quiet` having swallowed the error — a false failure reporting the opposite of what happened. The same hole is open to anyone running the suite as root.

## The fix

A regular file where a parent directory should be. Creating `…/not-a-directory/index.db` fails on every platform, and for a reason that is not about permissions at all, so the assertion the test exists for — a quiet run still says why it could not build — keeps running everywhere instead of being skipped on two.

I wrote it with a `runtime.GOOS == "windows" || os.Geteuid() == 0` skip first, matching the guards elsewhere in the package, then replaced it: skipping gives up the check on the platforms where it just broke, and this case does not need a platform-specific mechanism.

Green here with `go test ./cmd/deja/ -run 'TestIndexQuiet|TestIndexWithout|TestIndexStillRefuses'`, `go vet`, `staticcheck` and `gofmt -s`. Sorry for the red trunk.


vshulcz:
Better than the skip I had queued for this — a directory under a regular file fails everywhere, so the case keeps running on the leg that found it. This closes #3731, which I filed an hour ago for the same failure.

I'll drop my skip from #3730 and take this instead.

vshulcz:
Thanks — and for going after the premise rather than the assertion. A skip would have left the one platform that can hit this uncovered, which is where it came from.

PR #3742, opened 2026-09-18
docs: catch up on 0.20.2 and pin the counts that drifted

### LINKED ISSUE #3741: the spelled-out harness count is stale in three documents and unchecked in a fourth
The harness count is spelled out in words in eight places and three of them are wrong, in two different ways.

`#3649` added six stores and bumped the word by editing the first half of it, leaving the old number's tail behind:

```
llms-install.md:29:            the thirty-three-six harnesses deja installs into
docs/guide/harnesses.html:      the thirty-three-six coding agents deja indexes today   (×6, lede and three meta tags)
docs/ARCHITECTURE.md:8:         the thirty-three-five the loader registers
```

`TestReadmeSpellsTheHarnessCountTheRegistryHas` reads those files and did not see it: it removes the wanted word from the text before looking for stale ones, and removing `thirty-three` from `thirty-three-six` leaves `-six`, which is not a word the test knows. Three releases shipped it.

`docs/llms.txt` is wrong on its own terms and is checked by nothing — the word test cannot take it, because the file legitimately says "and twenty-seven more" beside six named agents, and the digit test only reads `.html` and `.md`:

```
docs/llms.txt:14:  the conversation history file each of the thirty-two agents writes
docs/llms.txt:70:  where each of the thirty-one agents stores its sessions
```

Two counts apart, in one file, both stale against the 33 in `docs/registry/registry.json`.

The GitHub repository description said "and 20 more coding agents" — 5 named, so 25 — which is what anyone searching sees before they reach any of these files.

vshulcz:
Two things the release leaves behind, plus the documentation for what it shipped.

The catch-up: `pinmanifests -version 0.20.2` over the release checksums, and `extension-drift --write` now that npm serves the new packages. `extension-drift` is clean again.

The counts: `#3649` bumped the spelled-out harness count by editing the first half of the word, so "the thirty-three-six coding agents" shipped for three releases in four documents, and the test that reads them could not see it — it strips the wanted word before looking for stale ones, and `thirty-three` out of `thirty-three-six` leaves `-six`. `docs/llms.txt` was checked by nothing and said thirty-two in one line and thirty-one in another. Both are pinned now, each verified by putting the old text back and watching the test fail. The GitHub description said "and 20 more coding agents"; that is fixed outside the repository.

The docs: prose-password redaction in the security model, the privacy page and both READMEs; line-level blame on the auditing page, with what it answers for and what it will not print; the project scope on `deja how` and `deja files`; `deja index --quiet` where someone keeping the index warm will read it. The architecture document's source table listed 24 of 33 stores and its redaction paragraph was four rules behind. The contributing guide now says when the Windows leg runs and why a directory mode bit is not a permission there — the trap two pull requests hit this week.

Fixes #3741.


vshulcz:
Added the fix for #3743 here rather than in its own request, because `main` is red on it and the windows leg is one run either way.

`internal/atomicfile` retried a rename a flat 20 times, 5 ms apart. A reader polling `os.ReadFile` in a loop blocks that rename on Windows — Go's `os.Open` does not ask for `FILE_SHARE_DELETE` — and a constant wait keeps step with a loop, so the writer woke inside the reader's next open every time. It doubles from 2 ms to 40 ms over 30 attempts now, with up to half of itself again at random; a test fails if the jitter becomes a constant, which I checked by making it one.

Labelled `windows` for the leg that found it.

Issue #2050, opened 2026-08-26
a session that only asked the question can outrank the one that answered it


vshulcz:
A session where the same thing was *asked* can outrank the session that
answered it, and `userRoleBoost` is what makes it win.

Found by asking an agent five questions about decisions taken earlier the same
day, through opencode on gpt-5.6-luna, from a directory outside the repository
so files and git history could not answer. One hit in five. Looking at what
recall returned for the misses:

```
$ deja search "claudeworkshop email who sends"
  opencode  "Who is going to send the email to claudeworkshop.com about their outd…
  claude    контакт: info@claudeworkshop.com …
```

The top hit is a transcript of that exact question being asked. The agent
answered from it and got the answer wrong.

Why it wins: a past question matches a query word for word — the strongest
possible lexical match — and every one of those terms is user text, so
`userRoleBoost` (1.3) applies to all of them. The session that settled the
question answers in *different* words, which is what an answer is, so it matches
on fewer terms and none of them boosted.

`TestASessionThatDecidedSomethingOutranksOneThatDidNot` already encodes that
this is wrong, and passes — the fixture separates the two by an assistant line
carrying a decision marker. It does not cover the case where the losing session
is a transcript of the question itself.

Two signals tried and rejected, both caught by that test within minutes:

- **user-only matches.** Demote a session where every query term appears only in
  user text. Wrong: the session that answered matches only in the user's problem
  statement, because the answer is phrased differently — "Root cause was
  transaction pooling" shares no query term. The rule demotes exactly the
  session it should promote.
- **short user line carrying every term.** Demote a user line that holds all
  query tokens and is not much longer than the query. Wrong at any bound the
  fixture allows: "prepared statements pgbouncer keeps failing" is five words
  for a three-token query and is a problem statement, not an echo.

So the separation is not length and not role. What distinguished it in the live
failure was the *shape of the session*: a transcript whose first user message is
the query and which contains no answer to it — a one-shot agent run. That is
closer to "this session concluded nothing" than to anything about the question.

Worth noting the store had been seeded with those one-shot transcripts by the
measurement itself, which is how the case became visible. It is not an artefact:
any agent asked a question writes that question to a transcript deja then
indexes, so every question asked twice has this shape by the second time.

vshulcz:
Dug into the cause behind this. The echo is a symptom; here is the measurement.

**69% of the recent index is an agent runtime's own scratch tree.** Of the 300
most recent sessions on this store:

```
[...]

PR #3745, opened 2026-09-18
fix(policy): reach a working directory folded into one directory name, and name the rule the cross-agent claim rests on

### LINKED ISSUE #3744: the cross-agent claim rests on one rule with no test naming it
The product's headline claim is that a fix found in one agent comes back in another. What carries it is not the project name: the harnesses do not agree on that, and nothing in the tree says so.

Measured on a real store of 2,331 sessions, read-only: of 170 pairs of sessions from **different** harnesses that edited the same absolute file, 26 agreed on the project string and 144 did not — one name contained the other in 84 cases (`src/pool` against `pool`), one was a path where the other was a bare name in 52, unrelated in 7.

The rule that makes the claim hold anyway is `metasWorkingUnder`: a session is admitted by the files it touched under this checkout, whatever project string it was recorded under. Over 515 handovers in that store (sessions per repository root, consecutive pairs within a week), the session before this one was never missing from the candidate pool — never-pooled = 0. The three the block shows carried the predecessor in 496 of 515, and of the 19 misses 13 are subagent runs the digest drops on purpose.

So the path works, and the only thing holding it up has no test naming this case. `TestRecentProjectsUnderNeedsACheckout` covers a session recorded under a name that does not reach the root, with one harness; nothing covers two harnesses disagreeing, which is 85% of the real pairs and the whole cross-agent story. Deleting the rescue leaves that test green.

Wanted: a test that builds the disagreement — Claude's encoded working directory decoding to `parent/base`, Codex's `cwd` giving the bare name — and asserts a session start in the checkout reaches both, with the premise checked so it cannot pass for the wrong reason if the two harnesses ever agree.

### LINKED ISSUE #3746: the job-tree rule misses a working directory encoded into one directory name
The one rule deja applies without being asked — keep the background-job tree out of recall, `*/.claude/jobs/*` — misses the harnesses that encode the working directory into a single directory name.

Measured on a real store of 2,721 rows: 317 name a job tree, and **12 of them are not hidden**. Every one is omp, and the path shape is the reason:

```
~/.omp/agent/sessions/-.claude-jobs-<id>-tmp-dsh-work/2026-08-21T19-03-26-101Z_<id>.jsonl   project=dsh/work
```

The working directory is one segment with dashes where the separators were, so `*/.claude/jobs/*` cannot match it, and the project the parser records keeps only the last two segments — `dsh/work` — which has no `jobs` in it either. Claude Code's own rows under the same tree are hidden because its decoded project string still carries the directory; omp's does not, and neither does any harness that names a session directory this way.

What those 12 rows are is the point: transcripts from a stand a background agent ran inside `.claude/jobs`, which is exactly the class #2050 measured — 207 of the 300 most recent sessions there, one of them a one-shot question that outranked the session that had settled it.

A fix that stays honest about over-reach: when matching a path, also match a directory segment that begins with `-` — the shape every one of these encoders produces, because the path it encodes starts with `/` — after replacing its dashes with separators. A file name with dashes in it (`my-jobs-list.jsonl`) does not begin with `-` and so cannot be decoded into a match.

vshulcz:
Two things from one audit of the real store, in one request so the windows leg is paid once.

**The job-tree rule misses an encoded working directory.** omp records a session under `-.claude-jobs-<id>-tmp-dsh-work` and calls the project `dsh/work`, so neither spelling contains what `*/.claude/jobs/*` matches: of 2,721 rows, 317 name a job tree and 12 were served, all that shape, all stand transcripts from inside `.claude/jobs` — the class #2050 measured at 207 of the 300 most recent sessions. A directory segment beginning with a dash is decoded before the literal match; a name that merely has dashes in it does not begin with one. After the change, on the same store: escapes 12 to 0, hidden rows 305 to 317, nothing hidden whose path and project never say jobs.

**The rule the cross-agent claim rests on now has a test naming it.** Two harnesses in one checkout disagree on the project string in 144 of 170 pairs that edited the same absolute file; what carries the handover is the rule admitting a session by the files it touched under this checkout. Over 515 handovers the predecessor was never absent from the pool, and the three the block shows carried it in 496. Nothing covered the two-harness disagreement, so deleting that rule left the suite green — I checked by returning nil from it. The test also checks its own premise, so it fails asking to be rewritten rather than passing for the wrong reason if the harnesses ever agree.

The audit behind both: 3,027 sessions on disk across 21 stores against 2,726 in the index, all 301 of the difference accounted for — 290 forgotten, 11 with two messages or fewer, none unexplained — and 0 rows whose session no longer exists on disk.

Fixes #3744, fixes #3746.


vshulcz:
Second commit here, from the coverage audit that followed: the job-tree rule misses a working directory a harness folded into one directory name.

omp records a session under `-.claude-jobs-<id>-tmp-dsh-work` and calls the project `dsh/work`, so neither spelling contains what `*/.claude/jobs/*` matches. On a real store of 2,721 rows, 317 name a job tree and 12 were served — every one of them that shape, and all of them stand transcripts from inside `.claude/jobs`, which is the class #2050 measured. A directory segment beginning with a dash is decoded before the literal match now; a name that merely has dashes in it does not begin with one, so a rule cannot reach past what its author meant. After the change on the same store: escapes 12 to 0, hidden rows 305 to 317, nothing hidden whose path and project never say jobs. `internal/policy` stays at 100%.

The audit itself is the reason to trust the number above: 3,027 sessions on disk across 21 stores against 2,726 in the index, with all 301 of the difference accounted for — 290 forgotten, 11 with two messages or fewer, none unexplained — and 0 rows whose session is no longer on disk.

Labelled `windows` since the rule matches paths, which is where that leg earns its keep.

PR #3749, opened 2026-09-19
fix(blame): say when a line spec is not a line, instead of ignoring it (#3738)

### LINKED ISSUE #3738: blame is silent about a line spec it could not use
`deja blame <path>:<line>` says which silence it is when it cannot answer for a line (#3726). It stays silent when the line spec itself is the problem, and answers about the file as if no line had been asked for.

Swept 18 specs against a three-line file in a scratch repository. `a.txt:99` is right: `the file has 3 lines`. What is silent:

- `a.txt:0` and `a.txt:99999999999999999999` — `cutLineSuffix` cuts the suffix and returns line 0 (`strconv.Atoi` overflows on the second), so the line path is never entered and the file-level answer prints with no word about the line.
- `a.txt:abc`, `a.txt:-1`, `a.txt:+3`, `a.txt:2.5` — the suffix is not cut at all, so the whole string is the path and the answer reads `no sessions mention a.txt:abc`.

Both are typed mistakes and the fallback is a correct file-level answer, so this is not urgent. It is the same rule #3726 landed on: a line deja cannot answer for gets said out loud rather than dropped. One sentence where `cutLineSuffix` gives up — the number is not a line number — would cover every case above.

vaibhav8a:
Closes #3738.

## What this changes

Every spec in your sweep now says something:

```
$ deja blame a.txt:0
deja: there is no line 0 — answering for the whole file

$ deja blame a.txt:99999999999999999999
deja: "99999999999999999999" is too large to be a line number — answering for the whole file

$ deja blame a.txt:abc
deja: "abc" is not a line number — answering for the whole file
```

`a.txt:-1`, `a.txt:+3` and `a.txt:2.5` read the same way. `a.txt:99` still says `the file has 3 lines`, and `a.txt`, `a.txt:12` and `a.txt:12:5` still say nothing.

## The two shapes, since they fail differently

- **Digits that name no line.** `trimLineSuffix` cuts them, then `strconv.Atoi` returns 0 or overflows, and the line path was never entered. The path is already correct here — the note is the only thing missing.
- **A suffix that is not digits.** Never cut, so the whole string stayed the path and the answer read `no sessions mention a.txt:abc` — which describes a *missing file*, not a mistyped line.

Both come out of `cutLineSuffix`, as the issue suggests, as a third return that `ResolveBlamePath` carries on `BlameTarget.LineNote`.

## Printed before the answer, not after

A footnote under a file-level answer reads as a footnote to a line-level one. The sentence goes to stderr ahead of the result, so the answer that follows is already framed as being about the file.

## What is searched does not change

This is the part I was careful about: a colon is legal in a unix filename, and `trimLineSuffix`'s own comment says `C:\src\main.go` has to survive. So the second note is narrow — it fires only when the part before the colon looks like a filename (it has an extension) and the part after holds no separator:

| input | note | searched as |
|---|---|---|
| `a.txt:abc` | yes | `a.txt:abc` (unchanged) |
| `notes:draft` | no | `notes:draft` |
| `C:\work\pool.go` | no | unchanged |
| `weird:name.go` | no | unchanged |

A file genuinely named `a.txt:abc` collects a sentence too many and is still searched for as written — a sentence, not a wrong answer. Say the word if you'd rather have it narrower still (requiring a digit in the tail would drop `abc`, at the cost of the case your sweep opens with).

## Not in scope

The MCP `blame` tool. `BlameTarget.LineNote` is populated for it, but surfacing it means adding a field to a payload whose budget logic is load-bearing, and that is a bigger change than this issue. Happy to follow up.

## Tests

`internal/search/blame_line_spec_note_test.go`, 4 cases: every spec from the sweep with its note and the path it still resolves to, silence for usable specs and bare paths, silence for the four colon-in-a-filename forms (with the path asserted unchanged), and the note reaching `BlameTarget`. The existing `TestCutLineSuffixKeepsTheLine` is updated for the third return and otherwise untouched.

`go test ./internal/search/ ./cmd/deja/ -race -count=1` green, `go vet ./...`, `staticcheck` and `gofmt -s` clean.


vshulcz:
Your version is better than mine in three places, and mine is the one that landed — that part is my mistake, not a judgement on this.

What happened, plainly: I filed #3738 out of a sweep, wrote a fix for it, and merged it in #3748 at 03:06 UTC. You opened this at 02:36 UTC. I did not look for an open request before merging my own, which is the one rule I have written down for exactly this situation.

Where yours wins, all three verified against my own sweep:

- `:0` and an overflowing number are different mistakes and yours says so — `there is no line 0` against `"99999999999999999999" is too large to be a line number`. Mine printed one sentence for both.
- The note goes to stderr ahead of the answer. Mine went to stdout, which puts a sentence a reader did not ask for into `deja blame … | …`.
- `mistypedLineNote` decides by shape — the head has an extension, the tail has no separator — where mine asked the disk whether the file exists. Yours leaves `notes:draft` and `C:\src\main.go` alone; mine accused `notes:draft` of being a mistyped line, and mine also needed a syscall to answer a question about a string.

So I rebased this onto main and added a commit that removes my version in favour of yours. Your commit is untouched; the CHANGELOG keeps your entry and drops mine. Tests pass here — `internal/search` green, and the sweep of eighteen specs reads the way your description says.

Merging once CI is green. Thank you for the third point in particular: a rule about a string that reaches for the filesystem is a rule that behaves differently on someone else's machine, and I had not seen it that way.

vshulcz:
Merged. The point worth keeping from this one: a rule about a string that reaches for the filesystem answers differently on somebody else's machine, and yours decides the same question by the shape of the input. I took that as the general lesson, not just for this function.

PR #3751, opened 2026-09-19
docs: the page search lands on — all 33 stores, in Chinese, plus one more query answered

### LINKED ISSUE #3750: the store table lists 25 of 33, and a spelled count anywhere in the guide is unchecked
`docs/guide/where-sessions-are-stored.html` is the page search sends people to — "where does Claude Code keep its history", "where are Cursor chats stored" — and its table lists **25 of the 33** stores deja reads. Missing: Kilo Code, Kiro, Cherry Studio, Command Code, ZCode, Kimchi Coding, gajae-code, Senpi.

The lede says "twenty-five agents, one table", which is the reason nothing caught it. `TestReadmeSpellsTheHarnessCountTheRegistryHas` pins the word in seven named files and this is not one of them; `TestPagesCountHarnessesInDigitsCorrectly` reads every page in the guide but only matches numerals. A count spelled out in prose anywhere else in the guide is unchecked, and the guide is where most of the prose is.

Extending that check to spelled counts finds twelve more stale claims in the same directory:

```
docs/guide/find-a-session.html       "twenty-five harnesses"  (×2)
docs/guide/lost-context.html         "twenty-five harnesses"
docs/guide/resume-a-session.html     "twenty-five harnesses"
docs/guide/memory-for-amp.html       "thirty agents"
docs/guide/memory-for-cline.html     "thirty agents"
docs/guide/memory-for-continue.html  "thirty agents"
docs/guide/memory-for-crush.html     "thirty agents"
docs/guide/memory-for-goose.html     "thirty agents"
docs/guide/memory-for-openclaw.html  "thirty agents"
docs/guide/memory-for-pi.html        "thirty agents"
docs/guide/memory-for-prime.html     "thirty agents"
docs/guide/memory-for-roo.html       "thirty agents"
```

Two shapes have to stay relative or the check is wrong: "and twenty-seven more agents" counts from the ones already named, and the Hermes page's "the other thirty-two agents" counts every store but its own. Both move with the total without equalling it.

Wanted: the eight rows, the ledes, and a check that reads a spelled count anywhere in the guide.

vshulcz:
Two halves of the same page — the one search sends people to.

**It listed 25 of the 33 stores.** Kilo Code, Kiro, Cherry Studio, Command Code, ZCode, Kimchi Coding, gajae-code and Senpi had no row, so the page whose whole purpose is completeness answered nothing for eight agents. Search is the biggest external source here — Google, Bing and ChatGPT together sent 153 of the last fortnight's 1,018 unique visitors, more than any forum — and this is where it lands.

Its lede said "twenty-five agents, one table", which is why nothing caught it: the word check pins seven named documents and this is not one, the digit check reads only numerals. Extending the digit check to spelled counts across the guide found twelve more stale claims — "twenty-five harnesses" on find-a-session, lost-context and resume-a-session, "thirty agents" on nine per-agent pages. Two relative shapes stay relative or the check would be wrong: "and twenty-seven more agents" counts from the ones already named, and the Hermes page's "the other thirty-two agents" counts every store but its own.

**And it now exists in Chinese.** `README.zh.md` is the second most visited path in the repository — 154 of the 707 unique visitors who reached the root over a fortnight, so 22% — with the DeepSeek Harness extension and its own Chinese document next, and linux.do the third referrer. Six of the guide's fifty-four pages were translated and this was not one of them. The table is generated from the English page rather than retyped, so the thirty-three paths cannot drift and only the format column is translated. `hreflang` is declared on both sides, both sitemaps carry the page, and the text sitemap was regenerated from the XML — that pair has its own test, and it caught me.

Fixes #3750.


vshulcz:
Third commit: a page for deleting VS Code Copilot Chat history. Search is the biggest external source here and the pages it lands on are problem-shaped — `delete Codex sessions`, `delete Cursor chat history` — two of those against thirty-three stores. This one answers a question with no documentation anywhere, from what #3637 established: two stores, which one a machine has after the upgrade, how `workspace.json` turns a hashed directory back into a project, and what deleting costs — 11 of 47 transcripts on a real machine have no user turn at all, so they are an agent run and the only record it happened.

My own snippet test caught two mistakes on the way: a 76-character title, and og:title/twitter:title filled with the description, which would have made the social card name a different page.

vshulcz:
Fourth commit, same lever. `Delete Gemini CLI chat history` answers what no documentation covers: the chats sit under `~/.gemini/tmp/<project-id>/chats/` in a directory named after an id rather than a path, so the page gives both ways to map it back — `projects.json` and the `.project_root` marker — names the prompt log beside them that deleting the chats leaves behind, and warns that the same `tmp` tree holds Antigravity's store on some installs, which is why deja reads the two apart rather than treating the directory as one store.

And `session-files-on-disk` — the page that answers how to delete Claude Code sessions and how large `~/.claude/projects` gets — now exists in Chinese with its numbers intact: 875 files, 1.6 GB, a 304 KB median against one 586 MB outlier, and the 30-day sweep nobody is told about.

Problem-shaped pages: two to four. Chinese guide: six pages to eight.

PR #3759, opened 2026-09-19
docs: a recorded demo for line-level blame

### LINKED ISSUE #3758: nothing shows line-level blame working
`deja blame <path>:<line>` has been in since #1181 and nothing shows it. The guide page has a pasted block of output, which is the weakest form of proof there is — it cannot be re-run and it looks the same whether the feature works or not.

The search demo (`assets/demo.gif`) has a generator behind it so it can be re-recorded from scratch. The line-level answer needs more than a corpus: git has to name a commit, and an indexed session has to hold an edit record whose replaced span matches what that commit deleted. That is two halves that have to line up, which is exactly why it is worth showing rather than describing.

Wanted: a `-blame` mode on `scripts/demo` that writes a small repository with two commits and the session that made the second, plus a tape. Synthetic, like the search corpus — recording attribution against a real checkout would put someone's code and their prompts in the same frame.

vshulcz:
`scripts/demo -blame` writes a two-commit repository and the Claude session that made the second commit, timestamped before it, with an edit record whose replaced span is the line the first commit wrote. That is what line attribution needs to line up, so the recording exercises the real join instead of showing a pasted block.

`blame.tape` asks git and then deja about the same line, side by side. `TestTheTapeAsksAboutTheLineTheStandWrites` keeps the two numbers in step with the generated file — adding an import to that source would otherwise move the line and the demo would show deja having nothing to say.

The GIF is on the auditing guide, above the paragraph that explains the rule. 16 seconds, 91 KB.

Fixes #3758.


vshulcz:
Caught while reviewing my own diff: the GIF was committed to `assets/` only, and the page asks for `../assets/blame.gif`, which on the published site is `docs/assets/`. It rendered locally from the README copy and would have 404'd on the site. Fixed, plus `TestEveryImageThePagesAskForExists`, which fails when a page asks for an image the published tree does not have — it catches the missing file, and it had nothing to say about the twenty-odd images that were already fine.

Issue #3747, opened 2026-09-18
ten transcripts the index has never read, and nothing says so


vshulcz:
The index on this machine has never read ten transcripts that the loaders list as theirs. Five are senpi sessions written on **26 July**; five are Copilot Chat transcripts from its second store. Measured against three snapshots of the live index taken over six and a half hours — 1,533 files tracked, the same ten absent in every one.

They are not unreadable, ignored, or forgotten: parsing each one by hand gives a session with one to six messages, `harnessForPath` names the store correctly, none is a symlink, no exclude file or tombstone covers them, and a fresh single-store build tracks all five senpi files and holds all five sessions.

What the ordinary path does with them, run against a copy of that same index:

```
before: tracked=1533 untracked=10 rows=2721
the search reported the index stale=false; after it: tracked=1543 untracked=0 rows=2726   (35s)
```

So one search would adopt them. The reason none has is the shape of the ten: Copilot Chat transcripts are not an appendable kind (`appendableKind("copilot-chat")` is false), and `canAppendIncremental` refuses the whole batch when any changed file is a new file of a kind with no offset parser. Every pass that sees them is therefore replacement-grade — the 35 seconds above — and the only surfaces that take that path are a CLI search, an MCP recall and `deja index`. The hooks never index at all.

And this machine is driven by hooks. Its usage log for the last two days: 101 events on 18 September and 50 on 17 September, every one of them `dejavu`, `tool` or `hook` — not a single recall or search. The sidecars tell the same story: `commands.gob`, `cooccur.gob`, `fixes.gob` and `buckets/` were last written 2026-09-18 00:40 while the manifest says 2026-09-19 01:31.

The mechanism past that point is a hypothesis, not a claim: something keeps the manifest moving without adopting these files, and the candidates are a scoped pass and an abandoned replacement.

What is worth building regardless of which it is: **`deja doctor` should say how many transcripts it has never read, per store.** Every fact above came out of a probe written by hand today; a store that arrived eight weeks ago and never entered the index is exactly the thing a person would want told, and nothing on any surface says it. `deja sources` prints sessions and redaction counts per harness and would have shown senpi at zero — it does — but zero reads as "nothing there yet" rather than "five files never read", and that is the difference between a quiet machine and a broken one.

vshulcz:
[...]

PR #3775, opened 2026-09-19
fix: freshness is measured from when deja read the stores

### LINKED ISSUE #3774: a sync import makes every store look freshly read, so doctor says there is nothing to index
The mechanism behind "nothing says so" in #3747, and it is not the ingest paths I was looking at.

`deja doctor` decides a store is behind by comparing its newest file against `index.ManifestBuiltAt` (`cmd/deja/doctor_report.go:796`). An import writes the manifest and bumps that time without reading a single local transcript — `appendImportedRecords` ends with `m.BuiltAt = time.Now()` (`internal/index/sync.go:1095`). So on a machine that syncs on a timer, every sync makes every store look freshly read, and `freshness up to date` is printed over transcripts deja has never opened.

That matches the evidence in #3747 exactly: the manifest moving at 01:31 while `commands.gob`, `cooccur.gob` and `fixes.gob` stood at 00:40. A sync import rewrites the manifest and touches none of the sidecars, because it read no sources.

This machine has a launchd agent that imports from a peer on a timer, which is why it showed up here and not in any test: nothing in the suite imports and then asks about freshness.

Fix: separate "when the manifest was written" from "when deja last walked the stores". A new `Manifest.SourcesReadAt` stamped by the three rebuild paths and the append path, carried untouched through an import, with `ManifestSourcesReadAt` falling back to `BuiltAt` for a store written before the field. No version bump — the field is additive and an old store answers as it does today.

The doctor lines change with it, because the old wording was the wrong question: `1 store changed since last build` becomes `changed since deja last read it`.

vshulcz:
`deja doctor` called every store fresh on a machine that syncs on a timer. It compares each store's newest file against `ManifestBuiltAt`, and `appendImportedRecords` ends with `m.BuiltAt = time.Now()` — an import rewrites the manifest without opening a local transcript. That is the surface half of #3747: ten transcripts the index had never read, and `freshness up to date` printed over them.

`Manifest.SourcesReadAt` is stamped by the three rebuild paths and the append path, carried untouched through an import, and `ManifestSourcesReadAt` falls back to `BuiltAt` for a store written before the field. Additive, so no version bump and no rebuild for anyone.

`TestAnImportDoesNotClaimTheStoresWereRead` asserts the mechanism both ways: the import moves the build time and leaves the read time alone, and a pass that walks the stores moves it. Nothing in the suite imported and then asked about freshness, which is why this lived.

The doctor lines follow the meaning: `1 store changed since last build` becomes `changed since deja last read it`.

Fixes #3774.


vshulcz:
Reviewed my own diff and found the same lie one level down: the manifest an import writes on a machine that has never indexed carries no files on purpose (#1307), so the BuiltAt fallback claimed the stores had been read at import time. `ManifestSourcesReadAt` now answers "never" for that shape — no files and no read time — and only for that shape, because a store written before this field has files and answering "never" there would call every store behind the moment someone upgrades. Third test covers it.

Benches after the change, from a clone outside the ignore rule: recall@5 1.00 at 0.36 ms, prompt 13/13 with 0 false fires, context 294 tokens median against 16,919 for full history. Nothing here touches the ranking path; that is the record, not a comparison.

vshulcz:
Third commit makes the fact visible rather than only correct. `deja doctor` now says it:

```
status   built (size=215.0 MB, updated=2026-09-19 15:00, stores read 2026-09-18 00:40)
```

— the second date only when it differs from the build time by more than an hour, so an ordinary machine keeps the one-line form, and `stores never read` for an index an import built. `--json` carries `index.sources_read_at` (RFC 3339, or the word `never`), documented in `docs/json-output.md` beside `stale_stores` with the reason the two fields exist, and added to the contract test's key list.

That is the part that would have made #3747 self-diagnosing: the machine could see the index being written daily and its own transcripts last read a day earlier.

PR #3777, opened 2026-09-19
fix: the extension drift check fails instead of reporting

### LINKED ISSUE #3776: two extension packages sit a release behind npm, and the CI check that knows is set to continue-on-error
`node scripts/extension-drift.mjs` reports two packages behind npm right now:

```
opencode-deja: this repository says 0.20.1, npm serves 0.20.2
@vshulcz/openclaw-deja: this repository says 0.20.1, npm serves 0.20.2
```

0.20.2 went out on 2026-09-18. The release publishes these packages and the repository's own `package.json` is left on the previous number until someone runs the script by hand.

CI already runs the check, and hides the result: the step in `.github/workflows/extensions.yml` carries `continue-on-error: true`, with a comment saying to drop it "once the release line has caught up" with dsh-deja. That has happened — the script now reports nothing stranded ahead, only these two behind. So the condition the comment names is met and the guard can start guarding.

Two parts: catch the two packages up, and drop `continue-on-error` so the next release that leaves a package behind fails the extensions workflow instead of printing into a log nobody reads. The workflow is path-filtered to `extensions/**` plus a weekly run, so a binding check lands on exactly the pull requests that should be current, and the weekly run is what notices after a release.

vshulcz:
`node scripts/extension-drift.mjs --write` catches the two packages up to what npm serves — 0.20.2 published them and left the repository on 0.20.1 — and the CI step that already knew this stops being `continue-on-error`.

The comment on that step said to drop it "once the release line has caught up" with dsh-deja. It has: the script reports nothing stranded ahead now, only those two behind. The workflow is path-filtered to `extensions/**` plus a weekly run, so a binding check lands on the pull requests that should be current, and the weekly run is what catches the next release. The failure prints the one command that fixes it.

Fixes #3776.


vshulcz:
Root cause, with timestamps: the follow-up ran its drift check at 13:59:05 after 0.20.2, and npm served the two packages at 14:01:54 and 14:02:50 — three minutes later. So it asked a registry that had not caught up, saw repository and npm both on 0.20.1, and reported that everything was fine. The weekly extensions run was the only other check and it was continue-on-error.

Second commit adds `--after <version>`: the follow-up waits for npm to serve the released version (20s apart, bounded at ten minutes) before deciding, and says which package it gave up waiting for. Three unit tests, injected lookups, no network.

PR #3778, opened 2026-09-19
docs add Chinese guide for finding a session

theluckystrike:
Closes #3765.

This adds a Simplified Chinese page at docs/zh/guide/find-a-session.html translating the English find-a-session guide, matching the structure and voice of the nine existing Chinese guide pages. Commands in code blocks stay byte-identical apart from the translated comment lines, and the page carries the same JSON-LD blocks with inLanguage set to zh-Hans.

To satisfy docs_hreflang_test.go the English find-a-session page now declares the hreflang triple pointing to the new Chinese page, and the new page declares it back. The Chinese URL was appended to docs/sitemap.xml, docs/sitemap.txt, and docs/llms.txt following the placement of the other zh rows.

Verified against the repo's own checks: hreflang pairings on both pages, sitemap.xml and sitemap.txt list equality, title and description rune budgets, og and twitter meta parity, one aria-current marker, and no future-dated JSON-LD timestamps.

PR #3780, opened 2026-09-19
docs: the resume count comes from the registry now

### LINKED ISSUE #3779: the resume claim says twenty of thirty-three, and names a harness that does resume
A translation caught this: #3778 rendered `find-a-session.html` into Chinese faithfully, including the sentence "twenty of the thirty-three harnesses … the remaining five — aider, Continue, VS Code Copilot Chat, DeepSeek Harness and Zed". Twenty and five do not add to thirty-three.

What the registry says — `capabilities.resume`, which `TestCapabilityRegistryMatchesCode` ties to what `resume.go` prints:

- **26** harnesses resume, not twenty.
- **7** do not: aider, Cherry Studio, Command Code, DeepSeek Harness, VS Code Copilot Chat, ZCode, Zed.
- **Continue does resume** — `cn --fork <id>` is wired and the registry records it — and the page told readers it could not.

`resume-a-session.html` carries it twice more: "Twenty of the twenty-five indexed harnesses" in the FAQ block, two counts old, and a body paragraph that explains the wrong five.

Three pages fixed, and the number pinned: the prose is read out of the registry now, both the count and the list of exceptions, so a harness that gains or loses a resume path fails the test instead of leaving the page to be translated into another language.

vshulcz:
The count and the exception list are read out of the registry now, so the three pages that carry the claim cannot drift apart from what `resume.go` prints: 26 harnesses resume, seven do not, and Continue is not one of the seven — `cn --fork <id>` is wired, and the pages said otherwise.

`resume-a-session.html` also said "Twenty of the twenty-five indexed harnesses" in its FAQ block, two counts old, and explained the wrong five in its body. Both fixed from the registry's own `gaps.resume.why`, so the reasons match what the format registry records.

The Chinese page from #3778 is fixed here too — the contributor translated the stale sentence faithfully, which is how this surfaced. Their PR is merged; this is the leftover that was mine.

`TestTheResumeClaimMatchesTheRegistry` fails on all three pages as they were.

Fixes #3779.


vshulcz:
Broadened the invariant to every page under `docs/` and it immediately found a fourth one: `lost-context.html` said **nineteen of thirty-three** in its FAQ block and **twenty of the twenty-five** in its body — two more numbers, neither matching the other two pages. So four pages carried the claim in three spellings and no two agreed.

The check now reads all three shapes: `N of the thirty-three harnesses`, `N of thirty-three harnesses` (no article, which is how the FAQ one hid), and the bare `N of the twenty-five do` with no noun after it. Bench corpus sizes like "thirty task chains" are left alone, which is what the trailing-word list is for.

PR #3782, opened 2026-09-19
docs: regenerate the harness matrix, and check it stays generated

### LINKED ISSUE #3781: the generated harness matrix is a store path behind the registry, and nothing checks it
`docs/guide/harnesses.html` is generated by `go run ./scripts/genmatrix`, and nothing checks that it was re-run. Today it is one store path behind the registry:

```
<vscode-User>/workspaceStorage/*/GitHub.copilot-chat/transcripts/*.jsonl
```

That is the path newer VS Code Copilot Chat builds actually write, added to the registry when the parser learned it. A reader on the harnesses page — the page the site links as the answer to "where does my agent keep its history" — is told to look in the older directory only.

`TestCapabilityRegistryMatchesCode` checks that README has a row per harness, which is why this did not show: the rows are all there, and the paths inside them are what drifted.

One test closes it: every `store_paths` entry in `docs/registry/registry.json` has to appear in the generated page. Measured before the fix, exactly one of 121 was missing, so the check is precise rather than noisy. (README's matrix carries no paths column, so it is not part of this.)

vshulcz:
`go run ./scripts/genmatrix` had not been re-run since the Copilot Chat parser learned the transcripts directory, so the harnesses page named the older path only. Regenerated, and pinned: every `store_paths` entry in the registry has to appear in the page.

Precise rather than noisy — measured against the page as it was, exactly one of 121 paths was missing, and that one is the bug. The existing capability test checks that a row exists per harness, which is why this hid: the rows were all there and the paths inside them drifted.

Fixes #3781.


vshulcz:
Second commit pins the other unchecked number in the same family: `capped at 1,536 bytes per prompt and about 4 KB per tool call` is written by hand on twenty-odd pages, twice each — prose and FAQ block. Both numbers are right today (`promptHookBudget`, `recallMCPBudget`); the reason to pin them is the four claims this week that were not, and this one would go wrong on every per-harness page at once. Mutation-checked: change one page to 1,024 and the test names it.

PR #3783, opened 2026-09-19
docs pin the resume counts on the site pages to the registry

### LINKED ISSUE #3779: the resume claim says twenty of thirty-three, and names a harness that does resume

theluckystrike:
The find-a-session and resume-a-session pages claim twenty of the indexed harnesses resume and list Continue among the refusals. The registry in docs/registry/registry.json says twenty-six resume and Continue does resume.

This fixes the prose on three pages so the counts and the refusal list match the registry, and adds a test that reads the counts back out of the registry and fails when the pages drift from it. Closes #3779.


vshulcz:
This collided with #3780, which was open fifteen minutes earlier and has just merged — and that is my fault twice over: I filed #3779 while already fixing it and did not say so in the issue, and I said I would do it in the #3778 thread rather than in the issue you would have read. I'll say "taken" on an issue I am working on from now on.

Your test covers something mine does not, and it is the interesting half: **the Chinese numeral**. `TestSiteResumeCountsMatchTheRegistry` reads `三十三个智能体里有二十六` and pins it, while mine only reads the English pages for the resume count — so today a translation could drift again and nothing would fail. Mine covers the other direction: it walks every page under `docs/` (which is how a fourth page turned up — `lost-context.html` said "nineteen of thirty-three" in its FAQ and "twenty of the twenty-five do" in its body), all three spellings including the bare form with no noun, and the exception list in both directions.

So the two are complementary rather than duplicates. What I would like, if you are up for it: rebase this onto main and keep only the Chinese check — the prose fixes are already in, and the English claims are pinned. If you would rather not spend another round on it, say so and I will lift that check across with the commit crediting you.

Either way: two PRs in an hour, both of them finding something real. Thank you.

vshulcz:
lifted the Chinese check across in #3808, and it failed the moment it ran on main: CodeWhale (#3804) made it twenty-seven of thirty-four, and 三十四个智能体里有二十六个 was still on the page in two places — the prose and the FAQ block. The half you found is the half that keeps drifting, and the English pages being right is exactly why nothing caught it.

two things changed on the way across. list membership is exact now: "pi" is a substring of "VS Code Copilot Chat", so a `strings.Contains` over that sentence reads it as naming a harness that resumes. and it lives in its own file — this PR rewrote `cmd/deja/capability_drift_test.go` in place, 335 lines down to 131, which would have taken `TestCapabilityRegistryMatchesCode` with it.

closing this as landed, with the commit naming your PR.

vshulcz:
and separately: thank you. the numeral was the one claim on that page no test read, in either language, and you found it inside a day of the translation landing.

Issue #3784, opened 2026-09-19
re-verify one harness store against its live client — nine entries are two months old


vshulcz:
Every registry entry carries `last_verified` — the day someone last checked the format against a real client rather than against its docs. Nine entries are now two months old, and the harnesses on that list are the ones people actually use:

```
64 days  opencode      2026-07-17
64 days  Cursor        2026-07-17
64 days  Antigravity   2026-07-17
57 days  Gemini CLI    2026-07-24
54 days  aider         2026-07-27
53 days  Kimi Code     2026-07-28
53 days  Hermes        2026-07-28
53 days  Cline         2026-07-28
36 days  Copilot CLI   2026-08-14
```

Pick **one** you have installed and re-verify it. Say which in a comment so two people do not check the same one.

What that means, concretely:

1. Run the client, have a short session in it — one question, one file read, one command.
2. Find the files it wrote and compare them against that harness's `store_paths` in `docs/registry/registry.json`. A path that moved, a new file beside the old one, a field that changed name — that is the finding.
3. Run `deja sources` and `deja doctor` and check the harness is listed with the sessions you just created.
4. Update `last_verified` to today, and `store_paths` if anything moved. Regenerate the pages: `go run ./scripts/genregistry` and `go run ./scripts/genmatrix`.
5. If a path did move, a fixture under `fixtures/registry/<id>/` showing the new shape is what makes the parser keep working — keep it small and make up the content.

Two things drifted this way already and both were found by accident rather than by this check: newer VS Code Copilot Chat builds write a `GitHub.copilot-chat/transcripts` directory the page did not name (#3781), and Cursor CLI has a second chat store nobody reads (#3772). A finding of "nothing moved, here is what I ran" is a good outcome too — it moves the date, which is the point.

Haklifesavr:
Taking opencode. Will run it and open a PR with what I find.

vshulcz:
Two additions, both learned from the first PR against this issue (#3789).

**The page keeps its own date.** `docs/registry/<id>.md` carries a `**Last verified:**` line in prose, and `TestRegistryPagesAgreeWithTheRegistry` holds it to `registry.json`. Moving one and not the other turns the tree red. So the step list gains one line: update the date in *both* files, then regenerate.

[...]

PR #3789, opened 2026-09-19
registry: re-verify opencode against live client (2026-09-19)

Haklifesavr:
## Summary

Re-verifies the opencode harness against a live client for #3784.

- Ran a short opencode session: read `test.txt`, ran `echo hello`
- Store still at `~/.local/share/opencode/opencode.db` (matches registry `store_paths`)
- `deja sources` / `deja doctor` both found the session (1 session, 2 messages)
- No path or schema drift — only change is `last_verified` → `2026-09-19`

## Test plan

- [x] Live opencode session (file read + command)
- [x] Confirmed files under `~/.local/share/opencode/`
- [x] `deja sources` lists opencode with the new session
- [x] `deja doctor` reports opencode store found
- [x] `go run ./scripts/genregistry` and `go run ./scripts/genmatrix` succeed

## Notes

Picked opencode on the issue so others do not duplicate it.


vshulcz:
Your claim holds: the store is where the registry says, `deja sources` and `deja doctor` both see the session, and nothing drifted at that path. Merging once CI comes back.

**Two things are on me, not you.**

First, the red check was a leftover I should have made impossible: `docs/registry/opencode.md` keeps its own `**Last verified:**` line in prose, and `TestRegistryPagesAgreeWithTheRegistry` holds the two to each other. Nothing in #3784 said so, and I hit the same wall an hour after writing that issue. I pushed the one-line fix to your branch rather than sending you back for it, and the issue now needs a line about the second file — that is my edit to make.

Second, and worth more than the PR: **your check found something neither of us was looking for.** While confirming "nothing drifted at that path", I looked at what else opencode writes next to the database, and there is a whole store deja has never read:

```
~/.local/share/opencode/storage/session_diff/ses_<id>.json
[{"file": …, "status": …, "additions": …, "deletions": …, "patch": "@@ …"}]
```

1,002 files here. Of 400 of those sessions, 384 are in the index and **17** hold any record of what the session changed — so for the rest this is the only account of it. That is #3791, and #3793 now reads it: `deja blame internal/pool/pool.go` answers with the replaced line for a file that is named nowhere but in those diffs, where before it said "no sessions mention pool.go".

So the re-verify issue paid for itself on the first try, and not in the way it was written. If you want the next one, the same question is open for every harness on that list — "what else does this client write beside the store we read" turns out to be the interesting half.

PR #3796, opened 2026-09-19
docs: add Cline history deletion guide

### LINKED ISSUE #3766: a guide page for deleting Cline task history
Six pages answer "how do I delete X history" — Codex, Cursor, VS Code Copilot Chat, Gemini CLI, opencode and Kilo Code (Claude Code's answer lives on `session-files-on-disk.html`). Cline has none, and its store is the awkward one: two trees, depending on whether you ran it as the extension or the CLI.

From `docs/registry/cline.html`:

- CLI: `${CLINE_SESSION_DATA_DIR:-${CLINE_DATA_DIR:-${CLINE_DIR:-~/.cline}/data}/sessions}/*/*.messages.json` — three environment variables deep, and worth stating in that order because the outermost one wins.
- Extension: `<vscode-globalStorage>/saoudrizwan.claude-dev/tasks/*/api_conversation_history.json`, one directory per task.
- A task directory holds more than the conversation, so deleting the directory takes the rest of that task's state with it. Say which file is the conversation.

`docs/guide/delete-opencode-session-history.html` is the template: where it is, how to delete one, how to delete all, what is lost, and the "keep it searchable instead" section at the end.

The page needs a row in `docs/sitemap.xml` and `docs/sitemap.txt`, a sidebar with exactly one `aria-current="page"`, a canonical link, and a line in `docs/llms.txt`. `TestGuidePagesShareTheirNavigation`, `TestTextSitemapMatchesTheXMLOne` and `TestGuideSnippetsFitWhatSearchShows` will name whatever is missing.

Voyagerroc-Lab:
Closes #3766\n\nAdds the Cline CLI and VS Code extension history deletion guide, links it in both sitemaps and llms.txt.\n\nVerified the shared navigation, sitemap parity, search snippets and date metadata tests.


vshulcz:
Read the whole page against `docs/registry/cline.html`, and the two things that decide whether a delete guide is safe are both right.

The CLI path carries the full chain — `${CLINE_SESSION_DATA_DIR:-${CLINE_DATA_DIR:-${CLINE_DIR:-~/.cline}/data}/sessions}` — in the order the outermost variable wins, which is the part that is easy to shorten and wrong to. And the extension half removes `api_conversation_history.json` rather than the task directory, with the sentence saying why: that directory holds more than the conversation. A guide that says `rm -r` there would cost somebody their task state.

Naming the VS Code-family hosts separately (Code, Insiders, VSCodium, Cursor, Windsurf) is a detail the other delete pages do not spell out and should.

The doc invariants pass here too. Merging once CI is back — it is a first-time contributor branch, so the runs needed approving; that is done.

vshulcz:
Thank you — the two details that make a delete guide safe rather than merely correct are both in it: the variable chain in the order the outermost one wins, and removing the conversation file rather than the task directory beside it. Merged.

Issue #441, opened 2026-07-28
Forgotten sessions come back if the tombstone file is lost


vshulcz:
`deja forget` drops matching sessions and records a tombstone so a later `deja index` cannot restore them from the source history, which is still on disk. The record lives in one place: `$XDG_CONFIG_HOME/deja/tombstones`.

Delete that one file — a wiped `~/.config`, a machine migration, an `XDG_CONFIG_HOME` that moved, someone tidying up — and the next reindex brings the forgotten sessions back. Verified:

```
deja forget --session f1     -> messages dropped: 1, tombstones added: 1
deja index --rebuild         -> still gone
rm ~/.config/deja/tombstones
deja index --rebuild         -> [claude] proj · Jan 2 · f1 — 1 matches
```

Forgetting is the one operation people use for things they specifically do not want kept, so the quiet failure mode is the wrong one to have. The record should be mirrored into the index directory and the two merged on read: losing both is a much narrower accident than losing either, and the index is exactly what the tombstone protects.

PR #3842, opened 2026-09-21
docs(day-zero): every tool the comparison scores links the repository it scored

### LINKED ISSUE #3839: day-zero compares six other tools and links none of them
The day-zero page names six other tools in its comparison table and none of them is a
link. A reader who wants to check a row — or just see which `ctx` or `funes` this is,
both names being taken several times over on GitHub — has to go search for it.

Current header, `docs/guide/day-zero.html:150`:

```html
<tr><th></th><th class="us">deja-vu</th><th>funes</th><th>ctx</th><th>CASS</th><th>agentmemory</th><th>MemPalace</th><th>claude-mem</th></tr>
```

The footnotes under the table do name versions (`ctx 1.3.1 (ctx.rs, …)`, `funes 1.3.0
(Hugging Face, …)`), so the identification is recoverable from prose, but the repository
is never given.

Expected: every compared tool in the header links to the repository that was measured.

The six, all verified 200 today:

```
$ for u in huggingface/funes ctxrs/ctx Dicklesworthstone/cass_memory_system \
           rohitg00/agentmemory MemPalace/mempalace thedotmack/claude-mem; do
    printf '%s %s\n' "$(curl -so /dev/null -w '%{http_code}' https://github.com/$u)" "$u"
  done
200 huggingface/funes
200 ctxrs/ctx
200 Dicklesworthstone/cass_memory_system
200 rohitg00/agentmemory
200 MemPalace/mempalace
200 thedotmack/claude-mem
```

Impact: cosmetic on its own, but this is the page that asks a reader to trust six rows
of numbers about other people's software. A row you cannot click is a row you cannot
check.

Follow-on from #3147 and #3152, which added the funes and ctx rows.

### LINKED ISSUE #3845: day-zero linked CASS to a different project by the same author
The comparison table header in #3842 linked CASS to `Dicklesworthstone/cass_memory_system`. The tool the row measures is `Dicklesworthstone/coding_agent_session_search` — the footnote cites version 0.7.1, issue #441 and commit 38f0412, and all three resolve only there:

```
$ curl -so /dev/null -w "%{http_code}\n" https://github.com/Dicklesworthstone/cass_memory_system/commit/38f0412
404
$ curl -so /dev/null -w "%{http_code}\n" https://github.com/Dicklesworthstone/coding_agent_session_search/commit/38f0412
200
$ curl -s https://api.github.com/repos/Dicklesworthstone/cass_memory_system/issues?state=all\&per_page=1 | grep -o "\"number\": [0-9]*" | head -1
"number": 80
```

The highest issue number in `cass_memory_system` is 80, so `#441` cannot be there either; its tags stop at v0.2.14.

Expected: the header links the repository the row scored. Actual: it linked a neighbouring project by the same author, which would read as scoring the wrong tool.

Caught before merge; fixed in #3842 along with a test that pins every column to its repository. Filed so the mistake is on the record rather than only in a diff.

Follow-on from #3839.

vshulcz:
The comparison table header now links each of the six other tools to the repository that was measured, and `TestDayZeroLinksEveryToolItCompares` pins every column to its repository — a bare name fails it, and so does a link to the wrong project.

That second check is here because the first version of this branch got CASS wrong: it linked `cass_memory_system`, a neighbouring project by the same author, while the version, issue and commit the footnote cites (0.7.1, #441, 38f0412) resolve only in `coding_agent_session_search`. #3845 has the evidence. Reverting the fix, the test says:

```
docs_day_zero_links_test.go:63: CASS links Dicklesworthstone/cass_memory_system, want Dicklesworthstone/coding_agent_session_search
```

Two more version strings on the same line are wrong and are not fixed here, because fixing them properly means re-running those two rows: #3844.

The sitemap and feed changes are `scripts/genregistry` and `scripts/genfeed` output — touching the page moves its `lastmod` to today, which reorders the feed.

Closes #3839, Closes #3845


vshulcz:
Review found the header regex would reject `<tr id="…">` or `<th data-…>`, and that the owner/repo check failed on a trailing slash. Both loosened in b5615ec6 — verified by editing the page to `<tr id="cmp"><th data-i="0">…` with `https://github.com/huggingface/funes/` and re-running: passes. Stripping the links back to bare names still fails and still names all six.

Left the `rel="noopener"` without `target="_blank"`: the site uses that on six other guide links and has no `target="_blank"` anywhere, so this follows the page next door rather than introducing a rule.

PR #3864, opened 2026-09-21
docs: the two LongMemEval numbers say why they differ, and day zero says what its corpus is

### LINKED ISSUE #3851: the site publishes 85.3% and 18% hit@1 for LongMemEval-S one click apart, with nothing connecting them
The site publishes two hit@1 numbers for the same dataset, on pages one click apart, and
neither says the other exists.

`docs/guide/benchmarks.html:142`:

> LongMemEval-S (session-level retrieval, 470 questions on the cleaned set): hit@1 85.3%,
> hit@5 95.5% … Both re-measured on 0.20.2

`docs/guide/day-zero.html`, table:

> hit@1 / 100 — 18

Same dataset name, same tool, 85.3% against 18%. The difference is the haystack: the
benchmark asks each question against the sessions it ships for that question — about fifty —
while day zero asks all 100 against the whole 19,195-session pile at once. The day-zero page
does explain the mechanism in prose:

> The absolute hit@1 is low for every tool because these are LongMemEval questions asked
> against nineteen thousand sessions rather than the fifty the benchmark ships per question

but it never names 85.3%, and the benchmarks page never mentions the day-zero regime at all.
The README (line 25) carries the 85.3% on its own, above the fold, with no qualifier.

Control that the two really are the same dataset and not a labelling accident: both read the
cleaned LongMemEval-S set, and the day-zero harness takes it by path —
`go run ./scripts/day0bench -data longmemeval_s_cleaned.json …` — while the benchmark harness
is `scripts/longmemeval`. Same file, two setups.

Expected: a reader who sees both knows within a sentence why they differ. Actual: the two
numbers sit four-and-a-half times apart with nothing connecting them, and the smaller one is
on the page that argues for trusting our measurements.

Impact: this is the first thing someone checking the claims will find, and "their own site
says 18%" is a cheap objection to hand out. It costs one sentence on each page to close.

Also worth deciding: whether the README's 85.3% should carry the qualifier that the benchmark
haystack is per-question. That is a wording call, not a fact call.

### LINKED ISSUE #3860: day-zero calls its corpus "coding-agent sessions"; 98% of it is personal-assistant chat in coding-agent file layouts
The day-zero page's structured data describes its corpus as "19,195 coding-agent sessions".
The sessions are LongMemEval-S conversations — a personal assistant asked about guitar amps,
trip plans and gift ideas — written to disk in the file layouts Claude Code and Codex use. The
layout is real; the content is not coding work, and the page never says so.

`docs/guide/day-zero.html:35`:

```
"description": "Seven memory tools installed on a machine with 19,195 coding-agent sessions already on disk, …"
```

and the lede, `:152`: "The machine already holds months of Claude Code, Codex and Cursor sessions".

What the sessions contain — first user turn of 400 sampled at random (seed 7) from the lay-down:

```
sessions sampled: 400; first user turn mentions code-ish terms: 6 (2%)
  - I'm thinking of buying a new guitar amp, but I'm not sure what to look for. I've been play
  - I'm thinking of planning a team-building event for my office and I was considering organiz
  - I'm planning a trip to San Francisco and I was wondering if you could recommend some good
  - In salesforce, how can I achieve a normalized data model and at the same time allow abtitr
  - I'm looking for some gift ideas for my family members. I've already got a few gifts recent
  - I'm planning a fundraising event and I'm looking for some ideas on how to make it more eng
```

The "code-ish" test is generous (it counts `api`, `error`, `bug`, `code` anywhere in the turn),
so 2% is an upper bound.

Control that the setup section is otherwise accurate: it does say the sessions come "from the
LongMemEval-S cleaned set", and anyone who knows that dataset knows what it holds. The problem is
the reader who does not, and the one sentence that tells them the opposite.

Expected: the page says what the corpus is — a public question-answering set with ground truth,
chosen so every question has exactly one answering session, laid out in the real transcript
formats — and does not describe it as coding work. Actual: the machine-readable description
says "coding-agent sessions" and the lede implies months of coding history.

Impact: this is the first thing a skeptical reader checks by opening a single file, and the
answer undercuts every number under it. It is also why the absolute scores are what they are:
general-chat vocabulary is not what a coding-session index is tuned for, which cuts against deja
as much as anyone, and is worth saying.

Related: #3851 (the page never connects its 18% to the benchmark page's 85.3%).

vshulcz:
Two pages one click apart published 85.3% and 19 hit@1 for the same dataset with nothing connecting them, and the day-zero page's machine-readable description called its corpus "coding-agent sessions" when it is LongMemEval-S chat in coding-agent file layouts.

- benchmarks.html gets a note: the harness asks each question against the ~50 sessions the dataset ships for it; the same 100 questions against 19,195 sessions score 19 at rank one, with the link.
- day-zero.html names the 85.3% and the reason (same questions, same index, haystack four hundred times larger), says what the corpus actually holds, and that a code-tuned index gets no help from guitar-amp questions — which cuts against us.

Fixes #3851, fixes #3860


vshulcz:
Duplicate — #3862 landed the same two fixes with a test that holds the pages together.

